Issue #374, opened 2015-04-09
Star Soldier  Vanishing Earth - Cutted Screen


Predator82Germany:
![unbenannt](https://cloud.githubusercontent.com/assets/6679921/7070848/e3adb1c0-dee1-11e4-95b7-6be84d4edacc.jpg)

gonetz:
Can't reproduce.

Predator82Germany:
It´s only the choose mode screen

gonetz:
It's looks like this for me:
![gliden64_star_soldier_000](https://cloud.githubusercontent.com/assets/8983092/7072422/1a5c4b4c-df0e-11e4-85d0-fb87faf38525.jpg)

Predator82Germany:
Again a NVIDIA problem?

gonetz:
It's hardly NVIdia problem. I've uploaded new build, please test with it.

Predator82Germany:
Still exist with 09Apr

gonetz:
Still can't reproduce it. May be it is NVidia issue indeed. Need reports from other NVidia users.

Predator82Germany:
Are there some?

baptiste0602:
same problem on my nvidia

gonetz:
OMG! What else can it be? Is anything in gliden64.log?

baptiste0602:
nope

Predator82Germany:
Was this sarcasm gonetz? ;)
No log

baptiste0602:
FYI, it didn't happen in previous build (week26)

gonetz:
It's not sarcasm. I need starting point to search the problem. So, it is something between Week 26 and Yesterday. I hope that Mupen64Plus build works the same. I'll prepare bunch of builds and upload them for you. You'll tell me, with which one the problem appears. Fastest way is to use Mupen64Plus command line. Unpack new plugin build and in command line prompt type/repeat mupen64plus.exe <rom file name>. I use that method when I need to find quickly a regression bug.

gonetz:
A pack with test builds uploaded.

baptiste0602:
It appears on 6_b623b6e

Predator82Germany:
Right 1-5 works 6 & 7 are broken

gonetz:
Fast answer, thanks!
I'll prepare new bunch of builds for commits between 5 and 6. We will catch it soon!

gonetz:
New pack uploaded.

[...]

Issue #363, opened 2015-04-09
Lylat Wars - Crash On ROM-Start


Predator82Germany:
With all regions & NVIDIA

baptiste0602:
I confirm with default settings.

No problem with these settings : 

[STARFOX64]
frameBufferEmulation\copyToRDRAM=0
frameBufferEmulation\validityCheckMethod=1

gonetz:
The game crashes on copy to RDRAM. Buffer height calculated incorrectly.

Predator82Germany:
With baptiste´s settings here it works but NVIDIA

gonetz:
Crazy game allocates buffer of less size than necessary. How it works on real console?!

gonetz:
Fixed.

baptiste0602:
I confirm.
It also fix crash with copyColorToRDRAM for Zelda Majora's Mask (E) and Operation Winback (E).

Great job !

Predator82Germany:
I didn´t have copyColorToRDRAM?

baptiste0602:
It's the same as "Copy frame buffer to RDRAM". In mupen64plus, the option is named "EnableCopyColorToRDRAM"

PR #379, opened 2015-04-10
Update GLideN64.ini NVIDIA-Optimized

Predator82Germany:
With added Good_Name & alphabetically sorted after this

### DIFF
--- a/GLideN64.ini
+++ b/GLideN64.ini
@@ -57,83 +57,153 @@ blendMode=0
 blurAmount=10
 blurStrength=20
 
-[DR.MARIO%2064]
-frameBufferEmulation\copyFromRDRAM=1
-
-[JET%20FORCE%20GEMINI]
-frameBufferEmulation\copyFromRDRAM=1
-
-[J%20F%20G%20DISPLAY]
-frameBufferEmulation\copyFromRDRAM=1
-
-[STAR%20TWINS]
-frameBufferEmulation\copyFromRDRAM=1
+[TWINE]
+Good_Name=007 - The World Is Not Enough (E)(U)
+frameBufferEmulation\N64DepthCompare=0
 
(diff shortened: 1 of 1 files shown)


GLideN64.ini
```diff
@@ -57,83 +57,153 @@ blendMode=0
...
+frameBufferEmulation\copyFromRDRAM=1
+
 [LEGORacers]
+Good_Name=Lego Racers (E)(U)
 frameBufferEmulation\detectCFB=1
 
+[STARFOX64]
+Good_Name=Lylat Wars (A)(E) / Star Fox 64 (J)(U)
+frameBufferEmulation\copyToRDRAM=0
+frameBufferEmulation\validityCheckMethod=1
```
baptiste0602: It's not necessary since #363 is fixed

baptiste0602:
Good work Predator.

Since the below options are default values, maybe it's not necessary to force them (what do you think @gonetz ?) : 
copyToRDRAM=1 (not in this gliden64.ini but option is check when you press "Resore Defaults")
copyDepthToRDRAM=1
N64DepthCompare=0

gonetz:
From the one side setting custom option to its default value is redundant change.
From the other side, it can make sense if you know, that that setting is essential for correct game work. Then the game will work correct even if defaults will be changed for some reason.

GLideN64.ini
Predator82Germany: Ok thanks

Predator82Germany:
N64DepthCompare=0 have I forced cause they normally set to 1 but NVIDIA don´t like this
So if this gets fixed anytime they can set to 1
Others are only forced cause needed & not defaul

gonetz:
Ok, lets try to fix depth compare after finish with #374

PR #405, opened 2015-04-14
All Japanese ROMs (No-Intro) added & little fixes

Predator82Germany:


Predator82Germany:
Cause these have a weird intern name which can GLideN64 only handle with a checksum

Predator82Germany:
Cause these have a weird intern name which can GLideN64 only handle with a checksum

gonetz:
So, Predator82Germany added them just in case you need to add a special setting for such a game.
We don't have standalone converter "Japanese rom name => checksum"

Issue #416, opened 2015-04-16
Pokemon Stadium 2 framebuffer validity check fails


purplemarshmallow:
There are many glitches with the default validity method (checksum). They are all gone if "fill RDRAM" is used instead. But after more careful testing I noticed that this may cause freezes when entering/exiting the Pokemon Lab. Using the pokedex in the lab always results in a freeze. The rest of the game works fine with "fill RDRAM". Entering the pokemon lab requires transfer pack emulation.
![1](https://cloud.githubusercontent.com/assets/7283660/7188807/a520d1fa-e47a-11e4-9a53-911180da7d15.png)
![2](https://cloud.githubusercontent.com/assets/7283660/7188810/a85b95b2-e47a-11e4-8aa0-b3c482bca001.png)
![3](https://cloud.githubusercontent.com/assets/7283660/7188811/ab17348c-e47a-11e4-90a0-466aa4e29eca.png)
![4](https://cloud.githubusercontent.com/assets/7283660/7188816/ad7cea8c-e47a-11e4-9e32-e329039c0c60.png)

The Pokemon Lab:
![5](https://cloud.githubusercontent.com/assets/7283660/7188984/fea9c1a4-e47b-11e4-8ba6-9e28fd791841.png)

gonetz:
I have an idea for new validity check method. It may replace the both methods which are currently used. I need volunteers to check it. If you are ready for extensive testing, drop me a line by email. I need your mail to send test builds.

purplemarshmallow:
sent you an email

purplemarshmallow:
I found out that "fill RDRAM" is safe for this game. With the early release build it only causes crashes if "copy frame buffer to RDRAM" is enabled at the same time. With the latest build it is always safe. Maybe because a bug has been fixed. Fingerprint method works most of the time but not always. So I recommend the following settings:

[POKEMON%20STADIUM%20G&S]
Good_Name=Pocket Monsters Stadium GS (J)
frameBufferEmulation\validityCheckMethod=1
frameBufferEmulation\copyToRDRAM=0
frameBufferEmulation\copyDepthToRDRAM=0

[POKEMON%20STADIUM%202]
Good_Name=Pokemon Stadium 2 (E)(F)(G)(U)
frameBufferEmulation\validityCheckMethod=1
frameBufferEmulation\copyToRDRAM=0
frameBufferEmulation\copyDepthToRDRAM=0

[...]

Issue #410, opened 2015-04-15
Pokemon Stadium 2 (F) freeze


purplemarshmallow:
Pokemon Stadium 2 (U) works perfectly. But with Pokemon Stadium 2 (F) the N64 logo appears and then the screen turns black and the game freezes. I don't have this Problem with glN64. Tested with 1964 and PJ64.

purplemarshmallow:
There are also problems with other regions. It doesn't freeze if both "copy frame buffer to RDRAM" and "copy depth buffer to RDRAM" are disabled. Add these settings to the ini?

Issue #358, opened 2015-04-09
Donkey Kong 64 (E) - Zipper Is Not Correct


Predator82Germany:
Happens only with (E)
May related to #355

gonetz:
Yes, zip effect requires copy fb to RDRAM.

baptiste0602:
Seems fixed, no ?

olivieryuyu:
please close it works

Issue #355, opened 2015-04-07
Donkey Kong 64 (E) crash at start and when loading a new game


baptiste0602:
Donkey Kong 64 (E) version crash after Rare logo when one of this option is enabled :
- copyDepthToRDRAM
- copyToRDRAM

And it crash when you start a new game when one of this option is enabled : 
- copyDepthToRDRAM
- copyToRDRAM
- detectCFB

My settings to boot DK without crash : 
![image](https://cloud.githubusercontent.com/assets/11183826/7032914/5c6aae46-dd76-11e4-94b0-1e8b3b677446.png)

My setting to start a new game without crash : 
![image](https://cloud.githubusercontent.com/assets/11183826/7032939/91c1d84e-dd76-11e4-9e3a-59ca1875d0b7.png)

No problem with (U) version.

baptiste0602:
Not fixed with week27 build with checksum option.

If you're interested, error log from cdx4 rsp with mupen64plus : 
https://drive.google.com/file/d/0ByOA3_oOu6O9bjhzeUdfMHhwZUE/view?usp=sharing

Same error log with hatcat's rsp and pj64 with an additionnal message box : 
![image](https://cloud.githubusercontent.com/assets/11183826/7051742/c2d2cd30-de2a-11e4-9f71-ec2e6d043bdf.png)

Problem happen with other RSP : PJ64 RSP, mupen64plus-rsp-hle.dll

gonetz:
Ok, I tested (U) version.

gonetz:
Fixed

Predator82Germany:
copyDepthToRDRAM or copyToRDRAM still crash on (E)

gonetz:
Yes, now it pass the logo, but crashes later.

baptiste0602:
Exactly, it pass the "donkey rap" :) but it crash after (previously it crashed after Rare logo).
And still crash when you start a new game.

Predator82Germany:
With both options off I can start a new game & see full intro

gonetz:
It's either bug in game code, or bug in VI emulation on emulator's side. Frame buffer height calculated by the plugin is higher than its real height. Out of buffer write causes crash. I can't make real fix for it, only hack. I don't want to add such hack before release. So, (E) version will remain incompatible with buffer copy.

baptiste0602:
OK

Predator82Germany:
Then we should set copyToRDRAM=0 til fix?

[...]

PR #428, opened 2015-04-27
New fb validity check

gonetz:


Predator82Germany:
[MARIOGOLF64]
Good_Name=Mario Golf (E)(J)(U)
frameBufferEmulation\validityCheckMethod=0

Can´t see differences with all 3 options (may I´m blind then correct me), so use default value?

Predator82Germany:
[THE%20LEGEND%20OF%20ZELDA]
Good_Name=The Legend Of Zelda - Ocarina Of Time (E)(U)(J)
frameBufferEmulation\validityCheckMethod=1

[ZELDA%20MASTER%20QUEST]
Good_Name=The Legend Of Zelda - Ocarina Of Time - Master Quest
frameBufferEmulation\validityCheckMethod=1

Should stay so, is needed to see Link in pause-screen

gonetz:
If a game works with no difference for any validityCheckMethod option, it's better to remove that setting from custom settings for this game at all.

validityCheckMethod=1 for Zelda OOT is essential with no doubt.
Have you checked other games, which had validityCheckMethod = Fill RDRAM with other options?
May be Zelda is the only bad case which need it? It would be great, because Fill RDRAM is dangerous.

I'd like to get results for #410 and #416
@purplemarshmallow, where are you?

purplemarshmallow:
I commented in #416 looks like validityCheckMethod = Fill RDRAM is needed but I think it's safe

Predator82Germany:
Have all tested which have validityCheckMethod in custom.ini while Pokemon

purplemarshmallow:
Fill RDRAM seems to work much better now. I tested #323 but I can't reproduce a crash with the latest build. But it would be risky to use it for every game.

purplemarshmallow:
But fingerprint method works really good. Only 2 games seem to have a problem with it. Zelda OOT and Pokemon Stadium 2. I can't find any other games.

baptiste0602:
@Predator82Germany : For Mario Tennis, Fill RDRAM is useful to fix some shadows.

With fingerprint : 
image

With fillRDRAM : 
image

gonetz:
Let me summarize feedback from this thread and from #416

1. Fill RDRAM is the most reliable method for detection frame buffer validity. purplemarshmallow also found that it's current use does not cause crash in games which crashed before. I guess that this improvement is caused by the fact that I use Fill RDRAM only for auxiliary frame buffers in that version. Main frame buffers are checked using pure Copy RDRAM method, which is safe.
2. Fingerprint gives less reliable result. Again, fingerprint method is used only for auxiliary frame buffers.
3. Plugin still can detect height of auxiliary buffer wrong. Fill RDRAM will cause crash in that case. We don't find such cases, but I can't believe that my code can handle all possible cases. N64 games are very insidious. Fingerprint fills only 2% of buffer area and thus has very little chance to spoil program data in RDRAM.

Conclusion: in fact, I've implemented 2 different methods for validity detection of auxiliary frame buffers: fill and fingerprint. Validity detection for main frame buffers silently uses safe copy method only. Thus, we need only these two methods in GUI. 0 for fingerprint, 1 for fill RDRAM
Since fingerprint is more safe and works fine for most games, I suggest to use it as the default one.
Few games, namely Zelda OOT , Mario Tennis and Pokemon games will get validityCheck=1 (fill) in custom ini. validityCheck option must be removed for other games in custom ini.
That solution will make the project most compatible and safe.

If everyone agree with me, let's finish that work and push it to release level.
I'll correct code and GUI.
@purplemarshmallow, @Predator82Germany and @baptiste0602 : please prepare custom.ini, which I can pull, check and push to release branch.

Predator82Germany:
Pull request done

purplemarshmallow:
Fill RDRAM causes a fatal error with knockout kings 2000
The framenuffer effect still does not work #350

gonetz:
:(

Predator82Germany:
Same crash message like with Donkey Kong 64 (E) #355 & #358

gonetz:
@Predator82Germany - does it crash with any option, or only with Fill RDRAM?

Predator82Germany:
Have all off & only fill on (crash) with fingerprint it works

New bad news: F-Zero X crash after cup select (all regions) with both validity checks

gonetz:
Ok, now we know cases, where Fill does not work. Since Fingerprint is default, it's not a problem.
We may force set it for the games, which crash with Fill option for safety.

gonetz:
I'll check F-Zero

gonetz:
Stupid mistake in my yesterday fixes. Will be fixed and uploaded soon

gonetz:
F-Zero fixed

Predator82Germany:
May we should force all knowed settings to get the best result for users, so they can do nothing wrong if they change settings in gui

gonetz:
Users still able to shot themselves in a leg if they want to: custom ini use is optional.

But the idea is good. What you suggest to add to custom settings?

Predator82Germany:
In this thread
http://gliden64.boards.net/thread/93/gliden64-custom-game-settings
I´ve stored all required settings for most playability til today
When something should forced than this

Predator82Germany:
You´ve send the wrong file?

gonetz:
When?

Predator82Germany:
F-Zero X fix has sam date / time how before

gonetz:
It is not so in my archive. dll file has CRC32=2B773240

gonetz:
Try download again

gonetz:
I downloaded the archive. It is correct.

Predator82Germany:
Have E0A43419

purplemarshmallow:
F-zero x still crashes with f zero x fix
so I think it's the wrong file

Predator82Germany:
lol we should report plugin errors & get download-isues :D

gonetz:
Reuploaded

Predator82Germany:
Works

gonetz:
As expected :)
No idea, what was wrong with previous upload.

Predator82Germany:
May filename
Before Knockout Kings 2000 crash the Rumble-Pack message disappears may its a rumble issue & not fill-fail

PR #462, opened 2015-05-02
Brazilian Portuguese Translation

DaEckster:
Complete Brazilian Portuguese translation, should be understandable for any Portuguese speaker (even non-Brazilians). I'm not native, but it's a good starting point and I can speak as well as a native, lived there for 2 years. Some technical words like "cache", "frame" "buffer", and "performance" weren't translated, but generally they aren't in Portuguese.


gonetz:
It works, thanks!

I put your commits into special 'translations' branch, not in master.

Issue #244, opened 2015-02-19
Perfect Dark frame buffer effects do not work when FSAA is on.


gonetz:
Cap Spy, Night Vision and other full-screen effects does not work with FSAA on

gonetz:
Technical issue. Hard to resolve.

gonetz:
Paper Mario, same issue

mudlord:
I thought with shaders you can resolve from multisampled render targets/FBOs. :/

gonetz:
Yes, it is possible. In fact you have to manually blend samples from multisampling texture in your shader. Sounds creepy, so I left it for better times.

purplemarshmallow:
AA breaks the rocket launcher in Conker's BFD and possibly other framebuffer effects in this game

thecoreyburton:
Would it be possible to add a bruteforce method of supersampling as an alternative to combat this?

I know you can set custom resolutions for the glide plugin, eg 960x720. Add two radio buttons above the anti aliasing slider, one that says "Multisampling (default)" and the other that says "Supersampling" (that way people use the original method by default but if they run into problems they can try the alternative).

As for the feature, if Supersampling is selected it takes the multiple and then multiplies the resolution you've selected by this number. If your resolution is 960x720 and your anti aliasing value 4x it'd render at 3840x2800. Then, add a Lanczos3 resize to that to resize back down to the initial resolution, so 960x720. That way the plugin is just rendering at a larger resolution, then resizing the output down back to the selected resolution and displaying at what the user selected.

It'd be performance intensive and I know a lot of systems could struggle with it, but high end systems could handle it and as hardware evolves it'd definitely become a standard.

I'm not sure how the current method works, nor am I sure how complicated it'd be on a programming level to do this. But would a method like that fix or help the issue?

mudlord:
No.
Solution is to resolve MSAA'd FBOs properly. 

An example pixel shader of how to resolve MSAA'ed FBOs is as follows:

vec4 sampleMS(sampler2DMS texture, ivec2 ipos)
{
   vec4 texel = vec4(0.0);	
    for (int i = 0; i < u_msaaSamples; i++)
    {
    	texel += texelFetch(texture, ipos, i);
    }
    texel /= float(u_msaaSamples);
    return texel;
}

void main(void)
{
	// MSAA sampling.
	vec2 framebufferTextureSize = vec2(textureSize(u_framebufferTexture));
[...]

Issue #249, opened 2015-02-20
"FACTOR 5 GAMES" AA emulation breaks Title Screen's


theboy181:
Black Screen at title screen when AA emulation is on.

Indiana Jones, Rouge Squadron,  Battle for Naboo.. ect..

theboy181:
Tested a little further today. When the FBE is used the MENU is rendered really dark, but if you use AERO transparencies you can see that the screen is getting rendered.

To see what I am talking about DRAG the window under the TASKBAR at the bottom of the screen while using AERO. It will show that the menu is actually getting rendered.  (screen grabs didn't show my findings)

purplemarshmallow:
fixed?
Rogue Squadron works perfectly here

theboy181:
Still black for me, looks like its an AA issue.. updated
@ gonetz

purplemarshmallow:
pull request #500 created

theboy181:
Issue still exists with AA on after DEV.

@gonetz

theboy181:
Screen now flashes with AA on .. could it be linked to interlacing?

purplemarshmallow:
But AA + LLE does not seem to work in general

theboy181:
How so?  It looked aliased to me during my rouge squadron test.

PR #500, opened 2015-05-04
disable AA in Factor 5 games #249

purplemarshmallow:


theboy181:
This is not a solution to simple turn off AA. These games look horrible without AA imo.

purplemarshmallow:
Yes it is not a solution
But I think AA should be disabled until this is fixed
#244 same here

purplemarshmallow:
But I noticed it also affects Conker's BFD
there are so many games where AA does not work
add conker?

mudlord:
Thats not a valid solution.
The valid solution is fixed AA.

gonetz:
#244

PR #514, opened 2015-05-08
Reduced Main Menu button translation size

DaEckster:
Changed "Acerca de" to "Sobre" since their translations are equivalent and it makes for a smaller menu button.


gonetz:
Pushed, thanks!
You just in time: new public release is almost ready.

PR #522, opened 2015-05-09
Initial Polish translation

micechal:
Spent some time on this one, hope it's alright. I couldn't find a way to upload a file directly rather than typing it manually in (I'm making this PR from a web browser) so I copied it over from Notepad++. Not sure if it made some incompatibiliteis, but as far as I can tell, the file didn't contain any non-printable characters or stuff like that so it should be fine.


gonetz:
Add to the repository.
I've compiled your translation file. It loaded with no problems. Thanks!
Do you planing to make corrections to it?
I may add it to the translations pack right now.

translation_pl

micechal:
I might want to change some strings I wasn't fully satisfied with at the time of translating later.

PR #536, opened 2015-05-16
always show cfb until first dlist

purplemarshmallow:
many pd roms and demos use cpu rendering only
now they are supported with default settings
also detectCFB=1 now isn't needed anymore for some games

### DIFF
--- a/src/RDP.cpp
+++ b/src/RDP.cpp
@@ -561,6 +561,7 @@ void RDP_ProcessRDPList()
 	if (dp_end <= dp_current) return;
 
 	RSP.bLLE = true;
+	RSP.bfirstDlist = true;
 
 	// load command data
 	for (u32 i = 0; i < length; i += 4) {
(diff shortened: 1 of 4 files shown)


purplemarshmallow:
don't merge yet
just noticed RSP_ProcessDList isn't called in LLE

purplemarshmallow:
now same behavior in LLE

gonetz:
The idea is right, the implementation is wrong. Your code is redundant. Current code already has CFB flag: (gSP.changed & CHANGED_CPU_FB_WRITE) == CHANGED_CPU_FB_WRITE.
Just need to use it properly in all places. I've made necessary changes: 
https://github.com/gonetz/GLideN64/commit/b219997c80011459ffd60f644dfc172d636f01c4
your commit used as reference.

purplemarshmallow:
Ok
I'll test which games still need detectCFB setting

gonetz:
The ones, which use CFB between demos, e.g Rayman 2.

PR #545, opened 2015-05-20
change validity check treshold level to 0,5%

purplemarshmallow:
change validity check treshold level to 0,5%
fixes pause menu in Mario Tennis


purplemarshmallow:
My mistake
pause menu seems to work now without this change
I need to do more testing on this

PR #569, opened 2015-05-31
Ptit seb master

ptitSeb:


ptitSeb:
Useless :( it's the same pull request

ptitSeb:
At last, it seems good this time.

gonetz:
Nope, no good:
-	const double ref_X =  95.047;        //Observer= 2°, Illuminant= D65
+	const double ref_X =  95.047;        //Observer= 2?, Illuminant= D65
redundant change. You removed unicode symbol just for nothing.You had to left that file untouched in the first commit.

gonetz:
Also, in Config.cpp:
+	#ifdef PANDORA
#ifdef should be without any indent before it.
Please fix. I'll check that your changes do not break other platforms.

ptitSeb:
Done.

gonetz:
Thanks! Pushed.

PR #580, opened 2015-06-17
modify VI width calculation

### LINKED ISSUE #529: cfb gfx issues
"Detect CPU writes to framebuffer" needs to be enabled
the GB tower is fully functional but gfx are broken
correct with angrylion's plugin
gliden64_pokemon_stadium_2_016

purplemarshmallow:
fixes #529


purplemarshmallow:
sorry I created this one too fast
It may cause problems with other games and it's not a proper fix

PR #593, opened 2015-06-26
Removed some useless header files + forced multisampling option to 0 when not supported

Gillou68310:
I accidently launched the GLES3 version with multisampling set to 2 which of course didn't end up very well, so I just forced multisampling to 0 when not supported to avoid such a case in the future ;-)


gonetz:
Thanks!
https://github.com/gonetz/GLideN64/commit/6be5ff63964e5b27d36d18d3fe9986c57c5b072a

PR #595, opened 2015-06-29
visual studio cleanup

purplemarshmallow:
use only one .sln file for all projects
add freetype to the repo
add dummy Revision.h to avoid build errors
fix build warning: "TargetName(GLideN64) does not match the Linker's
OutputFile property value"


gonetz:
1. Bad idea to put all changes in one commit. It should be separate commit per logical modification.
2. I don't like the idea to add freetype to the repo. It's better to create additional repo for external dependencies.
3. I don't want to put MSVS project file for GLideNUI to the repo. QT plugin for MSVS is necessary to build GLideNUI. With the plugin you can easily build MSVS project from GLideNUI.pro file. Besides, MSVS project usually must be re-created when .pro file changes. Thus, you need to remember to update MSVS project in repo after .pro modification.
4. "add dummy Revision.h to avoid build errors" - it's better to add pre-build step in MSVS project for Revision.h creation/update.

I agree that GLideN64 and GLideNHQ can be in one solution. GLideN64 must be made dependent on GLideNHQ in that case.
GLideNUI is too specific, and should live separately imo. If you able to install and use QT plugin, you may create GLideNUI project and add it to the solution in few clicks.

purplemarshmallow:
I made a new pull request

>It's better to create additional repo for external dependencies.
are you going to create an additional repo? currently libpng and zlib are included, freetype is missing

gonetz:
I have to clone myself to perform all necessary tasks :(
That is yes, I'm going to, but have no time for.

 libpng and zlib include is inherited. These libs included to GlideHQ too.

purplemarshmallow:
@gonetz I created a new repo
https://github.com/purplemarshmallow/GLideN64-win32-deps
should I correct the paths and make a pull request?

vgturtle127:
:+1:

PR #598, opened 2015-07-01
use dependencies from external repo

purplemarshmallow:
https://github.com/purplemarshmallow/GLideN64-win32-deps


gonetz:
Why win32-deps? What about Linux?
You removed pre-compiled Linux libs and killed Linux build.

Moreover: even if you add Linux libs to your repo and correct cmake project files, it will just add inconvenience to Linux users. Currently Linux user can install all necessary dependencies with one command from terminal. No need to manually clone additional repo.

In general: I don't like that user now must know about the external repo and manually clone it.
I like the approach, used by Mupen64Plus AE project:
https://github.com/mupen64plus-ae/mupen64plus-ae.git
The project uses sources and data from external repositories. It has scripts to fetch from these repositories. The script works like this: clone external repo into temp directory; remove .git; move everything from temp to destination. Check any pull-*.sh script from 'tools' sub-folder.

My suggestion:
- lets all externals be in GLideN64/externals
- you write script, e.g install_externals.sh, which fetch from your external repo to 'externals' subfolder. That is we will have:
externals/freetype/include - freetype includes
external/inc - libpng and zlib includes
externals/libs - all external libs
- project files need to be corrected accordingly.
- user will need to do the following steps
$ git clone https://github.com/gonetz/GLideN64.git
$ cd GLideN64
$ ./install_externals.sh

Note: since Linux does not need freetype in externals, it is better to create separate repo for Linux externals, which will contain only libpng and zlib headers and libs. Thus, 2 scripts will be necessary: install_externals_win32.sh and install_externals_linux64.sh

purplemarshmallow:
I suggest we use an additional repo only for windows and also add Qt. And we keep the png and zlib headers and pre-compiled Linux libs in this repo, I removed them by accident. So just one script will be necessary.

gonetz:
I see no point to add the png and zlib headers to external repo if "we keep the png and zlib headers and pre-compiled [Linux] libs in this repo". Also I'm not sure, how to add QT. My static QT build contains tons of files and weights 700+ MB. And this is only for release build. Static debug build takes circa 4GB. We can put pre-build GLideNUI.lib into external repo. It weights circa 100 MB for release build.

purplemarshmallow:
>I see no point to add the png and zlib headers to external repo if "we keep the png and zlib headers and pre-compiled [Linux] libs in this repo"

that's true
should I add instructions to the wiki on how to download freetype from the official site and on how to build it from sources on windows?
or put it in an external repo?

>We can put pre-build GLideNUI.lib into external repo. It weights circa 100 MB for release build. 

but this has some disadvantages
but debug build should work too. And if there are changes to the UI the external repo needs to be updated.

PR #602, opened 2015-07-01
Fix framebuffer texture rectangle coordinate in GLES2

Gillou68310:
Here's a screenshot of the issue:
capture

And here's a screenshot of the normal behaviour:
capture2


gonetz:
Thanks!
https://github.com/gonetz/GLideN64/commit/40da41d953a21a3cd5c624ea8626a7cc4de775f8

PR #613, opened 2015-07-13
fix copy color to RDRAM in LLE

### LINKED ISSUE #554: [Mupen64Plus] Pokemon Snap - LLE - empty Photos
Description:
In LLE (CXD4) the photos of the pokemon are empty (no pokemon detected).
snap-lle

Also if the settings will be modified (e.g. EnableDetectCFB) the results are the same.

System: Windows 7
Frontend: M64Py 2.2
Graphic Plugin: GLideN64 1.1
Graphic Plugin Setting: Default (also tried with modification of default settings)
RSP Plugin: CXD4 SSSE3 and RSP-HLE
Mupen64Plus Version: Build from 2015-05-09 (c7fa0de)
Game: Pokemon Snap (PAL-G)

purplemarshmallow:
fixes #554
this is also an optimization for games with multiple dlists per frame

### DIFF
--- a/src/RSP.cpp
+++ b/src/RSP.cpp
@@ -205,8 +205,6 @@ void RSP_ProcessDList()
 		}
 	}
 
-	if (config.frameBufferEmulation.copyToRDRAM)
-		FrameBuffer_CopyToRDRAM( gDP.colorImage.address );
 	if (config.frameBufferEmulation.copyDepthToRDRAM)
 		FrameBuffer_CopyDepthBuffer( gDP.colorImage.address );
 
--- a/src/RSP.cpp
+++ b/src/RSP.cpp
@@ -204,10 +204,6 @@ void RSP_ProcessDList()
 			RSP_CheckDLCounter();
 		}
 	}
-
-	if (config.frameBufferEmulation.copyDepthToRDRAM)
-		FrameBuffer_CopyDepthBuffer( gDP.colorImage.address );
-
 	RSP.busy = FALSE;
 	gDP.changed |= CHANGED_COLORBUFFER;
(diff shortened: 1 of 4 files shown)


purplemarshmallow:
indiana jones pause menu now works!
gliden64_indiana_jones_000

purplemarshmallow:
I thought it works for every game but after more testing I noticed that it does not work for Glover2 :(
Every other game seems to work

AmbientMalice:
Lovely work.

AmbientMalice:
On a related topic, do you have any idea why depth buffer is broken in LLE? Because I suspect Indiana Jones' remaining graphics issues are depth buffer related.

purplemarshmallow:
I fixed Glover2
>do you have any idea why depth buffer is broken in LLE?

I fixed depth buffer copy the same way but it does not work as good as in HLE, I don't know why

purplemarshmallow:
If I put it in ProcessRDPList depth buffer is copied mulitple times per frame but effects don't work at all. I think there's a different bug in the plugin that screws up values in depth buffer in LLE. Depth buffer emulation in HLE works as good as before after this change.

AmbientMalice:
You could make the optimisation a per-game setting. Preferably one adjustable from the plugin's cfg file.

purplemarshmallow:
I think #614 needs some investigations before a per-game setting is added. Maybe it's possible to find a solution that works for every game

gonetz:
Thanks!
https://github.com/gonetz/GLideN64/commit/77f342fc630a20b6e3724f4dda5ea4bb02cc4bf4

Issue #560, opened 2015-05-26
Yoshi's story fb validity check fails


purplemarshmallow:
fingerprint sometimes does not work
I need to check if it happens with fill RDRAM too but these errors are sometimes hard to reproduce...
![gliden64_yoshi_story_011](https://cloud.githubusercontent.com/assets/7283660/7809597/d2eea05e-039b-11e5-8b77-bfc37d5daa81.jpg)

purplemarshmallow:
fill RDRAM does not seem to help at all...

purplemarshmallow:
This should be fixed now, I can't reproduce it anymore

PR #619, opened 2015-07-19
remove condition from FrameBufferList::saveBuffer

purplemarshmallow:
With this condition some auxiliary buffers are not filled with a test value, this causes problems with validity checking in Pokemon Stadium 2 #416 and possibly also in other games. I konw some games may need this condition, so this pull request is more meant for discussion, not for merging.

### DIFF
--- a/ini/GLideN64.custom.ini
+++ b/ini/GLideN64.custom.ini
@@ -299,13 +299,11 @@ frameBufferEmulation\detectCFB=1
 Good_Name=Pokemon Stadium 2 (E)(F)(G)(I)(J)(S)(U)
 frameBufferEmulation\copyToRDRAM=0
 frameBufferEmulation\copyDepthToRDRAM=0
-frameBufferEmulation\validityCheckMethod=1
 
 [POKEMON%20STADIUM%20G&S]
 Good_Name=Pokemon Stadium Kin Gin (J)
 frameBufferEmulation\copyToRDRAM=0
 frameBufferEmulation\copyDepthToRDRAM=0
-frameBufferEmulation\validityCheckMethod=1
 
 [RAYMAN%202]
 Good_Name=Rayman 2 - The Great Escape (E)(U)
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -265,6 +265,9 @@ void FrameBuffer::copyRdram()
 			// Thus content of RDRAM on moment of buffer creation will be the same as when buffer becomes obsolete.
 			// Validity check will see that the RDRAM is the same and thus the buffer is valid, which is false.
 			// It can be enough to write data just little more than treshold level, but more safe to write twice as much in case that some values in buffer match our fingerprint.
+			FrameBuffer * pCurrentBuffer = frameBufferList().getCurrent();
+			if (pCurrentBuffer != NULL)
+				pCurrentBuffer->m_cleared = false;
 			const u32 twoPercent = dataSize / 200;
 			u32 start = m_startAddress >> 2;
 			u32 * pData = (u32*)RDRAM;
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -446,8 +446,9 @@ void FrameBufferList::saveBuffer(u32 _address, u16 _format, u16 _size, u16 _widt
 			if (m_pCurrent->m_width == VI.width)
 				gDP.colorImage.height = min(gDP.colorImage.height, VI.height);
 			m_pCurrent->m_endAddress = min(RDRAMSize, m_pCurrent->m_startAddress + (((m_pCurrent->m_width * gDP.colorImage.height) << m_pCurrent->m_size >> 1) - 1));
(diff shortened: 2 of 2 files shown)


purplemarshmallow:
I tracked down commits:
https://github.com/gonetz/GLideN64/commit/30d6792a176ce51de74274e4edef603de452e2c0
https://github.com/gonetz/GLideN64/commit/9935dab14769a4aaa81feecd031ee3b89c8424e5
https://github.com/gonetz/GLideN64/commit/f41bd934ef52051a85b78d5d5231f4624ae65254
https://github.com/gonetz/GLideN64/commit/d0073fe3a3ffbe76d8ee0d6fe5805fced4263db3
https://github.com/gonetz/GLideN64/commit/2454f70568cbd50f7b552738b1c507c99928c9b9

purplemarshmallow:
!config.frameBufferEmulation.copyFromRDRAM
This breaks Dr. Mario. GLideN64 fills only auxiliary buffers now so I think this is no longer needed and it’s save to remove it

!m_pCurrent->m_cleared
This breaks Pokemon Stadium 2 and I think also Yoshi's Story. It’s ancient, added in 2013 in https://github.com/gonetz/GLideN64/commit/971bd9d0ece6f681f9f684ef8005ce239e3cbf14 I think it can be removed.

everything else in this line should be checked as well, framebuffer emulation changed a lot and it there were two bugs in this line. Now it can be merged. It fixes #620 and most errors in Pokemon Stadium 2 #416 and hopefully also #560, I haven't noticed any problems when I tested it.

purplemarshmallow:
this also seems to fix validity checking errors in Mario Tennis pause menu

gonetz:
>> !config.frameBufferEmulation.copyFromRDRAM
>>This breaks Dr. Mario. GLideN64 fills only auxiliary buffers now so I think this is no longer needed and it’s save to remove it

Ok.

>>!m_pCurrent->m_cleared
>>This breaks Pokemon Stadium 2 and I think also Yoshi's Story. It’s ancient, added in 2013 in 971bd9d I think it can be removed.

m_pCurrent->m_cleared means that RDRAM area is filled with m_fillcolor. FrameBuffer::isValid() 
will check values in RDRAM, thus copyRdram() call is not needed. So, the condition is on its right place.
Removal of that condition will cause force fill of auxiliary buffers. May be this is what is needed for Pokemon Stadium 2, but I'd like to get explanation why the game does not work with that condition.

> remove m_pCurrent->m_RdramCopy.empty()
>    it breaks fingerprint method

Please explain me how it breaks fingerprint method? m_RdramCopy.empty() means that copyRdram() not called yet for the current buffer. copyRdram() meant to be called only once, after buffer allocation and before switch to another buffer. Removal of  m_pCurrent->m_RdramCopy.empty() condition will cause additional copyRdram() calls, which looks like unnecessary overhead.

purplemarshmallow:
>m_pCurrent->m_cleared means that RDRAM area is filled with m_fillcolor.

This condition works incorrect. Sometomes auxiliary buffers are not filled, but m_pCurrent->m_cleared is true. If I remove it all buffers are filled

>Please explain me how it breaks fingerprint method?

same problem. Sometimes m_RdramCopy.empty() is false but copyRdram() is not called yet for current buffer. If I remove it copyRdram() is called for all buffers

gonetz:
If it works wrong, need to find why. Otherwise your fix will just obscure that bug.

purplemarshmallow:
>Removal of that condition will cause force fill of auxiliary buffers. May be this is what is needed for Pokemon Stadium 2, but I'd like to get explanation why the game does not work with that condition.

I think I know what's going on. The game uses the same auxiliary buffers multiple times.
1: auxiliary buffer is created, the co-processor renders an image into it
2: auxiliary buffer is not visiblie in current screen, the CPU fills it with data
3: auxiliary buffer is visible again, the co-processor renders a different image into it

With GLideN64's HWFBE at (3) the buffer is not filled with a test value anymore, validity checking removes the buffer. If I force fill of auxiliary buffers the plugin fills it again with a test value and everything works correctly

gonetz:
>> I think I know what's going on. 

Is it Pokemon Stadium 2 scenario? Could you give me a savestate to check it?

purplemarshmallow:
>Is it Pokemon Stadium 2 scenario

Yes but I think in Yoshi's Story it's the same scenario
Here's a savestate: https://drive.google.com/file/d/0B7Y6r4SpC_QQdGdCdEZoRzRTQ0U/view?usp=sharing
And here's a suggested fix, but maybe it affects performance in games with many auxiliary buffers (Mario Tennis)
https://github.com/purplemarshmallow/GLideN64/commit/6fd1bc7acc038f30d0aa24cdada47461e9f4cec7

purplemarshmallow:
here's a new, clean pull request: #631

twarnercool2:
Here a version of project 64 that works with "Indiana Jones and the Infernal Machine"

https://www.dropbox.com/sh/siz5kf4vgpxqid6/AAA2iGXHBZptcwRD-6aYze42a?dl=0

PR #621, opened 2015-07-20
change fillcolor to something more original

purplemarshmallow:
together with #619 this fixes remaining problems with validity checking in Pokemon Stadium 2, #416 is fixed, but causes regressions in Mario Tennis. I dislike these lines, looks like a bug.
gliden64_mariotennis_003

gliden64_mariotennis_002


olivieryuyu:
well the sliding effect (1st screen) was not correct before: gfx are not the correct one

purplemarshmallow:
There's another much worse regression caused by this change
Replays in Mario Tennis can appear incorrectly and then the game freezes
gliden64_mariotennis_005

purplemarshmallow:
not needed anymore

PR #622, opened 2015-07-21
only check for fingerprint

purplemarshmallow:


purplemarshmallow:
I overlooked something, there are some regressions
I hope they this idea works and they can be fixed

AmbientMalice:
I hate to ask this, @purplemarshmallow but could you upload a build with this fix and also https://github.com/gonetz/GLideN64/pull/619 for me to test out? I'm still waiting for Windows build instructions.

purplemarshmallow:
https://drive.google.com/file/d/0B7Y6r4SpC_QQNFFhVzJoUGtLbWc/view?usp=sharing

AmbientMalice:
I was hoping this commit might have helped Vigilante 8, but the menus are still badly broken. Still, fixing some issues by removing bad code is encouraging. I imagine there have been subtle fixes in random games that I haven't seen yet.

purplemarshmallow:
I'm closing this because it does not always work, Zelda can write random data over the fingerprint

purplemarshmallow:
reopened because of #631
I think this can bring a performance boost

theboy181:
Speaking of performance I seems to be getting a lower fps on my ATI 295x vs my Nvidia 790 on my laptop. 

The laptop is i7 2.4 and the desktop is 4.6.

Any ideas?

purplemarshmallow:
maybe because of different drivers

theboy181:
Looks like I was mistaken. I was testing with banjo kazooie and the 60fps hack. Looks like pj64 needs a OC feature to get a full 60 fps. Maybe a CF=.5.  Lol

purplemarshmallow:
I finally found a way to fix the problem in Zelda OOT
here's a build if someone wants to help me with testing
https://drive.google.com/file/d/0B7Y6r4SpC_QQVWJqWnVnUWVmVzg/view?usp=sharing
my results (fingerprint validity check method):
Zelda OOT: perfect
Pokemon Stadium 2: works better, I can't explain why, but not perfect
Mario Tennis: pause menu looks ok
regression: scoreboards are broken. They use a hack, maybe that's the reason
gliden64_pokemon_stadium_2_000
gliden64_mariotennis_000

purplemarshmallow:
I fixed scoreboards in Mario Tennis
now fill RDRAM shouldn't be needed anymore

AmbientMalice:
Updated binary?

purplemarshmallow:
https://drive.google.com/file/d/0B7Y6r4SpC_QQWVpuaHBGaDRvUzQ/view?usp=sharing
with this build "write fingerprint to RDRAM" should always work as good or better than "fill buffer in RDRAM with test value". Next step would be to remove this option completely.

theboy181:
this build broke 1080 for me.

purplemarshmallow:
What is broken in 1080?

theboy181:
I had to change manually change the validity check.. the snow was black.

purplemarshmallow:
Ok thanks for testing
Can you give more details?
I can't reproduce this

theboy181:
Could be because I opened this game after playing another?  I'll test again tomorrow I know that pj64 can do some weird things if you don't close the emulator before opening a new rom.

AmbientMalice:
>Could be because I opened this game after playing another?

Most likely. With PJ64, this tends to cause all sorts of issues. I think GLideN64 is at fault, though.

Anyway, I'm not seeing any regressions.

gonetz:
I was offline last week. I'll check your changes asap.

src/FrameBuffer.cpp
```diff
@@ -264,12 +265,19 @@ void FrameBuffer::copyRdram()
...
 			const u32 twoPercent = dataSize / 200;
 			u32 start = m_startAddress >> 2;
 			u32 * pData = (u32*)RDRAM;
 			for (u32 i = 0; i < twoPercent; ++i)
-				pData[start++] = m_startAddress;
+			{
+				if (i < 4)
+					pData[start++] = fingerprint[i];
+				else
+					pData[start++] = 0;
```
gonetz: Why modify data below fingerprint if you check fingerprint only?
I mean the code after:
//check if our fingerprint is still there

purplemarshmallow: It's needed for Link's model in Zelda OOT in the equipment subscreen. This model is modified by the CPU. I'm pretty sure the CPU does some filtering. The CPU can modify the fingerprint and validity check will fail. If I clear the data below the fingerprint the CPU won't touch my fingerprint.

After adding this it also fixed validity checking problems in Pokemon Stadium 2.

gonetz: Ok, thanks for the explanation. Sounds weird, but fb emulation is full of weird things.

src/FrameBuffer.cpp
```diff
@@ -294,7 +302,19 @@ bool FrameBuffer::isValid() const
 				++wrongPixels;
 		}
 		return wrongPixels < (m_endAddress - m_startAddress) / 400; // treshold level 1% of dwords
-	} else if (!m_RdramCopy.empty()) {
+	}
+	else if (m_fingerprint) {
+			//check if our fingerprint is still there
+			const u32 stride = m_width << m_size >> 1;
+			const u32 height = _cutHeight(m_startAddress, m_height, stride);
+			const u32 dataSize = stride * height;
```
gonetz: All three constants do not used in further calculation.

gonetz:
https://github.com/gonetz/GLideN64/commit/8463c0347f21201b1de84324ccdc3f297c61c123

src/FrameBuffer.cpp
purplemarshmallow: I did some experiments. I think I forgot to remove them.

PR #631, opened 2015-07-28
 always fill auxiliary buffers

purplemarshmallow:


purplemarshmallow:
This is a clean version of #619 
To sum it up, my investigation in Pokemon Stadium 2 showed that the game uses the same auxiliary buffers multiple times. If they are not visible, the cpu can store data there.
 >1: auxiliary buffer is created, the co-processor renders an image into it
 2: auxiliary buffer is not visiblie in current screen, the CPU fills it with data
 3: auxiliary buffer is visible again, the co-processor renders a different image into it

>With GLideN64's HWFBE at (3) the buffer is not filled with a test value anymore, validity checking removes the buffer. If I force fill of auxiliary buffers the plugin fills it again with a test value and everything works correctly

Pokemon Stadium 2 might not be the only game that uses this mechanism, I think it affects Yoshi's Story as well and maybe other games, GLideN64 has to fill auxiliary buffers more than once to handle this case. I've opened #622 which suggests a faster way to fill auxiliary buffers (no need for RDRAM copy). I think this can compensate the performance loss

purplemarshmallow:
I made changes in #622 and Pokemon Stadium 2 looks better. This makes no sense to me, mabe I'm wrong here.

purplemarshmallow:
This seems to be not needed anymore because of #622 
I guess because games act different depending on data in RDRAM and now the first few values remain untouced

PR #636, opened 2015-07-29
Apply Quake II hack to Quake 64.

AmbientMalice:
I have no idea whether I got the syntax right thanks to not being able to build GLideN64. Assuming this works, this should fix underwater rendering in Quake 64.

### DIFF
--- a/src/RSP.cpp
+++ b/src/RSP.cpp
@@ -325,7 +325,9 @@ void RSP_Init()
 		config.generalEmulation.hacks |= hack_blastCorps;
 	else if (strstr(RSP.romname, (const char *)"SPACE INVADERS") != NULL)
 		config.generalEmulation.hacks |= hack_ignoreVIHeightChange;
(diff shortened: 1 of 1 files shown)


olivieryuyu:
i don't think we should go to too much hacks

AmbientMalice:
GLideN64 used to perform the necessary buffer swap check on all games, but it was causing rare performance issues so it was made a game specific hack - https://github.com/gonetz/GLideN64/commit/16e3af616c52db22638cacef60eb7dade73584bc

Predator82Germany:
Can hacks not be written in a special seperated section in the plugin code?

olivieryuyu:
i believe there should be a box in the GUI where we could deactivate ALL hacks. Would be nice

Predator82Germany:
Then a new tab were usefull

Predator82Germany:
Or both :)

gonetz:
https://github.com/gonetz/GLideN64/commit/b8d9d30ecc7f67bec60c43896850ee56a3803296

PR #647, opened 2015-08-04
hack to fix out of buffer write in Donkey Kong 64 (E)

### LINKED ISSUE #355: Donkey Kong 64 (E) crash at start and when loading a new game

purplemarshmallow:
fix #355

### DIFF
--- a/src/Config.h
+++ b/src/Config.h
@@ -126,6 +126,7 @@ struct Config
 #define hack_blastCorps				(1<<8)  //Blast Corps black polygons
 #define hack_ignoreVIHeightChange	(1<<9)  //Do not reset FBO when VI height is changed. Space Invaders need it.
 #define hack_VIUpdateOnCIChange		(1<<10) //Update frame if color buffer changed. Needed for Quake II underwater.
+#define hack_DonkeyKong				(1<<11) //Fix out of buffer write in Donkey Kong 64 (E)
 
 extern Config config;
 
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -929,6 +929,10 @@ void FrameBufferToRDRAM::CopyToRDRAM(u32 _address)
 {
 	if (VI.width == 0 || frameBufferList().getCurrent() == NULL) // H width is zero or no current buffer. Don't copy
 		return;
+
+	if ((config.generalEmulation.hacks & hack_DonkeyKong) !=0 && VI.width < (VI.height *1.18))
+		return;
+
 	FrameBuffer *pBuffer = frameBufferList().findBuffer(_address);
 	if (pBuffer == NULL || pBuffer->m_width < VI.width || pBuffer->m_isOBScreen)
 		return;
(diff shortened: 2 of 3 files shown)


AmbientMalice:
Wouldn't it be ideal to solve GLideN64's framebuffer bugs instead of temp-solving the crashes with hacks? I'm wary that fixes like this are going to backfire.

purplemarshmallow:
I think hacks should only be used if there's no other option. Gonetz came to this conclusion when he investigated this bug:
>It's either bug in game code, or bug in VI emulation on emulator's side. Frame buffer height calculated by the plugin is higher than its real height. Out of buffer write causes crash.

If it's a bug in game code or a bug on emulator's side only a hack can prevent a crash. Frame buffer height is mostly correct in Donkey Kong 64 (E) except for a few cases (after the monkey rap or when loading a game). If frame buffer height is wrong for a short time that's enough to cause a crash. It's easily possible that the missing level of synchronisation from the emulator's side causes these errors. Keep in mind Donkey Kong 64 is known to have timing issues with all emulators.

Everything this hack does is prevent a copy to RDRAM if the values are strange

AmbientMalice:
Vigilante 8 (E) crashes on boot. Likely same issue as DK64 (E). Plus I notice copying framebuffer from RDRAM is broken in a wide range of games. (For example, Mario Tennis crashes. Vigilante 8 crashes. Resident Evil 2 is even more broken.)

LegendOfDragoon:
@AmbientMalice , does the "wobble" occur in Mupen?

AmbientMalice:
@LegendOfDragoon Not exactly. For example, Quake 64 wobbles with PJ64, and "shimmers" with mupen64plus. It's a shifting, phasing effect. Like the colour is being washed out and then bleeding back in.

purplemarshmallow:
I tested Donkey Kong 64 (E) with Jabo's plugin and copy to RDRAM enabled. It does not crash. I wonder if he used a hack or a better formula.

gonetz:
>I wonder if he used a hack or a better formula.
Or hardcoded frame buffer size. IIRC, Jabo's plugin allows user to set frame buffer size manually.

purplemarshmallow:
I'm not sure but I think Jabo's plugin calculates it because it has wrong values as well. But it somehow manages to prevent a crash.
After loading a game there are are two zipper effects. The first one works fine with GLideN64 and Jabo's plugin. The second one is problematic. With GLideN64 my hack prevents a copy to RDRAM, the effect is missing. Without hack there would be a crash. With Jabo's plugin the effect is incorrect, but no crash. Here's a screenshot from Jabo's plugin:
donkey kong 64 e snap0005

purplemarshmallow:
>Jabo's plugin allows user to set frame buffer size manually

Jabo's plugin has "emulated width" and "emulated height" options. I've tested them but they seem to have no impact on copy to RDRAM. Jabo's formulas are very inaccurate, the puzzle effect in Banjo Kazooie (E) is wrong with Jabo's plugin and if you enable copy to RDRAM in Body Harvest (E) the emulator crashes

fzurita:
How about doing this instead at the beginning of the function?

_address = _address & 7F FFFF;

purplemarshmallow:
merge conflict and I'll close for now. I hope there's a better solution

Issue #630, opened 2015-07-28
Core dump in android X86 when enabling LOD


fzurita:
Edit: Improved Readability.

I'm getting a core dump using android x86 in Super mario 64. Here is a save state:

https://drive.google.com/file/d/0B57Ioy26LWegX0k2V2E4Q3JUWmc/view?usp=sharing

Here is a backtrace:
>********** Crash dump: **********
Build fingerprint: 'Android/full_fugu/fugu:5.1.1/LMY47V/cloudproject05011142:user/release-keys'
pid: 3393, tid: 3648, name: CoreThread  >>> org.mupen64plusae.v3.alpha <<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0
#00 pc 00039402  libglslcompiler.so
#01 pc 00018cfb  libglslcompiler.so
#02 pc 0002ae67  libglslcompiler.so
#03 pc 00065c89  libglslcompiler.so
#04 pc 0009debb  libglslcompiler.so (GLSLCompileToUniflex+507)
#05 pc 000547bf  libGLESv2_POWERVR_ROGUE.so
#06 pc 00054ff0  libGLESv2_POWERVR_ROGUE.so (glCompileShader+48)
#07 pc 00049b32  libmupen64plus-video-gliden64-gles31.so: Routine ShaderCombiner::ShaderCombiner(Combiner&, Combiner&, gDPCombine const&) at /mupen64plus-video-gliden64/src/OGL3X/GLSLCombiner_ogl3x.cpp:387
#08 pc 000096a0  libmupen64plus-video-gliden64-gles31.so: Routine CombinerInfo::_compile(unsigned long long) const at /mupen64plus-video-gliden64/src/Combiner.cpp:210
#09 pc 0000ae1d  libmupen64plus-video-gliden64-gles31.so: Routine CombinerInfo::setCombine(unsigned long long) at /mupen64plus-video-gliden64/src/Combiner.cpp:239
#10 pc 0002f3e8  libmupen64plus-video-gliden64-gles31.so: Routine OGLRender::_updateStates(OGLRender::RENDER_STATE) const at /mupen64plus-video-gliden64/src/OpenGL.cpp:550
#11 pc 0002fd7f  libmupen64plus-video-gliden64-gles31.so: Routine OGLRender::_prepareDrawTriangle(bool) at /mupen64plus-video-gliden64/src/OpenGL.cpp:680
[...]

PR #668, opened 2015-08-17
Fix core dump with PowerVR GPUs in android

fzurita:
Fixed core dump with PowerVR GPUs in android. This solution has been
tested with other android devices with no negative impacts.

This is against this issue:

https://github.com/gonetz/GLideN64/issues/630

### DIFF
--- a/src/OGL3X/Shaders_ogl3x.h
+++ b/src/OGL3X/Shaders_ogl3x.h
@@ -150,12 +150,14 @@ MAIN_SHADER_VERSION
 
 static const char* fragment_shader_header_common_variables =
 MAIN_SHADER_VERSION
+#ifndef GLESX
 "#ifdef GL_NV_fragdepth			\n"
 "    #extension GL_NV_fragdepth : enable \n"
 "#endif										\n"
 "#ifdef GL_OES_standard_derivatives			\n"
 "    #extension GL_OES_standard_derivatives : enable \n"
 "#endif										\n"
+#endif
 "uniform sampler2D uTex0;		\n"
 "uniform sampler2D uTex1;		\n"
 #ifdef GL_MULTISAMPLING_SUPPORT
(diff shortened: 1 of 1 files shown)


fzurita:
I tested this with other android devices in my possession and I haven't seen any negative impacts to doing this.

src/OGL3X/Shaders_ogl3x.h
```diff
@@ -150,12 +150,14 @@ MAIN_SHADER_VERSION
 
 static const char* fragment_shader_header_common_variables =
 MAIN_SHADER_VERSION
+#ifndef GLESX
```
gonetz: Bad idea. These are GL ES extensions. You will kill mip-mapping emulation on mobile devices.

fzurita: Hello Gonetz, it seems to be working without that defined. Are there any specific tests I can run to verify that mipmap is working correctly? I did test LOD by using the peach painting in Super Mario 64, that went from crashing to working correctly afterwards.

fzurita:
I may have found the actual solution to this problem. See below:

http://forum.apportable.com/t/new-kindle-fire-hd-6-7-4th-gen-failed/1383/24

The solution is to enable extensions before anything else, even before the version it seems.

I'll try this out tonight to see if it still works.

gonetz:
"the extension directives must occur before any non-preprocessor tokens."
I had shader compilation issues when #extension was not on top of my shader. #version is preprocessor directive, so when I put #extension under it shader compiled successfully. May be PowerVR shader compiler expects #extension to be the first, worth to check.

fzurita:
No luck, this causes shader compilation errors, like you said.

fzurita:
Hmm.... This seems to work ok:

```
"#extension all : warn \n"
```

I'm not sure if it's actually supposed to be enabling any extensions though.

gonetz:
I removed 
#extension GL_OES_standard_derivatives : enable \n"
and got no shader compilation errors on my GLES 3 compatible devices.
From one side this is strange, because I did not knew about this extension until I got shader compilation error on my Note 3.
From the other side, I upgraded Android version on Note 3 to 5.0 and now it works without that extension.

gonetz:
Your commit is wrong because it disables GL ES extensions for GL ES code.
I just removed the extensions:
https://github.com/gonetz/GLideN64/commit/60c8003dc9fb873937e1b9d5cbedfdf2e3b8bb8e

Need to be tested on all GL ES 3 compatible devices for possible regressions.

fzurita:
Ok, I wasn't sure if desktop GPUs would complain about the extensions missing since I'm not setup to build for windows.

fzurita:
This is interesting:

http://lists.freedesktop.org/archives/mesa-commit/2013-November/046218.html

Apparently these extensions are enabled by default in GLSL ES 3.0.

gonetz:
Interesting endeed.

PR #672, opened 2015-08-17
enable validity checking if copyFromRDRAM option is on

### LINKED ISSUE #620: Dr. Mario framebuffer validity check fails
it only fails if copy from rdram is enabled
fixed by #619 
if (!config.frameBufferEmulation.copyFromRDRAM && ...
why? This game needs copy from rdram and validity checking

purplemarshmallow:
fix #620


gonetz:
Pulled, thanks!

https://github.com/gonetz/GLideN64/commit/97468d8e2aa52e5337369fb2871b943970b576d4

PR #670, opened 2015-08-17
[ini] remove validityCheckMethod

purplemarshmallow:
#631 is merged, now there's one universal validity check method

PR #674, opened 2015-08-19
remove fill RDRAM

purplemarshmallow:
This just removes the functionality
It still needs to be removed from the GUI

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -100,7 +100,7 @@ FrameBuffer::FrameBuffer() : m_validityChecked(0), m_cleared(false), m_fingerpri
 FrameBuffer::FrameBuffer(FrameBuffer && _other) :
 	m_startAddress(_other.m_startAddress), m_endAddress(_other.m_endAddress),
 	m_size(_other.m_size), m_width(_other.m_width), m_height(_other.m_height), m_fillcolor(_other.m_fillcolor),
-	m_scaleX(_other.m_scaleX), m_scaleY(_other.m_scaleY), m_cleared(_other.m_cleared), m_changed(_other.m_changed), m_cfb(_other.m_cfb), m_isDepthBuffer(_other.m_isDepthBuffer),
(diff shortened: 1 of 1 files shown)


gonetz:
Sorry, I already lost in the modifications, related to validity checking. Lets clarify the situation:

Current master code contains two Validity Check Methods for auxiliary frame buffers: 

- Fill. The frame buffer is filled with some color. Validity check test passes if all values in the buffer are equal to the fill color.

- Fingerprint. 
    Initial implementation: fill beginning of the buffer (circa 2% of buffer's size) with some fingerprint.   Validity check test passes if the buffer contains the fingerprint.
    Your modification (currently used): write fixed size fingerprint (4 integers) to the beginning of the buffer and fill the rest of the buffer with zero. Validity check test passes if the buffer contains the fingerprint. Test does not check content of the buffer below the fingerprint.

You suggest to remove Fill method completely because current version of 
Fingerprint method works as well or better in all known situations. Is it correct?

purplemarshmallow:
>You suggest to remove Fill method completely because current version of 
 Fingerprint method works as well or better in all known situations. Is it correct?

Yes

gonetz:
Ok, thanks.

gonetz:
I put your changes from #670 and #674 to test branch remove_fill_check_method:
https://github.com/gonetz/GLideN64/commits/remove_fill_check_method
I want it to be tested by users before merge to master.

Test binary:
https://drive.google.com/file/d/0B0YqMPjGo3B2Um9YOFd3UGV0SzQ/view?usp=sharing

purplemarshmallow:
I am not able to find any regressions. @AmbientMalice also tested this and has not found any regressions. It would be nice to have a tutorial on how to compile GLideNUI. I'm still using the build uploaded by you compile the plugin. After this commit I need to compile GLideNUI again

purplemarshmallow:
>Up config version.

I think you forgot to change the config version in the ini.
[General]
version=3 shouldn't it be 4?

gonetz:
Yes, ini file need to be updated too: remove validityCheckMethod option and up version number.

I also see no regressions, but I want to get users a chance to find some :)

gonetz:
No new bug reports. Added to master:
https://github.com/gonetz/GLideN64/commit/59c771cb7be5db9cc877c075da728e9ef6995444

PR #689, opened 2015-08-29
Splitting up asserts

matthewharvey:
If these asserts are to be useful for debugging, they should really
be split up so that we can tell which one is failing.

I found this useful while investigating a texture cache corruption issue, which I'm going to do some more investigating on before creating an issue


gonetz:
Good point, thanks!

https://github.com/gonetz/GLideN64/commit/fa7e1c8137cace0236ee5da2f59e42f51a71003b

PR #694, opened 2015-08-31
Cleanup texture code

matthewharvey:
These are probably easier to review as separate commits.

I'm not sure how useful people find these cleanups, but I couldn't keep looking at this code without at least trying. Tell me what you think!

### DIFF
--- a/src/Textures.cpp
+++ b/src/Textures.cpp
@@ -487,9 +487,11 @@ void TextureCache::init()
 	glBindTexture(GL_TEXTURE_2D_MULTISAMPLE, m_pMSDummy->glName);
 
 #if defined(GLES3_1)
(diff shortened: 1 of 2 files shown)


gonetz:
@matthewharvey : I found your modifications pretty useful, thanks! I'm trying to do code cleanup when I see ugly parts of it, but I not always have time for that. Plus I seldom read code, which works :)
The only notice: "Adding assert on malloc" commit is locally right, but in general it is patch for a totally wrong approach. That C style memory handling is mostly legacy and need to be replaced by safe C++ constructions. STL classes and/or exceptions handling. My hands are itching to rewrite all such places, but I don't have time even for bug fixes. Don't you want to try to clean that too?

gonetz:
Added to master:
https://github.com/gonetz/GLideN64/commit/3e8ab7825893ab5153243c4e881fdf64a66157a0

PR #701, opened 2015-09-03
always reinit buffers during height correction

purplemarshmallow:
In the library in Pokemon Stadium 2 one image is always missing. Here's the fix
gliden64_pokemon_stadium_2_000


gonetz:
Good, thanks. The only request: please be more detailed in commit description for future commits.
Please describe not only what commit does, but also why and what it is fixes and in which games, with bug number, if fixes particular bug.

gonetz:
https://github.com/gonetz/GLideN64/commit/2847606877082a9b21262a066cd8d2d85b4e8564

Issue #415, opened 2015-04-16
pokemon stadium missing images


purplemarshmallow:
In the stadium some images are missing
![1](https://cloud.githubusercontent.com/assets/7283660/7172690/6b54228a-e3ec-11e4-9719-dd0fb3c1da56.png)
After a battle the picture of the trainer is missing
![2](https://cloud.githubusercontent.com/assets/7283660/7172730/e8eedbd6-e3ec-11e4-9d75-a8e549b88216.png)

gonetz:
Is it still actual?

purplemarshmallow:
Yes

purplemarshmallow:
fixed

purplemarshmallow:
My fix caused regressions. Maybe some buffers with 0 height must be created and others should not be created?

gonetz:
Fixed.

purplemarshmallow:
It's not fixed. Some portraits can be missing. correct with https://github.com/purplemarshmallow/GLideN64/commit/9c61274cd3ca395267890fa24aac6eb5ebd9d819
I tried to explain it in #703:
>But another problem is this formula: height = min(height, VI.height). First buffers are created correctly but later this formula sets the height to 0 which is incorrect.

Savestate: https://drive.google.com/file/d/0B7Y6r4SpC_QQRkRvUWk3NkF6aTQ/view?usp=sharing
![gliden64_pokemon_stadium_001](https://cloud.githubusercontent.com/assets/7283660/9981526/60f1669a-5fc0-11e5-9680-b25bbdcbc857.jpg)

gonetz:
Thanks, I'll check it.

gonetz:
Hmm, if I run the game not from the save state and select "Battle Now!" all images are shown.

purplemarshmallow:
That's how I made the savestate:
select "Battle Now!" 
give up
"Battle Now!" again

gonetz:
I have all portraits with this scenario too, if savestate is not used.
The problem exists indeed when game loaded from the savestate. I'll check your fix for the problem.

purplemarshmallow:
For me portraits are missing even without savestates. What about this? It would fix #718 as well
```
void FrameBufferList::saveBuffer(u32 _address, u16 _format, u16 _size, u16 _width, u16 _height, bool _cfb)
{
	if (VI.width == 0 || (_width != VI.width && _height == 0))
		return;

	OGLVideo & ogl = video();
```

[...]

PR #702, opened 2015-09-03
Revert "Don't discard frame buffers creation when _height == 0"

purplemarshmallow:
I think I just obscured the bug rather than fixing it. _height should not be 0
This opens #415 again


gonetz:
Ok:
https://github.com/gonetz/GLideN64/commit/1fcb64ca796ba77126c588999b8ba0587c407b90

PR #703, opened 2015-09-04
 fix for cases where Vi height is set to 0 

purplemarshmallow:


purplemarshmallow:
I can explain this further: The game sets VI height to 0 for a short time only. We have to store this buffer with 0 height for correct emulation or it will cause glitches. But another problem is this formula: height = min(height, VI.height). First buffers are created correctly but later this formula sets the height to 0 which is incorrect. These buffers will display wrong. We have to prevent that but this formula is needed. I believe this is the correct solution for #415 

I'm not sure about attaching the depth buffer. Is this needed for buffers with 0 height? I don't think so because they were just ignored before this change. Attaching a depth buffer to a buffer with 0 height causes a failed assertion.

gonetz:
Pokemon Stadium (U) (V1.0) crashes right on start with this fix.

purplemarshmallow:
I can't reproduce this, I also tested Pokemon Stadium (U) (V1.0)

gonetz:
It already does not matter. Unfortunately, your fix just obscures real problems in fb emulation code. 
However, it inspired me to spend some hours today to investigate #415. Your guess that reject to save buffer with zero height is the key to that problem helped me to find and fix several not obvious issues. Please check my latest commits. I close this pull request as incorrect, but it was actually very useful. Thanks!

PR #704, opened 2015-09-05
 Optimize Perfect Dark framebuffer effects 

purplemarshmallow:


oddMLan:
Do you have a build?

purplemarshmallow:
https://drive.google.com/file/d/0B7Y6r4SpC_QQQ1ZRVWRqWHgyRm8/view?usp=sharing

AmbientMalice:
Seems to work rather well, especially in high resolution mode. Wonder whether any other games are affected/improved?

purplemarshmallow:
It seems to affect only Perfect Dark during camspy or IR vision effects. I don't know any other game where this makes a noticeable difference. But maybe there are other games with the same problem.

purplemarshmallow:
But it breaks portraits in Pokemon Stadium 2 :/

AmbientMalice:
>But it breaks portraits in Pokemon Stadium 2 :/

Is there *anything* that doesn't break Pokemon Stadium 2?

purplemarshmallow:
I have a better idea. I should check once per frame

Issue #695, opened 2015-08-31
glTexImage2DMultisample being called with samples set to 0


matthewharvey:
This is an issue of great confusion, but after looking through the spec apparently 0 is not a valid value for samples, even though the man pages say it is. This starts failing on newer versions of the mesa driver, who have recently updated to be inline with the spec.
I'm just creating a bug so that I don't forget about this. If there is a way to assign bugs in Github, please assign it to me and I'll deal with it later.

matthewharvey:
Oh right. This is called in this way at least from Textures.cpp. I haven't seen any others.

gonetz:
Fixed with commit 18172a6

PR #706, opened 2015-09-05
Fixing multisampling

matthewharvey:
We were calling glTexStorage2DMultisample with a 'samples' parameter of 0.
Reading the GL 4.4 spec, this is specifically forbidden, but the
documentation pages from Chronos were incorrectly stating that it was
allowed, so many implementations probably accept it. Mesa, the graphics
driver for many Linux cards has recently been corrected so that it emits
a warning and fails when glTexStorage2DMultisample is called with
samples=0. The current release works, but the next one will fail when
multisampling is turned off in this plugin (with multisampling turned
on, it will work the same as before).

The documentation pages from Kronos have since been updated to reflect
the GL spec (I put in a bug to them). You can see the proper
changes at this address
(https://www.opengl.org/wiki/GLAPI/glTexStorage2DMultisample).

This fixes issue #695

### DIFF
--- a/src/Textures.cpp
+++ b/src/Textures.cpp
@@ -481,24 +481,26 @@ void TextureCache::init()
 	current[0] = current[1] = NULL;
 
 #ifdef GL_MULTISAMPLING_SUPPORT
-	m_pMSDummy = addFrameBufferTexture(); // we don't want to remove dummy texture
-	_initDummyTexture(m_pMSDummy);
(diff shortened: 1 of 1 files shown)


gonetz:
Merged, thanks!
https://github.com/gonetz/GLideN64/commit/18172a6c0b6f4f3753fc877503fbe5289e4410b0

PR #711, opened 2015-09-09
don't update NoiseTexture if noise is disabled 

purplemarshmallow:
With this change I get 210 VI/s in Star Wars Episode I - Racer (track select screen). With noise emulation enabled I get 95 VI/s. With noise emulation + copy color + copy depth I get 45 VI/s.


gonetz:
Added. Thanks!
https://github.com/gonetz/GLideN64/commit/e7024978acd4d734024cb4a40173b587ebfddcb0

PR #735, opened 2015-09-29
disable depth buffer emulation by default

purplemarshmallow:
for better performance and compatibility depth buffer emulation should be disabled by default

### DIFF
--- a/ini/GLideN64.custom.ini
+++ b/ini/GLideN64.custom.ini
@@ -133,13 +133,8 @@ frameBufferEmulation\N64DepthCompare=1
 Good_Name=Extreme-G (U)
 frameBufferEmulation\N64DepthCompare=1
 
-[EXTREME%20G%202]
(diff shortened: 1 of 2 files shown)


mudlord:
Would be nice if OGL had decent synchronization primitives for imagestore, or does it ?

AmbientMalice:
I know *why* you're doing this, but it seems LESS accurate from a "You really don't know when a game is going to use the depth buffer" perspective.

purplemarshmallow:
>Would be nice if OGL had decent synchronization primitives for imagestore, or does it ?

ARB_fragment_shader_interlock? Is that what you're looking for?

>I know why you're doing this, but it seems LESS accurate from a "You really don't know when a game is going to use the depth buffer" perspective.

I know I likely missed a game and now lens flares are visible through walls or something. But currently copyDepthToRDRAM breaks so many games I think there's no other option. It does not look like this is going to get fixed anytime soon

Issue #24, opened 2014-10-09
Pilots Wing/Twine wrong shadows


olivieryuyu:
Pilots Wing: the shadows are wrongly displayed in game (black polygons from
the aircraft to the ground) 

Only the shadow on the ground should be displayed. 

Twine: the shadow at the entrance of the building at the first level should
not be displayed (the issue exists in other levels). You will see a black texture where it should be displayed. I think there is a special trick used by the devs of the game in this respect.

Angrylion:

No, Pilotwings uses z mode 2 (write enables if (infront || max)). I don't
know what's causing wrong shadows there. I've never had this bug in my
plugin. It may be caused by the wrong emulation of "inverse alpha" BL
input. Here "inverse alpha" is used in the 2nd cycle of the BL, so it must
be an inverse of A mux from the 2nd cycle, which in this case is an inverse
of the combined alpha. But it definitely has nothing to do with coverage. »
It is the same bug with the shadows in TWINE

olivieryuyu:
TWINE fixed with enable depth compare. Only issue is the slowdow when using it (when shadows are used)

olivieryuyu:
new ticket open for pilotwings. twine closed!

PR #736, opened 2015-09-30
Add depth to RDRAM for games that need it.

AmbientMalice:
Knowing which games want depth buffer to RDRAM is really hard. For now, I think it should be enabled for the notorious suspects -- Rareware racing and platforming games.

### DIFF
--- a/ini/GLideN64.custom.ini
+++ b/ini/GLideN64.custom.ini
@@ -32,6 +32,17 @@ Good_Name=Bakushou Jinsei 64 - Mezase! Resort Ou (J)
 [BANJO-KAZOOIE]
 Good_Name=Banjo-Kazooie (E)(U) / Banjo To Kazooie No Daibouken (J)
 frameBufferEmulation\copyToRDRAM=1
+frameBufferEmulation\copyDepthToRDRAM=1
+
+[BANJO%20KAZOOIE%202]
+Good_Name=Banjo to Kazooie no Daibouken 2 (J)
+frameBufferEmulation\copyToRDRAM=1
+frameBufferEmulation\copyDepthToRDRAM=1
+
+[BANJO%20TOOIE]
+Good_Name=Banjo-Tooie (E)(U)
+frameBufferEmulation\copyToRDRAM=1
+frameBufferEmulation\copyDepthToRDRAM=1
 
 [BEETLE%20ADVENTURE%20RAC]
 Good_Name=Beetle Adventure Racing (E)(U)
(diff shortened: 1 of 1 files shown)


purplemarshmallow:
I think if you add a setting to the ini it should be for a reason not because you guess something. The number of games where copy depth to RDRAM breaks something is about as high as the number of games where it fixes something. Also it's performance intensive. It will be hard to know why you added this setting. I think there should be no guessing here

AmbientMalice:
Fair enough. I'll close for now and do some testing to ensure the games in question actually make use of the depth buffer. I am wondering, however, why depth buffer emulation is breaking games in the first place.

AmbientMalice:
So I looked over

purplemarshmallow:
>It's remarkable how many N64 games don't actually seem to need the depth buffer for their coronas and such.

It's only problematic if a game uses the cpu to read the depth buffer. Then you need the copyDepthToRDRAM option. If a game uses N64 depth compare modes you need the N64 depth compare option. When I tested I noticed copyDepthToRDRAM conflicts with N64 depth compare and it does not work proberly. Are you sure these games need both options?

AmbientMalice:
>When I tested I noticed copyDepthToRDRAM conflicts with N64 depth compare and it does not work proberly. Are you sure these games need both options?

You're quite right. I reverted that.

Any idea why The World is Not Enough uses N64 depth compare?

purplemarshmallow:
>Any idea why The World is Not Enough uses N64 depth compare?

It was added to fix the shadows see #24 

Can you remove Diddy Kong Racing? there are no settings for this game. Can you add Mickey's Speedway again? transition effects only seem to work with copyDepthToRDRAM

AmbientMalice:
Sorry, removing Mickey's depth buffer was a stupid mistake on my part. I'll correct when I'm back on my main PC in a few hours.

I'll remove the empty Diddy Kong Racing entry, too.

AmbientMalice:
Also, I'll skim through some more games to see whether they need depth buffer.

AmbientMalice:
@purplemarshmallow I'm not sure TWINE still needs N64 depth. I took this screenshot with it disabled.
gliden64_twine_000

purplemarshmallow:
with n64 depth compare
gliden64_twine_001
without
gliden64_twine_000

>removed blank entries. 

for some Japanese games there are blank entries with a checksum in case someone wants to add settings because you can't use the Internal name

AmbientMalice:
>for some Japanese games there are blank entries with a checksum in case someone wants to add settings because you can't use the Internal name

I'll revert that change.

gliden64_twine_002


Unfortunately, TWINE seems to perhaps have some sort of GPU-specific problem. This is how it looks on my Nvidia GTX 750Ti with N64 depth disabled.

purplemarshmallow:
>Unfortunately, TWINE seems to perhaps have some sort of GPU-specific problem. This is how it looks on my Nvidia GTX 750Ti with N64 depth disabled.

Yes shadows are visible without N64 depth compare. But there should be no shodows at this point the game uses N64 depth compare to hide the shadows when needed. No GPU specific problem.

AmbientMalice:
Ah, I misread your original post, and though the shadows were supposed to render. Gotcha.

gonetz:
Thanks! I have merged all commits to one and pushed it to master:
https://github.com/gonetz/GLideN64/commit/2b942b8d401f0ba614a5338aacc51aa599f2941c

PR #738, opened 2015-10-01
Fix "Shadres" typo.

AmbientMalice:
Since this will break the current GLideNUI, it shouldn't be merged until an updated GLideNUI file can be provided.


purplemarshmallow:
If you compile with the old version of GLideNUI it should still work. The only disadvantage is you'll have the shadres typo

gonetz:
Thanks! I put your changes in one commit:
https://github.com/gonetz/GLideN64/commit/eac8efc3515e5118fb7e27188ffa79056fafcac4

Issue #714, opened 2015-09-13
Flickering textures in Adreno android devices


fzurita:
I have two devices with adreno GPUs, the Nexus 7 2013 and the Xperia Z Ultra. Both of them have the same issues with textures behaving strangely.

For example:
In Banjo Kazooie: textures flicker.
In Legend of Zelda - Ocarina of time: textures turn strange colors on and off flickering, specially in Link's costume.
Super smash brothers: textures on characters flicker
In Super Mario 64: everything looks normal.

Here is a screenshot with super smash brother. Also, this may be related, notice that instead of having a black border when FBE is enabled, we get strange repeat of what is currently being displayed.

![screenshot_2015-09-13-01-09-56](https://cloud.githubusercontent.com/assets/11581687/9835333/65f4525e-59b4-11e5-9c3f-e598f9bc4ea2.png)

gonetz:
I may confirm the problem. Probably some texture format works incorrect on this device.

fzurita:
Another data point, the issue doesn't happen with the GL ES 2.0 version of the plugin, only the GL ES 3.0 one.

I tried using the GL ES 2.0 texture formats in OpenGL.cpp and Textures.cpp, but it didn't seem to make a difference.

fzurita:
After making pull request #739 the border now looks like static noise in Adreno devices.

PR #739, opened 2015-10-02
Fixes flickering textures in Adreno devices.

fzurita:
Through sheer luck, I think I found the issue. After applying this fix, flickering textures in Adreno devices are gone. This fixes this issue: https://github.com/gonetz/GLideN64/issues/714

You don't have to take this pull request as is since I'm dumb when it comes to OpenGL. 

Please see my comments.

### DIFF
--- a/src/OGL3X/UniformBlock.cpp
+++ b/src/OGL3X/UniformBlock.cpp
@@ -148,7 +148,17 @@ void UniformBlock::setColorData(ColorUniforms _index, u32 _dataSize, const void 
 		m_currentBuffer = m_colorsBlock.m_buffer;
 		glBindBuffer(GL_UNIFORM_BUFFER, m_colorsBlock.m_buffer);
 	}
-	glBufferSubData(GL_UNIFORM_BUFFER, m_colorsBlock.m_offsets[_index], _dataSize, _data);
+
+	OGLVideo & ogl = video();
+	OGLRender & render = ogl.getRender();
+	if (render.getRenderer() == OGLRender::glrAdreno)
+	{
+	        glBufferData(GL_UNIFORM_BUFFER, m_colorsBlockData.size(), m_colorsBlockData.data(), GL_STATIC_COPY);
+	}
+	else
+	{
+	        glBufferSubData(GL_UNIFORM_BUFFER, m_colorsBlock.m_offsets[_index], _dataSize, _data);
+	}
 }
 
 void UniformBlock::updateTextureParameters()
@@ -219,7 +229,18 @@ void UniformBlock::updateTextureParameters()
 		m_currentBuffer = m_textureBlock.m_buffer;
 		glBindBuffer(GL_UNIFORM_BUFFER, m_textureBlock.m_buffer);
 	}
-	glBufferSubData(GL_UNIFORM_BUFFER, m_textureBlock.m_offsets[tuTexScale], m_textureBlockData.size(), pData);
+
+
+        OGLVideo & ogl = video();
+        OGLRender & render = ogl.getRender();
+        if (render.getRenderer() == OGLRender::glrAdreno)
+        {
+                glBufferData(GL_UNIFORM_BUFFER, m_textureBlockData.size(), m_textureBlockData.data(), GL_STATIC_COPY);
+        }
+        else
+        {
+                glBufferSubData(GL_UNIFORM_BUFFER, m_textureBlock.m_offsets[tuTexScale], m_textureBlockData.size(), pData);
+        }
 }
 
 void UniformBlock::updateLightParameters()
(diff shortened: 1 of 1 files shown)


src/OGL3X/UniformBlock.cpp
```diff
@@ -148,7 +148,17 @@ void UniformBlock::setColorData(ColorUniforms _index, u32 _dataSize, const void
 		m_currentBuffer = m_colorsBlock.m_buffer;
 		glBindBuffer(GL_UNIFORM_BUFFER, m_colorsBlock.m_buffer);
 	}
-	glBufferSubData(GL_UNIFORM_BUFFER, m_colorsBlock.m_offsets[_index], _dataSize, _data);
+
+	OGLVideo & ogl = video();
+	OGLRender & render = ogl.getRender();
+	if (render.getRenderer() == OGLRender::glrAdreno)
+	{
+	        glBufferData(GL_UNIFORM_BUFFER, m_colorsBlockData.size(), m_colorsBlockData.data(), GL_STATIC_COPY);
```
fzurita: This alone fixes the flickering textures. glBufferSubData seems to be buggy in Adreno devices. I notice no performance loss after making this change.

src/OGL3X/UniformBlock.cpp
```diff
@@ -219,7 +229,18 @@ void UniformBlock::updateTextureParameters()
...
 		glBindBuffer(GL_UNIFORM_BUFFER, m_textureBlock.m_buffer);
 	}
-	glBufferSubData(GL_UNIFORM_BUFFER, m_textureBlock.m_offsets[tuTexScale], m_textureBlockData.size(), pData);
+
+
+        OGLVideo & ogl = video();
+        OGLRender & render = ogl.getRender();
+        if (render.getRenderer() == OGLRender::glrAdreno)
+        {
+                glBufferData(GL_UNIFORM_BUFFER, m_textureBlockData.size(), m_textureBlockData.data(), GL_STATIC_COPY);
```
fzurita: This I'm not too sure about.

src/OGL3X/UniformBlock.cpp
```diff
@@ -219,7 +229,18 @@ void UniformBlock::updateTextureParameters()
...
+        }
+        else
+        {
+                glBufferSubData(GL_UNIFORM_BUFFER, m_textureBlock.m_offsets[tuTexScale], m_textureBlockData.size(), pData);
```
fzurita: This takes an offset, but we still copy all the data in m_textureBlockData?

src/OGL3X/UniformBlock.cpp
```diff
@@ -237,7 +258,17 @@ void UniformBlock::updateLightParameters()
 		m_currentBuffer = m_lightBlock.m_buffer;
 		glBindBuffer(GL_UNIFORM_BUFFER, m_lightBlock.m_buffer);
 	}
-	glBufferSubData(GL_UNIFORM_BUFFER, m_lightBlock.m_offsets[luLightDirection], m_lightBlockData.size(), pData);
+
+        OGLVideo & ogl = video();
+        OGLRender & render = ogl.getRender();
+        if (render.getRenderer() == OGLRender::glrAdreno)
+        {
+                glBufferData(GL_UNIFORM_BUFFER, m_lightBlockData.size(), m_lightBlockData.data(), GL_STATIC_COPY);
```
fzurita: I'm not too sure about this one either.

src/OGL3X/UniformBlock.cpp
```diff
@@ -237,7 +258,17 @@ void UniformBlock::updateLightParameters()
...
+        }
+        else
+        {
+                glBufferSubData(GL_UNIFORM_BUFFER, m_lightBlock.m_offsets[luLightDirection], m_lightBlockData.size(), pData);
```
fzurita: Same as above, this takes an offset, but we still copy all the data in m_lightBlockData? Shouldn't it be copying a subset of it?

gonetz:
@fzurita your fix really cures the plague on adreno devices. Great discovery! I wonder, how you found it?
Regarding your questions: 
1. glBufferData vs glBufferSubData, cite official docs: "When replacing the entire data store, consider using glBufferSubData rather than completely recreating the data store with glBufferData. This avoids the cost of reallocating the data store."
2. TextureBlock - yes, the whole block needs to be updated. Offset points to the beginning of the block.
3. LightBlock - this could be optimized, because games seldom use all 8 lights. I need to know offset to uLightColor array for that. In that case I could use 2 glBufferSubData calls to update only changed part of data. From the other side, 2 GL calls can be more expensive than one call, which updates whole block.

fzurita:
Ok, the pull request was made just to show you what I did. You could take what actually is correct, make your own commit, and close my pull request.

Here is how I ran into the solution, I kind of followed my intuition:
* I did a google search for Adreno driver bugs. I run into this article: https://dolphin-emu.org/blog/2013/09/26/dolphin-emulator-and-opengl-drivers-hall-fameshame/

* This listed 3 GL calls that had bugs in Adreno drivers. This included glBufferSubData. Interestingly, glBufferSubData is supposed to work mostly correctly except that it introduced memory leaks and performance issues.

* I did a search for glBufferSubData in your code and first attempted to remove the call. This lead to most textures being black, so I thought that it could possibly be related.

* I did a google search for glBufferSubData adreno bugs and found that someone replaced glBufferSubData with glBufferData to fix a memory leak. I didn't think this would fix our problems because we had flickering textures, not a memory leak. I tried it anyways.

* After I replaced the call, flickering textures were gone. I was happy.

gonetz:
Probably glBufferSubData caused a memory leaks in drivers version of 2013. I use Android 5.0 with newer drivers, where glBufferSubData seems to be seriously broken.

fzurita:
It sounds like they only made it worse.

gonetz:
Flame in Zelda MM still works incorrect. Torches and logo look wrong.

fzurita:
It could be yet another driver bug which we haven't discovered yet. Or My implementation is not 100% correct, which I wouldn't be surprised about.

gonetz:
I made a minor changes in you commit and added it to master:
https://github.com/gonetz/GLideN64/commit/19668e2c200c31ffcbd441a87d8c989a6efe851a

gonetz:
Thanks to your discoveries (this one and for glBlitFramebuffer) I finally can use GLES3 version on my Note 3. Cool!

fzurita:
Yep, no problem. It's always fun and fulfilling to go bug hunting. Specially in fun stuff like this.

Issue #744, opened 2015-10-04
Texture Cache breaks on Beetle Adventure Racing


matthewharvey:
After playing Beetle Adventure racing for quite a while, I started to see some textures replaced by black, or some other background colour (one of them was white).
After a little investigation, I found that lowering the size of the texture cache caused the issue to happen much more quickly. What was happening was that the game was creating a whole bunch of textures that were used temporarily and then never using them again. This was filling up the cache. Then, when the cache was full, the _checkCacheSize function was removing textures from one "end" of the map of textures. I can only assume that the "end" to which new textures get inserted is platform dependent. On my platform the removal of textures from the cache was at the same end that they were being inserted, so when the texture cache became full, the textures that were currently on the screen were being removed from the cache, and when the emulator went to render those textures, they had already been removed from the cache (I can only guess on that last part, but it fits the behaviour).

Of course, this can be trivially fixed by changing the end of the cache that textures are removed from, but this will still have occasional corruption. Also, I believe that causes platform dependency issues. The proper solution is to convert the cache to a "least recently used" cache. That will ensure that as long as the cache is large enough, the texture will be properly loaded at render time. I have patches for both of these solutions. I will create a pull request for the lru cache solution in a moment.

purplemarshmallow:
If you're using Mupen64plus you should test with CPU interpreter I know there's a bug in the recompiler that can break this game

matthewharvey:
Thanks for the suggestion. I'm going to publish the pull request, and then give this a try later (I have to go to dinner. My girlfriend said she was hungry quite a while ago :D)

matthewharvey:
Exactly the same behaviour with the CPU interpreter and the cached interpreter.

purplemarshmallow:
I don't have this problem with Project64 how long do I need to play?

matthewharvey:
I moved my texture cache down to 30MB and was able to get the issue without even getting to into a race.

matthewharvey:
Also, I didn't provide platform details.

mupen64plus
Debian Sarge (testing)
Intel HD4000 graphics card using the Mesa OpenGL implementation

matthewharvey:
Fixed with merge request #745

Issue #534, opened 2015-05-15
Hangs and crashes over time, possibly due to cache


jimbo1qaz:
See https://github.com/gonetz/GLideN64/issues/509

Since you closed the issue, I have done more research.

It doesn't crash when I do nothing after playing the Song of Time, and just hang around the Clock Tower.
It does crash in Majora boss room.
These crashes are random and cannot be reliably triggered.

Lowering "cacheSize" causes the crash to happen faster. It may or may not be a separate crash bug.
Lowering it too much will make it lag and display corrupt textures as soon as the game begins.

M64Py console shows error 0xE06D7363.

thecoreyburton:
What hardware are you running?

jimbo1qaz:
Windows 7 x64
Intel i7 3570k not overclocked
Nvidia gtx 660
Latest graphics drivers (not at PC, will update)
M64py latest 0.23 I think
Gliden64 PubR1 and 1.1 tested to crash (most testing was done on 1.0)

AmbientMalice:
@jimbo1qaz Are you still having problems with latest version? If not, please close this issue.

jimbo1qaz:
BitDefender detects EmuCR as malware. Also, my main PC is inactive at the moment. Should I close until further testing?

gonetz:
Yes. You may reopen it later, if necessary.

PR #745, opened 2015-10-04
Changing the texture cache to be a proper LRU cache

matthewharvey:
I change the currently used 'map' of textures to a queue (implemented as a list),
and then remove from the end of the queue when textures need removing from
the cache, and adding to the beginning of the queue.
The other operation that is needed is moving the texture to the
front of the list whenever it is used, so that frequently used
textures are not deleted.
In order to make that last operation effecient, I created
a map of the locations of these textures, and keep that
properly updated. This makes the accessing of a texture still O(1).
Then, in order to have the iterators remain valid through insertions
and deletions, I needed to implement the queue as a list.

The map and queue implementation is a standard way to implement
an LRU cache, just FYI. Not something I thought of all on my own.

Fixes issue #744 

I'm hoping that this also fixes issue #534 

Finally, I may have broken framebuffer textures (I'm not sure how those work). What is a game that uses those so that I can make sure?

I know this doesn't clean up the texture code any, but I don't quite understand it all do a nice cleanup yet. Maybe later :P

### DIFF
--- a/src/Textures.h
+++ b/src/Textures.h
@@ -2,6 +2,7 @@
 #define TEXTURES_H
 
 #include <map>
+#include <deque>
 
 #include "CRC.h"
 #include "convert.h"
@@ -82,9 +83,12 @@ private:
 	void _initDummyTexture(CachedTexture * _pDummy);
 	void _getTextureDestData(CachedTexture& tmptex, u32* pDest, GLuint glInternalFormat, GetTexelFunc GetTexel, u16* pLine);
 
-	typedef std::map<u32, CachedTexture> Textures;
+	typedef std::list<CachedTexture> Textures;
+	typedef std::map<u32, Textures::iterator> Texture_Locations;
+	typedef std::map<u32, CachedTexture> FBTextures;
 	Textures m_textures;
-	Textures m_fbTextures;
+	Texture_Locations m_lruTextureLocations;
+	FBTextures m_fbTextures;
 	CachedTexture * m_pDummy;
 	CachedTexture * m_pMSDummy;
 	u32 m_hits, m_misses;
--- a/src/Textures.h
+++ b/src/Textures.h
@@ -2,7 +2,6 @@
 #define TEXTURES_H
 
 #include <map>
-#include <deque>
 
 #include "CRC.h"
 #include "convert.h"
--- a/src/Textures.cpp
+++ b/src/Textures.cpp
@@ -524,15 +524,15 @@ void TextureCache::_checkCacheSize()
 	if (m_cachedBytes <= m_maxBytes)
 		return;
 
-	Textures::const_iterator iter = m_textures.cend();
+	Textures::iterator iter = m_textures.end();
 	do {
 		--iter;
-		const CachedTexture& tex = *iter;
+		CachedTexture& tex = *iter;
 		m_cachedBytes -= tex.textureBytes;
 		glDeleteTextures(1, &tex.glName);
 		m_lruTextureLocations.erase(tex.crc);
 	} while (m_cachedBytes > m_maxBytes && iter != m_textures.cbegin());
-	m_textures.erase(iter, m_textures.cend());
+	m_textures.erase(iter, m_textures.end());
 }
 
 CachedTexture * TextureCache::_addTexture(u32 _crc32)
--- a/src/Textures.cpp
+++ b/src/Textures.cpp
@@ -508,10 +508,11 @@ void TextureCache::destroy()
 	current[0] = current[1] = NULL;
 
 	for (Textures::const_iterator cur = m_textures.cbegin(); cur != m_textures.cend(); ++cur)
-		glDeleteTextures( 1, &cur->second.glName );
+		glDeleteTextures( 1, &cur->glName );
 	m_textures.clear();
+	m_lruTextureLocations.clear();
 
-	for (Textures::const_iterator cur = m_fbTextures.cbegin(); cur != m_fbTextures.cend(); ++cur)
+	for (FBTextures::const_iterator cur = m_fbTextures.cbegin(); cur != m_fbTextures.cend(); ++cur)
 		glDeleteTextures( 1, &cur->second.glName );
 	m_fbTextures.clear();
 
@@ -522,11 +523,14 @@ void TextureCache::_checkCacheSize()
 {
 	if (m_cachedBytes <= m_maxBytes)
 		return;
+
 	Textures::const_iterator iter = m_textures.cend();
 	do {
 		--iter;
-		m_cachedBytes -= iter->second.textureBytes;
-		glDeleteTextures( 1, &iter->second.glName );
+		const CachedTexture& tex = *iter;
+		m_cachedBytes -= tex.textureBytes;
+		glDeleteTextures(1, &tex.glName);
+		m_lruTextureLocations.erase(tex.crc);
 	} while (m_cachedBytes > m_maxBytes && iter != m_textures.cbegin());
 	m_textures.erase(iter, m_textures.cend());
 }
@@ -538,16 +542,17 @@ CachedTexture * TextureCache::_addTexture(u32 _crc32)
 	_checkCacheSize();
 	GLuint glName;
 	glGenTextures(1, &glName);
-	m_textures.emplace(_crc32, glName);
-	CachedTexture & texture = m_textures.at(_crc32);
-	texture.crc = _crc32;
-	return &texture;
+	m_textures.emplace_front(glName);
+	Textures::iterator new_iter = m_textures.begin();
+	new_iter->crc = _crc32;
+	m_lruTextureLocations.insert(std::pair<u32, Textures::iterator>(_crc32, new_iter));
+	return &(*new_iter);
 }
 
 void TextureCache::removeFrameBufferTexture(CachedTexture * _pTexture)
 {
-	Textures::const_iterator iter = m_fbTextures.find(_pTexture->glName);
-	assert(iter != m_fbTextures.end());
+	FBTextures::const_iterator iter = m_fbTextures.find(_pTexture->glName);
+	assert(iter != m_fbTextures.cend());
 	m_cachedBytes -= iter->second.textureBytes;
 	glDeleteTextures( 1, &iter->second.glName );
 	m_fbTextures.erase(iter);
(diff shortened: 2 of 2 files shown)


purplemarshmallow:
I investigated both problems. I think #744 is caused by a bug in the mupen64plus recompiler and I tested majoras mask with low cacheSize and I was not able to reproduce this problem

matthewharvey:
Can you reproduce 744? I just reproduced using the "Pure Interpreter" option along with the recompiler. I think the issue might be platform specific, as it depends on how the platform implements std::map and on how the GL driver deletes textures (if the GL driver doesn't do the cleanup of deleted textures right away, then it might not happen).

purplemarshmallow:
>I moved my texture cache down to 30MB and was able to get the issue without even getting to into a race.

That's a different problem if you set the texture cache too low things will break happens for me too and happens in every game. But if the texture cache is big enough this does not happen for me. I can not reproduce this with Project64. I think it's a bug in mupen64plus or because of the GL driver.

My GPU: Geforce GT 525M
Windows 10 64 bit

matthewharvey:
I'm going to do a little more research to try to find more data that shows this is actually an issue, but for now I can say for certain that no matter what value I set my texture cache to this happens eventually.

And yes, I can set my texture cache to 8 or 5 MB or something like that where the game will simply not be able to load all of the textures. That is a different "issue" (not something that can be fixed) that manifests itself in a slightly different way.

AmbientMalice:
Does this texture cache fix have any affect on the Ocarina of Time text corruption on Intel GPUs?

matthewharvey:
I added instrumentation to the code to print to the screen whenever a texture was added to or removed from the cache, along with the crc of the texture. I then ran until the file got about one hundred thousand lines.

What I can see is that there are many textures that are used (either added or evicted) once or twice, and then a much smaller set that are used several hundred times in the cache, and then there is a set of 30 textures that are used several thousand times. So basically the histogram of the number of textures that were used some number of times follows a 1/x type function.

For my analysis I chose one of the textures that was moved in and out of the cache several thousand times. I searched through the file for that texture's crc. In the first half of the file, the texture was added and removed a handful of times, less than 5. Then, in the second half, there were a huge number of pairs of entries that were simply adding the texture and then removing it again as the next cache operation. This is consistent with what I thought the issue was. The texture cache (at least on my system) is evicting textures as soon as they are added, and keeping textures around that are no longer in use. A proper LRU cache has the potential to solve this.

For completeness, I performed the same operation with my cache modifications in this branch. Played to the same part in the game, where the old code had seen 129,000 additions and evictions from the cache, my code saw only 6,500. The texture that had seen the largest number of transactions in the cache had seen 4 transactions (2 adds, 2 removes), compared to 3982 transactions for the most used texture in the old code.

These were both with a 30MB cache.

So I'm pretty sure that this is an issue. It's quite possible that it is not an issue on your computer, but that leaves two possibilities.

1. Your computer implements std::map in the other direction, so when you ask for the 'end', is the opposite end to mine, and thus exhibits much better cache locality, as textures that are added are not immediately evicted. (I can send you a quick patch to test this if you'd be willing to apply and test with it)

2. Your computer does exactly the same thing as mine, but when the renderer tries to render with a texture that has been glDeleteTextured'd, it just deals with it and points to the old ("deleted" but not really) data, thus not simply showing black textures. That's nice, but off spec so we really shouldn't do that.

Seeing as caches are only as good as they are at increasing performance, is there a good way to measure performance for this plugin? I can't see any performance degradation while playing, but it seems that my computer runs a little hotter maybe (not rigorously), and I'd like to have some way of checking that the performance isn't hurt too much. I don't see an FPS counter or anything like that in the options.

matthewharvey:
@AmbientMalice I just took a look through that issue and don't believe they are related. :(

purplemarshmallow:
I tested Beetle Adventure Racing more and this time I managed to get corrupt textures with Project64. You are correct then. Good work tracking this down

gonetz:
Great to see such well-founded explanation of non-obvious code change. It is possible that behavior of MS implementation of std::map differs from g++ one. As I remember, I tested _checkCacheSize and it worked properly on my PC. But if it does not work right on other platforms, I don't mind to replace it by  a proper LRU cache. I don't have time to check and approve it today, so I want to ask other developers to pull this commit and test how it works. Zilmar specs emulators have VI/s counter, which gives some information about performance.

Few notices:
* "framebuffer textures" are textures, which are used for frame buffer emulation. The are mainly FBO targets. Such textures must not be freed when we run out of texture memory, so I put them in separate cache, which _checkCacheSize does not touch. You left them as they are and it is right.
* 30MB cache is too low when frame buffer emulation is used. We need to allocate several screen-size 32bit RGBA textures for color buffers, at least one texture for depth buffer plus several auxiliary texture buffers for copy operations. Also we need to store regular game textures somewhere.

src/Textures.h
```diff
@@ -2,6 +2,7 @@
 #define TEXTURES_H
 
 #include <map>
+#include <deque>
```
gonetz: why deque, if you use std::list for Textures?

src/Textures.h
```diff
@@ -2,6 +2,7 @@
...
```
gonetz: why deque, if you use std::list for Textures?

fzurita:
I'm getting build errors in android for this branch:

jni/./mupen64plus-video-gliden64/src/Textures.cpp: In member function 'void TextureCache::_checkCacheSize()':
jni/./mupen64plus-video-gliden64/src/Textures.cpp:535:49: error: no matching function for call to 'std::list<CachedTexture>::erase(std::list<CachedTexture>::const_iterator&, std::list<CachedTexture>::const_iterator)'
         m_textures.erase(iter, m_textures.cend());
                                                 ^
jni/./mupen64plus-video-gliden64/src/Textures.cpp:535:49: note: candidates are:
In file included from C:/Users/Francisco/android-ndks/sources/cxx-stl/gnu-libstdc++/4.8/include/list:64:0,
                 from jni/./mupen64plus-video-gliden64/src/GBI.h:4,
                 from jni/./mupen64plus-video-gliden64/src/gSP.h:5,
                 from jni/./mupen64plus-video-gliden64/src/OpenGL.h:63,
                 from jni/./mupen64plus-video-gliden64/src/Textures.cpp:6:
C:/Users/Francisco/android-ndks/sources/cxx-stl/gnu-libstdc++/4.8/include/bits/list.tcc:108:5: note: std::list<_Tp, _Alloc>::iterator std::list<_Tp, _Alloc>::erase(std::list<_Tp, _Alloc>::iterator) [with _Tp = CachedTexture; _Alloc = std::allocator<CachedTexture>; std::list<_Tp, _Alloc>::iterator = std::_List_iterator<CachedTexture>]
     list<_Tp, _Alloc>::
     ^
C:/Users/Francisco/android-ndks/sources/cxx-stl/gnu-libstdc++/4.8/include/bits/list.tcc:108:5: note:   candidate expects 1 argument, 2 provided
In file included from C:/Users/Francisco/android-ndks/sources/cxx-stl/gnu-libstdc++/4.8/include/list:63:0,
                 from jni/./mupen64plus-video-gliden64/src/GBI.h:4,
                 from jni/./mupen64plus-video-gliden64/src/gSP.h:5,
                 from jni/./mupen64plus-video-gliden64/src/OpenGL.h:63,
                 from jni/./mupen64plus-video-gliden64/src/Textures.cpp:6:
C:/Users/Francisco/android-ndks/sources/cxx-stl/gnu-libstdc++/4.8/include/bits/stl_list.h:1193:7: note: std::list<_Tp, _Alloc>::iterator std::list<_Tp, _Alloc>::erase(std::list<_Tp, _Alloc>::iterator, std::list<_Tp, _Alloc>::iterator) [with _Tp = CachedTexture; _Alloc = std::allocator<CachedTexture>; std::list<_Tp, _Alloc>::iterator = std::_List_iterator<CachedTexture>]
       erase(iterator __first, iterator __last)
       ^
C:/Users/Francisco/android-ndks/sources/cxx-stl/gnu-libstdc++/4.8/include/bits/stl_list.h:1193:7: note:   no known conversion for argument 1 from 'std::list<CachedTexture>::const_iterator {aka std::_List_const_iterator<CachedTexture>}' to 'std::list<CachedTexture>::iterator {aka std::_List_iterator<CachedTexture>}'
[armeabi-v7a] Compile++ arm  : mupen64plus-video-gliden64-gles20 <= GLSLCombiner_gles2.cpp

matthewharvey:
Hi @fzurita,

From the error message you gave, I tried to fix that error message. Thanks for your help!

Matt

src/Textures.h
matthewharvey: Because I was using them before :P
Thanks for the catch.

fzurita:
No problem, I'll re-test again later

purplemarshmallow:
@matthewharvey I tested Beetle Adventure Racing with your branch. In the middle of a race performance can go down somewhere in the 2nd or 3rd lap. Project64 shows the GFX plugin needs 30% CPU and around the 2nd lap it goes up to 60% for no reason. Not sure if its related with your changes. I need to do more testing. Maybe performance suffers if the texture cache is filled up. But I haven't got any corrupt textures with your branch

purplemarshmallow:
I benchmarked it TextureCache::_load eats all the performance :/
But only happens if you play for a while

fzurita:
I was able to build for android after the latest changes. I loaded up and played a few games and didn't notice anything wrong.

purplemarshmallow:
@matthewharvey This happens also with the current master it's not caused by your changes. With current master texture gliches appear at the moment when the game slows down. With your branch there are no texture glitches but there's still the slowdown

matthewharvey:
Graphical glitches should appear when the texture cache is full without my changes, so that would be consistent with a slowdown because the texture cache has to work harder.
I can't think of any way to avoid this. Because this particular game (and not any others I've encountered) creates so many temporary textures, it seems as though it might benefit from an occasional dump of the entire cache. Any other options would cause unacceptable stuttering (clearing out only part of the cache). But there would need to be logic to decide when to clear the cache, and even then building it back might cause stuttering.
You might be able to split the cache into two lists, and then move the older textures there during adding textures, and then clearing that out at some interval or something, but I can't imagine that making performance any better overall.

So are you seeing actual slowing down of the emulator below it's optimal rate? Or is it just increased CPU usage?

Thanks for all of the testing guys! It's really good to know that it's working on other games as well.

purplemarshmallow:
>So are you seeing actual slowing down of the emulator below it's optimal rate? Or is it just increased CPU usage?

I had the frame limiter and the RDRAM options disabled. The RDRAM options are needed for some effects in this game but they are performance intensive. I can imagine that the game can slow down below fullspeed on some systems (especially android).

I can think of a solution. What if the textures are checked before they are saved in the cache? The temporary textures must have something in common. And then we don't store these textures in the cache anymore.

gonetz:
Added to master:
https://github.com/gonetz/GLideN64/commit/0dfbd34b09101dfb1535dabed0563c2b30e01688

@matthewharvey Thanks for the good work!
@fzurita, @purplemarshmallow Thanks for the testing!

matthewharvey:
@purplemarshmallow that sort of game specific hack is probably beyond my skill level. I don't have an Android version to test with, but unless you think that my version is performing substantially worse, I think it's not worth it. Adding the code complexity of that sort would also require a pretty big justification.
@gonetz Thanks!

purplemarshmallow:
>unless you think that my version is performing substantially worse, I think it's not worth it.

performance is the same I think
>that sort of game specific hack is probably beyond my skill level

I wouldn't see that as a hack. If you store a texture in the cache or not it shouldn't matter for accuracy. I also wouldn't want to make it game specific I think any game can fill the cache with garbage. I would see it just as a performance optimization. But I'm not sure if the idea is good

Issue #604, opened 2015-07-05
GLideN64 not starting on android x86


fzurita:
GLideN64 does not seem to be starting up in android x86. I have tried debugging the issue and it's baffling me. Here is the log output when the plugin fails to start, it does not seem to be crashing:

07-05 00:11:04.029: I/GameSurface(11018): Creating GL context
07-05 00:11:04.029: V/GameSurface(11018): Found EGL display connection
07-05 00:11:04.029: V/GameSurface(11018): Initialized EGL display connection
07-05 00:11:04.030: V/GameSurface(11018): Found compatible EGL frame buffer configuration
07-05 00:11:04.036: V/GameSurface(11018): Created EGL rendering context
07-05 00:11:04.038: V/GameSurface(11018): Created EGL window surface
07-05 00:11:04.047: V/GameSurface(11018): Bound EGL rendering context to EGL window surface
07-05 00:11:04.047: I/GameSurface(11018): Created GL context OpenGL ES 3.1 build 1.4@3300288
07-05 00:11:04.047: D/GLideN64(11018): [gles2GlideN64]: Create setting videomode 1440x1080
07-05 00:11:04.047: D/GLideN64(11018): OpenGL version string: OpenGL ES 3.1 build 1.4@3300288
07-05 00:11:04.047: D/GLideN64(11018): OpenGL vendor: Imagination Technologies
07-05 00:11:04.047: D/GLideN64(11018): OpenGL renderer: PowerVR Rogue G6430
07-05 00:11:04.047: D/GLideN64(11018): OpenGL major version: 3
07-05 00:11:04.047: D/GLideN64(11018): OpenGL minor version: 1
07-05 00:11:04.047: D/GLideN64(11018): ImageTexture support: yes
07-05 00:11:04.047: D/GLideN64(11018): Max Anisotropy: 0.000000
07-05 00:11:05.078: I/GameSurface(11018): Destroying GL context
07-05 00:11:05.078: V/GameSurface(11018): Unbound EGL rendering context from EGL window surface
07-05 00:11:05.083: V/GameSurface(11018): Destroyed EGL window surface
07-05 00:11:05.086: V/GameSurface(11018): Destroyed EGL rendering context
07-05 00:11:05.086: V/GameSurface(11018): Terminated EGL display connection
07-05 00:11:05.091: D/Core(11018): Rom closed.

I have been trying to debug this issue and some strange things are happening.

[...]

PR #750, opened 2015-10-07
Add Raspberry Pi support

gizmo98:
Add support for raspberry pi's broadcom videocore IV (VC) gpu and raspbian linux distribution.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -199,15 +199,12 @@ else(GLES2)
   )
 endif(GLES2)
 
-if(CMAKE_COMPILER_IS_GNUCXX)
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
     #check for G++ 4.8+
     execute_process(COMMAND ${CMAKE_CXX_COMPILER} -dumpversion OUTPUT_VARIABLE G++_VERSION)
     if (G++_VERSION VERSION_LESS 4.8)
-        message(SEND_ERROR "You need at least G++ 4.8 to compile EmulationStation!")
+        message(SEND_ERROR "You need at least G++ 4.8 to compile GLideN64!")
     endif()
-endif()
-
-if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
 	SET(GCC_CPP11_COMPILE_FLAGS "-std=c++0x -static-libgcc -static-libstdc++")
 	SET( CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} ${GCC_CPP11_COMPILE_FLAGS}" )
 	SET(GCC_STATIC_LINK_FLAGS "-static-libgcc -static-libstdc++")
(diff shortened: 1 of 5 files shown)


AmbientMalice:
Without FB emulation, the list of games that will work grows significantly shorter -- although that's likely just the reality for Pi hardware.

gizmo98:
According to ticket #604 rpi does not support depth texture attachements. So rpi cannot use FB emulation atm.

Even without FB emulatin GLideN64 seems to be more compatible as any other plugin on the pi.

AmbientMalice:
I thought depth texture attachments only affected depth buffer emulation.

gizmo98:
I don't know. According to the above mentioned ticket depth buffer texture attachements seem to be necessary. If FB emulation is on screen stays black.

gonetz:
Yes, Frame Buffer Object needs both color and depth attachments. You will not get depth buffer functionality without depth attachment.

gizmo98:
#gonetz 
Thank you! I will modify unsigned, location of getDisplaySize and m_bFullscreen check. Do you have any other suggestions?
GCC-4.7 is not working. You use std::map::emplace http://en.cppreference.com/w/cpp/container/map/emplace. There is a "function emplace not found" error. You need at least GCC-4.8 to use this function. The current GNU compiler check does not work.

gizmo98:
Ok. I modified the mentioned code and recompiled everything. It still runs.

gonetz:
Good. I have added few requests to code appearance.
Plus, now your changes conflict with current master. I just merged bunch of modifications to it, including config files.
Please fix the conflicts. Also, please combine "Add gcc version check" and "cmake: merge gcc checks" into one commit. Same for "rpi: use screen resolution if fullscreen=1" and "rpi: move platform specific code into own function ".

gizmo98:
@gonetz 
I have added all changes to your last commit. https://github.com/gizmo98/GLideN64/commits/rpi-fix

There is a "undefined symbol: osal_is_directory" error now. I have seen you modified osal library position lately:
https://github.com/gonetz/GLideN64/commit/f212710ea752bc473566c493a8530f02d9b82827

gizmo98:
I fixed the "undefined symbol: osal_is_directory"  bug.

dankcushions:
In mupen64plus, the screenwidth/height config options are commented as being for windowed resolutions AND full screen resolutions. I tried using those for a pi3 (by adjusting https://github.com/gonetz/GLideN64/pull/750/commits/e1cb0fa4f576bb8e74e151c44b1641fbffe0455c to pick up the config height/width, rather than the current screen resolution), but it would mean there was only a small image displayed in the corner.

Is this just the way GLideN64 works for all hardware? It always want to use the current screen resolution, rather than the configured one?

gizmo98:
If fullscreen is enabled it uses your screen resolution if you have a raspberry pi. If you have a PC screen resolution will be switched to screenwidth/height config. Mupen64plus uses SDL2. As far as i know SDL2 does not support screen resolution switching on a pi.

PR #751, opened 2015-10-08
GlideNUI: Fix an out of bounds access in Settings.cpp

lioncash:


src/GLideNUI/Settings.cpp
```diff
@@ -78,7 +78,7 @@ void _loadSettings(QSettings & settings)
 	config.font.color[0] = fontColor.red();
 	config.font.color[1] = fontColor.green();
 	config.font.color[2] = fontColor.blue();
-	config.font.color[4] = fontColor.alpha();
```
gonetz: Ops!

PR #754, opened 2015-10-09
Fix build error for target GLES2

gizmo98:
On a GLES2 target build error output tells "GL_NUM_PROGRAM_BINARY_FORMATS was not declared in this scope". GL_NUM_PROGRAM_BINARY_FORMATS is not declared in gl2.h. Regression was introduced here: https://github.com/gonetz/GLideN64/commit/e126a51f6f8c6a4f47155a8d8e564b823d8d3bb9

### DIFF
--- a/src/Combiner.cpp
+++ b/src/Combiner.cpp
@@ -97,12 +97,16 @@ void CombinerInfo::init()
 {
 	m_pCurrent = NULL;
 	m_pUniformCollection = createUniformCollection();
+#if defined(GLES2)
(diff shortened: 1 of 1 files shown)


matthewharvey:
Shoot. Sorry about this. I don't have a GLES2 target to test with. I'll try to be more careful in the future.

gonetz:
My bad: I had to check Android compilation after any such change.
I feel sick from all these #ifdef GLES2 checks, so I fixed it a bit differently:
https://github.com/gonetz/GLideN64/commit/7303ce42865167838d1b0e0cd1471e09fd07b989

PR #756, opened 2015-10-10
GBI: Minor cleanup

lioncash:


gonetz:
Very clean and accurate commits. The code definitely looks better with them. Thanks!

src/GBI.cpp
```diff
@@ -142,6 +143,11 @@ bool GBIInfo::isHWLSupported() const
 	return true;
 }
 
+void GBIInfo::_flushCommands()
```
Narann: What about using `constexpr`?

lioncash: std::fill does not satisfy the constexpr requirements.

Narann: I wonder if the original version couldn't be `constexpr`'ed:
```cpp
constexpr void GBIInfo::_flushCommands()
{
    for (u32 i = 0; i <= 0xFF; ++i)
        cmd[i] = GBI_Unknown;
}
```

lioncash: Sure, but MSVC doesn't support extended constexpr, which is why I didn't do that.

PR #763, opened 2015-10-16
rename RSP.DList

purplemarshmallow:
I think it's confusing RSP.DList does not actually count the dlists. The amount of dlists and buffer swaps can be the same but it's not necessarily the case. Maybe this should be changed further and get moved away from RSP

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -287,7 +287,7 @@ void FrameBuffer::copyRdram()
 
 bool FrameBuffer::isValid() const
 {
-	if (m_validityChecked == RSP.buffer_count)
+	if (m_validityChecked == video().getBuffersSwapCount())
 		return true; // Already checked
 
 	const u32 * const pData = (const u32*)RDRAM;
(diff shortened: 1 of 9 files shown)


gonetz:
* Since RSP.DList is modified in OGLVideo::swapBuffers(), it will be more logical to make it as OGLVideo class private member and provide access to it via getter method
* I use camel case for variables and class/struct members. All these uc_start, uc_dstart are legacy and this style must not be used in new code. Class members have m_ prefix.
* buffer_count is not much better than DList. I think frameCount is better.

So, please change it as following:
* remove DList from RSPInfo.
* add private member m_frameCount to OGLVideo
* add public method
u32 getFrameCount() const {return m_FrameCount;}
* replace RSP.DList by getFrameCount()

purplemarshmallow:
>•buffer_count is not much better than DList. I think frameCount is better.

I was thinking of that but if hack_VIUpdateOnCIChange is enabled there are dummy swapbuffers. I think the number of buffer swaps can be higher than the number of frames

gonetz:
Ok, m_buffersSwapCount, getBuffersSwapCount

purplemarshmallow:
done

gonetz:
Thanks!
https://github.com/gonetz/GLideN64/commit/c661ace24ec7c88968f9cd0359e3af224cd48a3f

PR #770, opened 2015-10-19
Make Qt optional

purplemarshmallow:
Release_noUI configuration now builds the plugin without UI
I think it's a good thing to seperate the essential parts of the plugin and the optional parts. Unfortunately Qt currently reads the ini files and writes the screenshots so this functionality will be missing. Only the default settings can be used

### DIFF
--- a/projects/msvc12/GLideN64.sln
+++ b/projects/msvc12/GLideN64.sln
@@ -20,6 +20,7 @@ Global
 		MinSizeRel|Win32 = MinSizeRel|Win32
 		Release_dll|Win32 = Release_dll|Win32
 		Release_mupenplus|Win32 = Release_mupenplus|Win32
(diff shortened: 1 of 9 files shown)


gonetz:
Why?
GUI is essential part of zilmar spec plugin.
I see no point to use plugin without that functionality.
I may add this commit to a separate branch in case some devs need it, but not in master.

purplemarshmallow:
This is more a suggestion
I think it's useful and it makes building the project easier

gonetz:
It does not make building the functional project easier. If you want to simplify life for those devs, who have troubles with QT, make dummy implementation of GLidenHQ. Main project will remain the same, without these ugly #ifdef NO_UI. Just switch to dummy lib in project properti

gonetz:
*properties, and you will get no-ui versulion.

Issue #725, opened 2015-09-24
Pokemon Stadium 2 validity check fails in LLE in the library


purplemarshmallow:
In LLE the game messes up in the library. Does not happen in HLE. Strange... the framebuffer code should work the same in LLE and HLE...
![gliden64_pokemon_stadium_2_007](https://cloud.githubusercontent.com/assets/7283660/10086959/8de3db98-6311-11e5-80ac-e817043b009b.jpg)

Issue #723, opened 2015-09-24
Pokemon Stadium 2 library images


purplemarshmallow:
It can still happen that images in the library appear incorrect. I overlooked this because it's hard to find this problem. To reproduce it you have to:
1 Open a book from the "Pokemon" category. It must have over 24 images or you can't scroll down
2 Scroll down
3 Go to the "Egg groups" category and under "check by group" open a page
The library is problematic because the game reuses the same buffers multiple times. If buffers are not visible the game can store data there
![gliden64_pokemon_stadium_2_011](https://cloud.githubusercontent.com/assets/7283660/10086053/a3632172-630c-11e5-8003-94296d68f204.jpg)

Issue #699, opened 2015-09-02
Pokemon Stadium missing framebuffer effect in Pokedex


purplemarshmallow:
The same auxiliary buffer is loaded multiple times. Correct with Glide64 but Glide64 writes the buffer to RDRAM, it does not use hardware framebuffer emulation in this game.
savestate: https://drive.google.com/file/d/0B7Y6r4SpC_QQeE1DU1U5enJHT0E/view?usp=sharing
![gliden64_pokemon_stadium_000](https://cloud.githubusercontent.com/assets/7283660/9630968/f87dbb7c-517d-11e5-84d1-f76e2d598218.jpg)
![glide64_pokemon_stadium_01](https://cloud.githubusercontent.com/assets/7283660/9630969/fb723c40-517d-11e5-99c5-8b0c88a35d66.png)

purplemarshmallow:
fixed in FB to RDRAM branch
![gliden64_pokemon_stadium_000](https://cloud.githubusercontent.com/assets/7283660/10677402/719a4ce0-790c-11e5-96c5-e01745e0be5b.jpg)

oddMLan:
![GLideN64_POKEMON_STADIUM_000](https://user-images.githubusercontent.com/7278372/68209236-2608df80-ff90-11e9-9317-9feb132bf3aa.png)
correct with copy aux color buffers from rdram

Issue #617, opened 2015-07-16
copy color to RDRAM overwrites data written by the CPU


purplemarshmallow:
If the CPU writes data to colorbuffer, copy color to RDRAM overwrites this data. This does not happen with software gfx plugins (Jabo's, angrylion's).

Example: Superman64 websites
The CPU draws the website one time and then this part of the screen is not updated anymore. An image is rendered on top of that. This works with software gfx plugins. With GLideN64 this effect works too, copy color from RDRAM must be enabled because it's a CPU based effect (screenshot). But if copy color to RDRAM is enabled at the same time the plugin fills the colorbuffer with black pixels, the website written by the CPU is lost (screenshot).

I suspect this problem causes some issues where software gfx plugins work, but in these cases turning copy color off is not a solution because the CPU needs this data

Savestate for Superman: https://drive.google.com/file/d/0B7Y6r4SpC_QQUkxqZkJjWl9pY3M/view?usp=sharing
![gliden64_superman_000](https://cloud.githubusercontent.com/assets/7283660/8725638/958498c4-2bd5-11e5-879e-fb5ebf2aebc1.jpg)
![gliden64_superman_001](https://cloud.githubusercontent.com/assets/7283660/8725644/a085bb86-2bd5-11e5-890d-c37676128635.jpg)

AmbientMalice:
This kinda reminds me of the character select screen in Vigilante 8, with a frame that briefly flashes onscreen before going black. (However, you can get it back by scrolling left and right, so it may not be related.)

olivieryuyu:
the beta of superman brings you easily to this screens. i cannot make them work indeed

purplemarshmallow:
Mostly fixed in fb_fixes branch
copy color from RDRAM is no longer needed
The first word is flickering
Copy Color to RDRAM still breaks the websites
![gliden64_superman_000](https://user-images.githubusercontent.com/7283660/46829051-fa232680-cd9c-11e8-83ea-c9ca922bb642.jpg)

oddMLan:
bug is worse now, first line constantly flickers, and copy aux color buffers from rdram doesnt help at all

PR #773, opened 2015-10-23
implement framebuffer to RDRAM

purplemarshmallow:
At the moment for auxiliary buffers only. async is disabled for now. This needs a config option.
My idea is to write all auxiliary buffers at fullsync and then remove them. Works very good in a number of games. Copying main framebuffers still needs to be implemented. 

Current state: 
Pokemon Stadium games: #415 is back. Some bugs are fixed like #725 #723 #699
Mario Tennis: works at the beginning but crashes after a while (?)
Pokemon Snap: camera detection works
Zelda OOT: works but the subscreen delay fix must be disabled
Yoshi's Story: Water effects are correct
Conker/Jet Force Gemini: Crash when shadows are used
Body Harvest: just freezes

screenshots:
gliden64_mariotennis_001
gliden64_the_legend_of_zelda_004
gliden64_pokemon_stadium_2_009
gliden64_pokemon_snap_000

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -23,15 +23,15 @@ class FrameBufferToRDRAM
 {
 public:
 	FrameBufferToRDRAM() :
-		m_bSync(true), m_FBO(0), m_pTexture(NULL), m_curIndex(0)
+		m_FBO(0), m_pTexture(NULL)
 	{
-		m_aPBO[0] = m_aPBO[1] = 0;
+		m_PBO[0] = m_PBO[1] = m_PBO[2] = 0;
 	}
 
 	void Init();
 	void Destroy();
 
-	void CopyToRDRAM(u32 _address);
+	void CopyToRDRAM(u32 _address, u32 readPBO, u32 writePBO);
 
 private:
 	union RGBA {
@@ -41,11 +41,9 @@ private:
 		u32 raw;
 	};
 
-	bool m_bSync;
 	GLuint m_FBO;
 	CachedTexture * m_pTexture;
-	u32 m_curIndex;
-	GLuint m_aPBO[2];
+	GLuint m_PBO[3];
 };
 
 class DepthBufferToRDRAM
@@ -537,7 +535,7 @@ void FrameBufferList::toRDRAM()
 {
 	for (FrameBuffers::iterator iter = m_list.begin(); iter != m_list.end(); ++iter) {
 		if (iter->m_width != VI.width && iter->m_height != VI.height) {
-			FrameBuffer_CopyToRDRAM(iter->m_startAddress);
+			FrameBuffer_CopyToRDRAM(iter->m_startAddress, 3, 3);
 			m_list.erase(iter);
 		}
 	}
@@ -942,14 +940,16 @@ void FrameBufferToRDRAM::Init()
 	glBindFramebuffer(GL_DRAW_FRAMEBUFFER, 0);
 
 	// Generate and initialize Pixel Buffer Objects
-	glGenBuffers(2, m_aPBO);
-	glBindBuffer(GL_PIXEL_PACK_BUFFER, m_aPBO[0]);
+	glGenBuffers(3, m_PBO);
+	glBindBuffer(GL_PIXEL_PACK_BUFFER, m_PBO[0]);
 	glBufferData(GL_PIXEL_PACK_BUFFER, m_pTexture->textureBytes, NULL, GL_DYNAMIC_READ);
-	glBindBuffer(GL_PIXEL_PACK_BUFFER, m_aPBO[1]);
+	glBindBuffer(GL_PIXEL_PACK_BUFFER, m_PBO[1]);
+	glBufferData(GL_PIXEL_PACK_BUFFER, m_pTexture->textureBytes, NULL, GL_DYNAMIC_READ);
+	glBindBuffer(GL_PIXEL_PACK_BUFFER, m_PBO[2]);
 	glBufferData(GL_PIXEL_PACK_BUFFER, m_pTexture->textureBytes, NULL, GL_DYNAMIC_READ);
 	glBindBuffer(GL_PIXEL_PACK_BUFFER, 0);
-	m_bSync = config.frameBufferEmulation.copyToRDRAM == Config::ctSync;
-	m_curIndex = 0;
+	gDP.colorImage.sync = config.frameBufferEmulation.copyToRDRAM == Config::ctSync;
+	gDP.colorImage.curIndex = 0;
 }
 
 void FrameBufferToRDRAM::Destroy() {
@@ -962,11 +962,11 @@ void FrameBufferToRDRAM::Destroy() {
 		textureCache().removeFrameBufferTexture(m_pTexture);
 		m_pTexture = NULL;
 	}
-	glDeleteBuffers(2, m_aPBO);
-	m_aPBO[0] = m_aPBO[1] = 0;
+	glDeleteBuffers(3, m_PBO);
+	m_PBO[0] = m_PBO[1] = m_PBO[2] = 0;
 }
 
-void FrameBufferToRDRAM::CopyToRDRAM(u32 _address)
+void FrameBufferToRDRAM::CopyToRDRAM(u32 _address, u32 readPBO, u32 writePBO)
 {
 	const u32 numPixels = VI.width * VI.height;
 	if (numPixels == 0 || frameBufferList().getCurrent() == NULL) // Incorrect buffer size or no current buffer. Don't copy
@@ -999,13 +999,9 @@ void FrameBufferToRDRAM::CopyToRDRAM(u32 _address)
 	glBindFramebuffer(GL_READ_FRAMEBUFFER, m_FBO);
 	glReadBuffer(GL_COLOR_ATTACHMENT0);
 #ifndef GLES2
-	// If Sync, read pixels from the buffer, copy them to RDRAM.
-	// If not Sync, read pixels from the buffer, copy pixels from the previous buffer to RDRAM.
-	//m_curIndex ^= 1;
-	//const u32 nextIndex = m_bSync ? m_curIndex : m_curIndex^1;
-	glBindBuffer(GL_PIXEL_PACK_BUFFER, m_aPBO[0]);
+	glBindBuffer(GL_PIXEL_PACK_BUFFER, m_PBO[readPBO]);
 	glReadPixels(0, 0, pBuffer->m_width, pBuffer->m_height, GL_RGBA, GL_UNSIGNED_BYTE, 0);
-	PBOBinder binder(GL_PIXEL_PACK_BUFFER, m_aPBO[0]);
+	PBOBinder binder(GL_PIXEL_PACK_BUFFER, m_PBO[writePBO]);
 	GLubyte* pixelData = (GLubyte*)glMapBufferRange(GL_PIXEL_PACK_BUFFER, 0, numPixels * 4, GL_MAP_READ_BIT);
 	if(pixelData == NULL)
 		return;
@@ -1052,10 +1048,10 @@ void FrameBufferToRDRAM::CopyToRDRAM(u32 _address)
 }
 #endif // GLES2
 
-void FrameBuffer_CopyToRDRAM(u32 _address)
+void FrameBuffer_CopyToRDRAM(u32 _address, u32 readPBO, u32 writePBO)
 {
 #ifndef GLES2
-	g_fbToRDRAM.CopyToRDRAM(_address);
+	g_fbToRDRAM.CopyToRDRAM(_address, readPBO, writePBO);
 #else
 	if ((config.generalEmulation.hacks & hack_subscreen) == 0)
 		return;
@@ -1218,7 +1214,7 @@ bool FrameBuffer_CopyDepthBuffer( u32 address ) {
 	FrameBuffer * pCopyBuffer = frameBufferList().getCopyBuffer();
 	if (pCopyBuffer != NULL) {
 		// This code is mainly to emulate Zelda MM camera.
-		g_fbToRDRAM.CopyToRDRAM(pCopyBuffer->m_startAddress);
+		g_fbToRDRAM.CopyToRDRAM(pCopyBuffer->m_startAddress, 3, 3);
 		pCopyBuffer->m_RdramCopy.resize(0); // To disable validity check by RDRAM content. CPU may change content of the buffer for some unknown reason.
 		frameBufferList().setCopyBuffer(NULL);
 		return true;
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -191,8 +191,14 @@ void FrameBuffer::init(u32 _address, u32 _endAddress, u16 _format, u16 _size, u1
 	m_width = _width;
 	m_height = _height;
 	m_size = _size;
-	m_scaleX = ogl.getScaleX();
-	m_scaleY = ogl.getScaleY();
+	if (m_width != VI.width && config.frameBufferEmulation.copyAuxiliary == 1) {
+		m_scaleX = 1;
+		m_scaleY = 1;
+	}
+	else {
+		m_scaleX = ogl.getScaleX();
+		m_scaleY = ogl.getScaleY();
+	}
 	m_fillcolor = 0;
 	m_cfb = _cfb;
 	m_needHeightCorrection = _width != VI.width && _width != *REG.VI_WIDTH;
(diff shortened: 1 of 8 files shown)


AmbientMalice:
>Pokemon Snap: camera detection works

I notice Snap's red dot needs FB to texture, which has been broken on the Zilmar versions forever.

>Zelda OOT: works but the subscreen delay fix must be disabled

I assume that once main framebuffer is implemented, the fix won't be needed in this mode? I assume this'll also properly fix Animal Crossing and such?

>Mario Tennis: works at the beginning but crashes after a while (?)

I was hoping this would fix the buggy menu transitions. But maybe implementing main framebuffer will fix that.

purplemarshmallow:
>I notice Snap's red dot needs FB to texture, which has been broken on the Zilmar versions forever.

camera detection works but the red dot is missing. Copy from RDRAM does not work here. Not in the zilmar spec version and not with mupen64plus.
Also Pokemon Snap needs hack_VIUpdateOnCIChange. Otherwise there will be depth problems and flickering

>I assume that once main framebuffer is implemented, the fix won't be needed in this mode? I assume this'll also properly fix Animal Crossing and such?

I hope but not sure. It worries me that Jabo's software renderer has the subscreen delay

>I was hoping this would fix the buggy menu transitions. But maybe implementing main framebuffer will fix that.

Mario Tennis draws things in a strange way. If I attempt to remove a framebuffer and copy it to RDRAM nearly everything turns black. Pokemon Snap has a similar problem.

gonetz:
Strange solution, which kills main advantage of hardware fb emulation: high-res frame buffers. Link in Zelda OOT subscreen looks ugly.

It can be used if for some reason standard method can't be applied. But I need to be sure that copy to RDRAM  is the last resort.

AmbientMalice:
>Strange solution, which kills main advantage of hardware fb emulation: high-res frame buffers. Link in Zelda OOT subscreen looks ugly.

I'm no expert, but this is pretty normal behavior in the sense that emulators such as Dolphin have been unable to fix certain framebuffer effects without doing native resolution RDRAM copies. Granted, there may be clever methods, but they would tend to be per-game hacks akin to what Nintendo seems to be doing with their own N64 VC emulators..

purplemarshmallow:
I don't know why copy to RDRAM should be avoided at all cost. It's an accurate way of framebuffer emulation and I think better than a hack.

>I assume that once main framebuffer is implemented, the fix won't be needed in this mode? I assume this'll also properly fix Animal Crossing and such?

Bad news here. I tried copying all buffers instantly. Nothing works. I always have the subscreen delay

gonetz:
>>  this is pretty normal behavior in the sense that emulators such as Dolphin have been unable to fix certain framebuffer effects without doing native resolution RDRAM copies.

GLideN64 uses exactly the same approach: if certain framebuffer effect can't be emulated with HWFBE, copy to RDRAM is used.

>> I don't know why copy to RDRAM should be avoided at all cost.

Because image quality degraded.

>> It's an accurate way of framebuffer emulation and I think better than a hack.

Acceptable when there is no other way. We consciously use HWFBE, which is hack by its nature to get best possible image quality. Now you say "accuracy is better than a hack". Users who want accuracy use software render. The goal of hardware render is to get nice high-res picture. Low-res approaches are used only when we face the choice: low-res picture or no picture at all.

purplemarshmallow:
Dolphin's approach is a bit different. They put most effort into the copy to RAM option. Not so much effect is put into framebuffer to texture. Framebuffer to texture is still default for most games. If an effect does not work they tell users: "disable fb to texture to see effect xy".

I'll agree that hardware framebuffer emulation should be default. But I think the focus should still go more towards accurate emulation. Even if it's not the default setting. Writing special code for hardware framebuffer emulation should not be as important.

purplemarshmallow:
Async buffer reads are working again. And I added an option in the mupen64plus settings. I think this can be merged.

purplemarshmallow:
I noticed there are failed assertions in debug build. This should be fixed before merging

AmbientMalice:
@purplemarshmallow Would you mind uploading a build with this branch and also the Pokemon VI hack enabled?

gonetz:
>Dolphin's approach is a bit different. They put most effort into the copy to RAM option... I think the focus should still go more towards accurate emulation.  

I don't know about difficulties in implementation of copy to RAM for GC emulation, but copy to RAM for N64 is simple as a stick. It does not need any special code besides the already implemented. I don't see what to focus on there. If you did it right, it will work for everything.

purplemarshmallow:
I fixed the problems with the debug build

>copy to RAM for N64 is simple as a stick

I'm not sure yet how copying main frame buffers can be implemented properly

>I don't see what to focus on there. If you did it right, it will work for everything.

Mario Tennis freezes randomly. And there are strange objects on top of some buffers. Conker/Jet Force Gemini freezes when shadows are used. I noticed in Pokemon Snap the camera can sometimes trigger incorrectly even if there's no Pokemon on the screen. I got Body Harvest working. I had to remove VI width calculation. Collisions start working somehow but still incorrect. I think the game uses auxiliary buffers for collision detection but not sure

Not sure if something is wrong with my implementation or if copying to RDRAM revealed problems with the framebuffer code

>@purplemarshmallow Would you mind uploading a build with this branch and also the Pokemon VI hack enabled?

Currently I use mupen64plus. I can only upload a mupen64plus build

gonetz:
>Not sure if something is wrong with my implementation or if copying to RDRAM revealed problems with the framebuffer code

Copy buffer to RDRAM has one problem: buffer height is unknown. Game set only buffer width. This problem has no safe solution; we may only guess about the height. Main buffer somehow correlates with VI.height. Any assumption about height of auxiliary buffer can be wrong. If real height is higher than our assumption, we will get incomplete picture. If real height is less, copy to RDRAM will corrupt game data  and cause glitches or crashes.

purplemarshmallow:
That's a bad problem I hope it's possible to solve this. Maybe there's a better method to get the buffer height?

>I noticed in Pokemon Snap the camera can sometimes trigger incorrectly even if there's no Pokemon on the screen

Also happens with angrylion's plugin so it's unlikely a problem with GLideN64

purplemarshmallow:
I made a fix. Now scoreboards in Mario Tennis work properly. Time to remove the ugly scoreboard hack

src/FrameBuffer.cpp
```diff
@@ -1208,7 +1228,7 @@ bool FrameBuffer_CopyDepthBuffer( u32 address ) {
 	FrameBuffer * pCopyBuffer = frameBufferList().getCopyBuffer();
 	if (pCopyBuffer != NULL) {
 		// This code is mainly to emulate Zelda MM camera.
-		g_fbToRDRAM.CopyToRDRAM(pCopyBuffer->m_startAddress);
+		g_fbToRDRAM.CopyToRDRAM(pCopyBuffer->m_startAddress, 3, 3);
```
gonetz: 3, 3 ?

gonetz:
1. Please rearrange commits chain. Current set of "implement-fix-fix..." can be replaced by one-two accurate commits.
2. Fix  "int i; if (i) .. if (!i)"  by "if (i !=0) .. if (i ==0)" to avoid implicit cast to bool.
3. Fix indexes in CopyToRDRAM(..., 3, 3)
4. gDP.colorImage.sync and gDP.colorImage.curIndex  is a bad architectural solution. sync and curIndex are internal frame buffer knowledge, and gDP must not know about it.

>I made a fix. Now scoreboards in Mario Tennis work properly. Time to remove the ugly scoreboard hack

As I understand, this behavior depends on config.frameBufferEmulation.copyAuxiliary and will copy and clear all aux buffers, including shadows and lots of other things. This is bad for performance and quality.

purplemarshmallow:
>As I understand, this behavior depends on config.frameBufferEmulation.copyAuxiliary and will copy and clear all aux buffers, including shadows and lots of other things. This is bad for performance and quality.

Not at all it's meant to be a general solution that works for all buffers. I hoped it would help with #617 as well but it does not help. I don't know why it should affect performance. But it causes a regression. The buffers with the koopas can fill with garbage. I managed to solve this one as well. What do you think?

```
if (c.a)
    ptr_dst[(x + y*pBuffer->m_width) ^ 1] = ((c.r >> 3) << 11) | ((c.g >> 3) << 6) | ((c.b >> 3) << 1 | 1) ;
else
    ptr_dst[(x + y*pBuffer->m_width) ^ 1] = ((ptr_dst[(x + y*pBuffer->m_width) ^ 1] >> 1) << 1);
```

gonetz:
>Not at all it's meant to be a general solution that works for all buffers.

Wait. You suggest to copy all aux buffers to RDRAM and load textures from RDRAM. That is hi-res buffer is converted to low-res. It is ok for original 2D images, like scoreboard or koopas. However, it means quality degradation for buffers with 3D images and for buffers, which use filtered textures or textures from hd textures packs. 
Then, aux buffers for dynamic shadows changed every frame. Thus you need to copy them to RDRAM every frame in sync mode. It is not for free.

>I managed to solve this one as well. What do you think?

If incoming pixel's alpha is not zero, you copy the pixel as usual. Ok. But, if it is zero, you discard incoming pixel and use currently stored pixel instead, but only with alpha turned to zero, and you selected very weird way to make it zero. I don't understand that decision.
Do you understand format of 16bit RGBA? All you need to do is to convert incoming 32RGBA to 16bit one. Alpha conversion looks like this: if (c.a > 0) alpha =1 else alpha = 0. Alpha is the first bit of 16bit rgba (5-5-5-1).

purplemarshmallow:
>Wait. You suggest to copy all aux buffers to RDRAM and load textures from RDRAM.

Because that's how real hardware works

>That is hi-res buffer is converted to low-res. It is ok for original 2D images, like scoreboard or koopas. However, it means quality degradation for buffers with 3D images and for buffers, which use filtered textures or textures from hd textures packs. 

I think compatibility and accuracy is way more important than hi-res, texture packs, filters and all that stuff. Where's the benefit if games don't work properly?

>All you need to do is to convert incoming 32RGBA to 16bit one.

No I also need to prevent the scoreboard from being overwritten. I need to blend the numbers with the scoreboard. And I need to prevent garbage in the buffers with the koopas. This code does this

AmbientMalice:
I agree with @purplemarshmallow that accuracy and compatibility should trump niceties such as increased resolution when those nicities fundamentally break accuracy. However, I think GLideN64 should strive to achieve a balance. Might it be possible to use this more accurate mode as a default, while also allowing users to disable it and revert to the old behavior if they wish? Kinda like how GLideN64 defaults to mipmapping on by default -- it can look ugly (Beetle Adventure Racing signposts look hideous), but it's simply more accurate. Give it a scary sounding name like "Accurate RDRAM emulation" so that users won't be inclined to disable it instinctively, but people who wanna use high resolution texture packs and such will have the option.

purplemarshmallow:
I think hardware framebuffer emulation should never be removed from the plugin. But you have to accept that certain things can't be emulated properly with this method. It makes no sense to try to fix it up with hacks. Adding things like the ugly scoreboard hack is a bad solution. It barely works and causes glitches. If you want to see effects like this you have to enable the accurate option. I see no reason to hack around here

AmbientMalice:
>I think hardware framebuffer emulation should never be removed from the plugin.
Oh, sure. Especially since so many cases can be more or less accurately emulated with it. I suppose the problem is deciding whether GLideN64 should default to hardware and enable software FB on a per-game basis or default to software and enable hardware on a per-game basis. Likely the former.

>If you want to see effects like this you have to enable the accurate option. I see no reason to hack around here

Agree here.

purplemarshmallow:
>It is regrettable that there is no apparent way to solve the issue without resorting to native resolution aux FB effects.

maybe there's a way to hack around it. But I don't think it's worth it. If you have a general solution it will work everywhere. Hacks make the code ugly and can cause bugs

>the problem is deciding whether GLideN64 should default to hardware and enable software FB on a per-game basis or default to software and enable hardware on a per-game basis.

GLideN64 does not support software FB. Jabo's plugin does this. But Jabo added tons of hacks around it

AmbientMalice:
>GLideN64 does not support software FB.
Sorry, I meant the opposite of hardware framebuffering. Bad terminology on my part.

gonetz:
Current GLideN64 policy is to use HWFBE where it csn be used; switch to other methods where it falls. Copy aux buffers to RDRAM is an useful-for-some-cases technique, like current copy main fb to RDRAM, which is used to emulate effects incompatible with HWFBE.

>I think compatibility and accuracy is way more important than hi-res, texture packs, filters and all that stuff. 

All these things are important. I also prefer to see game perfectly emulated. However, let not forget that hardware rendering is imperfect by design, and we choose it to get better picture than original hardware offers.


>Where's the benefit if games don't work properly?

Sure, if game works imperfect with one method, we should find another one to fix it. If it means have-some-glitch VS no-glitches-but-looks-ugly, let it for user to decide.

>No I also need to prevent the scoreboard from being overwritten. I need to blend the numbers with the scoreboard. And I need to prevent garbage in the buffers with the koopas. This code does this

Sorry, but your code does weird things, and I tried to point on that. If you need to blend numbers with scoreboard, use blending equation: in_color*in-alpha+mem_color*(1-in_alpha). Since alpha either zero or full in case of 16 bit RGBA, you just write incoming pixel if its alpha us not zero. If it does not help with garbage in koopas images then need to find the origin of that garbage. Probably alpha testing is used to draw them. And again, if you need to make zero the first bit of a int number, it can be done easier way.

purplemarshmallow:
>However, let not forget that hardware rendering is imperfect by design

it's imperfect but it can still be very accurate. Other emulators already proved that

>and we choose it to get better picture than original hardware offers.

I think the main advantage of hardware rendering is speed, not the better picture. 

>If it does not help with garbage in koopas images then need to find the origin of that garbage

So you think setting alpha of the current pixel to 0 is the wrong solution?

gonetz:
>I think the main advantage of hardware rendering is speed, not the better picture.

Very rare opinion :) It explains a lot though

>So you think setting alpha of the current pixel to 0 is the wrong solution?

Yes, it is wrong. Let's consider scoreboard example. The game loaded scoreboard texture image. It is correct texture image. When you copy aux buffer data, you make all texels in that texture transparent. That is, you break the correct original data. If that texture will be rendered with blending or with alpha test, it will be invisible. If you see it after the copy, that means that it is rendered without blending. We are lucky there, but it will not work always.

Actually, the answer to the question "which alpha should have pixels in a frame buffer" is not that simple. When N64 writes pixel to buffer it calculates alpha as a function of pixel's coverage. We can ignore that fact for now because we don't have coverage information anyway. Then there are two situations:
* whole buffer content was updated after allocation. E.g. koopas. In that case, when you copy that buffer to RDRAM, you do it almost the same as for main buffer. Just need to set correct pixel's alpha: alpha = (c.a != 0) ? 1 : 0;
It is *normal* case, " that's how real hardware works".
* only part of the buffer was modified. Example: scoreboard's numbers. If you did not copy scoreboard background to the aux buffer when it was allocated (the hack is disabled) then in that case aux frame buffer contains only part of necessary information. If you will copy whole the buffer to RDRAM, you will erase the scoreboard data. Thus, you need to apply special code (**a hack**!) to avoid it. Namely, force clear buffer with zero on allocation and copy only non-zero pixels to RDRAM.

As an alternative, you may **always** load data from RDRAM to the buffer on buffer allocation. In that case you will have complete buffer data and it will be safe to write it back to RDRAM using normal method. Note that buffer initialization with RDRAM data also will make HWFBE scoreboard working correctly without scoreboard hack. I even tried to do it when I worked on MT emulation, but it caused strange glitches, so I removed it and made the hack. We can try to return to this solution and polish it.

purplemarshmallow:
>When N64 writes pixel to buffer it calculates alpha as a function of pixel's coverage. We can ignore that fact for now because we don't have coverage information anyway.

:/

>As an alternative, you may always load data from RDRAM to the buffer on buffer allocation.

And this can work correctly even if we don't have the pixel coverage information?

gonetz:
Yes, it will work because it works now. As I remember, coverage info is used only for 16bit rgba buffers. More details can be found in Angrylion's  soft plugin. Alpha calculated as function of coverage is the same as pixel's alpha in most cases. I met only one case where it was not so. That it we should know and remember about that problem, but not overestimate its importance.

purplemarshmallow:
>Yes, it will work because it works now

I understand. The numbers must be rendered inside the scoreboard texture. And then the whole texture needs to be upadted

>I met only one case where it was not so

Pokemon Stadium 2 credits

purplemarshmallow:
One problem is CopyToRDRAM does not support 8 bit buffers. That's why it just crashes in many games

AmbientMalice:
8bit buffers are used for shadows, correct? I mention this because all the hardware plugins suffer shadows that are hard-edged and black when they should be opaque with soft edges.

purplemarshmallow:
8bit auxiliary buffers are often used as shadows. But the background in Yoshi's Story also uses 8 bit auxiliary buffers.

gonetz:
> The numbers must be rendered inside the scoreboard texture. And then the whole texture needs to be updated

Correct.

> Pokemon Stadium 2 credits

Good memory. I tried to recall it myself, but without success.

> One problem is CopyToRDRAM does not support 8 bit buffers. That's why it just crashes in many games

CopyToRDRAM designed for main color buffers, which are displayed on screen. 8bit buffers not used for that. If you want to copy 8bit buffers to RDRAM, you need to remember that GL texture for 8bit buffer is also 8bit (for GLES2 it can be 16bit). Thus, you need to use glReadPixels with proper format and destination buffer.

purplemarshmallow:
>If you want to copy 8bit buffers to RDRAM, you need to remember that GL texture for 8bit buffer is also 8bit 

For 16 bit buffers colors are converted. For 8 bit buffers I don't need that? GL texture and N64 texture have the same format?

>Thus, you need to use glReadPixels with proper format and destination buffer.

GL_RGB? What needs to be different about the destination buffer?

purplemarshmallow:
Here's an update on the current state of this feature:
* no game freezes anymore if it's enabled (Conker, Jet Force Gemini, Mario Tennis all fixed)

* Mario Tennis: Menu wipe transition effects are working now (If I remove VI width calculation). They are incorrect with hardware framebuffer. Shadows are missing (8 bit buffers not supported)

* Pokemon Stadium games: all fb effects are working (also #699 #723 ). No more validity checking problems. Some problems with wrong alpha. Fixing them may require pixel coverage emulation unfortunately.

* Knockout Kings: one portrait shows up (#350 )

purplemarshmallow:
Top gear hyper bike looks so much better with auxiliary to RDRAM. Completing it and adding 8 bit support is really important I think. Games like Paper Mario and Yoshi's Story won't require hacks anymore. And who knows what other benefits it has.

theboy181:
Please upload for testing. :)

purplemarshmallow:
mupen64plus version
https://drive.google.com/file/d/0B7Y6r4SpC_QQT3ZYMGJLSUpTNG8/view?usp=sharing

theboy181:
That's just cruel. No zilmar spec version?

purplemarshmallow:
I rather finish this and get it merged before dealing with Qt again
new PR: #781

gonetz:
Can we close this one?

Issue #609, opened 2015-07-07
Perfect Dark flickering lights


Filipianosol:
Since Public Release 1.2/one of the July 3rd commits the lights started to shimmer uncontrollably like so:
![ezgif-3842154586](https://cloud.githubusercontent.com/assets/12639838/8552781/8a2d6dae-24df-11e5-84e9-d28555e85b3e.gif)
Compared to:
![ezgif-2885995329](https://cloud.githubusercontent.com/assets/12639838/8552803/ae23855e-24df-11e5-93e6-1759d214c3dc.gif)
Before it used to be negligible since the shimmer was rare and only showed upon certain angles/lights like the blue light in the beginning:
![ezgif-1638080750](https://cloud.githubusercontent.com/assets/12639838/8554147/4fda79ea-24e8-11e5-98ac-54107deb3cf1.gif)
Compared to:
![ezgif-109060030](https://cloud.githubusercontent.com/assets/12639838/8553869/a383b91e-24e6-11e5-85b4-7244b6938308.gif)
Messing with frame buffer settings didn't seem to help.
Save state: https://drive.google.com/file/d/0Bz3WFRZJ1ZbRNUtyMzdqcVh3OUU/view?usp=sharing

baptiste0602:
I confirm

gonetz:
Fixed

PR #774, opened 2015-10-23
add settings for Pokemon Snap

purplemarshmallow:
hack_VIUpdateOnCIChange to fix depth problems and flickering
copyToRDRAM=1 (fb sync) for correct pictures

### DIFF
--- a/ini/GLideN64.custom.ini
+++ b/ini/GLideN64.custom.ini
@@ -299,6 +299,10 @@ frameBufferEmulation\copyDepthToRDRAM=1
 Good_Name=Pokemon Puzzle League (E)(F)(G)(U)
 frameBufferEmulation\detectCFB=1
 
+[POKEMON%20SNAP]
+Good_Name=Pokemon Snap (U)
+frameBufferEmulation\copyToRDRAM=1
+
 [POKEMON%20STADIUM%202]
 Good_Name=Pokemon Stadium 2 (E)(F)(G)(I)(J)(S)(U)
 frameBufferEmulation\copyToRDRAM=0
--- a/src/RSP.cpp
+++ b/src/RSP.cpp
@@ -337,7 +337,8 @@ void RSP_Init()
 	else if (strstr(RSP.romname, (const char *)"QUAKE II") != NULL ||
 		strstr(RSP.romname, (const char *)"Quake") != NULL ||
 		strstr(RSP.romname, (const char *)"Perfect Dark") ||
-		strstr(RSP.romname, (const char *)"PERFECT DARK")
+		strstr(RSP.romname, (const char *)"PERFECT DARK") ||
+		strstr(RSP.romname, (const char *)"POKEMON SNAP")
 		)
 		config.generalEmulation.hacks |= hack_VIUpdateOnCIChange;
 	else if (strstr(RSP.romname, (const char *)"MASK") != NULL) // Zelda MM


purplemarshmallow:
@gonetz what do you think about screenupdate on CI change? ```if (gDP.colorImage.changed != 0) {```

screenupdate on VI origin change ```*REG.VI_ORIGIN != VI.lastOrigin``` is very bad. It does not work if double buffering is not used (Quake). It does not work properly in Pokemon Snap and also Taz Express there are strange geometry glitches.

screenupdate on  VI origin change && CI change is also bad it creates performance problems on Android.

screenupdate on CI change seems to be ideal. The screen is only updated once per frame and still every game works. No more need for hack_VIUpdateOnCIChange

gonetz:
> screenupdate on VI origin change *REG.VI_ORIGIN != VI.lastOrigin is very bad.

Three problem game. Not enough for "very bad" score.

So, you suggest swap buffer on gDP.colorImage.changed != 0 instead of *REG.VI_ORIGIN != VI.lastOrigin? Will it work correct for games, which use 2 dlists per frame? Paper Mario sewers by example? As I remember, Paper Mario does not flicker only when buffers swap on origin change.

purplemarshmallow:
>So, you suggest swap buffer on gDP.colorImage.changed != 0 instead of *REG.VI_ORIGIN != VI.lastOrigin?

Yes

>Will it work correct for games, which use 2 dlists per frame? 

Glover 2 has multiple dlists per frame. Quake II also. Both games look fine

>Paper Mario sewers by example? As I remember, Paper Mario does not flicker only when buffers swap on origin change.

I get no flickering but I haven't tested the sewers
However I noticed there is indeed a problem. During the menu transition effects in Beetle Adventure Racing the color image does not change. The effect is missing.

>Three problem game. Not enough for "very bad" score.

5 known games. Perfect Dark, Quake, Quake II, Taz Express, Pokemon Snap

LegendOfDragoon:
>Three problem game. Not enough for "very bad" score.

Agreed. I haven't tested Pokemon Snap much or Taz Express at all, but I've tested quite a few games. I basically experimented with VI emulation in Angrylion's plugin a while back, to see if I can speed it up. I ended up making it update only if VI_ORIGIN changed. This alone worked fine with many games. I only ran into like 2 problems. Certain demos like Absolute Crap Intro 2 didn't display anything, since VI_ORIGIN never changes. I noticed it was interlaced so I just added a check to see if it's interlaced, since I assume interlacing means the screen should change every frame. The other issue was the Nintendo logo in DK64. I fixed that by forcing screen to update on the next call to UpdateScreen, after ViStatusChanged. It might not be a proper solution, but it's quick and easy.

@purplemarshmallow can you elaborate the issue you encountered with Quake II? I just re-tested that game and it does use double buffering, if I'm not mistaken. Also I'd like to test out Perfect Dark and Pokemon Snap. What is the problem with these games?

purplemarshmallow:
@LegendOfDragoon the problem here is this if I understand it right: There are 60 VI/s (50 PAL). But almost no game runs at 60 fps. We want to update the screen only once per frame, not once per VI. Whenever a new frame is rendered we need to update.

>I ended up making it update only if VI_ORIGIN changed

VI_ORIGIN will not change if a game uses single buffering that's the problem.

>Certain demos like Absolute Crap Intro 2 didn't display anything, since VI_ORIGIN never changes. I noticed it was interlaced so I just added a check to see if it's interlaced, since I assume interlacing means the screen should change every frame. 

I think Absolute Crap Intro 2 is rendered by the CPU. The problem here is likely single buffering. I don't think interlacing means you must update every frame. I think interlaced games can run below 60 fps

>can you elaborate the issue you encountered with Quake II?

underwater it uses single buffering

>Also I'd like to test out Perfect Dark and Pokemon Snap. What is the problem with these games?

In Perfect Dark lights are flickering #609. Pokemon Snap has strange geometry glitches

LegendOfDragoon:
>I think Absolute Crap Intro 2 is rendered by the CPU. The problem here is likely single buffering. I don't think interlacing means you must always update. I think interlaced games can run below 60 fps

I know that just because it's interlaced doesn't necessarily mean it's running at 60 fps. However, so far, I haven't seen a case where it wasn't 60 fps, so it's really just a temporary quick fix for me, since I actually run that demo to benchmark VI emulation. I'll try testing more games out today. 

Does the Nintendo logo appear with this plugin, in DK64? Last time I tried, it didn't.

>underwater it uses single buffering
In Perfect Dark lights are flickering #609. Pokemon Snap has strange geometry glitches

That is very interesting! Thanks, I will certainly check it out asap.

purplemarshmallow:
>Does the Nintendo logo appear with this plugin, in DK64?

It's correct in the (U) version but in the (E) version you only see pixel garbage. I think it's a different problem

LegendOfDragoon:
>It's correct in the (U) version but in the (E) version you only see pixel garbage. I think it's a different problem

I see. Anyway, I started testing Perfect Dark and can confirm the issue you brought up. I'll try and see if I can find a good solution.

gonetz:
I tested Paper Mario sewers. No problems with 'gDP.colorImage.changed' condition.
So, it can be used. Since it causes problem with BAR, it can't replace 'origin change' condition.

If we would follow "no-hacks" strategy, we would need no tests for buffer change.
No-hack solution is to copy buffer with *REG.VI_ORIGIN address to main buffer and swap buffers on each VI_UpdateScreen() call. It must not cause any problems for desktop version. Mobile version may get slower. In that case we can make swap condition setting: 1) on VI_UpdateScreen 2) on VI origin change 3) On CI change.

purplemarshmallow:
>Since it causes problem with BAR, it can't replace 'origin change' condition

Maybe 'origin change' condition can be replaced by something different? Something that indicates that there are VI manipulations going on?

LegendOfDragoon:
@purplemarshmallow , I did further testing in perfect dark with angrylion's. It turns out that just relying on origin change condition was actually good enough. I set a breakpoint to see if the screen ever changes while the origin value didn't and I haven't seen it happen.

Anyway, I'll test out Quake II next I guess.

gonetz:
>Maybe 'origin change' condition can be replaced by something different? Something that indicates that there are VI manipulations going on?

Sounds hackish and unreliable.

purplemarshmallow:
>It turns out that just relying on origin change condition was actually good enough.

Then there must be a different problem. The flickering never stops it only gets less noticeable. 

>No-hack solution is to copy buffer with *REG.VI_ORIGIN address to main buffer and swap buffers on each VI_UpdateScreen() call. It must not cause any problems for desktop version.

Not sure, I want to compare performance. Would be better to update once per frame.

gonetz:
> Not sure, I want to compare performance. Would be better to update once per frame.

N64 video Interface updates picture on screen 50/60 times per second, no matter was the buffer updated or not. VI emulation in the plugin is  incomplete and hackish. "update once per frame" is a hack for performance.

purplemarshmallow:
The desktop version can handle it so it's possibly better to update 50/60 times like real hardware

gonetz:
I can do it in a feature branch.

gonetz:
Approved:
https://github.com/gonetz/GLideN64/commit/65dda8058e62cd458dbe49cbbe258939e2efa888

Sorry for delay with code review.

Issue #650, opened 2015-08-07
Body Harvest: incorrect/missing framebuffer effects: inventory/map


olivieryuyu:
This game has errors or completely missed framebuffer effects in game

For instance the inventory:

gliden64

![sdfdf](https://cloud.githubusercontent.com/assets/8987490/9143549/6c2ac01a-3d47-11e5-88d8-b66f91ce3846.jpg)

angrylion

![dsfsdffdfs](https://cloud.githubusercontent.com/assets/8987490/9143554/76347cb8-3d47-11e5-9e11-293120c8e9e1.jpg)

For instance the map:

gliden64:

![sdfsdffffffd](https://cloud.githubusercontent.com/assets/8987490/9143564/7f5b88e0-3d47-11e5-8833-be3504494b65.jpg)

angrylion

![map](https://cloud.githubusercontent.com/assets/8987490/9143572/8936f8b8-3d47-11e5-950e-51ccc39c2ad4.jpg)

olivieryuyu:
it works with Glide64!!

purplemarshmallow:
same problem as #529
VI.width calculation is wrong. If you remove one line from VI.cpp this is fixed:
VI.width = min((u32)floorf((hEnd - hStart)*xScale + 0.5f), *REG.VI_WIDTH);
but this might have side effects in other games. This line also causes a crash in Mario Tennis if "render frame buffer as texture" is enabled

AmbientMalice:
>but this might have side effects in other games. This line also causes a crash in Mario Tennis if "render frame buffer as texture" is enabled

Are you aware of the "fix" breaking anything?

purplemarshmallow:
Glide64 does not calculate VI.width, it uses REG.VI_WIDTH. GLideN64 calculates it but if I remove this line it uses REG.VI_WIDTH, like Glide64. I think REG.VI_WIDTH can be wrong if there's a bug in the emulator or in certain games, but I don't know a specific case. If VI.width is too high and copy frame buffer to RDRAM is enabled the emulator will crash. I think min is used here to make copy frame buffer to RDRAM more safe.

There are two ways to fix this:
Calculate VI.width more accurate or do not calculate it at all.

olivieryuyu:
angrylion plugin does the correct calculation for VI width, so it can be made more accurate. That should be the goal anyhow.

purplemarshmallow:
>angrylion plugin does the correct calculation for VI width

[...]

PR #780, opened 2015-10-30
remove VI width calculation

purplemarshmallow:
I see no reason why this is needed. REG.VI_WIDTH always has the correct
width. This fixes the pole position minigame in namco museum and the gb
tower in Pokemon Stadium 2 #529. This fixes transition effects in Body
Harvest  #650. This fixes problems with the copy auxiliary option in
Body Harvest and Mario Tennis

### DIFF
--- a/src/VI.cpp
+++ b/src/VI.cpp
@@ -46,8 +46,8 @@ void VI_UpdateSize()
 		}
 		if (VI.real_height % 2 == 1)
 			--VI.real_height;
-	} else if (hEnd != 0 && *REG.VI_WIDTH != 0)
-		VI.width = min((u32)floorf((hEnd - hStart)*xScale + 0.5f), *REG.VI_WIDTH);
+	} //else if (hEnd != 0 && *REG.VI_WIDTH != 0)
+		//VI.width = min((u32)floorf((hEnd - hStart)*xScale + 0.5f), *REG.VI_WIDTH);
 
 	VI.PAL = (*REG.VI_V_SYNC & 0x3ff) > 550;
 	if (VI.PAL && (vEnd - vStart) > 478) {


AmbientMalice:
Will this also fix FB to RAM crashes?

AmbientMalice:
Will this also fix FB to texture crashes? (Usually triggered when the resolution changes. Menu transitions, high resolution mode, etc.)

purplemarshmallow:
>Will this also fix FB to texture crashes?

I don't think so FB to texture seems to work correctly even if VI width is wrong

>Usually triggered when the resolution changes. Menu transitions, high resolution mode, etc.

None of these should cause a crash

AmbientMalice:
Regardless, FB to texture crashes currently. Turning on Perfect Dark's high resolution mode, which in NTSC widens the horizontal FB resolution from 320 to 640, immediately crashes the emulator if FB to texture is enabled.

purplemarshmallow:
>Feature has heaps of problems including flickering and missing graphics, but it at least doesn't drag the emulator down with it.

You mean copy from rdram has problems? That's true.

gonetz:
>> I see no reason why this is needed. REG.VI_WIDTH always has the correct
width.

It is needed because REG.VI_WIDTH is not always correct. I can't quickly recall where use of REG.VI_WIDTH causes problems, but I remember for sure that I not use it for the reason.

purplemarshmallow:
I don't know a case where REG.VI_WIDTH is incorrect. I know many cases where the current formula is incorrect. I can name them: Mario Tennis (menus), Body Harvest (ingame), Pokemon Stadium 2 (gb tower), Namco Museum (pole position minigame)

gonetz:
Well, VI emulation code was modified many times. If current code causes so many issues, it definitely need to be corrected. Still, I'm not sure that use of  REG.VI_WIDTH is the right solution. I need time to find, how I came to the current solution. I also may accept this change and wait for regression reports :) No regressions - your version is correct. But I don't like to act blindly. I want to know, what is going on there. Due to my current load it will take time.

gonetz:
You're right, current VI.width calculation is incorrect. I tracked it down: (hEnd - hStart)*xScale inherited from the original glN64 code. I modified that code several times, but the method itself remain the same. 

I did not find any problem with VI.width = *REG.VI_WIDTH, so merged.
Thanks for info about the problem!

Issue #746, opened 2015-10-05
Extreme-G wrong framebuffer effect


purplemarshmallow:
after selecting a bike
![gliden64_extremeg_000](https://cloud.githubusercontent.com/assets/7283660/10271158/588356b0-6b07-11e5-84a0-e187eb4c58e4.jpg)
z64gl looks more correct but still wrong
![1](https://cloud.githubusercontent.com/assets/7283660/10271159/5c39b7cc-6b07-11e5-986f-3bdaff2db2f6.jpg)

purplemarshmallow:
Fixed in  an experimental branch. Framebuffer to RDRAM
![gliden64_extremeg_002](https://cloud.githubusercontent.com/assets/7283660/10674726/cb9928c0-78fc-11e5-8641-3c477ad5011c.jpg)

AmbientMalice:
What was the difference between the framebuffer branch that fixed this, and the one that got merged? As i understand it, that one was auxiliary, this is primary framebuffer? But isn't the primary framebuffer already sent to RDRAM?

purplemarshmallow:
The branch that showed this image copied everything and this was really slow. Also it did not always work correctly. I had to switch window/fullsceen to get rid of hardware framebuffers

You can try removing this line
https://github.com/gonetz/GLideN64/blob/master/src/FrameBuffer.cpp#L525

And this line
https://github.com/gonetz/GLideN64/blob/master/src/FrameBuffer.cpp#L527

>But isn't the primary framebuffer already sent to RDRAM?

Only if it's address is gDP.colorImage.address. Then the copy color option will copy it. But for example the pause menu background in OOT also does not get copied. The plugin always uses hardware framebuffer textures in this case but in ExtremeG the hardware framebuffer code does not work correctly

gonetz:
Seems to be fixed. Please test

AmbientMalice:
Seems fixed.

PR #781, opened 2015-10-31
implement framebuffer to RDRAM

purplemarshmallow:
new, clean PR
accurate commits

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -49,6 +49,7 @@ void Config::resetToDefaults()
 #endif
 	frameBufferEmulation.copyDepthToRDRAM = ctDisable;
 	frameBufferEmulation.copyFromRDRAM = 0;
+	frameBufferEmulation.copyAuxToRDRAM = 0;
 	frameBufferEmulation.copyToRDRAM = ctAsync;
 	frameBufferEmulation.detectCFB = 0;
 	frameBufferEmulation.N64DepthCompare = 0;
--- a/src/Config.h
+++ b/src/Config.h
@@ -71,6 +71,7 @@ struct Config
 
 	struct {
 		u32 enable;
+		u32 copyAuxToRDRAM;
 		u32 copyToRDRAM;
 		u32 copyDepthToRDRAM;
 		u32 copyFromRDRAM;
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -269,7 +269,7 @@ void FrameBuffer::copyRdram()
 	const u32 dataSize = stride * height;
 
 	// Auxiliary frame buffer
-	if (m_width != VI.width) {
+	if (m_width != VI.width && config.frameBufferEmulation.copyAuxToRDRAM == 0) {
 		// Write small amount of data to the start of the buffer.
 		// This is necessary for auxilary buffers: game can restore content of RDRAM when buffer is not needed anymore
 		// Thus content of RDRAM on moment of buffer creation will be the same as when buffer becomes obsolete.
@@ -537,6 +537,29 @@ void FrameBufferList::saveBuffer(u32 _address, u16 _format, u16 _size, u16 _widt
 	m_pCurrent->m_isDepthBuffer = _address == gDP.depthImageAddress;
 	m_pCurrent->m_isPauseScreen = m_pCurrent->m_isOBScreen = false;
 	m_pCurrent->m_postProcessed = false;
+}
+
+void FrameBufferList::copyAux()
+{
+	for (FrameBuffers::iterator iter = m_list.begin(); iter != m_list.end(); ++iter) {
+		if (iter->m_width != VI.width && iter->m_height != VI.height)
+			FrameBuffer_CopyToRDRAM(iter->m_startAddress);
+	}
+}
+
+void FrameBufferList::removeAux()
+{
+	for (FrameBuffers::iterator iter = m_list.begin(); iter != m_list.end(); ++iter) {
+		while (iter->m_width != VI.width && iter->m_height != VI.height) {
+			if (&(*iter) == m_pCurrent) {
+				m_pCurrent = NULL;
+				glBindFramebuffer(GL_DRAW_FRAMEBUFFER, 0);
+			}
+			iter = m_list.erase(iter);
+			if (iter == m_list.end())
+				return;
+		}
+	}
 }
 
 void FrameBufferList::removeBuffer(u32 _address )
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -23,15 +23,15 @@ class FrameBufferToRDRAM
 {
 public:
 	FrameBufferToRDRAM() :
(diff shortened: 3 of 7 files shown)


purplemarshmallow:
Resident evil 2 now works (LLE + copy auxiliary option)

AmbientMalice:
Yea, I was about to mention RE2. I've been doing some testing. Game has some deeper issues, I think, Aux to RDRAM makes the game display about as much graphics as completely disabling FB, but with less glitches. None of the "background" layer textures render - splash screens, background layers of menus, etc.

I notice that in LLE mode, the spinning N64 logo renders. HLE renders nothing. But LLE mode seems really unstable in mupen for this game.

purplemarshmallow:
Yes RE2 still needs work.

>I notice #746 seems to have come back?

It's not an auxiliary buffer it can not work with this option. Copying main frame buffers needs to be implemented. But I can't just remove the buffers because they can be displayed on screen. You will see nothing. One solution I can think of is to just copy the buffers without removing them. This should help to get rid of some hacks like hack_skipVIChangeCheck or hack_blurPauseScreen. But it won't help ExtremeG.

>Also, I notice the performance overhead is pretty atrocious for some games. Mupen64plus don't have proper fps/vi/cpu usage meters, but games like Mario Tennis stutter constantly. I guess that's probably to be expected, but it will severely limit the feature's appeal if it causes unplayable performance drops in the games it most helps

Only games that update buffers once per frame are problematic like Mario Tennis. But I think it's slower than it should be. There are not so many pixels to read. Maybe it can be optimized. The rendering pipeline is stalled multiple times maybe that's a problem.

gonetz:
Please correct the commit so we can move forward.

purplemarshmallow:
I have to replace
```PBOBinder binder(GL_PIXEL_PACK_BUFFER, m_PBO[nextIndex]); ```
 with 
```glBindBuffer(GL_PIXEL_PACK_BUFFER, m_PBO[nextIndex]); ``` 
to get async buffer reads working. Not sure why. Is this a problem?

**Comment by gonetz** [comment]
If you don't know why, it is the problem.
May be you allocated the binder inside some 'if' clause, and it get destroyed after the 'if'?
PBOBinder releases bound buffer in its destructor.

**Comment by purplemarshmallow** [comment]
Yes it's inside an if
if (!_sync)
I've corrected my commits. Another problem left: glBlitFramebuffer is redundant if scale is 1 but I don't know how to get rid of it: https://github.com/purplemarshmallow/GLideN64/commit/57729b4303a9ffc79fa05bf6b4bdbedf35f46b67

**Comment by purplemarshmallow** [comment]
@gonetz 
when I removed my hack for Mario Tennis scoreboards ```if (c.a)``` it caused many regressions. not only Vigilante8 started to crash now but also Knockout Kings. I think games can allocate buffers but render nothing inside. Data must stay 100% intact. I don't think buffer allocation with RDRAM content is a good way to fix this problem. It's hard to stay 100% correct and important game data can be overwritten.

I can think of a different solution. At buffer creation fill buffers with a test color. And later don't copy pixels to RDRAM if they still have the test color

**Comment by gonetz** [comment]
>glBlitFramebuffer is redundant if scale is 1 but I don't know how to get rid of it: 

Something like this:
```
 	if (config.video.multisampling != 0) {
 		pBuffer->resolveMultisampledTexture();
 		glBindFramebuffer(GL_READ_FRAMEBUFFER, pBuffer->m_resolveFBO);
 	} else
 		glBindFramebuffer(GL_READ_FRAMEBUFFER, pBuffer->m_FBO);
	if (pBuffer->m_scaleX != 1) {
	 	glBindFramebuffer(GL_DRAW_FRAMEBUFFER, m_FBO);
 		glScissor(0, 0, pBuffer->m_pTexture->realWidth, pBuffer->m_pTexture->realHeight);
		glBlitFramebuffer(
			0, 0, video().getWidth(), video().getHeight(),
			0, 0, VI.width, VI.height,
			GL_COLOR_BUFFER_BIT, GL_NEAREST
		);
	 	glBindFramebuffer(GL_DRAW_FRAMEBUFFER, frameBufferList().getCurrent()->m_FBO);
	 	glBindFramebuffer(GL_READ_FRAMEBUFFER, m_FBO);
	}
	
	#ifndef GLES2
	...
```

That is use pBuffer->m_FBO as input buffer

gonetz:
>I think games can allocate buffers but render nothing inside. Data must stay 100% intact.

Don't you think that it is sounds crazy?

I think, that the problem is in incorrect aux buffer's height. If height is wrong, write to RDRAM may alter program data. Change in one bit of program data can have fatal consequences. 

>I don't think buffer allocation with RDRAM content is a good way to fix this problem. It's hard to stay 100% correct and important game data can be overwritten.

Yep, data may change when converted to buffer and read back. If we replace program data with altered data, program will crash.

>I can think of a different solution. At buffer creation fill buffers with a test color. And later don't copy pixels to RDRAM if they still have the test color

Since we don't have robust way to detect aux buffer height, I have to agree with you there. Ok, let's  zero pixel (0, 0, 0, 0)  be the test color.

purplemarshmallow:
>I think, that the problem is in incorrect aux buffer's height.

Yes most likely that's the real problem.

>Since we don't have robust way to detect aux buffer height, I have to agree with you there. Ok, let's zero pixel (0, 0, 0, 0) be the test color.

I think pixels in a buffer are already always 0 by default. I tested it and added ```if (c.r||c.b||c.g||c.a)``` The old garbage gfx problem with the Koopas in Mario Tennis returned. I think zero is a bad test color. It should be a color not used by any game. Maybe a 32 bit color to make sure

gonetz:
> I tested it and added if (c.r||c.b||c.g||c.a)

OMG! if (c.raw != 0)

>The old garbage gfx problem with the Koopas in Mario Tennis returned.

It is not garbage. Garbage looks like a random noise. Background color in Koopas on your screen shot is solid white. Looks like a part of other buffer, which is filled with white color, e.g. shadows.

>I think zero is a bad test color. It should be a color not used by any game.

How will you find a color, which not used by any game?

> Maybe a 32 bit color to make sure

We use 32bit rendering, so all pixels in the buffers are 32bit.


Well, zero is a bad test color, because it is often used as clear color and thus must not be discarded.
Try to pick other test color. Clear allocated buffer with that test color before use.

purplemarshmallow:
>Background color in Koopas on your screen shot is solid white. Looks like a part of other buffer, which is filled with white color, e.g. shadows.

The bug with the white bar at the top is fixed. I think https://github.com/purplemarshmallow/GLideN64/commit/57729b4303a9ffc79fa05bf6b4bdbedf35f46b67 fixed it. Only the garbage gfx problem remains. 0 is a bad test color. It looks like this now:

purplemarshmallow:
@gonetz I've corrected my commit https://github.com/purplemarshmallow/GLideN64/commit/eec2c8d70acabc3f9fb73f6d26dea8a3647e03be
Can you merge this? Every commit should be correct now. I think work can continue in master

gonetz:
Looks good, thanks!
Merged:
https://github.com/gonetz/GLideN64/commit/142ad0bc3ca986735791ec6d89247227a6a29034

PR #784, opened 2015-11-02
Added an option to force gamma correction level

Gillou68310:


AmbientMalice:
Is this going to be added to the Zilmar version down the track? If so, where would gamma override be located?

gonetz:
>Is this going to be added to the Zilmar version down the track?

I doubt that it is necessary. If somebody will find it useful, it can be added to GUI also.

Gillou68310:
I thinks @gonetz is right this option has been initialy added to fix a limitation on android where you can't change gamma level via the OS.

gonetz:
Looks good. Merged. Thanks!

Issue #562, opened 2015-05-28
Bio F.R.E.A.K.S. intro transitions


purplemarshmallow:
There are CPU based effects, the plugin needs to switch to CPU rendering mode, detectCFB is needed. The problem is that some effects last for under a secound, the plugin currently switches to CPU rendering mode after 25 frames, that's too slow. The effect is already over. This value needs to be different for every game, it should be possible to set it in the ini file.

Same problem in Lego racers #362 
In this game the value needs to be higher to avoid unsharp image, Glide64 uses a built-in hack. But I think it's better to introduce a new setting in the ini file since every game needs a different value. It's impossible to find one value that works for every game

AmbientMalice:
Reminds me of this - https://github.com/project64/project64/issues/72

It might be time to re-examine better methods for handling CPU framebuffer writes. (Not to mention FB in general, because GLideN64 is hitting some nasty snags with eccentric games.)

purplemarshmallow:
with detectCFB this game will never be perfect, if the value is too low you see the unsharp image ingame, it's only possible to find a compromise. with proper notification from the emlator it should work.

I'm not sure about lego racers anymore. I thought it's needed for the map but the map works without it. Someone must have enabled it for a reason and it's enabled in Glide64 as well but I can't find a point in the game where it's needed

purplemarshmallow:
@gonetz do you think it's possible to implement a better method to detect CPU writes? the plugin could save a copy of the data in the colorbuffer. And later it could compare with this copy and if there's any difference the plugin knows that the CPU wrote something there. Do you think this can work?

gonetz:
It may work if the CPU will write to previously allocated color buffer. It's not necessary the case. But it may be used as additional check, which may trigger faster than my 25 frames barrier.

purplemarshmallow:
fixed

Issue #362, opened 2015-04-09
LEGO Racers (All) - Sharp To Unsharp


Predator82Germany:
Happens in every menü screen til ingame

purplemarshmallow:
fixed

olivieryuyu:
fixed please close

PR #788, opened 2015-11-10
remove detectCFB hack

purplemarshmallow:
if there is no buffer RDRAM content is displayed. If there's an invalid
buffer the buffer is removed and then RDRAM content is displayed. No
game needs the detectCFB hack anymore. This fixes the transitions in BIO
F.R.E.A.K.S #562 and also #362

### DIFF
--- a/src/VI.cpp
+++ b/src/VI.cpp
@@ -86,8 +86,6 @@ void VI_UpdateSize()
 
 void VI_UpdateScreen()
 {
-	static u32 uNumCurFrameIsShown = 0;
-
 	if (VI.lastOrigin == -1) // Workaround for Mupen64Plus issue with initialization
 		isGLError();
 
@@ -109,12 +107,21 @@ void VI_UpdateScreen()
 	}
 
 	if (config.frameBufferEmulation.enable) {
+
+		FrameBuffer * pBuffer = frameBufferList().findBuffer(*REG.VI_ORIGIN);
+		if (pBuffer == NULL)
+			gDP.changed |= CHANGED_CPU_FB_WRITE;
+		else if (!pBuffer->isValid()) {
+			gDP.changed |= CHANGED_CPU_FB_WRITE;
+			if (config.frameBufferEmulation.copyToRDRAM == 0)
+				pBuffer->copyRdram();
+		}
+
 		const bool bCFB = (gDP.changed&CHANGED_CPU_FB_WRITE) == CHANGED_CPU_FB_WRITE;
 		const bool bNeedUpdate = (bCFB ? true : (*REG.VI_ORIGIN != VI.lastOrigin)) || ((config.generalEmulation.hacks & hack_VIUpdateOnCIChange) != 0 && gDP.colorImage.changed != 0);
 
 		if (bNeedUpdate) {
 			if (bCFB) {
-				FrameBuffer * pBuffer = frameBufferList().findBuffer(*REG.VI_ORIGIN);
 				if (pBuffer == NULL || pBuffer->m_width != VI.width) {
 					if (!bVIUpdated) {
 						VI_UpdateSize();
@@ -134,25 +141,13 @@ void VI_UpdateScreen()
 				FrameBuffer_CopyFromRDRAM(*REG.VI_ORIGIN, config.frameBufferEmulation.copyFromRDRAM && !bCFB);
 			}
 			frameBufferList().renderBuffer(*REG.VI_ORIGIN);
-
-			if (gDP.colorImage.changed)
-				uNumCurFrameIsShown = 0;
-			else if (config.frameBufferEmulation.detectCFB != 0) {
-				uNumCurFrameIsShown++;
-				if (uNumCurFrameIsShown > 25)
-					gDP.changed |= CHANGED_CPU_FB_WRITE;
-			}
 			frameBufferList().clearBuffersChanged();
 			VI.lastOrigin = *REG.VI_ORIGIN;
 #ifdef DEBUG
 			while (Debug.paused && !Debug.step);
 			Debug.step = FALSE;
 #endif
-		} else if (config.frameBufferEmulation.detectCFB != 0) {
-			uNumCurFrameIsShown++;
-			if (uNumCurFrameIsShown > 25)
-				gDP.changed |= CHANGED_CPU_FB_WRITE;
-		}
+		} 
 	}
 	else {
 		if (gDP.changed & CHANGED_COLORBUFFER) {


AmbientMalice:
Maybe I'm doing something wrong, but this doesn't seem to work.

purplemarshmallow:
Can you be more precise? To see the transitions in BIO F.R.E.A.K.S you have to set copy color to sync

AmbientMalice:
Okay, so using sync mode has restored the proper functionality. DK64 renders its start screen correctly, that sorta thing. But there's a catch. 

Quake II's loadscreen no longer renders the load bar.
Vigilante 8 is behaving weirdly in the menus.

purplemarshmallow:
In Quake II the area modified by the CPU is very small only the load bar. It's not working because of the threshold level. I have an Idea how it can be fixed. In Vigilant8 there's something weird but I think it's a mupen64plus core issue. I should test with PJ64 again.

olivieryuyu:
it works with RE2 movies, hexen, tonic trouble,etc ... ?

purplemarshmallow:
Sure and it works much better. There's no 25 frame delay

gonetz:
It works. Great idea!
I will wait for Quake II's loadscreen fix.
BTW, current threshold level is not result of experiments. It guaranties that buffer definitely changed, but it probably can be lowered.

theboy181:
can hardly wait for the next release!!

@gonetz  I wrote on the forum.

purplemarshmallow:
I made a fix for conker. It's war cutscene is now correct. But in the Uga Buga chapter cpu write is now triggered as well. But I think it's better than seeing this guys penis.


>I will wait for Quake II's loadscreen fix.

@gonetz It won't be a quick fix. I think you can merge it. My idea is to generate a map when comparing buffers and then copy from RDRAM according to this map. It will take time and testing to get it working. Not sure if I can get it working at all. But it should work perfectly for all games in theory. But Superman64 is problematic. It works the other way round. First the CPU draws something and then the RDP draws over it. This and copy color is incorrect in this game

>BTW, current threshold level is not result of experiments. It guaranties that buffer definitely changed, but it probably can be lowered.

I don't think it can be lowered much. In Pokemon Snap and Jet Force Gemini the CPU draws some details. I guess the rain in JFG is already too much.

gonetz:
>But I think it's better than seeing this guy's penis

First screen is correct. The game looks as this on N64.

> It won't be a quick fix. I think you can merge it. 

Ok.

purplemarshmallow:
>First screen is correct. The game looks as this on N64.

Would be better to copy only the modified pixels from RDRAM

gonetz:
> Would be better to copy only the modified pixels from RDRAM

Would be better to have support from emulator's side for that.

tony971:
> Would be better to have support from emulator's side for that.

@Project64 @cxd4 @LegendOfDragoon

purplemarshmallow:
>Would be better to have support from emulator's side for that.

Yes it should be less hackish. And it should also give a great speedup.

gonetz:
BTW, you made a research in that direction. Did you get any practical results?

purplemarshmallow:
Not so much. I got some read/write notifications from 1964 and mupen64 not much more
I noticed Rice Video + 1964 supports copying from RDRAM with emulator notification. Haven't figured out yet how to implement it in GLideN64 using the existing copy from RDRAM code. The implementation must be different. Rice Video uses DirectX and it's not perfect.

LegendOfDragoon:
>Rice Video uses DirectX and it's not perfect.

Does the OpenGL version of Rice not support this feature?

purplemarshmallow:
No the OpenGL version does not support framebuffer emulation. The framebuffer code only works properly in the DirectX8 version of Rice Video. In the DirectX9 version most framebuffer code is broken.

gonetz:
I've made frameBufferEmulation.bufferSwapMode option. By default it swaps buffers on each VI_UpdateScreen() call. QII works with no hacks.
I found, that my code broke CFB emulation, so I added this your commit too.
I put the changes into new_buffer_swap branch:
https://github.com/gonetz/GLideN64/tree/new_buffer_swap

I've fixed Pokemon Stadium 2 crashes: 93f659c8a
Vigilante 8 crash in the menus still exists, but since it crashes with master also, it is not regression caused by your changes.

Binaries for those, who want to test it:
https://drive.google.com/file/d/0B0YqMPjGo3B2ZjBXOXUtX2szeUk/view?usp=sharing

olivieryuyu:
i can see in the code a hack for the map for Lego Racers

Is it still needed?

olivieryuyu:
omg this genius, bottom of the 9th is nearly entirely fixed!

sdfdsf

purplemarshmallow:
>bottom of the 9th is nearly entirely fixed!

Nice to see that it works well in games I haven't tested!

>Vigilante 8 crash in the menus still exists, but since it crashes with master also, it is not regression caused by your changes.

Vigilante 8 works with master, no crashing. It crashes because of this change. But it does not make sense to me. Project64 gives unknown memory action errors. How can copying from RDRAM possibly cause a problem like this?

purplemarshmallow:
>i can see in the code a hack for the map for Lego Racers. Is it still needed?

I'd say no

gonetz:
I have crash on master with Vigilante 8. May be emulator related.
I agree that mere copy from RDRAM must be safe as far as you read inside RDRAM bounds. It can show garbage if you read from a wrong place, but not a crash.

LegendOfDragoon:
>No the OpenGL version does not support framebuffer emulation.

You're right. Does the m64p version of rice also not support frame buffer emulation?

>Project64 gives unknown memory action errors. How can copying from RDRAM possibly cause a problem like this?

Try cpu interpreter.

purplemarshmallow:
>You're right. Does the m64p version of rice also not support frame buffer emulation?

No framebuffer emulation. I don't know why they added the not working fb options to mupen64plus

>Try cpu interpreter.

Yes CPU interpreter is possibly better for testing things like this. No more unknown memory action. But it freezes in the menus. I think the problem is incorrect buffer size. Vigilante8 switches resolutions during the menu transitions. Back when VI width calculation was incorrect in some games copy from RDRAM always crashed. This should not happen.

>I have crash on master with Vigilante 8. May be emulator related.

The game works on master. I don't know about any emulator issues. It's a known problem the (E) version crashes with copy to RDRAM.

olivieryuyu:
"QII works with no hacks"

I have no bar in Q2. Do i do something incorrect?

purplemarshmallow:
Quake 2 loading bar does not work

PR #799, opened 2015-11-18
copy only modified pixels to RDRAM

purplemarshmallow:
I know this solution is not perfect. But without this change copying to RDRAM can cause crashes. And that's something which should never happen. I noticed this fixes Donkey Kong 64 (E)!

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -1066,7 +1066,8 @@ void FrameBufferToRDRAM::CopyToRDRAM(u32 _address, bool _sync)
 		for (u32 y = 0; y < height; ++y) {
 			for (u32 x = 0; x < pBuffer->m_width; ++x) {
 				c.raw = ptr_src[x + (height - y - 1)*pBuffer->m_width];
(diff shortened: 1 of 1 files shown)


purplemarshmallow:
But only sync fb copies will work in Donkey Kong 64 (E)

gonetz:
Does it cause regressions we should know about?

purplemarshmallow:
The only regression I know about is the auxiliary buffers with the Koopas in Mario Tennis fill with garbage if copy auxiliary to RDRAM is enabled

gonetz:
OK.
https://github.com/gonetz/GLideN64/commit/84611502c6d6994edbe35f3e571be05a510d99c3

purplemarshmallow:
Now I think async fb is a bad default setting since it can cause crashes. I think sync is also bad as a default setting because it costs a lot of performance. It's possibly best to disable copy color by default. Copy with emulator notification must be implemented anyway. This is the only good solution.

gonetz:
Agreed. I suppose that custom ini will need corrections if copy color buffer will be disabled.

purplemarshmallow:
No. It has sync enabled for games that need it and async enabled for games compatible with it. Donkey Kong 64 needs to be corrected.

PR #801, opened 2015-11-19
disable copy color by default 

purplemarshmallow:
+typo fix

### DIFF
--- a/ini/GLideN64.custom.ini
+++ b/ini/GLideN64.custom.ini
@@ -106,14 +106,13 @@ frameBufferEmulation\copyDepthToRDRAM=1
 
 [DONKEY%20KONG%2064]
 Good_Name=Donkey Kong 64 (E)(J)(U)
-; set to 1 for correct HUD in pause menu
-frameBufferEmulation\copyToRDRAM=2
+frameBufferEmulation\copyToRDRAM=1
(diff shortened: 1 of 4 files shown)


AmbientMalice:
Is there any particular reason for this change? This will mean having to re-enable RDRAM copies for every single game that needs them.

purplemarshmallow:
Yes see https://github.com/gonetz/GLideN64/pull/799
async fb copies can be dangerous

AmbientMalice:
Shouldn't it be set back to sync copies by default, then? Or simply set to sync on the games where async causes crashes/problems?

purplemarshmallow:
async can't be default. The default settings should be safe. Maybe another game has the same problem as Donkey Kong 64. Enabling sync by default would be an option. But I think it's bad. It costs a lot of performance and many users complained about bad performance. Perfect solution would be to have emulator notifications.

AmbientMalice:
>Perfect solution would be to have emulator notifications.
That would be ideal, yes. It would also cure related problems.

purplemarshmallow:
Not sure what's the right thing to do. The current ini file has many entries of games that need RDRAM copies.

purplemarshmallow:
I'll colse for now. If copy to RDRAM is disabled by default this possibly needs more testing

PR #813, opened 2015-12-05
Fixing FBInfo build on gcc (5.2.1)

matthewharvey:
There was a conflict between signed and unsigned types to std::min.
Explicitly setting the literal as unsigned keeps all math as unsigned,
which I think is the best case here (assuming that 1 + y0 - y1 will never
be negative).
Also adding std:: prefix to make it clearer where these functions come from.


gonetz:
Added, thanks!

0ab7e1133

PR #821, opened 2015-12-17
Adds noisy warning if we can't find Texture Image support

matthewharvey:
This was confusing me for a while as it would simply fail to find the
proper support and I was confronted with a blank screen in certain areas.
This warning (and the associated change of default. I'd like users to see
warnings as well) will help others figure out why there is a blank screen.

My driver supports the necessary extensions (but not 4.3), so if I force it to
advertise 4.3 support (and glsl 420 support) it will properly render.
I tried to create a patch that only looks for the proper extension, and
it works for me, but I feel as though it would depend on behaviour that
most GL implementations would not implement the same way so I've put
it in a separate branch and probably won't request to merge it.
Mesa should get 4.3 support soon enough anyway, so it shouldn't
be an issue for much longer.


src/OpenGL.cpp
```diff
@@ -1327,6 +1327,10 @@ void OGLRender::_initExtensions()
 	m_bImageTexture = false;
 #endif
 	LOG(LOG_VERBOSE, "ImageTexture support: %s\n", m_bImageTexture ? "yes" : "no");
+	if (m_bImageTexture == false)
+	{
+		LOG(LOG_WARNING, "Image Textures not supported. Without OpenGL >= 4.3 or GLES >= 3.1, depth buffer emulation will not work");
```
gonetz: >> depth buffer emulation will not work

it is incorrect. "N64 depth compare" option will not work.
Depth buffer copy to RDRAM does not use Image Textures. Thus, effects based on depth buffer check in RDRAM will work.

matthewharvey: I think I've got my terminology wrong. I associate the fog effect in Beetle Adventure Racing with depth textures (from your blog post), and I forget what the name of the buffer used for that is (framebuffer texture?), but we definitely depend on image texture support for that type of fogging seeing as I get a black screen on those tracks that make use of it. Also there were a whole bunch of functions that simply return without doing any work if the image texture support wasn't available.

The N64 depth compare uses some other feature that my GL driver actually doesn't have, and thus I fail to run it with an error about something to do with memoryBarrier or something like that being called from a shader (I'm not at my computer right now otherwise I'd know the exact name).

I'll have a look later (maybe not the next couple of days. Flying on Friday) and clear up which features require which GL extensions and write down my evidence as to why.

gonetz: BAR fog was initially implemented with Image textures, but then I remade it and now it uses standard FBO depth texture. However it still uses Image texture for depth LookUpTable. I checked BAR with m_bImageTexture=false, it really turns black on levels with fog. 

So, "N64 depth compare and depth based fog will not work without  Image Textures".

matthewharvey:
Closing this because I clearly messed something up in git

PR #838, opened 2016-01-26
Typo fix

AmbientMalice:
Hope this doesn't break anything, translation-wise. It's just that typo has been bugging me, and I've finally figured out how to make unified pull requests.


gonetz:
Added to master.
The change is correct, but I'm afraid there are more such mistakes. The whole GUI will need to be proof-read before next release, and update the translation files.

PR #851, opened 2016-02-04
Add Japanese translation

Nekokabu:


gonetz:
Thank you for your work! I have one question: why almost all translations have type="unfinished" ?

Nekokabu:
Text of the portion of the marked is not registered. 
Could you update the ts files, create a template file?
glide_1
glide_2
glide_3
glide_4

gonetz:
I have added your commit to master.
Also, I've updated the .ts files with texts from current state of UI. Thus, you may add translation for new  elements, like 'Force gamma correction'. I recommend to use Qt Linguist for editing translation files:
https://github.com/gonetz/GLideN64/wiki/Internationalization

PR #874, opened 2016-02-28
Updating Spanish Translation

IlDucci:
Updating Spanish translation, adding the newest strings and editing some old ones.


gonetz:
Added

PR #893, opened 2016-03-06
Fixing what is almost certainly a memory leak

matthewharvey:
I've looked through it again since commit and this is most definitely leaking.

I know we've discussed before that this is not the correct way to be doing this, but for right now we should at least close the leak.

### DIFF
--- a/src/Textures.cpp
+++ b/src/Textures.cpp
@@ -820,6 +820,7 @@ void TextureCache::_loadBackground(CachedTexture *pTexture)
 	}
 	if (m_curUnpackAlignment > 1)
 		glPixelStorei(GL_UNPACK_ALIGNMENT, m_curUnpackAlignment);
+	free(pSwapped);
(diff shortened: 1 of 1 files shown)


gonetz:
> I've looked through it again since commit and this is most definitely leaking.

malloc without free. Yes, it is a leak.

>I  know we've discussed before that this is not the correct way to be doing this

I don't remember it. Why freeing memory is not the correct way to be doing this ?

gonetz:
Added, thanks!

matthewharvey:
I wasn't sure if the memory wasn't freed elsewhere, but after doing a thorough check it isn't

Back when I was doing texture work, we both agreed that a more C++ style memory management should be the answer. This will do for now though.

gonetz:
Can you check for memory errors with valgrind? Background textures can be checked with Starcraft. That games uses BG command for almost all graphics.

Issue #895, opened 2016-03-07
Nexus 6p fragment shader compilation error


fzurita:
Shaders fail to compile in the OpenGL ES (2.0, 3.0, and 3.1) version of the plugin on a Nexus 6p phone. The nexus 6p has an Adreno 430 GPU running driver release V127. Here is an example shader that fails to compile:

```
#version 310 es 
uniform sampler2D uTex0;		
uniform sampler2D uTex1;		
uniform lowp sampler2DMS uMSTex0;	
uniform lowp sampler2DMS uMSTex1;	
uniform lowp ivec2 uMSTexEnabled;	
layout (std140) uniform ColorsBlock {
  lowp vec4 uFogColor;			
  lowp vec4 uCenterColor;		
  lowp vec4 uScaleColor;		
  lowp vec4 uBlendColor;		
  lowp vec4 uEnvColor;			
  lowp vec4 uPrimColor;		
  lowp float uPrimLod;			
  lowp float uK4;				
  lowp float uK5;				
};								
uniform mediump vec2 uScreenScale;	
uniform lowp int uAlphaCompareMode;	
uniform lowp int uAlphaDitherMode;	
uniform lowp int uColorDitherMode;	
uniform lowp int uFogUsage;	
uniform lowp ivec2 uFbMonochrome;		
uniform lowp ivec2 uFbFixedAlpha;
uniform lowp int uSpecialBlendMode;
uniform lowp int uEnableAlphaTest;	
uniform lowp float uAlphaTestValue;
uniform mediump vec2 uDepthScale;	
in lowp vec4 vShadeColor;	
in mediump vec2 vTexCoord0;
in mediump vec2 vTexCoord1;
in mediump vec2 vLodTexCoord;
in lowp float vNumLights;	
in mediump float vFogFragCoord;
lowp vec3 input_color;			
out lowp vec4 fragColor;		
															
lowp float snoise();						
void calc_light(in lowp float fLights, in lowp vec3 input_color, out lowp vec3 output_color);
mediump float mipmap(out lowp vec4 readtex0, out lowp vec4 readtex1);		
lowp vec4 readTex(in sampler2D tex, in mediump vec2 texCoord, in lowp int fbMonochrome, in bool fbFixedAlpha);	
lowp vec4 readTexMS(in lowp sampler2DMS mstex, in mediump vec2 texCoord, in lowp int fbMonochrome, in bool fbFixedAlpha);	
bool depth_compare();										
[...]

PR #897, opened 2016-03-10
Fix OpenGL ES 3.0 build in Nexus 6p

fzurita:
This fixes issues with Nexus 6p in OpengGL ES 3.0 mode. It doesn't fix OpenGL ES 2.0 or 3.1. I suspect OpenGL ES 2.0 is the same fix. I'm still not sure what is wrong with the 3.1 version.

The fix is basically to not declare functions that are not defined later on in the GLSL.

Partial fix for this:

https://github.com/gonetz/GLideN64/issues/895


gonetz:
What about this commit:
https://github.com/mupen64plus-ae/mupen64plus-ae/commit/db87ffdf77a1029e0437cfd7ffaf3e507c33cba2

Is it unnecessary?

fzurita:
No, I took that out, it didn't make a difference.

gonetz:
Added:
https://github.com/gonetz/GLideN64/commit/25d9ae1fb85279e57d16129b5aadf3855bf727a6

I've corrected your code a bit to make it compliant with project's code style.

fzurita:
Cool, thanks

gonetz:
Thank you! I'm getting not so many useful code fixes. It's great to get help :)

PR #898, opened 2016-03-10
Update pokemon snap settings

tony971:
Attempt two since I'm not good enough with github to correct the original.

### DIFF
--- a/ini/GLideN64.custom.ini
+++ b/ini/GLideN64.custom.ini
@@ -308,7 +308,10 @@ frameBufferEmulation\detectCFB=1
 
 [POKEMON%20SNAP]
 Good_Name=Pokemon Snap (U)
-frameBufferEmulation\copyToRDRAM=1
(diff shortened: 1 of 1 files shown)


AmbientMalice:
The problem is frameBufferEmulation\copyFromRDRAM=1, which makes the game unstable and causes constant flickering because the feature is broken in a variety of ways.

tony971:
But the red dot won't appear without it.

tony971:
I think you might have meant copyToRDRAM=1

AmbientMalice:
>But the red dot won't appear without it.

That's my point. Using the framebuffer as a fullscreen texture makes quite a few games unstable. At least, it did last time I checked. Snap seems to crash with it enabled.

tony971:
It's the copyToRDRAM that causes constant flickering and crashing for me. copyFromRDRAM hasn't shown any negative side effects on my end.

gonetz:
Added

Issue #950, opened 2016-03-30
LoZ: Ocarina of Time randomly crashes when opening in-game menu


fzurita:
When opening the in-game menu within the LoZ: ocarina of time, the core will randomly crash and I think it's happening in the plugin. This is happening a lot in my android Shield TV.

Voron3DD:
Strange. I tried to run the game with other plugins. Crashes still happening.

gonetz:
@fzurita Can you run it under debugger to see, where exactly it crashes?
Or at least check adb log?

fzurita:
Yeah, once I get the time I'll do that to narrow it down to at least a  method.

Voron3DD:
Problem solved. Emulator incorrectly defined version of rom. And this has led to crash the game.

fzurita:
Interesting, the problem hasn't gone away for me. I'll dig further to see where we are crashing.

Voron3DD:
The more stable the emulator, the better) Thank you for your work)

retrobenny:
My thoughts on it with OOT MQ debug,the 3.1 variant causes it to crash outright when attempting to open the menu while the 3.0 variant merely freezes,but enabling the "subscreen delay fix" code stops the freezing though doesn't fix the 3.1 variant's crashing.

budtz:
I also get this crash. Using a build from 28 feb it is stable. Using anything newer and it crashes

fzurita:
Ok, I got a back trace from the debug version of LoZ Master quest. Here it is:

```
********** Crash dump: **********
Build fingerprint: 'NVIDIA/foster_e/foster:6.0/MRA58K/41937_667.2671:user/release-keys'
pid: 32338, tid: 5311, name: CoreThread  >>> org.mupen64plusae.v3.alpha <<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x3c
#00 pc 00015924  /data/app/org.mupen64plusae.v3.alpha-2/lib/arm/libmupen64plus-video-gliden64-gles31.so: Routine FrameBuffer_ActivateBufferTexture(short, FrameBuffer*) at ..\mupen64plus-ae/jni/./mupen64plus-video-gliden64/src/FrameBuffer.cpp:979
[...]

PR #959, opened 2016-04-11
Fix zelda subscreen crash

fzurita:
This commit makes the crash go away.

What I'm seeing is that there are multiple buffers with an start address of 0x00400e80 being generated even though there should only be one (I think). This is happening because this line
```
m_pCurrent->m_endAddress = min(RDRAMSize, m_pCurrent->m_startAddress + (((m_pCurrent->m_width * gDP.colorImage.height) << m_pCurrent->m_size >> 1) - 1));
```
is giving an end address that is equal to startAddress - 1. This is happening because gDP.colorImage.height is zero. gDP.colorImage.height is zero because m_prevColorImageHeight never got a chance to be set to anything.

I think because we have multiple buffers with the same start address, FrameBuffer_ActivateBufferTexture ends up getting called on a buffer that has not had a m_pLoadTile set.

https://github.com/gonetz/GLideN64/issues/950

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -542,10 +542,15 @@ void FrameBufferList::saveBuffer(u32 _address, u16 _format, u16 _size, u16 _widt
 				m_prevColorImageHeight = gDP.colorImage.height;
 			else if (gDP.colorImage.height == 0)
 				gDP.colorImage.height = m_prevColorImageHeight;
+
 			gDP.colorImage.height = min(gDP.colorImage.height, VI.height);
-			m_pCurrent->m_endAddress = min(RDRAMSize, m_pCurrent->m_startAddress + (((m_pCurrent->m_width * gDP.colorImage.height) << m_pCurrent->m_size >> 1) - 1));
-		} else if (m_pCurrent->m_needHeightCorrection && gDP.colorImage.height != 0) {
-			m_pCurrent->m_endAddress = min(RDRAMSize, m_pCurrent->m_startAddress + (((m_pCurrent->m_width * gDP.colorImage.height) << m_pCurrent->m_size >> 1) - 1));
+		}
+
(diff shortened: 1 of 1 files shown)


src/FrameBuffer.cpp
```diff
@@ -538,6 +538,11 @@ void FrameBufferList::saveBuffer(u32 _address, u16 _format, u16 _size, u16 _widt
 	if (m_pCurrent != NULL) {
 		// Correct buffer's end address
 		if (!m_pCurrent->isAuxiliary()) {
+			if(m_prevColorImageHeight == 0)
+			{
+				m_prevColorImageHeight = VI.height;
+			}
+			
 			if (gDP.colorImage.height > 200)
```
gonetz: m_prevColorImageHeight is ok to be zero.
The problem is that resulted gDP.colorImage.height is not tested against zero

fzurita: What should it use if ```gDP.colorImage.height``` is zero? Essentially, this is using ```VI.height``` instead.

gonetz: This is buffer height correction code. Initially buffer's height is VI.height for main buffer. It is good assumption.
If we calculated buffer height more precisely during rendering and gDP.colorImage.height is not zero, we can correct actual end address of the buffer. It is very important for fb emulation to know actual address space of the buffer.
If gDP.colorImage.height is zero, m_endAddress will not be corrected. It is ok.

fzurita:
Ok, gotcha. I'll make the correction and try it out.

src/FrameBuffer.cpp
fzurita: Ok, gotcha. I'll make the correction and try it out.

fzurita:
Commit has been updated using your suggestion. Also, I performed some code consolidation when correcting the end address between normal and auxiliary buffers for areas that were the same.

gonetz:
Looks good, thanks!

gonetz:
Added

Issue #968, opened 2016-04-21
[Android] Banjo-Tooie Groggy crash.


retrobenny:
If you enter the area where Groggy is,my case the Inferno area,it crashes with "unfortunately" error.
Doing so from my savestate made it exit outright,so try hard saving and reaching the area manually.
Both the savestate and .eep save files to help,annoyingly had to compress them both in zip form.
[inferno crash.zip](https://github.com/gonetz/GLideN64/files/230525/inferno.crash.zip)
[Banjo-Tooie (U) [!].eep.zip](https://github.com/gonetz/GLideN64/files/230526/Banjo-Tooie.U.eep.zip)

fzurita:
I got core dumps in 2 different places in the code:

```
********** Crash dump: **********
Build fingerprint: 'NVIDIA/foster_e/foster:6.0/MRA58K/41937_695.6426:user/release-keys'
pid: 20736, tid: 20785, name: CoreThread  >>> org.mupen64plusae.v3.alpha <<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x30
#00 pc 00030308  libmupen64plus-video-gliden64-gles31.so: Routine PostProcessor::doBlur(FrameBuffer*) at ../jni/./mupen64plus-video-gliden64/src/PostProcessor.cpp:583 (discriminator 1)
#01 pc 000120d4  libmupen64plus-video-gliden64-gles31.so: Routine FrameBufferList::renderBuffer(unsigned int) at ../jni/./mupen64plus-video-gliden64/src/FrameBuffer.cpp:843 (discriminator 4)
#02 pc 0003a940  libmupen64plus-video-gliden64-gles31.so: Routine VI_UpdateScreen() at ../jni/./mupen64plus-video-gliden64/src/VI.cpp:104
#03 pc 000b0d84  libmupen64plus-core.so: Routine vi_vertical_interrupt_event at ../jni/./mupen64plus-core/src/vi/vi_controller.c:99
#04 pc 0004543c  libmupen64plus-core.so: Routine gen_interupt at ../jni/./mupen64plus-core/src/r4300/interupt.c:536
#05 pc 000b12a4  libmupen64plus-core.so: Routine .E1 at ../linkage_arm.S:541
Crash dump is completed
```

and 

```
********** Crash dump: **********
Build fingerprint: 'NVIDIA/foster_e/foster:6.0/MRA58K/41937_695.6426:user/release-keys'
[...]

PR #969, opened 2016-04-22
Fix PostProcessor crash

fzurita:
See this issue:

https://github.com/gonetz/GLideN64/issues/968

Somehow the local static variable is being re-constructed. I added some logging to the constructor to verify and it's indeed getting called multiple times. I have no idea how this is happening since I thought the language would guarantee that it would only be called once. All I can think is compiler bug, let me know if you have any better ideas.

By the way, the change I did in this pull request fixes it even though I can't understand why the original code doesn't work.

### DIFF
--- a/src/PostProcessor.cpp
+++ b/src/PostProcessor.cpp
@@ -24,6 +24,10 @@
 #define FRAGMENT_SHADER_END "\n"
 #endif
 
+#ifdef ANDROID
+PostProcessor PostProcessor::processor;
+#endif
+
 static const char * vertexShader =
 SHADER_VERSION
 "#if (__VERSION__ > 120)						\n"
@@ -469,7 +473,9 @@ void PostProcessor::destroy()
 
 PostProcessor & PostProcessor::get()
 {
+#ifndef ANDROID
 	static PostProcessor processor;
+#endif
 	return processor;
 }
 
--- a/src/PostProcessor.h
+++ b/src/PostProcessor.h
@@ -18,6 +18,7 @@ public:
 private:
 	PostProcessor();
 	PostProcessor(const PostProcessor & _other);
+
 	void _initCommon();
 	void _destroyCommon();
 	void _initGammaCorrection();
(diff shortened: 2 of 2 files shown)


gonetz:
Looks like compiler bug indeed. The original code is correct. New one is less safe.
Try to make a build with lesser compiler optimizations. Debug one for instance.

fzurita:
Debug build with no optimizations also crashes. The odd thing is that it always crashes in the same spot, so it's possible that we are corrupting memory somehow and this is how it's showing up.

fzurita:
Also, just a note. When doing a debug build in android TextureFIlters_xbrz.cpp doesn't build because the debug build defines NDEBUG. The Android compiler doesn't recognize these calls:

```
int debugPixelX = -1;
int debugPixelY = 12;
__declspec(thread) bool breakIntoDebugger = false;
```

gonetz:
> Debug build with no optimizations also crashes. The odd thing is that it always crashes in the same spot, so it's possible that we are corrupting memory somehow and this is how it's showing up.

Well, this is possible. Bad things is that when I tried to run mupen64plus under valgrind, it crashed.

fzurita:
Until whatever is happening shows up in a more obvious way, do you want to merge my pull request? What do you think is unsafe about what I'm doing?

gonetz:
> Until whatever is happening shows up in a more obvious way, do you want to merge my pull request? 

I want to test it on Linux first. May be it is a general issue with g++

> What do you think is unsafe about what I'm doing?

Your fix changes the time when static objects will be created. It is not unsafe, it's just not right.
I'll merge it if I will not find a better solution.

fzurita:
I can always change the PostProcessor member attribute to a pointer that is only newed up when the "get" method is called the first time around.

fzurita:
Even though this doesn't fix the root cause of the problem, can you still merge it to prevent the crash in android? As far as I can tell, when the issue happens, nothing else is affected.

gonetz:
Please make it as Android-specific fix. I don't want it be common.

fzurita:
Ok

fzurita:
I updated this, hopefully I did what you wanted for this to be accepted. If not, let me know.

PR #980, opened 2016-05-07
Fix Adreno driver crash with new blending

fzurita:
I think I finally found the root cause of the crash, although I have no idea why it works but it doesn't seem to affect the new blending output. By the way, a clamp of 0.0 to 1.0 didn't work, only slightly higher than zero and slightly lower than 1.

Also, this may make the original issue that you had with crashing with the Nvidia device go away as well.

Any idea why this works? As far as I can tell the output should already be between 0.0 and 1.0.


fzurita:
Grrr.... never mind, it still crashes. Although, it completely stopped crashing in Majora's mask, but it still crashes in Mario Kart. Although, removing this line from ShaderUtils.cpp makes the problem go away, although, blending doesn't work very well after that.

```
"    color2 = clamp(blend2.rgb, 0.0, 1.0);
```

fzurita:
I updated my pull request, I found a combination that works 100% for everything so far.

gonetz:
I can't accept this. [0, 1] is valid range for color attributes.
If this ugly hack helps adreno devices, ok, but please enable it for adreno only.
We already has such an option for adreno specific hacks.

fzurita:
@gonetz , please try 3 point texture filtering with your Adreno device. I just realized that if you disable 3-point filtering, the driver crashes go away.

gonetz:
I have enabled everything:  fb emulation, 3 point filtering, noise, lod, HW lighting, widescreen - plugin works slow but does not crash.

fzurita:
Replacing this line in the shader:
```
"  lowp vec4 texColor = uTextureFilterMode == 0 ? texStandard : tex3Point;	\n"
```

with:
```
"  lowp vec4 texColor = tex3Point;	\n"
```

Also fixes the crashes in adreno.

Why do we need the ```uTextureFilterMode``` uniform? Can we set filtering when generating the shader instead?

fzurita:
Ok, good news. Crashes completely go away after changing the ```readTex``` method from

```
"  lowp vec4 texStandard = texture(tex, texCoord); 							\n"
"  lowp vec4 tex3Point = filter3point(tex, texCoord); 						\n"
"  lowp vec4 texColor = uTextureFilterMode == 0 ? texStandard : tex3Point;	\n"
```

to this:
```
"  lowp vec4 texColor = vec4(0.0);                          				\n"
"  if (uTextureFilterMode == 0) texColor = texture(tex, texCoord);      	\n"
"  else texColor = filter3point(tex, texCoord);                             \n"
```

These should both provide the same result. I'm thinking that the Adreno driver is optimizing in a bad way.

fzurita:
Pull request updated.

fzurita:
I give up on this. So after I got rid of all discards, I can't get it to crash any more, but graphics don't look right and who knows if that would fix it under all games. I'm just going to say that this is a driver bug and live with it.

gonetz:
>> Why do we need the uTextureFilterMode uniform? Can we set filtering when generating the shader instead?

Good point

fzurita:
It's a moot point. It doesn't fix the issue. Although, some how, the 3 point filtering is connected to the corrupt uScreenScale uniform.

gonetz:
I'd suggest to stop wasting your time on fight with bad drivers. As I said, plugin works ok on my device with the same GPU. My real concern is performance. I noticed that GLES3 build works much faster when fb emulation is off. This is strange, but I'm too busy with bugfixes to investigate that. Could you check, what breaks the performance? I suspect that glBlitFramebuffer is slow on Adreno.

gonetz:
I'm I right that this pull request does not fix the problem and can be ignored and closed?

fzurita:
Yeah, I have been trying to investigate that a little bit, but haven't put much time into it. As you said, I'm going to stop wasting my time with bad drivers.

fzurita:
I just did

fzurita:
FYI, I just made a 64 bit build of your plugin for Android arm64 devices like the shield tv. Everything seems to work fine. Here it is, if you want to try it:

https://github.com/mupen64plus-ae/mupen64plus-ae/tree/arm64

gonetz:
Thanks for the info! I don't have arm64 device yet.

PR #983, opened 2016-05-10
Fix Linux compilation

MaximeMorel:
std::sqrt and fabsf need cmath include to build on Linux.


gonetz:
Ok. Which Linux you are using? I'm asking because I've no problems with compilations on Ubuntu.

MaximeMorel:
Oh funny!
I'm using Archlinux x64, gcc 6.1.1.
You merged in the end? It was failing on some other systems?

gonetz:
No, but if it fixes compilation for you it should be merged.
Ubuntu 14.04 still uses gcc 4.x. Looks like the issue with new gcc.

PR #988, opened 2016-05-16
Framebuffer copy performance improvements

fzurita:
I've been trying to investigate where the performance loss is when enabling frame buffer emulation and one big hitter that I found is the ```_writeToRdram``` method in FrameBuffer.cpp

I noticed that in that method we are copying either 1 byte or 2 bytes at a time for the whole frame buffer. This does seem inefficient to me since most architectures now a day are 32 bits, so you can write 4 bytes in one operation.

I refactored the code so that we would write 4 bytes at a time to memory. I did this be creating ```_writeToRdram8``` and ```_writeToRdram16``` functions.

I want to make sure the changes are correct. What are some good games to verify this in?

Also, with CopyDepthBufferToRDRAM setting enabled, the Majoras's mask spinning N64 logo intro went from ~38fps to ~52fps in my device.

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -1369,15 +1369,21 @@ void FrameBufferToRDRAM::_copy(u32 _startAddress, u32 _endAddress, bool _sync)
 	if (m_pCurFrameBuffer->m_size == G_IM_SIZ_32b) {
 		u32 *ptr_src = (u32*)pixelData;
 		u32 *ptr_dst = (u32*)(RDRAM + _startAddress);
-		_writeToRdram<u32, u32>(ptr_src, ptr_dst, &FrameBufferToRDRAM::_RGBAtoRGBA32, 0, 0, width, height, numPixels, _startAddress, m_pCurFrameBuffer->m_startAddress, m_pCurFrameBuffer->m_size);
+		std::vector<u32> srcBuf(width * height);
+		memcpy(srcBuf.data(), ptr_src, width * height*sizeof(u32));
+		_writeToRdram<u32, u32>(srcBuf.data(), ptr_dst, &FrameBufferToRDRAM::_RGBAtoRGBA32, 0, 0, width, height, numPixels, _startAddress, m_pCurFrameBuffer->m_startAddress, m_pCurFrameBuffer->m_size);
 	} else if (m_pCurFrameBuffer->m_size == G_IM_SIZ_16b) {
 		u32 *ptr_src = (u32*)pixelData;
 		u16 *ptr_dst = (u16*)(RDRAM + _startAddress);
-		_writeToRdram<u32, u16>(ptr_src, ptr_dst, &FrameBufferToRDRAM::_RGBAtoRGBA16, 0, 1, width, height, numPixels, _startAddress, m_pCurFrameBuffer->m_startAddress, m_pCurFrameBuffer->m_size);
+		std::vector<u32> srcBuf(width * height);
+		memcpy(srcBuf.data(), ptr_src, width * height*sizeof(u32));
+		_writeToRdram<u32, u16>(srcBuf.data(), ptr_dst, &FrameBufferToRDRAM::_RGBAtoRGBA16, 0, 1, width, height, numPixels, _startAddress, m_pCurFrameBuffer->m_startAddress, m_pCurFrameBuffer->m_size);
 	}	else if (m_pCurFrameBuffer->m_size == G_IM_SIZ_8b) {
 		u8 *ptr_src = (u8*)pixelData;
 		u8 *ptr_dst = RDRAM + _startAddress;
-		_writeToRdram<u8, u8>(ptr_src, ptr_dst, &FrameBufferToRDRAM::_RGBAtoR8, 0, 3, width, height, numPixels, _startAddress, m_pCurFrameBuffer->m_startAddress, m_pCurFrameBuffer->m_size);
+		std::vector<u8> srcBuf(width * height);
+		memcpy(srcBuf.data(), ptr_src, width * height*sizeof(u8));
+		_writeToRdram<u8, u8>(srcBuf.data(), ptr_dst, &FrameBufferToRDRAM::_RGBAtoR8, 0, 3, width, height, numPixels, _startAddress, m_pCurFrameBuffer->m_startAddress, m_pCurFrameBuffer->m_size);
 	}
 
 	m_pCurFrameBuffer->m_copiedToRdram = true;
@@ -1590,7 +1596,10 @@ bool DepthBufferToRDRAM::_copy(u32 _startAddress, u32 _endAddress)
 
 	f32 * ptr_src = (f32*)pixelData;
 	u16 *ptr_dst = (u16*)(RDRAM + _startAddress);
-	_writeToRdram<f32, u16>(ptr_src, ptr_dst, &DepthBufferToRDRAM::_FloatToUInt16, 2.0f, 1, width, height, numPixels, _startAddress, m_pCurDepthBuffer->m_address, G_IM_SIZ_16b);
+
+	std::vector<f32> srcBuf(width * height);
+	memcpy(srcBuf.data(), ptr_src, width * height*sizeof(f32));
+	_writeToRdram<f32, u16>(srcBuf.data(), ptr_dst, &DepthBufferToRDRAM::_FloatToUInt16, 2.0f, 1, width, height, numPixels, _startAddress, m_pCurDepthBuffer->m_address, G_IM_SIZ_16b);
 
 	m_pCurDepthBuffer->m_cleared = false;
 	FrameBuffer * pBuffer = frameBufferList().findBuffer(m_pCurDepthBuffer->m_address);


AmbientMalice:
Tested on PC with mupen64plus, and I'm not seeing any significant performance differences.

fzurita:
It's possible that the compiler you are using is already performing such optimization. It seemed to make a difference in Android. Another thing I'm thinking about doing is making this multi-threaded, but it may hurt rather than increase performance. I have to do more research.

fzurita:
Wow, my code formatting looks horrible in Github. Android studio doesn't do well with tabs/spaces.

LegendOfDragoon:
>Also, with CopyDepthBufferToRDRAM setting enabled, the Majoras's mask spinning N64 logo intro went from ~38fps to ~52fps in my device.

I guess it makes more of a difference on your device, as opposed to desktops then. Doing 4 byte writes instead of 2 or 1 generally is faster, but writing to RDRAM isn't much of a bottleneck for desktops. Which is probably why AmbientMalice does not see a noticeable difference.

Since you see a difference on your device, it's worth optimizing.

src/FrameBuffer.cpp
```diff
@@ -1281,18 +1282,126 @@ void _writeToRdram(TSrc* _src, TDst* _dst, TDst(*converter)(TSrc _c), TSrc _test
...
+			c1 = _src[x + heightOffset*_width];
+		    c2 = _src[x + 1 + heightOffset*_width];
+			c3 = _src[x + 2 + heightOffset*_width];
+		    c4 = _src[x + 3 + heightOffset*_width];
+		    val1 = c1 != _testValue ? converter(c1) : _dst[numStored + 3];
+		    val2 = c2 != _testValue ? converter(c2) : _dst[numStored + 2];
+		    val3 = c3 != _testValue ? converter(c3) : _dst[numStored + 1];
+		    val4 = c4 != _testValue ? converter(c4) : _dst[numStored];
+
+            _dst32[dst32index] = (val1 << 24) | (val2 << 16) | (val3 << 8) | val4;
```
fzurita: I really hope I don't have this reversed, so I need find a game that triggers this to make sure it's right.

src/FrameBuffer.cpp
```diff
@@ -1281,18 +1282,126 @@ void _writeToRdram(TSrc* _src, TDst* _dst, TDst(*converter)(TSrc _c), TSrc _test
...
+
+	for (; y < _height; ++y) {
+		for (u32 x = 0; x < _width && numStored < _numPixels; x += 2) {
+		    heightOffset = _height - y - 1;
+			c1 = _src[x + heightOffset*_width];
+		    c2 = _src[x + 1 + heightOffset*_width];
+		    val1 = c1 != _testValue ? converter(c1) : _dst[numStored + 1];
+		    val2 = c2 != _testValue ? converter(c2) : _dst[numStored];
+
+            _dst32[dst32index] = (val1 << 16) | val2;
```
fzurita: I'm hoping I don't have this reversed... I need a game that triggers this to make sure it's right.

mudlord:
On writes on desktops: not a problem. The problem has always been the read.

fzurita:
I found that the read from the GPU FB, at least in android, doesn't take very long. It could be because of the shared memory architecture.

gonetz:
I specially wrote template function to avoid duplication of code. You not only returned code duplication, but made it even worse. It is unacceptable in its current state.
From the other side, the code seems to be working. I tested only 16bit buffers copy, I don't remember, which game uses 32bit buffers and need copy buffer to RDRAM.

gonetz:
@fzurita "Also, with CopyDepthBufferToRDRAM setting enabled, the Majoras's mask spinning N64 logo intro went from ~38fps to ~52fps in my device."

I wonder, what is performance difference in game, not on logo?

gonetz:
_writeToRdram8 looks incorrect: xor == 3 for 8bits, but you wrote it as  _dst32[dst32index] = (val1 << 24) | (val2 << 16) | (val3 << 8) | val4;

gonetz:
@fzurita please test my patch:
https://drive.google.com/file/d/0B0YqMPjGo3B2YTBtWEYxS1gyRms/view?usp=sharing

it also does 32bit writing to destination buffer, but without awful code duplication.

fzurita:
I submitted the pull request mainly to get the idea of there. I was open updating it duplicate code less. I'll try your patch tonight to see if I see if I get the same improvements.

fzurita:
I had time to do a test. For the same test u was running:
Your patch = 41-44fps
My pull request = 49-53fps
Original= ~38fps

fzurita:
If you want, I can update my pull request to do the same thing with no code duplication.

gonetz:
I'd like to get result of more representative test, e.g Mario head in SM64 or/and Zelda MM demo

fzurita:
Mario's head with the hand out and game speed set to 250%:

My pull request = 23-26fps
Your patch = 20-23fps

I took out the hand and waited fir the stars to go away since they cause a lot of variability.

gonetz:
I have an idea to test. Write to RDRAM must not be a bottleneck because RDRAM data is in conventional memory. The bottleneck is read from source buffer, which is mapped to video memory data. Even if it is in shared memory, it is not directly accessible. That's why it is important to access it as fast as possible. Your code reads two 32bit number per cycle:
```
		c1 = _src[x + heightOffset*_width];
	    c2 = _src[x + 1 + heightOffset*_width];
```
It probably makes reads faster on mobile GPU.

I suggest to try to read whole source buffer into a temporal u32 buffer and put that temporal buffer to unmodified _writeToRdram. 
```
u32 *ptr_src = (u32*)pixelData;
u32 * srcBuf = new u32[width * height];
for (u32 i = 0; i < width * height; ++i)
  srcBuf[i] = ptr_src[u32];
_writeToRdram
```

fzurita:
Sure, I will try that, I ran out of time this morning though, it will have to try this after I get back from my job.

gonetz:
good, thanks.

fzurita:
One additional thing I found recently glReadPixels causes no slowdown for the depth buffer, but it causes a very big slow down for the color buffer. I figured they would be the same, but they are not. 

I've been doing a good amount of research online and it's very common for glReadPixels to be very slow for some reason even with PBOs. One solution I'm seeing very often is to use the EGL_KHR_image_base extension instead and that seems to be a lot faster. I'm going to see if I can implement that for the glReadPixels call for the color buffer.

fzurita:
Your proposed change didn't seem make a significant difference at least in Mario's face. It seems to be back down to the original speed there. Although, at the Majora's mask intro n64 logo it's doing around 50fps.

fzurita:
The above number was after doing what you said right before the _writeToRdram when copying the depth buffer.

After I did the above with before all the calls to _writeToRdram and used memcpy instead of a for loop, the fps at Mario's face went to 31 to 32fps . That is a huge boost.

I'm going to check now if I combine both solutions if it will make a difference.

fzurita:
I updated the pull request request with what gives me the best performance.

src/FrameBuffer.cpp
```diff
@@ -1263,6 +1263,9 @@ bool FrameBufferToRDRAM::_prepareCopy(u32 _startAddress)
 template <typename TSrc, typename TDst>
 void _writeToRdram(TSrc* _src, TDst* _dst, TDst(*converter)(TSrc _c), TSrc _testValue, u32 _xor, u32 _width, u32 _height, u32 _numPixels, u32 _startAddress, u32 _bufferAddress, u32 _bufferSize)
 {
+	const u32 np = sizeof(u32) / sizeof(TDst);
+	TDst buf[sizeof(u32) / sizeof(TDst)];
+	u32* dst32 = reinterpret_cast<u32*>(_dst);
```
gonetz: Does that trick with 32bit write really helps?
It does not much difference, write values to a buffer and then to destination, or right to destination.

src/FrameBuffer.cpp
```diff
@@ -1369,15 +1380,25 @@ void FrameBufferToRDRAM::_copy(u32 _startAddress, u32 _endAddress, bool _sync)
 	if (m_pCurFrameBuffer->m_size == G_IM_SIZ_32b) {
 		u32 *ptr_src = (u32*)pixelData;
 		u32 *ptr_dst = (u32*)(RDRAM + _startAddress);
-		_writeToRdram<u32, u32>(ptr_src, ptr_dst, &FrameBufferToRDRAM::_RGBAtoRGBA32, 0, 0, width, height, numPixels, _startAddress, m_pCurFrameBuffer->m_startAddress, m_pCurFrameBuffer->m_size);
+		u32 srcBuf[width * height];
```
gonetz: Only g++ can do that trick. MSVS gets error:
error C2131: expression did not evaluate to a constant

If you don't want to mess with new/delete, you may do like this:
```
std::vector<u32> srcBuf(width * height);
u32 * pSrc = srcBuf.data();
memcpy(pSrc, ptr_src, width * height*sizeof(u32));
....
```

src/FrameBuffer.cpp
fzurita: If it helps, it's not by much. Biggest different is doing the memcpy of the buffer before passing it to _writeToRdram.

src/FrameBuffer.cpp
fzurita: I'm surprised that doesn't work with MSVS. It's allowed by the language. I can change it to that.

fzurita:
I updated the pull request once again.

gonetz:
Looks good. Merged.

PR #992, opened 2016-05-19
Add support for building without GLideNHQ

loganmc10:
Building GLideNHQ doesn't work properly on GLES2 targets. This commit:

https://github.com/loganmc10/GLideN64/commit/12102146f13afc24000cdf9dbf99043ded8e7351

added support for building without GLideNHQ, but the CMake system doesn't currently support building without GlideNHQ.

This commit adds a -DNOHQ option to cmake, allowing the project to be built without GlideNHQ


gonetz:
Merged

PR #993, opened 2016-05-19
Fix shaders on the Raspberry Pi

loganmc10:
Tested and working on my Raspberry Pi 3, I haven't tested this on any other platforms.


gonetz:
Good that you fixed that. However, that patch is useless for other platforms, and it looks ugly. 
IIRC, we have a define for pi platform. 
Please move your variant of blending shaders to separate .h file and use it instead of main ones for pi.

loganmc10:
Ok I've made that change

gonetz:
Thanks! Merged.

PR #996, opened 2016-05-22
Fix multisampling when the VI origin is not 0

matthewharvey:
This happens with Beetle Adventure Racing when the movie reel
effect is at work. It threw up a whole bunch of errors without
taking into account whether or not the framebuffer is multi-sampled.


gonetz:
Thanks! I corrected your commit a bit (post filter was missing also) and merged it.

matthewharvey:
Excellent. Thanks for the correction.

Issue #979, opened 2016-05-04
Frame buffer emulation without having to copy from GPU memory


fzurita:
With current mobile devices copying GPU memory to system memory is terribly slow and we seem to be doing a lot of that during frame buffer emulation. The question is, why are we doing so many copies from GPU memory to system memory when the same memory is shared between the GPU and CPU in  a mobile device?

I was reading a little bit about this and there is an OpenGL ES extension that could potentially help with this. The extension is OES_EGL_image_external but it may only apply to shaders reading system memory.

I think this is something worth investigating.

fzurita:
@gonetz Here is kind of a guide. I may try to take a shot at this? How doable do you think it is? I guess I would be converting the glReadPixels calls to using memory directly.

https://vec.io/posts/faster-alternatives-to-glreadpixels-and-glteximage2d-in-opengl-es

gonetz:
It would be great to improve buffer read speed,  by I doubt that it is possible.  GLES3 implementation already uses PBO. The problem with PBO is that it has little to no performance gain when you use buffer right after read, that is in sync mode. It is possible to use async mode for color buffer read in many games, and it is really many times faster than sync mode. However,  depth buffer emulaton can't use async buffer read, which creates performance issue on PC as well. You may try EGLImage extension, may be it uses faster platform specific data transfer.

fzurita:
For reference, from Gonetz:

> 
GLES2 uses FBO as well.
You may disable frame buffer emulation and compare speed. GLES3 shaders are more complex, it can be the reason. 
Regarding FB emulation: the main difference is that GLES3 uses glBlitFramebuffer to draw FBO on screen, while GLES2 renders textured rect with glDrawArrays. glBlitFramebuffer should work faster. Again, you may try to use GLES2 method with GLES3: replace
```
#ifndef GLES2
void FrameBufferList::renderBuffer(u32 _address)

by

#if 0
void FrameBufferList::renderBuffer(u32 _address)
```
>

fzurita:
@gonetz I found why glReadPixels is so slow on Adreno devices even when using PBOs. We are some how passing the wrong color format to the glReadPixels call. When hard coding the color format to GL_RGB, glReadPixels turns into a non-blocking call!

Now to find out why it's getting set wrong....

fzurita:
[...]

PR #998, opened 2016-05-24
Android CopyColorToRDRAM performance improvements

fzurita:
When in CopyColorToRDRAM async mode, we will now use the OES_EGL_image_external OpenGL ES extension to copy the color buffer into system memory. This provides large performance  improvements in many Android devices.

This was tested using an Adreno and a Shield TV device.

See this issue: https://github.com/gonetz/GLideN64/issues/979

Also, I'm only setup to compile for Android. Please verify that this builds on other platforms.

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -18,6 +18,19 @@
 #include "Debug.h"
 #include "PostProcessor.h"
 #include "FrameBufferInfo.h"
+#include "Log.h"
+
+#ifdef ANDROID
+#include "ui/GraphicBuffer.h"
+#include <android/native_window.h>
+#include <EGL/egl.h>
+#include <EGL/eglext.h>
+
+typedef void (GL_APIENTRY* PFNGLEGLIMAGETARGETTEXTURE2DOESPROC) (GLenum target, EGLImageKHR image);
+typedef void (GL_APIENTRY* PFNGLEGLIMAGETARGETRENDERBUFFERSTORAGEOESPROC) (GLenum target, EGLImageKHR image);
+
+using namespace android;
+#endif
 
 using namespace std;
 
@@ -51,6 +64,9 @@ private:
 	};
 
 	bool _prepareCopy(u32 _startAddress);
+
+	GLubyte* getPixels(GLint _x0, GLint _y0, GLsizei _width, GLsizei _height, bool _sync);
+	void cleanUpPixels(GLubyte* pixelData);
 	void _copy(u32 _startAddress, u32 _endAddress, bool _sync);
 
 	// Convert pixel from video memory to N64 buffer format.
@@ -65,6 +81,12 @@ private:
 	u32 m_frameCount;
 	u32 m_startAddress;
 	GLuint m_PBO[3];
+
+#ifdef ANDROID
+	GraphicBuffer* m_window;
+	EGLImageKHR m_image;
+	PFNGLEGLIMAGETARGETTEXTURE2DOESPROC m_glEGLImageTargetTexture2DOES;
+#endif
 };
 
 class DepthBufferToRDRAM
@@ -680,7 +702,7 @@ void FrameBufferList::saveBuffer(u32 _address, u16 _format, u16 _size, u16 _widt
 		//Also, before making any adjustments, make sure gDP.colorImage.height has a valid value.
 		if((!m_pCurrent->isAuxiliary() || m_pCurrent->m_needHeightCorrection) && gDP.colorImage.height != 0)
 		{
-		    m_pCurrent->m_endAddress = min(RDRAMSize, m_pCurrent->m_startAddress + (((m_pCurrent->m_width * gDP.colorImage.height) << m_pCurrent->m_size >> 1) - 1));
+			m_pCurrent->m_endAddress = min(RDRAMSize, m_pCurrent->m_startAddress + (((m_pCurrent->m_width * gDP.colorImage.height) << m_pCurrent->m_size >> 1) - 1));
 		}
 
 		if (!m_pCurrent->_isMarioTennisScoreboard() && !m_pCurrent->m_isDepthBuffer && !m_pCurrent->m_copiedToRdram && !m_pCurrent->m_cfb && !m_pCurrent->m_cleared && m_pCurrent->m_RdramCopy.empty() && gDP.colorImage.height > 1) {
@@ -1172,6 +1194,16 @@ void FrameBufferToRDRAM::Init()
 	assert(checkFBO());
 	glBindFramebuffer(GL_DRAW_FRAMEBUFFER, 0);
 
+#ifdef ANDROID
+	m_window = new GraphicBuffer(m_pTexture->realWidth, m_pTexture->realHeight,
+		PIXEL_FORMAT_RGBA_8888, GraphicBuffer::USAGE_SW_READ_OFTEN | GraphicBuffer::USAGE_HW_TEXTURE);
+
+	EGLint eglImgAttrs[] = { EGL_IMAGE_PRESERVED_KHR, EGL_TRUE, EGL_NONE, EGL_NONE };
+	m_image = eglCreateImageKHR(eglGetDisplay(EGL_DEFAULT_DISPLAY), EGL_NO_CONTEXT,
+		EGL_NATIVE_BUFFER_ANDROID, (EGLClientBuffer)m_window->getNativeBuffer(), eglImgAttrs);
+
+	m_glEGLImageTargetTexture2DOES = (PFNGLEGLIMAGETARGETTEXTURE2DOESPROC)eglGetProcAddress("glEGLImageTargetTexture2DOES");
+#else
 	// Generate and initialize Pixel Buffer Objects
 	glGenBuffers(3, m_PBO);
 	for (u32 i = 0; i < 3; ++i) {
@@ -1180,6 +1212,7 @@ void FrameBufferToRDRAM::Init()
 	}
 	glBindBuffer(GL_PIXEL_PACK_BUFFER, 0);
 	m_curIndex = 0;
+#endif
 }
 
 void FrameBufferToRDRAM::Destroy() {
@@ -1192,8 +1225,13 @@ void FrameBufferToRDRAM::Destroy() {
 		textureCache().removeFrameBufferTexture(m_pTexture);
 		m_pTexture = NULL;
 	}
+
+#ifdef ANDROID
+	eglDestroyImageKHR(eglGetDisplay(EGL_DEFAULT_DISPLAY), m_image);
+#else
 	glDeleteBuffers(3, m_PBO);
 	m_PBO[0] = m_PBO[1] = m_PBO[2] = 0;
+#endif
 }
 
 bool FrameBufferToRDRAM::_prepareCopy(u32 _startAddress)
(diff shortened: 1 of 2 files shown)


fzurita:
I want to try this for the depth buffer as well, but I have no idea what to look for there. What are some games that depend on accurate depth buffer emulation?

gonetz:
Coronas in Zeldas or Perfect Dark is a good example.

fzurita:
Is there anything else I need to do? I can get coronas on Zelda to show no matter what. I'm looking straight at the sun.

gonetz:
Coronas around torches. Wait for the night :)

fzurita:
I looked for them there as well. I'm not seeing any coronas. I'll send a picture.

fzurita:
No matter the angle, I'm not seeing them. Maybe Adreno devices have a bug 
screenshot_2016-05-24-07-48-28
with the depth buffer emulation

gonetz:
My device renders coronas when depth buffer copy disabled. Of course, it renders all coronas, through all walls. When I enable depth buffer copy, no coronas. I don't remember, did I test depth buffer copy on Android or not.

fzurita:
I don't seem to be getting coronas with depth buffer enabled or disabled.

gonetz:
It is hardly possible. Your drivers can't be broken **so** much.

gonetz:
Check bombers hideout in Zelda MM:
screenshot_2016-05-24-22-22-12

src/FrameBuffer.cpp
```diff
@@ -680,7 +702,7 @@ void FrameBufferList::saveBuffer(u32 _address, u16 _format, u16 _size, u16 _widt
 		//Also, before making any adjustments, make sure gDP.colorImage.height has a valid value.
 		if((!m_pCurrent->isAuxiliary() || m_pCurrent->m_needHeightCorrection) && gDP.colorImage.height != 0)
 		{
-		    m_pCurrent->m_endAddress = min(RDRAMSize, m_pCurrent->m_startAddress + (((m_pCurrent->m_width * gDP.colorImage.height) << m_pCurrent->m_size >> 1) - 1));
+			m_pCurrent->m_endAddress = min(RDRAMSize, m_pCurrent->m_startAddress + (((m_pCurrent->m_width * gDP.colorImage.height) << m_pCurrent->m_size >> 1) - 1));
 		}
```
gonetz: useless modification

gonetz:
@fzurita your modifications look ok, except for the fact that you use spaces instead of tabs, as in rest of the code. The general problem is that the code is now heavy sliced by #ifdef/#endif and becomes unreadable. It is not your fault. I'll accept it when after build and test. I'll do necessary refactoring later.

Are your changes in mupen64plus-ae repo? Which branch may I build to test it?

src/FrameBuffer.cpp
fzurita: Whoops, that must had snuck in there when I converted all my spaces to tabs.

fzurita:
My changes are not there. I'll submit the latest to the master branch within a few minutes.

fzurita:
@gonetz, do you have a save state for bomber's hide out?

fzurita:
Also, I committed these changes to the latest master branch here:

https://github.com/mupen64plus-ae/mupen64plus-ae/commits/master

gonetz:
>do you have a save state for bomber's hide out?

here:
https://drive.google.com/file/d/0B0YqMPjGo3B2TzF6QWhiQlVqTDQ/view?usp=sharing

gonetz:
@fzurita I noticed strange thing with current master. First, I build from your previous commit, video: GLideN64 optimizations test. I decided to test depth buffer emulation on my Shield tablet. To my surprise, depth buffer worked perfectly (no poke thorough walls) with no slowdown. I checked emulation profile settings and found that depth buffer copy is set off! I know, that it it impossible - coronas will not work without depth copy. I enabled depth copy - the same result. I disabled them again - the same.
I decided to rollback to previous commit, video-gliden64: Update to commit 5d45a45.
That build works the same! The only option, which can change depth buffer copy is 'Disable buffers read/write with FBInfo'. That option was on. Normally, that option being enabled disables support of FBInfo functionality. It added specially for mupen64plus, because FBInfo is broken in it.  I disabled that function and plugin started to work as it should work without depth buffer copy, that is as on my screen shot from adreno: all coronas poke through walls. I enabled depth buffer copy - again, no reaction, all coronas are visible. I set 'Disable buffers read/write with FBInfo' on again - depth buffer copy started to work.
Conclusion: 
* something is wrong with GLideN64 fb settings, at least on Shield.
* my current implementation of depth buffer copy works without noticeable slowdown on Shield (at least for Zelda MM)

gonetz:
I built latest mupen64plus-ae master. No visible changes in depth buffer emulation. No coronas on adreno with depth buffer copy enabled. No problems with depth buffer emulation on Shield.

fzurita:
Thanks for the bug report. It seems like the depth buffer emulation setting has no effect. I will have to investigate. I wonder if somehow I have the Disable FBInfo and Depth buffer emulation settings swapped. This is probably unrelated to this change.

How much of a performance difference are you seeing in your Adreno device with this change?

fzurita:
@gonetz The save state you gave me was for ocarina of time.

fzurita:
@gonetz I found out why you couldn't disable EnableCopyDepthToRDRAM. It's because of GLideN64.custom.ini. It overrides any values picked in the profiles screen.

fzurita:
Ok, no need for a save state. I was able to progress to an area with torches. Depth buffer emulation works correctly in my device. It's only about a10%hit in performance in Android.

gonetz:
I have built latest mupen64plus-ae master. No changes in depth buffer emulation on my adreno: coronas do not work in Zeldas and Perfect Dark when depth buffer copy enabled.

gonetz:
> The save state you gave me was for ocarina of time.

Sorry. It was after midnight for me, so I overlooked that. Here Zelda MM saves, just in case:
https://drive.google.com/file/d/0B0YqMPjGo3B2Tnl4T0ZNTDZvX00/view?usp=sharing

gonetz:
Forgot to say: copy color buffer to RDRAM works fine. Tested with Perfect Dark (blurred pause screen) and with Mario Kart.

fzurita:
Do you see any speed improvements in your Adreno device with the async color buffer to RDRAM?

gonetz:
I can't tell for sure. FPS is unreliable indicator. I want to measure buffer reads with time functions and output result to Android log.

fzurita:
I see a big difference in Banjo Kazzoie in my Adreno device. Same for super smash brothers.

gonetz:
You see depth buffer working, while I see that it does not work at all on my adreno.

BTW, could you give me your apk builds, one for old buffer copy and one with your modifications? To be sure that we testing the same things.

fzurita:
Current build I'm testing with. Based of master branch:

https://drive.google.com/file/d/0B57Ioy26LWegby1fUHktODZOQTg/view?usp=sharing

I'll try to get an earlier one soon.

fzurita:
Ok, here is an earlier one based off this https://github.com/mupen64plus-ae/mupen64plus-ae/commit/8f55f3bcc5dfa61630b0493b8a5f81b7973d1bb9:

https://drive.google.com/file/d/0B57Ioy26LWegQVFqd2NkX2lCZjA/view?usp=sharing

Keep in mind that this one will not gonor settings overriden in GLideN64.custom.ini.

fzurita:
One thing I change that you may not be doing. I always set the polygon offset to -3. It's the only one that works well with me. Try setting that in your device as well. I think I'm going to make that the default since I always find that it works well for all my devices so far.

fzurita:
I think I know of a way to fix many of the nasty #ifdefs in the code. If we switch to a GL function loader, I think can replace them with run-time checks.

With android, it's easy, EGL will do all the work for us for function loading. In addition, EGL supports full desktop openGL on devices that support it, like NVidia Shield.

I'll give it an initial shot by switching Android to support full OpenGL.

gonetz:
Run time checks will not improve readability. #ifdef will be just replaced by if (_suported_) ... else ...
I'll do classes with platform-specific implementation, as it was made for OGLVideo

gonetz:
> Keep in mind that this one will not honor settings overridden in GLideN64.custom.ini.

How to disable custom ini settings?

fzurita:
> How to disable custom ini settings?

There is no way in that build. You will have to use a game that does not have that value overridden.

gonetz:
Yes, Banjo Kazzoie at least twice as faster with new version on my adreno device. Great result!
Depth buffer copy does not work in old and new version: no coronas in Zelda MM and Perfect Dark.
It is not regression though.

gonetz:
I've put this commit to fzurita-copyColorToRDRAMAndroidPerformance dev branch.
I need to think, how to refactor that code.

fzurita:
ok, that's fine.

fzurita:
I added full OpenGL support to the android version. Please see this branch:

https://github.com/mupen64plus-ae/mupen64plus-ae/tree/EGL

Multisampling works in android with this, it seems.

gonetz:
I can't build it with Eclipse:
```
error: cannot find -lGLESv3
jni/./mupen64plus-video-gliden64/src/Combiner.cpp:401: error: undefined reference to 'glGetString'
jni/./mupen64plus-video-gliden64/src/Combiner.cpp:104: error: undefined reference to 'glGetIntegerv'
jni/./mupen64plus-video-gliden64/src/DepthBuffer.cpp:47: error: undefined reference to 'glDeleteFramebuffers'
...
```

fzurita:
Hmm.... I know what's wrong... My commit missed a few files. I'll commit the missing files in a sec.

fzurita:
Ok, branch has been updated.

fzurita:
As I'm thinking about it, using a GL function loader would be very beneficial for android. It would allow us to get rid of the 3 (potentially 4) versions of the plugin in android and replace them with one that does run time checks.

You could force a different OpenGL mode by using run config files instead that GLideN64 reads.

gonetz:
Full OpenGL works on my Shield and speed is nice. Cool!

> As I'm thinking about it, using a GL function loader would be very beneficial for android.

You probably right. I'm not sure how to check on GLES2, but it should be solvable.

> You could force a different OpenGL mode by using run config files instead that GLideN64 reads.

Please explain. I don't understand what is "using run config files instead that GLideN64 reads"

fzurita:
That was a typo on my end. I meant to say that the OpenGL mode ( GLES 2.0/3.0/3.1 or full OpenGL) could be selected using the mupen64plus.cfg file in the [Video-GLideN64] section.

fzurita:
I will make another pull request with Full OpenGL support for android once I clean the code up a little bit.

gonetz:
The main reason that current code is sliced by #ifdef GLES3 conditions is that GLES headers misses many GL functions, which I use on desktop. If it possible to compile for GLES3 using glext.h then it really will be better to replace all GLES specific code with special functions.

fzurita:
We don't have to use those headers. We can have our own custom header that defines everything we need.

loganmc10:
Is this even possible? Don't you still need to link it to the proper GL library when you compile it?

fzurita:
In android, you only need to link against EGL and with that, you can access any version of OpenGL/GLES as long as you use eglGetProcAddress to get to the functions.

fzurita:
Hmmm.... eglGetProcAddress may not be reliable against all Android versions/devices, specially against GLES core functions. dlSym can be used under those circumstances instead though... need to do more research.

gonetz:
@fzurita I've finished buffers copy related code refactoring. Check fzurita-copyColorToRDRAMAndroidPerformance branch. I compiled that version on Windows, Linux and Android, but tested only on Windows. Since I added new files, you need to update mupen64plus-video-gliden64.mk:
```
    $(SRCDIR)/mupenplus/OpenGL_mupenplus.cpp        \
    $(SRCDIR)/BufferCopy/ColorBufferToRDRAM.cpp     \
    $(SRCDIR)/BufferCopy/ColorBufferToRDRAMAndroid.cpp \
    $(SRCDIR)/BufferCopy/DepthBufferToRDRAM.cpp     \
    $(SRCDIR)/BufferCopy/RDRAMtoColorBuffer.cpp     \
    $(SRCDIR)/TextDrawer.cpp                        \

```
Please check. If it works for you, I'll put it to master.

gonetz:
@loganmc10 Please check that it compiles on Raspberry Pi
commit e08f531d1e

loganmc10:
Ok I'll compile it later today and let you know

loganmc10:
I've compiled the latest version of the fzurita-copyColorToRDRAMAndroidPerformance branch on the Raspberry Pi (that includes the render buffer stuff), it all compiles, but I won't have a chance o actually test it out until sometime tomorrow

fzurita:
Just tested. Everything is working as expected in GLES 3.0. I should probably test 2.0 as well.

fzurita:
GLES 2.0 works the same as before as well.

fzurita:
I may have to make another pull request for this if I cant figure out this issue. PowerVR doesn't like how I'm using the OES_EGL_image_external extension. I may have to add the FBO method of reading pixels back just for that device. So we could have three methods to read pixels in Android: Sync, async, and, "External image"?

gonetz:
>> It's good to merge I think.

Good,  thanks! 

>> Sync, async, and, "External image"?

External image also has sync/async modes, has not it?

fzurita:
I don't believe I saw that option. Actually, I'm not entirely sure if the  external image extension is sync or async.

Are there any games that would break easily with the async option? I could use that to test the extension.

gonetz:
F-1 Pole Position
async copy:
f1_pole_position_64-000
sync mode:
f1_pole_position_64-001

gonetz:
> I don't believe I saw that option.

My mistake. Your code uses "External image" when not in sync mode and plain glReadPixels otherwise.

fzurita:
It's definitely async. This game doesn't produce the correct effect in that menu with the "External image" extension.

gonetz:
Ok. Any progress with support of that feature on PowerVR? Current code need to be changed to add possibility to have both PBO and External image.

fzurita:
No, no progress. I think I will wait for you to merge this to master before making any additional changes.

gonetz:
Ok

gonetz:
Done

fzurita:
By the way, PowerVR is very sensitive to shader code changes. After the new blending was introduced, frame rate went down to half of original. I tried using the GLES 2.0 shader code in GLES 3.0 when using PowerVR and it caused Super Mario 64 in game to go from 12fps to 60fps(max) while fast forwarding.

Something in the GLES 3.0 shaders must be causing a massive slowdown with PowerVR. I'm going to try to find what it is. After I fix the async code in that device as well (Nexus Player).

PR #1003, opened 2016-05-28
Fix ReadPixels2 on Android

fzurita:
This seems to fix readScreen2 API call on Android. I used Rice as an example.

As far as I can tell, this should work correctly the original way it's written, but Adreno doesn't like glReadPixels calls that don't match the frame buffer format.


gonetz:
Added

PR #1004, opened 2016-05-28
Add support for Full OpenGL in Android through EGL

fzurita:
This adds support for full OpenGL in the android build for devices that support it, like the NVidia Shield TV or NVidia Shield tablet.

This also needs some changes to the FrameBuffer.h and FrameBuffer.cpp but I can't include them here since it's not in the master branch. The change is to use normal PBOs if in EGL mode.

### DIFF
--- a/src/OpenGL.h
+++ b/src/OpenGL.h
@@ -28,6 +28,13 @@ typedef char GLchar;
 #define GL_IMAGE_TEXTURES_SUPPORT
 #define GL_MULTISAMPLING_SUPPORT
 #define GL_UNIFORMBLOCK_SUPPORT
+#elif defined(EGL)
+#include <GL/glcorearb.h>
+#include "common/GLFunctions.h"
+#include <GL/glext.h>
+#define GL_IMAGE_TEXTURES_SUPPORT
+#define GL_MULTISAMPLING_SUPPORT
+#define GL_UNIFORMBLOCK_SUPPORT
 #else
 #if defined(OS_MAC_OS_X)
 #define GL_GLEXT_PROTOTYPES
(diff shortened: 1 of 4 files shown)


gonetz:
May be it will be better to add fzurita-copyColorToRDRAMAndroidPerformance to master first and then apply these changes over it?

fzurita:
I can make the pull request against that branch if you want. I know you were waiting to do some refactoring.

gonetz:
I'm waiting for your testing and approval:
https://github.com/gonetz/GLideN64/pull/998#issuecomment-222352155

fzurita:
Ok, i will great that when I can. Hopefully by tonight. Also, maybe we should consider making mupen64plus-video-gliden64.mk part of your sources since you have build files for other platforms there add well. The standard format is to have an Android.mk file in your directory.

gonetz:
> Also, maybe we should consider making mupen64plus-video-gliden64.mk part of your sources

I don't mind. You may add it to proper place and make pull request.

fzurita:
Ok, will do.

fzurita:
I did a rebase on my branch. It should be good to go. For the EGL version, I will use ColorBufferToRDRAMDesktop.

gonetz:
I don't like two GLFunctions files: I'll have to update both when necessary. Thus I added a patch to your commit. Please check fzurita-EGL branch.

gonetz:
I've built mupen64plus-ae with code from fzurita-EGL branch. It works. I had to use ColorBufferToRDRAMDesktop.cpp for EGL.

fzurita:
Cool, thanks

PR #1012, opened 2016-06-04
Fix PBOs in Android

fzurita:
I was finally able to get PBOs working in android. The issue is, at least with Adreno, that ```glReadPixels``` with PBOs doesn't like reading parts of the texture. The whole texture must be read, otherwise ```glReadPixels``` will end up converting the data.

Also, this doesn't mean that we should get rid of ColorBufferToRDRAMAndroid.cpp. This is because PBOs don't work with GLES 2.0 but EGLImageKHR will work with GLES 2.0 even in non-Android devices. That method will be useful when copy color to RDRAM is ported to GLES 2.0.

### DIFF
--- a/src/BufferCopy/ColorBufferToRDRAM.cpp
+++ b/src/BufferCopy/ColorBufferToRDRAM.cpp
@@ -9,6 +9,7 @@
 #include <Config.h>
 #include <N64.h>
 #include <VI.h>
+#include <Log.h>
 
 ColorBufferToRDRAM::ColorBufferToRDRAM()
 	: m_FBO(0)
@@ -16,7 +17,12 @@ ColorBufferToRDRAM::ColorBufferToRDRAM()
 	, m_pCurFrameBuffer(nullptr)
 	, m_frameCount(-1)
 	, m_startAddress(-1)
+	, m_lastVIWidth(-1)
+	, m_lastVIHeight(-1)
 {
+	m_allowedRealWidths[0] = 320;
+	m_allowedRealWidths[1] = 480;
+	m_allowedRealWidths[2] = 640;
 }
 
 ColorBufferToRDRAM::~ColorBufferToRDRAM()
@@ -30,6 +36,22 @@ void ColorBufferToRDRAM::init()
 	// generate a framebuffer
 	glBindFramebuffer(GL_DRAW_FRAMEBUFFER, 0);
 	glGenFramebuffers(1, &m_FBO);
+
+	_init();
+}
+
+void ColorBufferToRDRAM::destroy() {
+	_destroyFBTexure();
+	_destroy();
+	glBindFramebuffer(GL_DRAW_FRAMEBUFFER, 0);
+	if (m_FBO != 0) {
+		glDeleteFramebuffers(1, &m_FBO);
+		m_FBO = 0;
+	}
+}
+
+void ColorBufferToRDRAM::_initFBTexture(void)
+{
 	glBindFramebuffer(GL_DRAW_FRAMEBUFFER, m_FBO);
 
 	m_pTexture = textureCache().addFrameBufferTexture();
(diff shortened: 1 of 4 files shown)


gonetz:
You killed copyChunkToRDRAM functionality. The idea was to read only necessary part of the image.
You not only killed that, you also forced to read 640x580 texture, even when actual buffer size is 320x240. Very bad solution.

I allocated 640x580 texture because it is maximal possible frame buffer size. Desktop drivers have no problems with reading part of that texture. If Android drivers can read efficiently only whole texture, you need first to allocate m_pTexture  of VI.width x VI.height size, blit main buffer into it (ColorBufferToRDRAM::_prepareCopy does that) and then read this texture. If you will allocate new m_pTexture only when VI buffer size changed, overheads will be minimal. Also, fbInfoReadColorChunk option must be off.

fzurita:
Yeah, sure, I can make adjustments.

fzurita:
Also, this seems to be the same speed as the EGLImageKHR method.

fzurita:
So, a texture size of 292x212 results in slow PBOs again while a texture size of 640x580 is fine.... Android is way too picky.

fzurita:
I updated my solution. This works ok in Android. As far as I can tell, the height doesn't have to match, only the width.

The only difference now from the original solution is that we will read realWidth instead of _width. The memcpy operation later gets rid of the trailing space at the end of _width.

This should work fine with FBOInfo and it works fine on Android as well.

gonetz:
This is better, but still not good. m_pTexture->realWidth == 640. In most case we will read two times more data than needed. Read from video memory is not free. Also, how this code will work in case of native resolution, when main buffer width is 320?

I suggest to allocate m_pTexture with current buffer size. When buffer size changes, re-allocate m_pTexture . In that case you will not read more than necessary; also you will not read copy data from pixelData to pixelDataAlloc.

fzurita:
You are right, it doesn't work at 1x native resolution. I don't quite understand why... is the _width attribute still valid?

I tried what you are suggesting and it doesn't work well with Android. PBOs worth Adreno seen to only be fast at certain texture widths. A width of 292 for example in Banjo causes the glReadPixels call to stall again. I think the best I can do for Android is to read the whole texture width.

Another option is to have a list of allowed texture widths that are not slow. I can then always pick the smallest texture width possible. There would still be times where we read a little extra, but it won't be much more.

For example, a texture width of 320 is fast in Android. In Banjo Kazzoie, the VI width is 292, so I would pick 320 as the real texture width. In that case I would be reading 28 extra pixels.

gonetz:
I guess, only multiple of 160 work fast: 320, 480, 640.
You may use these values. Note, that games with buffer width less than 320 (e.g. Banjo Kazzoie) will not work in native mode: you can't read more data than buffer actually has.

Another note: copy data from pixelData to pixelDataAlloc still can be avoided.
writeToRdram just needs additional parameter _stride, which defines actual number of texels per line.

fzurita:
For Android, it's better to not pass the mapped memory directly to writeToRdram. This memory is slow. For Android at least, it's better to first copy the memory another buffer and then let writeToRdram perform operations on that buffer.

fzurita:
Ok, pull request has been updated again. Let me know if this is closer to what you want or if I need to make further changes.

fzurita:
What's a game that will switch VI size mid game?

gonetz:
> What's a game that will switch VI size mid game?

Many of them. Pokemon Stadium 1/2 for example.

gonetz:
The code is more complicated and entangled now, but it is ok until it works. Speaking of which: have you tested it with Banjo Kazzoie in native resolution? Will not it crash in that case?

fzurita:
Do you mean 1x native resolution setting? The 2x and up work fine, but with the 1x native resolution setting frame buffer effects like the banjo Kazzoie puzzle pieces and Mario kart screens seem to stop working. It doesn't crash though.

gonetz:
cite: "In Banjo Kazzoie, the VI width is 292"
Minimal width you pass to glReadPixels is 320. That is you read 320 pixels from buffer, which has only 292 pixel width. glReadPixels's specification seems to allow that: "Values for pixels that lie outside the window connected to the current GL context are undefined." However, actual implementation may fail.
The same situation may arise when VI.width is 400 and thus read width is 480.

Please check, does Banjo Kazzoie puzzle work in native mode, if you read VI.width pixels with glReadPixels. Or does it work with the current master.

fzurita:
By native mode, do you mean 1x native resolution?

gonetz:
Yes, 1x native resolution. I tested MK64 on my device with 1x native, monitor does not work.

gonetz:
>Yes, 1x native resolution. I tested MK64 on my device with 1x native, monitor does not work.

Android specific bug: desktop build works correct.

fzurita:
Ok. Can you test these changes on the desktop?

gonetz:
Your changes works on desktop. MK64 monitor works ok in 1x native resolution.
I guess that EGLImageKHR code does not work in 1x.

fzurita:
I tried PBOs and EGLImageKHR, neither work correctly in Android at 1x. Not sure why.

Issue #1010, opened 2016-06-03
A few issues with latest "fix_2D" code


loganmc10:
I get this error when I run it (Raspberry Pi):
```
shader_link error: ERROR:SEMANTIC-52 (fragment shader, line 2) Initializer for uniform
```

It doesn't like these lines:
https://github.com/gonetz/GLideN64/blob/master/src/GLES2/Shaders_gles2.h#L481
https://github.com/gonetz/GLideN64/blob/master/src/GLES2/Shaders_gles2.h#L495

If I take out the "uniform" then it works, but I'm not sure what the ramifications of that are.

Once that is done, I get this error:
```
mupen64plus: /home/pi/GLideN64/src/OpenGL.cpp:403: void OGLRender::TexrectDrawer::init(): Assertion `m_depthScaleLoc >= 0' failed.
```

It looks like that assertion was added in this commit: https://github.com/gonetz/GLideN64/commit/cefffa7c24f0899abda425d974e29afd5005b223 but I haven't been able to figure out how to fix it yet.

loganmc10:
When I run this on a different GLES2 device I get:
```
shader_compile error: 0:37(32): error: cannot initialize uniform uTestColor in GLSL ES 1.00 (GLSL 1.20 required)
```

fzurita:
I think you can fix that by making this modification to the lines you pointed out:
```
"uniform lowp vec4 uTestColor;    \n"
```

You will then have to manually set the uniform from ```ShaderCombiner::update``` method.

From there, you can set it manually to ```(4.0/255.0, 2.0/255.0, 1.0/255.0, 0.0)``` using m_uniforms.uTestColor.set(

loganmc10:
Does it need to be a uniform though? Can't it just be:
```
lowp vec4 uTestColor = vec4(4.0/255.0, 2.0/255.0, 1.0/255.0, 0.0);
```

fzurita:
You are right. I never see that uniform being updated outside the GLSL shader.  It could probably be made into not a uniform.

You should try it!

loganmc10:
I tried it, it worked but now I get:
```
mupen64plus: /home/pi/GLideN64/src/OpenGL.cpp:403: void OGLRender::TexrectDrawer::init(): Assertion `m_depthScaleLoc >= 0' failed.
```
So I'm still trying to figure that part out

gonetz:
[...]

PR #1014, opened 2016-06-06
Fix some assertions on GLES2

loganmc10:
GLSL ES 1.00 doesn't like when you initialize a uniform, and uDepthScale is never used in GLES2, and it was giving me an assertion error

### DIFF
--- a/src/OpenGL.cpp
+++ b/src/OpenGL.cpp
@@ -399,8 +399,10 @@ void OGLRender::TexrectDrawer::init()
 	assert(m_textureBoundsLoc >= 0);
 	m_enableAlphaTestLoc = glGetUniformLocation(m_programTex, "uEnableAlphaTest");
 	assert(m_enableAlphaTestLoc >= 0);
+#ifndef GLES2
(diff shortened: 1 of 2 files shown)


loganmc10:
This fixes https://github.com/gonetz/GLideN64/issues/1010

src/OpenGL.cpp
```diff
@@ -399,8 +399,10 @@ void OGLRender::TexrectDrawer::init()
 	assert(m_textureBoundsLoc >= 0);
 	m_enableAlphaTestLoc = glGetUniformLocation(m_programTex, "uEnableAlphaTest");
 	assert(m_enableAlphaTestLoc >= 0);
+#ifndef GLES2
 	m_depthScaleLoc = glGetUniformLocation(m_programTex, "uDepthScale");
 	assert(m_depthScaleLoc >= 0);
+#endif
 	glUseProgram(0);
```
gonetz: Please move these changes to another commit. Different issues should be fixed by separate commits. Both commits can be in one pull request.

loganmc10:
Ok I've split it into 2 commits

gonetz:
Thanks. Merged.

Issue #1015, opened 2016-06-06
Texture memory leak in some games


loganmc10:
I believe I am experiencing a memory leak in some games, the 2 games I consistently have this problem with are GoldenEye and Worms Armageddon, other games work fine, even if I play them for a long time.

I have tried to adjust the texture cache size, it seems to make no difference. The Raspberry Pi GPU has ~400MB of memory.

When I play Worms Armageddon, the GPU memory on my Raspberry Pi slowly fills up (about 1MB/sec or two). It only happens once I'm "in-game", not in the menu. Once the problem starts, the textures start to swap with each other and everything looks all garbled, here are some pictures from GoldenEye showing what happens (sorry for the cell phone pics, once this happens the program I use to take screenshots no longer works).

![vrhkr3e](https://cloud.githubusercontent.com/assets/848146/15829537/6d4163fe-2bd1-11e6-833c-555b0b6ecea8.jpg)
![46cbfle](https://cloud.githubusercontent.com/assets/848146/15829541/72926b78-2bd1-11e6-938c-eb1b6fec3a23.jpg)

loganmc10:
Perhaps this is somewhat related to this comment? https://github.com/gonetz/GLideN64/pull/745#issuecomment-146402996

I'm not really sure

loganmc10:
By the way, this happens before my device runs out of memory. This is what my GPU details show when the problem happens:

(In this case I gave the GPU 680M of memory, the problem occurred when the free memory reached 342M). When everything is good, the "alloc_fails" number is 0, when the problem begins, alloc_fails starts to climb, and the textures start to get messed up. My texture cache is at the default of 500MB

```
Relocatable heap version 4 found at 0x13400000
total space allocated is 680M, with 678M relocatable, 2.3M legacy and 0 offline
1 legacy blocks of size 2359296

small_allocs                  : 239
allocs                        : 32813
alloc_fails                   : 94860
legacy_block_fails            : 0
compactions                   : 0
discard_compactions           : 0
aggressive_compactions        : 0
aggressive_compaction_waits   : 0
aggressive_compaction_timeouts: 0
locks                         : 0
small_locks                   : 0

free list at 0x3d97d6a0
342M free memory in 17 free block(s)
largest free block is 342M bytes
```

loganmc10:
Ok, so this happens 100% of the time when "allocs" reaches ~32800 (I assume the actual "problem number" is 32767).

[...]

PR #1017, opened 2016-06-07
Fix Texture Cache on Raspberry Pi

loganmc10:
The Raspberry Pi only supports 32768 GPU memory allocations
see https://github.com/raspberrypi/firmware/issues/611

Each texture takes 2 allocs, so the max is around 16000, I've set it to 15000 to be safe

Fixes https://github.com/gonetz/GLideN64/issues/1015

It still seems odd that games are allocating so many textures, they must be duplicates. If there is a better way to solve this problem I am open to it. I can't notice any performance difference once it goes over the 15000 limit (tested GoldenEye and Worms Armageddon)

### DIFF
--- a/src/Textures.cpp
+++ b/src/Textures.cpp
@@ -522,6 +522,16 @@ void TextureCache::destroy()
 
 void TextureCache::_checkCacheSize()
 {
+#ifdef VC
+	if (m_textures.size() > 15000) {
+		CachedTexture& piTex = m_textures.back();
+		m_cachedBytes -= piTex.textureBytes;
+		glDeleteTextures(1, &piTex.glName);
+		m_lruTextureLocations.erase(piTex.crc);
+		m_textures.pop_back();
+	}
+#endif
+
 	if (m_cachedBytes <= m_maxBytes)
 		return;
 


gonetz:
> It still seems odd that games are allocating so many textures, they must be duplicates.

Depends on what you call duplicates. For example, **C**olor **I**ndexed texture is texture + palette. Change palette and you will get different resulted texture.
2D games like Worms Armageddon may use large 2D background images, larger than screen size. You see part of such image at time. That part is sliced on tiles, which fit N64 texture memory. These tiles stored in plugin's texture cahce. When your view is changed and you see another part of the background image, you get another set of tiles. Some of them may overlap with previously allocated tiles, but they are different.

gonetz:
> Each texture takes 2 allocs, so the max is around 16000

Why each texture takes 2 allocs? Where?

loganmc10:
It just seems like Worms Armageddon is allocating new textures not stop, I think if you left the game running it would never stop allocating new textures, but i could be wrong.

I got the 2 allocs number by watching the GPU log, every time a texture is allocated this shows up:
```
[29874] 0x2c551520: used  16K (refcount 1 lock count 0, size    12288, align 4096, data 0x2c552000, d1rual) 'Texture blob'
[29873] 0x2c555540: used  608 (refcount 1 lock count 0, size      540, align    4, data 0x2c555560, d0rual) 'GLXX_TEXTURE_T'
```
I confirmed this by printing out the value of m_textures.size() every time checkCacheSize ran, and it was using 2 allocs for each texture. It might just be a Raspberry Pi thing

gonetz:
Texture cache consist of two containers. When new texture allocated, new element created in both containers. However, GPU memory allocated only once per texture. Thus, if Pi supports 32768 GPU memory allocations, you may set the limit to, say, 30000.

loganmc10:
The 15000 limit is based on experimentation. I tried 30000 at first and I still had the same problem. The problem starts when m_textures.size() reaches around 16500, and the output of "sudo vcdbg recloc stats" on the Raspberry Pi shows about 32800 allocs.

I watched the numbers and the "allocs" field in "vcdbg reloc stats" jumped 2 every time m_textures.size() went up one.

gonetz:
Ok, understand.

loganmc10:
Has been merged

PR #1019, opened 2016-06-08
Update ColorBufferToRDRAMStub.cpp to match recent commits

loganmc10:
This file needs to be updated to fix compiling.


gonetz:
Merged

PR #1027, opened 2016-06-20
Allow OpenGL.h to use libretro GL State Machine

loganmc10:
I am working on porting this (properly) to libretro. Libretro uses its own GL state machine code and GL headers, so this commit create the #ifdef's necessary to get it working, it doesn't affect the current operation of GLideN64 for anyone else.

### DIFF
--- a/src/OpenGL.h
+++ b/src/OpenGL.h
@@ -10,7 +10,10 @@
 #include "winlnxdefs.h"
 #endif
 
-#ifdef GLES2
+#ifdef __LIBRETRO__
+#include <glsm/glsmsym.h>
+#include <GLideN64_libretro.h>
+#elif GLES2
 #include <GLES2/gl2.h>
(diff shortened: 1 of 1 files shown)


gonetz:
So, this is work-in-progress, right?
I'll put it to separate dev branch until all works will be done, ok?

loganmc10:
The port itself is still work-in-progress yes, things unrelated to GLideN64 like savestates and other things need to be sorted out, but this code (the integration of GLideN64 into the libretro core) has been tested.

I was hoping to be able to track the master branch of GLideN64 since this is the only change necessary to get the graphics working for me, but if you want to wait that is OK too. There is no need to put it into a separate branch, I'll just manually copy in my modified OpenGL.h when I compile my project and when it's all done I'll re-submit this request.

gonetz:
Understood.

PR #1031, opened 2016-06-24
Fix framebuffer emulation issues in Raspberry pi and PowerVR devices

fzurita:
These changes seem to fix frame buffer emulation in Raspberry Pi and PowerVR devices.

The glClear fixes PowerVR
The depth buffer invalidate fixes Raspberry Pi.

Edit: Added additional commit that speeds up PowerVR.

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -2,6 +2,10 @@
 #include <math.h>
 #include <algorithm>
 #include <vector>
+#ifdef GLES2
+#define GL_GLEXT_PROTOTYPES
+#include <GLES2/gl2ext.h>
+#endif
 #include "OpenGL.h"
 #include "FrameBuffer.h"
 #include "DepthBuffer.h"
@@ -591,6 +595,11 @@ void FrameBufferList::saveBuffer(u32 _address, u16 _format, u16 _size, u16 _widt
 		} else {
 			m_pCurrent->m_resolved = false;
 			glBindFramebuffer(GL_FRAMEBUFFER, m_pCurrent->m_FBO);
+			
+			//Needed for some devices
+			if (_address == gDP.depthImageAddress)
+				glClear(GL_DEPTH_BUFFER_BIT);
+			 
 			if (m_pCurrent->m_size != _size) {
 				f32 fillColor[4];
 				gDPGetFillColor(fillColor);
@@ -707,6 +716,14 @@ void FrameBufferList::attachDepthBuffer()
 {
 	if (m_pCurrent == nullptr)
 		return;
+
+	//Needed by some devices
+	const GLenum discards[]  = {GL_DEPTH_ATTACHMENT};
+#ifdef GLES2
+	glDiscardFramebufferEXT(GL_FRAMEBUFFER, 1, discards);
+#else
+	glInvalidateFramebuffer(GL_FRAMEBUFFER, 1, discards);
+#endif
 
 	DepthBuffer * pDepthBuffer = depthBufferList().getCurrent();
 	if (m_pCurrent->m_FBO > 0 && pDepthBuffer != nullptr) {
--- a/src/OGL3X/UniformBlock.cpp
+++ b/src/OGL3X/UniformBlock.cpp
@@ -148,7 +148,9 @@ void UniformBlock::setColorData(ColorUniforms _index, u32 _dataSize, const void 
 		glBindBuffer(GL_UNIFORM_BUFFER, m_colorsBlock.m_buffer);
 	}
 
-	if (m_renderer != OGLRender::glrAdreno)
+	if (m_renderer != OGLRender::glrAdreno &&
+		m_renderer != OGLRender::glrPowerVR &&
+		m_renderer != OGLRender::glrMali)
 		glBufferSubData(GL_UNIFORM_BUFFER, m_colorsBlock.m_offsets[_index], _dataSize, _data);
 	else
 		glBufferData(GL_UNIFORM_BUFFER, m_colorsBlockData.size(), m_colorsBlockData.data(), GL_STATIC_DRAW);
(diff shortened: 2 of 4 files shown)


loganmc10:
The Raspberry Pi only supports GLES2, so i don't know if the glInvalidateFramebuffer call is necessary.  If you take that out then the glDiscardFramebufferEXT call is still wrapped around an #ifdef GLES2, it could be #ifdef VC (same with the gl2ext.h include), but I guess we can leave it up to gonetz to decide what he'd prefer.

src/FrameBuffer.cpp
```diff
@@ -591,6 +595,11 @@ void FrameBufferList::saveBuffer(u32 _address, u16 _format, u16 _size, u16 _widt
 		} else {
 			m_pCurrent->m_resolved = false;
 			glBindFramebuffer(GL_FRAMEBUFFER, m_pCurrent->m_FBO);
+			
+			//Needed for some devices
+			if (_address == gDP.depthImageAddress)
+				glClear(GL_DEPTH_BUFFER_BIT);
```
gonetz: It is wrong functionality. If it fixes issues with some devices, enable it for these devices only.

src/FrameBuffer.cpp
```diff
@@ -708,6 +717,14 @@ void FrameBufferList::attachDepthBuffer()
...
 		return;
 
+	//Needed by some devices
+	const GLenum discards[]  = {GL_DEPTH_ATTACHMENT};
+#ifdef GLES2
+	glDiscardFramebufferEXT(GL_FRAMEBUFFER, 1, discards);
+#else
+	glInvalidateFramebuffer(GL_FRAMEBUFFER, 1, discards);
+#endif
+
```
gonetz: It is wrong functionality. If it fixes issues with some devices, enable it for these devices only.

fzurita:
About your comments. I was thinking that the changes don't affect performance or graphics in other devices. So to not have the code littered with a bunch of if(renderer) should we really have those checks?

src/OpenGL.cpp
```diff
@@ -1723,6 +1723,10 @@ void OGLRender::_initExtensions()
 	const GLubyte * strRenderer = glGetString(GL_RENDERER);
 	if (strstr((const char*)strRenderer, "Adreno") != nullptr)
 		m_oglRenderer = glrAdreno;
+	else if (strstr((const char*)strRenderer, "PowerVR") != nullptr)
+		m_oglRenderer = glrPowerVR;
```
gonetz: m_isBufferSubDataSupported = false;

src/OpenGL.cpp
```diff
@@ -1723,6 +1723,10 @@ void OGLRender::_initExtensions()
...
+	else if (strstr((const char*)strRenderer, "Mali") != nullptr)
+		m_oglRenderer = glrMali;
```
gonetz: m_isBufferSubDataSupported

src/OGL3X/UniformBlock.cpp
```diff
@@ -253,7 +263,9 @@ void UniformBlock::updateLightParameters()
 		glBindBuffer(GL_UNIFORM_BUFFER, m_lightBlock.m_buffer);
 	}
 
-	if (m_renderer != OGLRender::glrAdreno)
+	if (m_renderer != OGLRender::glrAdreno &&
+		m_renderer != OGLRender::glrPowerVR &&
+		m_renderer != OGLRender::glrMali)
```
gonetz: too many ifs.
use 
bool OGLRender::isBufferSubDataSupported() {return m_isBufferSubDataSupported;}

src/OGL3X/UniformBlock.cpp
```diff
@@ -148,7 +148,9 @@ void UniformBlock::setColorData(ColorUniforms _index, u32 _dataSize, const void
 		glBindBuffer(GL_UNIFORM_BUFFER, m_colorsBlock.m_buffer);
 	}
 
-	if (m_renderer != OGLRender::glrAdreno)
+	if (m_renderer != OGLRender::glrAdreno &&
+		m_renderer != OGLRender::glrPowerVR &&
+		m_renderer != OGLRender::glrMali)
```
gonetz: too many ifs.
use 
bool OGLRender::isBufferSubDataSupported() {return m_isBufferSubDataSupported;}

src/OGL3X/UniformBlock.cpp
```diff
@@ -231,10 +234,17 @@ void UniformBlock::updateTextureParameters()
...
-		glBufferSubData(GL_UNIFORM_BUFFER, m_textureBlock.m_offsets[tuTexScale], m_textureBlockData.size(), pData);
-	else
-		glBufferData(GL_UNIFORM_BUFFER, m_textureBlockData.size(), m_textureBlockData.data(), GL_STATIC_DRAW);
+	
+	if(temp != m_textureBlockData)
+	{
+		m_textureBlockData = temp;
+		if (m_renderer != OGLRender::glrAdreno &&
+			m_renderer != OGLRender::glrPowerVR &&
+			m_renderer != OGLRender::glrMali)
```
gonetz: too many ifs.
use 
bool OGLRender::isBufferSubDataSupported()

src/OGL3X/UniformBlock.cpp
gonetz: too many ifs.
use 
bool OGLRender::isBufferSubDataSupported()

fzurita:
Ok, I can make that change.

gonetz:
>I was thinking that the changes don't affect performance or graphics in other devices. So to not have the code littered with a bunch of if(renderer) should we really have those checks?

I'm not sure that the changes in FrameBufferList::attachDepthBuffer() will break anything, but changes in FrameBufferList::saveBuffer definitely can break things and thus must not be unconditionally enabled.

fzurita:
Is there a good place I can put the glClear for the depth buffer where it won't break anything?

gonetz:
Buffer must be cleared when N64 program clears it. If you force clear buffer for your own reasons you risks to break things.

fzurita:
Gotcha, when is the depth buffer being cleared right now?

fzurita:
Some of this was made redundant by https://github.com/gonetz/GLideN64/pull/1032.

I'll make a new pull requests that split out some of these commits.

PR #1032, opened 2016-06-25
Enable Framebuffer Emulation on the Raspberry Pi

loganmc10:
renderbuffer code was written by gonetz, not me

Since https://github.com/gonetz/GLideN64/pull/1031 isn't quite ready to merge I though I'd submit this, since it's scope is a little different.

I have tested this using Perfect Park, Pokemon Stadium, Legend of Zelda: OoT and it all works very well.

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -707,16 +707,24 @@ void FrameBufferList::attachDepthBuffer()
 {
 	if (m_pCurrent == nullptr)
 		return;
+#ifdef VC
+	const GLenum discards[]  = {GL_DEPTH_ATTACHMENT};
+	glDiscardFramebufferEXT(GL_FRAMEBUFFER, 1, discards);
+#endif
 
 	DepthBuffer * pDepthBuffer = depthBufferList().getCurrent();
 	if (m_pCurrent->m_FBO > 0 && pDepthBuffer != nullptr) {
 		pDepthBuffer->initDepthImageTexture(m_pCurrent);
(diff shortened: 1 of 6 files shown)


fzurita:
This makes my other pull request kind of redundant. I'll close it and create a different pull request with other specific issues being fixed.

src/FrameBuffer.cpp
```diff
@@ -707,16 +707,24 @@ void FrameBufferList::attachDepthBuffer()
 {
 	if (m_pCurrent == nullptr)
 		return;
+#ifdef VC
+	const GLenum discards[]  = {GL_DEPTH_ATTACHMENT};
+	glDiscardFramebufferEXT(GL_FRAMEBUFFER, 1, discards);
+#endif
```
fzurita: Does this really need to be specific to the Raspberry pi?

loganmc10: It is only known to benefit the Raspberry Pi, gonetz said this in regards to this line in the other pull request:

> It is wrong functionality. If it fixes issues with some devices, enable it for these devices only.

loganmc10:
Just a thought @fzurita , maybe you could try the glDiscardFramebufferEXT fix while using the -DUSE_DEPTH_RENDERBUFFER flag/code? Our problem seemed so similar, that is the only difference can think of in how the 2 devices are doing framebuffer emulation

fzurita:
Yeah, sure, once this pull request gets merged I'll give that a shot

gonetz:
Merged.

fzurita:
Yeah, no luck with the Nexus Player.

PR #1033, opened 2016-06-25
Fix internal depth format for GLES 2.0 Android devices

fzurita:
Fix internal depth format some GLES 2.0 Android devices. This change was originally made for Raspberry Pi, but it doesn't work with all Android GLES 2.0 devices.

Instead I added an #ifdef for VC


gonetz:
Merged

PR #1034, opened 2016-06-26
PowerVR performance improvement

fzurita:
Use BufferData instead of BufferSubData for PowerVR.

Also, only update texture uniform block if it has changed.


gonetz:
Merged

Issue #1037, opened 2016-06-27
Missing GLN64 ARM Neon optimizations


fzurita:
The GLN64 arm neon optimizations are missing. See this file to see what I mean:

https://github.com/mupen64plus-ae/mupen64plus-ae/blob/master/jni/mupen64plus-video-gln64/src/3DMathNeon.cpp

I'm sure these would provide a nice performance boost in Android/Raspberry Pi

Edit: So I don't forget, here is another file that has Neon optimization in GLN64 that are missing from GLideN64.

https://github.com/mupen64plus-ae/mupen64plus-ae/blob/master/jni/mupen64plus-video-gln64/src/gSPNeon.cpp

And here is a listing of all optimizations that GLN64 for Android uses that may be missing from GLideN64 as well: https://github.com/mupen64plus-ae/mupen64plus-ae/blob/master/jni/mupen64plus-video-gln64/Android.mk

    __CRC_OPT
    __HASHMAP_OPT
    __TRIBUFFER_OPT

gonetz:
@fzurita  GLideN64 code already has legacy win32 asm code. I don't mind to add neon asm as well, but I don't have time for that. Could you take this task?

fzurita:
Yep, I was actually planning to do it. I created the issue so I wouldn't forget.

fzurita:
So I don't forget, here is another file that has Neon optimization in GLN64 that are missing from GLideN64.

https://github.com/mupen64plus-ae/mupen64plus-ae/blob/master/jni/mupen64plus-video-gln64/src/gSPNeon.cpp

fzurita:
3DMath neon optimizations were easy to port. I don't see much of an improvement in performance though.

I'm porting the gSP neon optimizations now, those are a little trickier.

fzurita:
The gSP Neon optimizations may actually be helping. Doesn't seem by much though, maybe 5% improvement.

fzurita:
The __TRIBUFFER_OPT optimizations looks interesting, but much harder to port. I'll probably try running GLN64 without them to see if it makes any difference.

gonetz:
@fzurita Neon code is in master. Do you plan to do anything else for this ticket?

fzurita:
I was planning to look at the __TRIBUFFER_OPT compiler define in glN64, but I can make another ticket for that.

PR #1039, opened 2016-06-28
Port GLN64 for Android NEON optimizations

fzurita:
These optimizations provide a small performance boost in Android. They can also be potentially used in the Raspberry Pi with the right CMake changes.

See this issue: https://github.com/gonetz/GLideN64/issues/1037

### DIFF
--- a/src/3DMath.cpp
+++ b/src/3DMath.cpp
@@ -3,26 +3,19 @@
 
 void MultMatrix(float m0[4][4], float m1[4][4], float dest[4][4])
 {
-    int i;
-    for (i = 0; i < 4; i++)
-    {
-        dest[0][i] = m0[0][i]*m1[0][0] + m0[1][i]*m1[0][1] + m0[2][i]*m1[0][2] + m0[3][i]*m1[0][3];
-        dest[1][i] = m0[0][i]*m1[1][0] + m0[1][i]*m1[1][1] + m0[2][i]*m1[1][2] + m0[3][i]*m1[1][3];
-        dest[2][i] = m0[0][i]*m1[2][0] + m0[1][i]*m1[2][1] + m0[2][i]*m1[2][2] + m0[3][i]*m1[2][3];
-        dest[3][i] = m0[3][i]*m1[3][3] + m0[2][i]*m1[3][2] + m0[1][i]*m1[3][1] + m0[0][i]*m1[3][0];
-    }
-}
-
-void MultMatrix2(float m0[4][4], float m1[4][4])
-{
-    float dst[4][4];
-    MultMatrix(m0, m1, dst);
-    memcpy( m0, dst, sizeof(float) * 16 );
+	int i;
+	for (i = 0; i < 4; i++)
+	{
+		dest[0][i] = m0[0][i]*m1[0][0] + m0[1][i]*m1[0][1] + m0[2][i]*m1[0][2] + m0[3][i]*m1[0][3];
+		dest[1][i] = m0[0][i]*m1[1][0] + m0[1][i]*m1[1][1] + m0[2][i]*m1[1][2] + m0[3][i]*m1[1][3];
+		dest[2][i] = m0[0][i]*m1[2][0] + m0[1][i]*m1[2][1] + m0[2][i]*m1[2][2] + m0[3][i]*m1[2][3];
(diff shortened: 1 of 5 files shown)


fzurita:
Also, I converted leading spaces to tabs in files I modified. This makes it seem like I made more changes than I actually did.

fzurita:
Looks like u broke something, lighting is all messed up in Zelda

fzurita:
After learning enough assembly to figure out what was happening, I was able to come up with a fix. O updated the pull request.

fzurita:
Here is the updated Android makefile if you want to try this out:

[mupen64plus-video-gliden64.zip](https://github.com/gonetz/GLideN64/files/340629/mupen64plus-video-gliden64.zip)

gonetz:
I don't like bunch of new asm code in gSP.cpp
Please move it to another file.

fzurita:
Ok

fzurita:
I updated the pull request. retrobenny's issues were cleared by not using the shader cache. We need to do something about to prevent the different OpenGL/GLES versions sharing shader cache.

fzurita:
@gonetz is there anything that I need to do that is holding this up?

gonetz:
I was busy with other things and had no time to check your commit, sorry. I want to build it and test on my devices. BTW, do you have a mupen64plus-ae branch with your Neon fixes?

fzurita:
Sorry, I don't mean to make you feel rushed. Take your time, as long as you are not waiting on me.

Yeah, right here is a branch with the changes: https://github.com/mupen64plus-ae/mupen64plus-ae/commits/neon_test

gonetz:
I have built neon_test branch. Zelda MM screenshot from my device:
screenshot_2016-07-05-16-45-57

retrobenny:
You have been met with a terrible fate.

I wonder what is happening exactly this time? Glad to see something after a handful of slow days.

fzurita:
Ok, thanks for testing that. It looks like one of the functions didn't port over correctly. I just have to figure out which it is and fix it. 

Can you by any chance give me a save state?

fzurita:
It's also possible that GLN64 has a buggy lighting function. Did you have to make any mayor changes to GLN64 for Zelda MM lighting to work correctly?

I guess once I narrow down which function is broken, it will be easy to tell.

gonetz:
Autosave  for Zelda MM (U)

https://drive.google.com/open?
id=0B0YqMPjGo3B2clVzQW1ReWNnTGc

gonetz:
Zelda MM has different lighting function ind

fzurita:
Also, I need to cleanup my tab spacing as well. I see that it looks off in several places.

fzurita:
After replacing gSPLightVertex4 with the default version instead of the neon version, the lighting looks correct again in Zelda MM. Yet, when I look at the implementations, they look the same. Hmmm.... Any idea why the glN64 version of this function wouldn't work?

fzurita:
The more I look at it, the more I'm convincing myself that gSPLightVertex4NEON is wrong. I'm going to remove this neon function.

fzurita:
Pull request updated. Please try the latest version.

gonetz:
> Any idea why the glN64 version of this function wouldn't work?

Actually, I'm not sure that default implementation of gSPLightVertex4 is correct. I don't use it for desktop builds. When I modified gSPLightVertex I did similar changes in gSPLightVertex4 blindly, without tests.

fzurita:
The default version gSPLightVertex4 seems to give correct lighting in the save state you provided.

gonetz:
Good to know. I'm building updated neon branch.

gonetz:
Works ok. Merged.

retrobenny:
Sorry to bring this back open,but I ran into texture issues after the 2nd run with shader storage on a newer build,it still loses proper textures,so the actual entire issues aren't fixed,only some of them.
This time started with OOT file select missing most of its UI,then on reset,the initial logo had a familiar blue rectangle on it as I have seen in previous builds.

I do have a GS code for now if anyone wants to play it without stutters when shader storage is disabled.
It also uses a required delay fix code to prevent crashing and conveniently makes pausing near instant.
You also need audio synced to make it run at the full framerate,but it shows -1 the normal fps too.

Stutter Stopper
80006463 0004
80006493 0004
D01C7021 0002
801DA5CB 0002
D01C7021 0002
801C7021 0003

fzurita:
This could be related to the texture cache changes.

PR #1040, opened 2016-06-28
ColorBufferToRDRAMStub.cpp is actually a GLES2 requirement

loganmc10:
Since Android doesn't use cmake, any GLES2 device is going to require the stub file

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -93,16 +93,6 @@ if(MUPENPLUSAPI)
 	mupenplus/OpenGL_mupenplus.cpp
   )
 
-  if(BCMHOST)
-    set(GLideN64_SOURCES_UNIX ${GLideN64_SOURCES_UNIX}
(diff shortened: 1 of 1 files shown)


gonetz:
Ok.
BTW, color buffer copy currently supported on Android GLES2 devices. It uses GLES specific extension to read data from buffer in async mode. If your device supports this extension, you may try to adopt the code for your system.

loganmc10:
Unfortunately it uses an Android specific extension:

https://www.khronos.org/registry/egl/extensions/ANDROID/EGL_ANDROID_image_native_buffer.txt

There is an equivalent for Broadcom devices I'l probably look into one day

fzurita:
The sync version should actually work fine on his device. As long as he's just uses glReadPixels.

loganmc10:
Probably, I just assumed it would be too slow, I should try it sometime

fzurita:
Yep, it is slow.

PR #1041, opened 2016-06-30
Don't load textures so much (GLES2)

loganmc10:
I've tested this and it seems to work quite well. If you put a counter inside the else{} block you'll see that this happens quite a bit (finding duplicate textures).

It's only enabled for GLES2 right now, and only for regular textures (not highres/filtered), because I only have a GLES2 device to test.

Basically, glTexImage2D is a very expensive call, and it seems to be happening for duplicate textures. The CRC calculations aren't perfect, but they can happen very early so they are helpful. This happens further along the process, once we know the value of pDest. If it determines that pDest is a duplicate, it will "back out" of the texture loading, and just switch to using the "original" texture, thus saving us from having to do glTexImage2D.

They only thing left to implement is deleting the pDest_Locations value inside _checkCacheSize(), to do that I need to add a value to the CachedTexture struct, but it segfaults when I do that, I'm guessing the size of CachedTexture is being assumed somewhere?

I'm hoping someone else can test this to check for any texture corruption/issues. @fzurita I'm especially curious how this does on mobile devices, just anecdotally, the performance in GoldenEye seems to be much better (I need to figure out how to measure FPS on the Raspberry Pi).

### DIFF
--- a/src/Textures.cpp
+++ b/src/Textures.cpp
@@ -835,8 +835,6 @@ void TextureCache::_loadBackground(CachedTexture *pTexture)
 		} else {
 			Textures::iterator iter = locations_iter->second;
 			glDeleteTextures(1, &pTexture->glName);
-			m_lruTextureLocations.erase(pTexture->crc);
-			m_textures.pop_front();
 			*pTexture = *iter;
 			m_cachedBytes -= pTexture->textureBytes;
 		}
@@ -1146,8 +1144,6 @@ void TextureCache::_load(u32 _tile, CachedTexture *_pTexture)
 			} else {
 				Textures::iterator iter = locations_iter->second;
 				glDeleteTextures(1, &_pTexture->glName);
-				m_lruTextureLocations.erase(_pTexture->crc);
-				m_textures.pop_front();
 				*_pTexture = *iter;
 				m_cachedBytes -= _pTexture->textureBytes;
 			}
--- a/src/Textures.cpp
+++ b/src/Textures.cpp
@@ -513,6 +513,7 @@ void TextureCache::destroy()
 		glDeleteTextures( 1, &cur->glName );
 	m_textures.clear();
 	m_lruTextureLocations.clear();
+	pDest_Locations.clear();
 
 	for (FBTextures::const_iterator cur = m_fbTextures.cbegin(); cur != m_fbTextures.cend(); ++cur)
 		glDeleteTextures( 1, &cur->second.glName );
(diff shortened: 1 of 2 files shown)


fzurita:
Any reason why you are only doing it for GLES2? I would imagine this would apply to all OpenGL versions.

gizmo98:
Mupen64plus has a ShowFPS OSD option. It does not work because OGLFT.h uses glu.h:
https://github.com/mupen64plus/mupen64plus-core/blob/master/src/osd/OGLFT.h

loganmc10:
@fzurita I just did the GLES2 code for now because I can only test this on a GLES2 device, if this proves successful then it would make sense to use it all over, although there might be a cleaner solution than expanding every single glTexImage2D, I haven't thought much about it at this point.

Because GLES3+ does mipmapping (basically this stuff: https://github.com/loganmc10/GLideN64/blob/c6c754f609ca9ac144c430dc8aa536288e76b844/src/Textures.cpp#L1155-L1180), the glTexImage2D call doesn't happen exactly the same, so I wasn't sure that the results would be the same just by copying the code.

fzurita:
Gotcha, only GLES3.1+ does mipmapping. I'll try this out tonight and let you know how it goes.

loganmc10:
So after testing this a bit more it's definitely not stable. The part that gets messy is cleaning up the old texture (removing it from m_textures and such). I'll admit I'm not much of a C++ programmer, maybe someone who is a little better at this sort of thing can take a stab at it? I really think there is a pretty good opportunity to improve performance here if this is implemented properly.

fzurita:
Yeah, sounds like you found a good opportunity for optimization. I'll see how much it helps my device and help clean it up a little bit if it makes a big difference.

loganmc10:
I've updated the pull request with a commit that should fix the crashing so you can test the performance difference, the final version (if it is in fact much better) should do that kind of cleanup though.

fzurita:
I'm trying several games, but I'm not seeing a big performance difference. What games have you tried that are faster?

loganmc10:
I'm mostly testing GoldenEye since it seems to load 1000's of textures, try an area where you might run into a ton of baddies or a level that typically has some lag for you (if there are any?)

fzurita:
I get a lot of lag in the first level. This doesn't seem to be making a big difference though. I'm using the FPS counter to tell. I'll play through the first level and see if it feels any faster.

loganmc10:
Maybe it's all in my head, the fact that you can do a trick like this without any visible texture corruption tells me that something must be up though.

fzurita:
I'm actually seeing texture corruption.

fzurita:
The doors in the first level seem to have corrupt textures.

fzurita:
I'm not seeing a big difference. But GoldenEye is a hard game to measure objectively even with an FPS counter.

loganmc10:
Perhaps, maybe this is the wrong way to go, I'll submit a pull request with an improved CTC calculation (stolen from @gizmo98 in the Raspberry Pi's gln64 port) that should improve the texture catching problem and I'm a lot more confident in the stability of that change

loganmc10:
Since this is pretty rough im going to close it and focus on #1043, which should provide some improvement for this issue and is much easier to implement

PR #1042, opened 2016-07-01
Fixes crash at GoldenEye startup

fzurita:
It seems that once the EGLImageKHR buffer is created, it's then managed by Android itself, so it's a bad idea to destroy it.

This fixes a crash in GoldenEye startup.

### DIFF
--- a/src/BufferCopy/ColorBufferToRDRAM_GLES.cpp
+++ b/src/BufferCopy/ColorBufferToRDRAM_GLES.cpp
@@ -26,7 +26,6 @@ private:
 	bool _readPixels(GLint _x0, GLint _y0, GLsizei _width, GLsizei _height, u32 _size, bool _sync)  override;
 	void _cleanUp()  override;
 	void _initBuffers(void) override;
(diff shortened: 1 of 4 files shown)


src/BufferCopy/ColorBufferToRDRAM_GLES.cpp
```diff
@@ -62,17 +62,16 @@ void ColorBufferToRDRAM_GLES::_initBuffers(void)
...
 }
 
 void ColorBufferToRDRAM_GLES::_destroyBuffers(void)
 {
-	if(m_image != 0)
-	{
-	    eglDestroyImageKHR(eglGetDisplay(EGL_DEFAULT_DISPLAY), m_image);
-	    m_image = 0;
-	}
 }
```
gonetz: This makes _destroyBuffers method useless. Please remove it from ColorBufferToRDRAM and all its successors.

fzurita:
Ok, I updated it.

gonetz:
Merged

Issue #997, opened 2016-05-23
GLideN64 Keeps Dumping Duplicate Textures


waltzz92:
Hi, I noticed that the same textures for a game keep getting dumped every time it shows up while playing. It seems the plugin does not check whether is has the texture dumped already. I have observed this in The Legend of Zelda Ocarina of Time, Majora's Mask, and Mario Party.

Steps to reproduce in Ocarina of Time:

1. Enable Texture Dumping on game start. Keep an eye on the dumped texture folder.
2. Observe the textures of Hyrule Field being dumped from the Title screen and when you start the game.
3. End emulation and close Project64 all the way.
4. Repeat steps 1-3. You should see the same textures being dumped, just with a different name.

This is easier to produce in Mario Party. If you play a Mini-Game while Texture Dumping is on, the game's textures are dumped every time you play the Mini-Game.

loganmc10:
Does this happen with every game? I'm experiencing a memory leak on my Raspberry Pi when I play GoldenEye or Worms Armageddon, but not other games. The GPU memory is filling up with textures and eventually I run out of memory, I wonder if it is related to this

waltzz92:
Yes, this happens with every game. When using GLideN64's texture dumping feature, textures are dumped every time they appear in-game, creating many duplicate textures. It makes it difficult to make my own HD Texture Packs.

matthewharvey:
The different names might indicate that the textures have different CRC values, so the emulator is not counting them as the same texture.

Sarkie:
I wonder if 
txHresAltCRC=1
Don't mind looking into this if I have a bit of time
would help with this and #1015 

Are the files exactly the same if you compare by CRC on disk?

loganmc10:
So I tested this by enabling texture dumping on Worms Armageddon, and then doing an md5sum on all the dumped textures and looking for duplicates, there were lots.

Here are the md5sums:
[md5sums.txt](https://github.com/gonetz/GLideN64/files/307227/md5sums.txt)

Here is a list of the duplicate md5's:
[duplicates.txt](https://github.com/gonetz/GLideN64/files/307230/duplicates.txt)

Here is the textures I dumped:
[gliden.tar.gz](https://github.com/gonetz/GLideN64/files/307232/gliden.tar.gz)

Sarkie:
Cool.

Using this 
http://www.freeware-guide.com/rareware/DupDetector.html

[...]

PR #1043, opened 2016-07-01
Faster and less strict CRC calculator

loganmc10:
Taken from https://github.com/ricrpi/mupen64plus-video-gles2n64

Fixes some issues noted in https://github.com/gonetz/GLideN64/issues/1015

I've tested this and I can't see any texture corruption, but i welcome othet testing. This fixes many (but not all) duplicate textures I noticed in issue #1015 

According to @gizmo98 (https://github.com/ricrpi/mupen64plus-video-gles2n64/commit/4ab0d6fe39fc8ffee4de2a4ad8f57b8b39250de1) this should be much faster, that combined with the fact that we aren't loading as many duplicate textures should provide performance improvements. GoldenEye amd Smash Bros seem a little snappier to me using this, but that might be in my head.

### DIFF
--- a/src/CRC.cpp
+++ b/src/CRC.cpp
@@ -6,77 +6,79 @@ unsigned int CRCTable[ 256 ];
 
 u32 Reflect( u32 ref, char ch )
 {
-	 u32 value = 0;
+	u32 value = 0;
 
-	 // Swap bit 0 for bit 7
-	 // bit 1 for bit 6, etc.
-	 for (int i = 1; i < (ch + 1); ++i) {
-		  if(ref & 1)
+	// Swap bit 0 for bit 7
+	// bit 1 for bit 6, etc.
+	for (int i = 1; i < (ch + 1); i++)
+	{
+		if(ref & 1)
 			value |= 1 << (ch - i);
-		  ref >>= 1;
-	 }
-	 return value;
+		ref >>= 1;
+	}
+	return value;
 }
 
 void CRC_BuildTable()
 {
 	u32 crc;
 
-	for (int i = 0; i < 256; ++i) {
+	for (int i = 0; i < 256; i++)
+	{
 		crc = Reflect( i, 8 ) << 24;
-		for (int j = 0; j < 8; ++j)
+		for (int j = 0; j < 8; j++)
 			crc = (crc << 1) ^ (crc & (1 << 31) ? CRC32_POLYNOMIAL : 0);
 
 		CRCTable[i] = Reflect( crc, 32 );
 	}
 }
 
-u32 CRC_Calculate( u32 crc, const void * buffer, u32 count )
+u32 CRC_Calculate(void *buffer, u32 count)
 {
 	u8 *p;
-	u32 orig = crc;
+	u32 crc = 0xffffffff;
 
 	p = (u8*) buffer;
+
 	while (count--)
 		crc = (crc >> 8) ^ CRCTable[(crc & 0xFF) ^ *p++];
 
-	return crc ^ orig;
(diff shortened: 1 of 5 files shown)


gizmo98:
To honour the right one this piece of code comes from Nebuleons glesn64 repo.
https://github.com/Nebuleon/mupen64plus-video-gles2n64/commit/6bea91ef2cdc7daa4e98d0169fcbcd90f13fd051

AmbientMalice:
So this affects the desktop version as well?

loganmc10:
Yes this change would happen for every kind of device

Sarkie:
This affects #997 also ?

loganmc10:
@Sarkie yes it should help, but I don't think it will eliminate the problem

loganmc10:
I've been testing this for a few days now and it does seem to improve performance, I also don't see any issue with the textures.

gonetz:
This commit conflicts with my texture cache fixes I currently working on. Please wait until I complete it. It will take some time because N64 texturing is tricky and any optimization may break things.

I  looked at your code. Basically, you suggest to replace CRC32 by faster checksum calculation algorithm. It may give some speed for mobile CPU, but for desktops there will be no difference. Since CRC32 is a robust algorithm, which proved to be good for N64 texturing, I prefer to keep it. I don't mind if you will do another implementation for functions in CRC.h in another .cpp file and do corrections in the makefile to use that source file instead of CRC.cpp for platforms with ARM cpu.

loganmc10:
Ok no problem, I wasn't aware you are working on this, I'll wait until you merge your changes.

Once @fzurita's Neon changes are merged, I was planning on modifying the Makefile to compile in those files if -DOPT=On is set and a NEON capable CPU is detected. I'll probably submit something similar for the faster CRC algorithm but I'll wait until you update the texture caching.

gonetz:
Done. Most fixes are in commit 5efd89ccd223347 
These changes should reduce number of duplicated textures. I tried to optimize few other things, but other changes eventually cause regressions.

Also, if you want to submit to faster checksum algorithm, it would be right to measure how fast it is.
You may write simple test which calculates checksum of large amount of random data using both methods and measure time required for each of methods. If difference is more than 100% it is worth to use.

loganmc10:
Ok I did a test using this branch:
https://github.com/loganmc10/GLideN64/tree/CRC2

And Google's perftools (https://github.com/gperftools/gperftools) to profile the performance. Using the current CRC algorithm, CRC_Calculate made up 11.2% of the CPU usage. Using the new CRC algorithm it used 2.9% of the CPU time.

I tested using an in-game savestate of Super Smash Bros, I just load it and let it play for 4 minutes without touching the controller or anything. Watching the Pi's CPU usage, using the current CRC, mupen64plus averaged about 65% CPU. Using the new CRC it averaged about 58-59%.

I tested a few different games and I haven't noticed any difference in the textures.

I might get around to writing a more "scientific" test later this week like you suggested, this was just a quick way for me to see what the difference might be in-game.

gonetz:
Please test this algorithm with Knife Edge - Nose Gunner. Ingame text in this game is a good test for checksum algorithm. If you will not find errors in text messages, the algorithm is ok, most likely.

loganmc10:
snapshot

It looks good to me, interestingly that game makes the Framebuffer depth problems on the Raspberry Pi resurface. Every other game I've tried works well, except sometimes Goldeneye will have those issues (you see the sky everywhere) after a while. Anyway that problem showed up using both CRC methods, so it is unrelated.

gonetz:
Ok

PR #1050, opened 2016-07-06
Add NEON code to cmake system

loganmc10:
This allows cmake to use the new neon code if GLideN64 is compiled with "OPT" set to on. It assumes that all ARM systems have neon, is that a fair assumption?

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -178,9 +178,25 @@ FIND_PACKAGE( Freetype REQUIRED )
 include_directories( ${FREETYPE_INCLUDE_DIRS} )
 
 if(OPT)
-	add_definitions(
-	-D__VEC4_OPT
-	)
+  add_definitions(
+    -D__VEC4_OPT
+  )
+  if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+    EXEC_PROGRAM(cat ARGS "/proc/cpuinfo" OUTPUT_VARIABLE CPUINFO)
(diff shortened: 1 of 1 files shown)


dankcushions:
pi 1 and zero don't have neon. not sure about other arm devices. mupen64plus makefile uses param "NEON=1" to handle neon specific stuff so maybe that could work?

loganmc10:
That would work, its just a lot of cmake flags, there are people out there that are doing automatic neon detection in cmake like this:

https://github.com/torch/torch7/blob/master/lib/TH/cmake/FindARM.cmake

@gonetz what are your thoughts?

gonetz:
I think you must check for neon support with cmake to avoid troubles.

fzurita:
Another option is to always compile neon and do a run time check.

dankcushions:
if it's useful, pcsx has a simple check in the configure file that could be used maybe: https://github.com/notaz/pcsx_rearmed/blob/master/configure#L214
```
have_arm_neon=`check_define __ARM_NEON__ && echo yes/`
```

gizmo98:
Is this working? According to cmake documentation CMAKE_SYSTEM_PROCESSOR is "uname -p". raspbian uname -p reports "unknow".

loganmc10:
@gizmo98 It works on my system, the neon files get compiled in. "uname -p" also returns "unknown", so it must not be using that. I am using Arch however, not Raspbian.

Either way, I'm going to get rid of that anyway and replace it with a NEON test.

loganmc10:
Ok I've updated the pull request with a check that specifically checks for a neon capable processer

gonetz:
Merged.

PR #1051, opened 2016-07-07
Fixes corrupt graphics in PowerVR

fzurita:
Gonetz, would this be an ok place to clear the depth buffer without negatively affecting the N64 depth buffer? This fixes PowerVR graphics as well.

### DIFF
--- a/src/DepthBuffer.cpp
+++ b/src/DepthBuffer.cpp
@@ -398,7 +398,10 @@ void DepthBufferList::saveBuffer(u32 _address)
 
 		m_pCurrent = &buffer;
 	}
-
+#ifdef ANDROID
+	//Fixes issues with PowerVR devices and potentially other Android devices
+	glClear( GL_DEPTH_BUFFER_BIT );
+#endif
 	frameBufferList().attachDepthBuffer();
 
 #ifdef DEBUG


loganmc10:
I haven't compared the difference, but it seems like glClear isn't always trivial:

http://www.mesa3d.org/brianp/sig97/perfopt.htm#Clearing

https://www.opengl.org/discussion_boards/showthread.php/181833-Performance-problem-with-glClear(GL_DEPTH_BUFFER_BIT)

It should probably be device specific. I tested this with the Raspberry Pi and it doesn't fix the Framebuffer Emulation graphics issue for that device.

fzurita:
Yeah, I can make this device specific to PowerVR, but still, I don't want to ruin graphics for PowerVR at the same time.

gonetz:
> Gonetz, would this be an ok place to clear the depth buffer without negatively affecting the N64 depth buffer? 

We already discussed that. Plugin must clear buffer only when game clears it via FillRect command. Your fix **may** break things. I can't say that it **will** break things, testing needed.

What if you will clear buffer only for new depth buffer textures, that is
```
	buffer.initDepthBufferTexture(pFrameBuffer);
        glClear(GL_DEPTH_BUFFER_BIT);

```
will it help?

fzurita:
Nope, it doesn't help. Really I'm just trying to find the least intrusive way to insert the glClear.

gonetz:
I wonder why glClear helps. It looks suspicious. May be you need to clear main depth buffer on PowerVR to see anything? It is cleared when fb emulation off, but not touched when it is on.

fzurita:
That is possible. Is this depth buffer used for anything when software depth buffer emulation is enabled?

gonetz:
Software render does not use any GL functions.  

>if depth buffer emulation is disabled, is the depth buffer used?

I don' understand, what you mean by depth buffer emulaton disabled. There is an option to disable depth buffer copy to RDRAM, but depth buffer still will be used. There is N64 depth compare option, which disables use of standard depth buffer.

> glFinish also fixes the problem. These tells me that I'm purely working around driver issues.

If it is really drivers issue, the fix must be GPU specific.

fzurita:
I guess what I mean is that if we are emulating the N64 depth buffer purely in software, like in software copy depth buffer to RDRAM, do we still need to do anything with the GPU depth buffer?

gonetz:
Software depth buffer has no relation to GPU. Of course, we still need depth buffer for depth compare on GPU level.

fzurita:
Ok, I understand.

fzurita:
I updated the pull request. I left it as an ANDROID #ifdef. That way if something breaks due to this, I should see it quickly.

gonetz:
Merged

PR #1053, opened 2016-07-07
Add Android make files

fzurita:
Add Android make files. I'm getting tired of having to recreate the make files whenever I switch to older versions. This would allow for switching version while at the same getting the corresponding Android make files for that version.


gonetz:
Merged

Issue #1011, opened 2016-06-04
PowerVR corrupt graphics


fzurita:
@gonetz, so while trying to debug PowerVR issues, I updated the Nexus Player to Android N preview. After doing this, it must had updated the PowerVR drivers because I'm seeing a brand new issue. See this screenshot:

![super_mario_64-000](https://cloud.githubusercontent.com/assets/11581687/15797455/04a5150c-29e7-11e6-98e3-d152511b4b95.png)

I have tried GL ES 2.0/3.0/3.1 with no effect. It seems like only 2D is rendered correctly. ADB has no errors. I have the feeling that is has to do with depth since the tile based rendered will not draw any polygons that are behind everything else.

Other plugins seem to work ok. Any ideas as to what I could try?

gonetz:
First, you may test that this is depth buffer related. Disable depth compare:
https://github.com/gonetz/GLideN64/issues/994#issuecomment-222589793

Trees and grass textures look very garbled. May be color/alpha dither or alpha compare work wrong. Try to disable noise. You may also comment alpha compare in shaders. In ShaderUtils.cpp remove
```
	_strShader.append(
		"  if (uEnableAlphaTest != 0) {							\n"
...

```

fzurita:
Sure, I'll try those things.

fzurita:
Ok, I tried what was in the comment in issue #994, no luck. Also tried disabling noise, also no luck. Disabling alpha compare didn't do much either. I guess I'll try stripping the shader to the bare minimum and see if I can get any geometry to show.

fzurita:
My shader is down to this and it's still the same issue:
```
void main()						
{							
  fragColor = vShadeColor;	
}                        
```

fzurita:
Ok, I don't know why I didn't check it earlier, but once I disable frame buffer emulation completely, the issue goes away.

loganmc10:
I wonder if this is the same issue as I'm having in https://github.com/gonetz/GLideN64/issues/994. If you look at first screenshots I posted you see that 2D elements like that are rendered fine, and it only happens with Framebuffer emulation (althoufh it's a little different because I am using that depth renderbuffer stuff)

fzurita:
It's possible, I'll try to narrow down which call is causing the issue, once i find it, you can try it as well.

fzurita:
Ok, I found the specific line that is causing this issue... It's this one:

[...]

PR #1055, opened 2016-07-07
Fix some more graphical glitches with FBE on the Raspberry Pi

loganmc10:
On the Raspberry Pi some games were still experiencing issues as described in:

https://github.com/gonetz/GLideN64/issues/1011

For instance:
https://github.com/RetroPie/RetroPie-Setup/pull/1563
https://github.com/gonetz/GLideN64/pull/1043#issuecomment-230656938

Adding these lines clears up all the graphical glitches. All 3 instances (these 2 and the original one) of ```glDiscardFramebufferEXT(GL_FRAMEBUFFER, 1, discards);``` are required or some portion of the screen won't display correctly.


gonetz:
Merged

PR #1056, opened 2016-07-08
Add optimized CRC calculator when OPT is on

loganmc10:
Since this algorithm hasn't been tested extensively I thought it would make sense to only enable it if OPT is on.

Original discussion about this change was here:

https://github.com/gonetz/GLideN64/pull/1043

In my testing on the Raspberry Pi this resulted in a reduction in CPU usage of about 6-7%.

@gonetz If you would prefer that this be enabled just for ARM systems instead of OPT I'm fine with that as well.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -181,6 +181,12 @@ if(OPT)
   add_definitions(
     -D__VEC4_OPT
   )
+  list(APPEND GLideN64_SOURCES
+    CRC_OPT.cpp
+  )
+  list(REMOVE_ITEM GLideN64_SOURCES
+    CRC.cpp
+  )
   if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
     EXEC_PROGRAM(cat ARGS "/proc/cpuinfo" OUTPUT_VARIABLE CPUINFO)
     STRING(REGEX REPLACE "^.*(neon).*$" "\\1" NEON_THERE ${CPUINFO})
--- a/src/CRC_OPT.cpp
+++ b/src/CRC_OPT.cpp
@@ -1,0 +1,41 @@
+#include "Types.h"
+
+void CRC_BuildTable()
+{
+}
+
+u32 CRC_Calculate( u32 crc, const void * buffer, u32 count )
+{
+        unsigned int i;
+        u32 *data = (u32 *) buffer;
+
+        count /= 4;
+        for(i = 0; i < count; ++i) {
+                crc += data[i];
+                crc += (crc << 10);
+                crc ^= (crc >> 6);
+        }
+
+        crc += (crc << 3);
+        crc ^= (crc >> 11);
+        crc += (crc << 15);
+        return crc;
+}
+
+u32 CRC_CalculatePalette( u32 crc, const void * buffer, u32 count )
+{
+	unsigned int i;
+	u16 *data = (u16 *) buffer;
+
+	count /= 4;
+	for(i = 0; i < count; ++i) {
+		crc += data[i << 2];
+		crc += (crc << 10);
+		crc ^= (crc >> 6);
+	}
+
+	crc += (crc << 3);
+	crc ^= (crc >> 11);
+	crc += (crc << 15);
+	return crc;
+}


gonetz:
I don't mind it to be enabled for OPT parameter.
I don't like that it is implemented via #ifndef __CRC_OPT #else #endif
This code slicing reduces its readability and I'm trying to avoid it.
Please move alternative implementation of checksum functions to separate .cpp file and include it instead of CRC.cpp when OPT enabled.

loganmc10:
Ok I've updated the pull request

gizmo98:
Is it possible to add crc32 cpu instructions to speedup things? The raspberry pi 3 armv8 core should support crc32 and crc32c instructions. PCs with SSE4.2 should have crc32c instructions as well.  
https://wiki.linaro.org/LEG/Engineering/OPTIM/CRC
http://stackoverflow.com/questions/31184201/how-to-implement-crc32-taking-advantage-of-intel-specific-instructions

gizmo98:
Or crc intrinsics: 
https://gcc.gnu.org/onlinedocs/gcc-4.9.0/gcc/ARM-ACLE-Intrinsics.html
uint32_t __crc32b (uint32_t, uint8_t) 
uint32_t __crc32h (uint32_t, uint16_t) 
uint32_t __crc32w (uint32_t, uint32_t) 
uint32_t __crc32d (uint32_t, uint64_t) 

https://gcc.gnu.org/onlinedocs/gcc-4.9.1/gcc/X86-Built-in-Functions.html#X86-Built-in-Functions
unsigned int __builtin_ia32_crc32qi (unsigned int, unsigned char)
unsigned int __builtin_ia32_crc32hi (unsigned int, unsigned short)
unsigned int __builtin_ia32_crc32si (unsigned int, unsigned int)
unsigned long long __builtin_ia32_crc32di (unsigned long long, unsigned long long)

gizmo98:
Just wrote something:
https://github.com/gizmo98/GLideN64/commit/287c7a6023d178d07c3752ef7aaec05eec03fbaf

Needs  `-march=armv8-a+crc` instead of `-mcpu=cortex-a53`. Seems to run. No glitches in Zelda OoT.

gizmo98:
This rough implementation is 25% faster than the normal implementation. Don't know if it is worth to add it.

loganmc10:
I think this is a good idea. It should replace the current CRC_OPT, the current one isn't stable enough. It causes some graphical glitches and some games dont even start, like Perfect Dark. It complains about some microcode error if I remember right

gizmo98:
I tried to implement the linux kernel solution because it uses the eight byte intrinsic as long as possible. You need less loops to compute a crc. It runs even slower because __get_unaligned_leXX(p) functions seem to take too much time shifting bytes to the right position. Without all these alignment functions it was 3.5x faster but the crc was always wrong. So the easiest solution would be to implement just __crc32b(x,y).

```
u32 CRC_CalculateARM( u32 crc, const void * buffer, int count )
{
        u8 *p;
        u32 orig = crc; 

        p = (u8*) buffer;
        while ((count -= 8) >= 0) {
                crc = __crc32d(crc, __get_unaligned_le64(p));
                p += 8;
        }

        if (count == 4) {
                crc = __crc32w(crc, __get_unaligned_le32(p));
                p += 4;
        }
        if (count == 2) {
                crc = __crc32h(crc, __get_unaligned_le16(p));
                p += 2;
        }
        if (count == 1) 
                crc = __crc32b(crc, *p);

        return crc ^ orig;
}

http://lxr.free-electrons.com/source/tools/include/tools/le_byteshift.h#L50

**Comment by gizmo98** [comment]
This one runs quite good. 1.7x faster. Does not use u64 intrinsics and alignment functions. 
 
```
u32 CRC_CalculateARM( u32 crc, const void * buffer, int count )
{
        u8 *p;
        u32 orig = crc;

        p = (u8*) buffer;
        while ((count -= 4) >= 0) {
                crc = __crc32w(crc, (p[0]  | p[1] << 8 | p[2] << 16 | p[3] << 24) );
                p += 4;
        }
        if (count == 2) {
                crc = __crc32h(crc, (p[0]  | p[1] << 8) );
                p += 2;
        }
        if (count == 1)
                crc = __crc32b(crc, *p);

        return crc ^ orig;
}
```

loganmc10:
Nice, if it doesn't cause any glitches I would submit a pull request getting rid of CRC_OPT and adding this to NEON_OPT or something

gizmo98:
I tested it a little bit and had always the same CRC as the pure C implementation.

It only works if the cpu supports a crc extensions. So rpi2 will not benefit from this feature. rpi3 does only benefit if -march=armv8-a+crc is set (enables __ARM_FEATURE_CRC32 macro). Don't set -mcpu=cortex-a53 because gcc complains this cpu does not support crc. SSE4.2 crc intrinsics could also be implemented (just replace arm intrinsics with sse4.2 intrinsics).

gizmo98:
@loganmc10 could you test if cpu usage is lower under real conditions? You can checkout my `ARM-CRC-OPT` branch and compile it with `-march=armv8-a+crc -mtune=cortex-a53`.

PR #1058, opened 2016-07-09
cmake: separate optimization options

gizmo98:
The new CRC optimization breaks Zelda OoT fairy glow on a rpi. To get at least some optimizations it would be better to enable/disable them separately.
dsc01628


loganmc10:
You should update the documentation in https://github.com/gonetz/GLideN64/blob/master/projects/cmake/Readme.txt as well

gizmo98:
Ok. done.

PR #1059, opened 2016-07-09
Fixes menu on Castlevania and Rayman 2 with FBE enabled on Raspberry Pi

loganmc10:
Probably fixes other FBE issues as well

Sorry I know this might seem a bit random but it is based on experimentation based on user testing, see:

https://github.com/RetroPie/RetroPie-Setup/pull/1564


gonetz:
This commit reverts your changes from f2babed36d2c0
Are you sure?

loganmc10:
Yes, I have been testing this with another user here:

https://github.com/RetroPie/RetroPie-Setup/pull/1564

It's still not 100%, but it is some kind of GPU driver bug so it's hard to get it 100% fixed, and this commit makes it much better.

gonetz:
Ok. Merged

Issue #1063, opened 2016-07-20
__VEC4_OPT creates incorrect lighting for CBFD


fzurita:
With __VEC4_OPT enabled:
![conker_bfd-000](https://cloud.githubusercontent.com/assets/11581687/16975347/6031d518-4e13-11e6-9674-caae04246e5b.png)

Without __VEC4_OPT:
![conker_bfd-001_correct](https://cloud.githubusercontent.com/assets/11581687/16975503/2dcb843c-4e15-11e6-9bb8-ea78bbe47ad7.png)

Check the tail.

Also, this game uses special versions of the gSPLightVertex4 and gSPPointLightVertex4 which don't have any neon optimizations.

I haven't noticed this with any other game yet.

gonetz:
I guess, I made a mistake in gSPPointLightVertex4 or gSPLightVertex4 for this game.

fzurita:
Yeah, I'm looking into it. I'm not seeing any obviously incorrect differences from the non-optimized versions of these functions.

gonetz:
BTW, do you see any positive effect from VEC4 optimization?

fzurita:
The neon optimizations require this enabled and I do hear very good performance improvements with the neon optimizations enabled, but only under specific circumstances.

fzurita:
I do see difference that doesn't make sense in gSPCBFDVertex, 

```
vtx.a = vertex->color.a * 0.0039215689f;
```

This is being set twice.

fzurita:
It definitely has to do with this piece of code:

```
#ifdef __VEC4_OPT
		for (; i < n - (n%4) + v0; i += 4) {
			u32 v = i;
			for(int j = 0; j < 4; ++j) {
				SPVertex & vtx = render.getVertex(v+j);
				vtx.x = vertex->x;
				vtx.y = vertex->y;
				vtx.z = vertex->z;
				vtx.s = _FIXED2FLOAT( vertex->s, 5 );
				vtx.t = _FIXED2FLOAT( vertex->t, 5 );
				if (gSP.geometryMode & G_LIGHTING) {
					const u32 normaleAddrOffset = ((v0+v+j)<<1);
					vtx.nx = (float)(((s8*)RDRAM)[(gSP.vertexNormalBase + normaleAddrOffset + 0)^3]);
					vtx.ny = (float)(((s8*)RDRAM)[(gSP.vertexNormalBase + normaleAddrOffset + 1)^3]);
[...]

PR #1064, opened 2016-07-20
Fix __VEC4_OPT issue for CBFD in gSPCBFDVertex.

fzurita:
This fixes incorrect lighting with VEC4_OPT enabled in CBFD.

This addresses this issue:

https://github.com/gonetz/GLideN64/issues/1063


fzurita:
I checked other places where the __VEC4_OPT was used and everything else looks correct at first glance.

PR #1068, opened 2016-07-22
RPi HW Lighting

OrN:
Fixes hw lighting for RPi.

It doesn't seem to work on some games with 3 point filtering enabled and legacy blend mode disabled, I'm guessing this is because of the shader being to long?


loganmc10:
How does this perform? This is a pretty complicated shader for the rpi

OrN:
I didn't notice any performance decrease while playing all of my games, but I havn't tested it further than that

PR #1069, opened 2016-07-23
Fixes some depth issues in Android

fzurita:
The previous change was causing depth issues in Android. Moving the glClear to saveBuffer seems to fix those issues.

### DIFF
--- a/src/DepthBuffer.cpp
+++ b/src/DepthBuffer.cpp
@@ -398,10 +398,7 @@ void DepthBufferList::saveBuffer(u32 _address)
 
 		m_pCurrent = &buffer;
 	}
-#ifdef ANDROID
-	//Fixes issues with PowerVR devices and potentially other Android devices
-	glClear( GL_DEPTH_BUFFER_BIT );
-#endif
+
 	frameBufferList().attachDepthBuffer();
 
 #ifdef DEBUG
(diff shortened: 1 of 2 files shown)


fzurita:
Hmm, Zelda OOT was having depth issues in the previous location of the glClear. In the new location now CBFD has depth issues. I need to find a better place for this that won't cause issues.

fzurita:
Ok, this is the strangest driver bug ever. I must clear the depth buffer with glDepthMask( FALSE ) and then again with glDepthMask( TRUE ) otherwise it doesn't work. Anyways.... this fix works for everything and it doesn't produce depth buffer issues anywhere it seems.

fzurita:
@gonetz Ok, this is the strangest driver bug ever. I must clear the depth buffer with glDepthMask( FALSE ) and then again with glDepthMask( TRUE ) otherwise it doesn't work. Anyways.... this fix works for everything and it doesn't produce depth buffer issues anywhere it seems (That I have found yet).

fzurita:
@loganmc10 I found the problem with my driver at least that is causing the depth issues. Look at the change I made. I wonder if you have a similar problem with your driver. You could try adding the discard right in the same place where I clear the depth buffer.

loganmc10:
Nope, doesn't fix the problem for me on the Raspberry Pi (I tried the discard statement and your GL clear, doesn't work)

fzurita:
Ok. It sounds like the raspberry pi is having problems clearing  the depth buffer at the right time as well. I wonder if there is a more controlled way to do a depth buffer clear on the raspberry pi.

PR #1076, opened 2016-07-29
Fix shader compilation error in GLES 3.0 mode.

fzurita:
This fixes a shader compilation error in GLES 3.0 mode.


gonetz:
Merged

PR #1077, opened 2016-07-29
Don't allow different OpenGL versions to share shader cache

fzurita:
Don't allow different OpenGL versions to share shader cache. This is causing some graphic glitches in the Shield TV.

### DIFF
--- a/src/Combiner.cpp
+++ b/src/Combiner.cpp
@@ -351,7 +351,16 @@ void getStorageFileName(wchar_t * _fileName)
 		if (osal_mkdirp(strShaderFolderPath) != 0)
 			pPath = strCacheFolderPath;
 	}
-	swprintf(_fileName, PLUGIN_PATH_SIZE, L"%ls/GLideN64.%08lx.shaders", pPath, std::hash<std::string>()(RSP.romname));
+
+#ifdef GLES3
+	const wchar_t* strOpenGLType = L"GLES3";
+#elif GLES3_1
+	const wchar_t* strOpenGLType = L"GLES3_1";
+#else
+	const wchar_t* strOpenGLType = L"OpenGL";
+#endif
+
+	swprintf(_fileName, PLUGIN_PATH_SIZE, L"%ls/GLideN64.%08lx.%ls.shaders", pPath, std::hash<std::string>()(RSP.romname), strOpenGLType);
 }
 
 u32 CombinerInfo::_getConfigOptionsBitSet() const


fzurita:
Also, I see that glGetString(GL_VERSION) is part of the shader validity check, but this is always returning full opengl for the Shield TV it seems.

fzurita:
Another problem I'm seeing. The ```operator<<``` for GLSLCombiner_ogl3x is returning early on the very first shader that it writes. 

For some reason the first glGetProgramBinary call is producing a isGLError of true.

fzurita:
This commit allows all shaders that succeed to still be loaded.

fzurita:
Ok, one more commit. With this last one, we will save the correct length of number of shaders saved if one fails to generate a program binary.

gonetz:
>  if one fails to generate a program binary.

How it is possible?

fzurita:
I'm not sure how this is happening. But it is happening on the new Android device that I have. It may be related to the failed compilation error that I fixed on a separate PR.

fzurita:
Still, should we write incorrectly if we fail to compile one shader?

Also, should we throw away all the shaders that we read from the file if we fail to find the number of shaders stated on the file?

gonetz:
I never met situation when shader's binary representation does not work for some shaders. It either works for everything or does not work at all. Thus the code written that way. If situation, when some shaders can fail to be stored, is valid, your corrections are necessary. The only notice: old code based on  symmetric operators '<<' and '>>'. It was easy to control what is written and what should be loaded. You changed only write function and symmetry is gone. Please rewrite read function the same way.

fzurita:
One thing I notice is that it's always only the first compiled shader that does this. I don't understand why.

fzurita:
Pull request updated.

gonetz:
> One thing I notice is that it's always only the first compiled shader that does this. I don't understand why.

Let me guess: you got isGLError triggered after the first glGetProgramBinary, right?
In that case most likely the error occurred **before** glGetProgramBinary call. GL error remains triggered until glGetError call. Try to put assert(!isGLError()) at the beginning of  operator<< and before glGetProgramBinary to check it.

gonetz:
I usually put if (isGLError) to RSP_ProcessDList() after
`			GBI.cmd[RSP.cmd](RSP.w0, RSP.w1);
`
that way you mas see which command produced the error and put isGLError inside that command.

fzurita:
Ok, I will try that tomorrow night. Tonight I ran out of time. It would be good to find the root cause of the problem.

gonetz:
Yep. #1080 probably will be fixed also.

fzurita:
I'm hoping.

fzurita:
So I found the source of the GL error. It's this specific function in F3D.cpp

```F3D_Tri1( u32 w0, u32 w1 )```

**Comment by fzurita** [comment]
Ok. I don't the the GL Errors have anything to do with the hardware lighting saved shader not loading correctly. The issue is fixed if I revert the return statement of _loadShadersStorage to ``` !isGLError() ```

Even though isGLError is returning false....

**Comment by gonetz** [comment]
I don't understand, what is going on. Why we load corrupt shader? How the shader got corrupted?

File: src/Combiner.cpp [hunk 1]
```diff
@@ -422,6 +422,9 @@ void CombinerInfo::_saveShadersStorage() const
 
 	u32 totalWritten = 0;
 	std::vector<char> allShaderData;
+	
+	//Clear the GL error flag so that follow up check don't falsely return errors.
+	isGLError();
```
gonetz: We need to find, what caused glError , not just clear it.

fzurita:
Not sure. The load _loadShadersStorage function is returning a GL error because of this line in ShaderCombiner::updateFrameBufferInfo:

```
	m_uniforms.uMSTexEnabled.set(nMSTex0Enabled, nMSTex1Enabled, _bForce);
```

Also, note, everything works correctly with Hardware lightning disabled.

src/Combiner.cpp
fzurita: @gonetz With hardware lighting enabled, it's this causing the error flag:

```
void UniformSet::_updateLightUniforms(UniformSetLocation & _location, bool _bForce)
{
	for (s32 i = 0; i <= gSP.numLights; ++i) {
		_location.uLightDirection[i].set(&gSP.lights[i].x, _bForce);
		_location.uLightColor[i].set(&gSP.lights[i].r, _bForce);
	}
}
```

I did not try to dig to see what is causing the error flag hardware lighting is disabled.

fzurita:
I'm thinking I could remove the clearing of the GL error flag from this pull request. I can make another pull request to actually find all the sources of GL Errors. What do you think?

gonetz:
I hope that GL error is not related to shader storage functionality and when you will find the source of GL Error, that change will not be necessary.

gonetz:
> Can you try this pull request with shader storage enabled in your Adreno device? Do you get the same issue with hardware lighting enabled?

Next next week. I'll be offline again for a week.

fzurita:
I found the fix. Correcting the GL error in ''''UniformSet::_updateLightUniforms''' fixes the problem with shader storage and hardware lighting.

I now have to find the source of the GL error when  hardware lighting is disabled.

fzurita:
Ok, I finally found all the problems. I made a separate pull request:

https://github.com/gonetz/GLideN64/pull/1083

I still think that the first commit is valid. Should I rebase to remove all follow up commits?

fzurita:
Ok, this pull request is just one commit now.

gonetz:
Merged: c17b1f2e111db

PR #1082, opened 2016-08-08
Fix GL Error on GLES3

loganmc10:
uMSTexEnabled is only used when GL_MULTISAMPLING_SUPPORT is set, and it is not supported with GLES3. This line was throwing a GL Error since uMSTexEnabled wasn't present in the shader for GLES3

### DIFF
--- a/src/OGL3X/GLSLCombiner_ogl3x.cpp
+++ b/src/OGL3X/GLSLCombiner_ogl3x.cpp
@@ -745,8 +745,9 @@ void ShaderCombiner::updateFrameBufferInfo(bool _bForce) {
 	}
 	m_uniforms.uFbMonochrome.set(nFbMonochromeMode0, nFbMonochromeMode1, _bForce);
 	m_uniforms.uFbFixedAlpha.set(nFbFixedAlpha0, nFbFixedAlpha1, _bForce);
+#ifdef GL_MULTISAMPLING_SUPPORT
 	m_uniforms.uMSTexEnabled.set(nMSTex0Enabled, nMSTex1Enabled, _bForce);
-
+#endif
 	gDP.changed &= ~CHANGED_FB_TEXTURE;
 }
 


fzurita:
Ok, I just made the exact same change in this pull request:

https://github.com/gonetz/GLideN64/pull/1083

Also, I fixed a few more GL errors as well.

loganmc10:
OK thanks, by the way, this is off topic. Have you ever run into "flickering" with Framebuffer Emulation enabled on Android?

I am working on porting mupen64plus+GLideN64 to RetroArch (https://github.com/loganmc10/GLupeN64) and pretty much everything works now, but when I enabled Framebuffer Emulation in Android, the whole image flickers with a black screen, almost as if every 2nd frame is completely black. It happens using GLES2 or GLES3, and it doesn't happen on the Raspberry Pi (GLES2) or my laptop (running as GLES2 or full OpenGL).

fzurita:
I think I do recall seeing something like this at one point when the BufferSwapMode was either 0 or 1. But I haven't seen it in a while.

loganmc10:
Ya I tried different buffer swap modes but it didn't seem to make any difference, I noticed that mupen64plus-ae has BufferSwapMode defaulted to 2, whereas the normal default is 0, so I tried that, but no difference.

fzurita:
Compare one of the default profile configuration in mipen64plus-ae to your configuration. Maybe if you make everything the same, it will point to something.

You can find the generated config file for let's say Mario 64 at this directory:
/sdcard/mupen64plus/GameData/SUPER MARIO 64 (U) 20B854B239203BAF6C961B850A4A51A2/CoreConfig/mupen64plus.cfg

loganmc10:
Closing because this is included in #1083

loganmc10:
It's crazy to me how different every phone can be. The flickering only occurs on my Xperia XA (Mali T860). My Nexus 4 (Adreno 320 I believe) renders the intro to Majora's Mask properly (Framebuffer Emulation), but crashes at the first menu of Pokemon Stadium.

The Xperia XA works in GLES3 mode fine, the Nexus 4 shows the wrong colors and alpha for all the textures in GLES3 mode, even though they are both supposed to support GLES3... I don't get it.

loganmc10:
These are issues I experience with RetroArch, not mupen64plus-ae

fzurita:
It sounds like you are missing something

loganmc10:
Ya I know it's got to be something wrong with the app, it's just weird to me that the 2 phones react so differently

Issue #1080, opened 2016-08-04
Per pixel lighting broken with shader cache with Adreno and Android


fzurita:
When shader cache that has per pixel lighting enabled is loaded, we get an incorrect result:

![super_mario_64-000](https://cloud.githubusercontent.com/assets/11581687/17392797/d3bfecd2-59ed-11e6-8744-fb86a736d30a.png)

This is the case with at least Android and Adreno.

fzurita:
It almost looks like the per pixel lighting computation is not happening at all in the shader.

gonetz:
I can't reproduce it. You probably loaded obsolete shader. I not always bump shader storage version, so it can happen. Please clear 'shaders' folder and test again.

fzurita:
Already tried that. It's not helping. Probably an Adreno only issue.

fzurita:
This may be related to not being able to generate binaries for all shaders issue that I put a workaround for.

fzurita:
On an unrelated note, I put a beta version of mupen64plus for Android on the playstore; https://play.google.com/store/apps/details?id=org.mupen64plusae.v3.fzurita

That means your plugin is in the playstore now.

gonetz:
> That means your plugin is in the playstore now.

Cool! How often you are planning to update it?

fzurita:
I'm updating whenever I think something significant enough has been fixed or changed.

Right now this is quite often since a lot of issues are being reported. So I'm fixing things and updating as I go along.

fzurita:
Problem is on my end.

fzurita:
Ok, the problem is in the code. In the tcUniformSet bindWithShaderCombiner we have this check:

```
GBI.isHWLSupported()
```

This check is being performed before any microcodes have been loaded. This is because the shader cache is loaded before the game starts running.

fzurita:
Ok, I came up with a better fix. I added the HW Lighting enabled state of the shader to the m_inputs of the shader. So along these lines:

```
	const bool bUseHWLight = config.generalEmulation.enableHWLighting != 0 && GBI.isHWLSupported() && usesShadeColor();

	if( bUseHWLight )
	    m_nInputs |= 1 << HW_LIGHT;
```

This seems to work ok.

fzurita:
See my last 6 commits here:

https://github.com/mupen64plus-ae/mupen64plus-ae/commits/GLideN64_GL_Errors

[...]

PR #1083, opened 2016-08-09
Fix various errors and HW lighting with shader storage

fzurita:
This fixes all GL errors encountered in my device along with HW lighting when shader storage is enabled.

See more details about the HW lighting issue here: https://github.com/gonetz/GLideN64/issues/1080


gonetz:
Merged. Thanks!

PR #1085, opened 2016-08-10
Save/load shader cache on errors

fzurita:
This pull request correctly saves and loads shader cache when there are errors. I split it off from https://github.com/gonetz/GLideN64/pull/1077

That way we can decide separately if want to include it or not. I'm ok if you don't want to include it.


gonetz:
I've pushed it to separate branch, 64688521e, so it can be included to master when necessary.

PR #1086, opened 2016-08-12
Fix CBFD lighting in Android

fzurita:
This fixes CBFD lighting in Android. It looks a lot more correct now.

### DIFF
--- a/src/gSP.cpp
+++ b/src/gSP.cpp
@@ -258,7 +258,7 @@ static void gSPPointLightVertex4_CBFD(u32 v, float _vPos[4][3])
 		for (u32 l = 0; l < gSP.numLights-1; ++l) {
 			const SPLight & light = gSP.lights[l];
 			intensity = DotProduct( &vtx.nx, &light.x );
-			if (intensity < 0.0f)
+			if ((light.r == 0.0 && light.g == 0.0 && light.b == 0.0) || intensity < 0)
 				continue;
 			if (light.ca > 0.0f) {
 				const f32 vx = (vtx.x + gSP.vertexCoordMod[ 8])*gSP.vertexCoordMod[12] - light.posx;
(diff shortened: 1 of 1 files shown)


AmbientMalice:
Is the lighting correct-ish in the PC version before this?

fzurita:
I'm not sure, it was definitely not correct in android. I haven't tried the PC version.

fzurita:
I have compared this while running in LLE mode, and to be truthful, it's not close to that. But this gets rid of some very annoying random flickering in Android.

fzurita:
By the way, the reason why I made this change is because often omg of the lights always had all fields zeroed out. Trying to apply these lights cause the conker model to be fully lot momentarily.

One thing I'm noticing is that we often have negative intensity for lights. I see that we have checks for that, but the game must be generating those lights for a reason. Should really be ignoring them?

weinerschnitzel:
Hey could you post screenshots of your progress, experiments, discoveries, and comparisons in CBFD lighting? 

Using the negative values for something sounds like a good lead.

Are there any cheat codes that can disable some lighting to make it easier to confirm how point lighting works?

fzurita:
The effects are hard to capture with screenshots. A video would be better suited, but I'm I'll equipped for that. Too much setup.

I have the feeling that when the point lights are zeroed out, that we are instead supposed to use one of the non-point lights to replace it with.

The reason I think this, is because when running in LLE mode, when we would normally get this zeroed out lights, at these points in time Conker's model lighting matches the background lighting.

Frank-74:
By your reasoning above, Conker's model lighting should allways use the background lighting when point lights are zeroed.

A video or screenshot from a real N64 is needed to see how it's supposed to look.

fzurita:
Angrylion shows the effect correctly I believe as well. It's very hard to find footage of a real N64 in YouTube.

gonetz:
At which points of the game the lighting looks a lot more correct now?

fzurita:
I have not played the whole game but intro scene in the throne and in-game. Without this fix, Conker's lighting keeps flickering to full brightness over and over. With this fix, his lighting stays consistent.

Either way, I don't think we should be applying a light with all fields zeroed out.

gonetz:
I tested this fix a bit and did not notice much difference. At least it does not break anything, so merged: b81fb2336

fzurita:
Did you try with Android?

gonetz:
Not yet

fzurita:
Ok. I thought that you may have been using project 64 and the behavior may not have been the same there.

PR #1094, opened 2016-08-23
Fix potential segfault in isExtensionSupported

loganmc10:
glGetStringi can return 0 if there is an error getting the value. If "name" is 0 (happens in RetroArch using GLES3 for whatever reason), then the strcmp segfaults.


gonetz:
merged: c9e1f95890

PR #1095, opened 2016-08-24
GLSL Improvements

loganmc10:
Putting ```uFbFixedAlpha[1] != 0``` in the definition of the function actually create a "shader computed variable", at least in Mesa, when Mesa expands the function.

Basically it forces the GLSL compiler to create an if statement like:
```
bool variable
if uFbFixedAlpha[1] !=0
   variable = true
else
  variable = false
```

Since the values of uFbFixedAlpha[] can only be 0 or 1, this simplification works. I also removed a needless "else if" statement, since there is only one other possible value.


gonetz:
Looks good, thanks. Merged: 05d2f150

PR #1096, opened 2016-08-26
VBO implementation (except for GLES)

loganmc10:
I figure I would submit this as a pull request so people can review it and test it if they'd like.

This implements VBO's for OpenGL (not OpenGL ES). This allows the OpenGL core profile to be used. Everything looks good to me, but it does affect the rendering pipeline, and I'm not sure how to "trigger" every glDrawArrays call, for instance, drawLines, native texrects, DMA triangles, LLE triangles, etc... could all be affected by this change.

### DIFF
--- a/src/OpenGL.cpp
+++ b/src/OpenGL.cpp
@@ -512,6 +512,7 @@ void OGLRender::TexrectDrawer::add()
 		m_max_lry = max(m_max_lry, m_lry);
 	}
 
+	render.updateVBO(&render.rect_vbo, sizeof(GLVertex) * 4, pRect);
 	glDrawArrays(GL_TRIANGLE_STRIP, 0, 4);
 
 	RectCoords coords;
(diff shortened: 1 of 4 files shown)


AmbientMalice:
Aside from the obvious benefit of Core Profile, should this have any other immediate benefits?

loganmc10:
No I don't think so, one day there might be some way to rearrange the draw calls to better utilize VBO's (right now the VBO is updated right before each draw call, but the real purpose of a VBO is that you keep information in that buffer and re-use it over and over, but that is difficult with emulators).

gonetz:
@loganmc10 Thanks for working solution of the problem! I put it on dev branch loganmc10-vbo. Two reasons for that:
- I don't want to make major changes in master until next major Release.
- The code needs polishing.

fzurita:
Since GLES does support this functionality but it's just slower with it, I think it's better to make it into a config parameter rather than using #ifdefs.

That way, if for some reason they're are devices that are better with VBOs,  we don't have to recompile to use the functionality.

I think #ifdefs only make sense for functionality that won't even compile without it.

loganmc10:
Well just GLES3+, GLES2 doesn't support this

fzurita:
Oh, gotcha

loganmc10:
FYI I added a commit to https://github.com/loganmc10/GLideN64/tree/vbo

https://github.com/loganmc10/GLideN64/commit/c811fb86c83f5d147c06edad63e649d65c838fb6

Fixes native res texrects

PR #1098, opened 2016-08-27
Fix typo with LICENCE

orbea:
`LICENCE` should be spelled `LICENSE`. :)

Reference - https://en.wikipedia.org/wiki/License


gonetz:
Thanks :)

orbea:
Well, the the entire contents of the file has `License` so my thinking the file name and file itself should be consistent. Simply renaming the file is easiest. :)

PR #1106, opened 2016-09-04
CRC.cpp: add ARMv8 crc intrinsics

gizmo98:
Speedup:
CRC_Calculate() up to 3,5x
CRC_CalculatePalette() up to 2,1x

Will be only used if the right compiler flags are set.
rpi3: -march=armv8-a+crc -mtune=cortex-a53

https://gcc.gnu.org/onlinedocs/gcc-4.9.0/gcc/ARM-ACLE-Intrinsics.html
http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0801f/pge
1427897662265.html

If CRC32 can be replaced with CRC32C SSE4.2 CRC32C intrinsics could be
added as well.
https://github.com/gonetz/GLideN64/pull/1056#issuecomment-244530166

### DIFF
--- a/projects/cmake/Readme.txt
+++ b/projects/cmake/Readme.txt
@@ -5,6 +5,7 @@ cmake [-DCMAKE_BUILD_TYPE=Debug] [-DVEC4_OPT=On] [-DCRC_OPT=On] [-DNEON_OPT=On] 
 -DCMAKE_BUILD_TYPE=Debug - optional parameter, if you want debug build. Default buid type is Release
 -DVEC4_OPT=On  - optional parameter. set it if you want to enable additional VEC4 optimization (can cause additional bugs).
 -DCRC_OPT=On  - optional parameter. set it if you want to enable additional CRC optimization (can cause additional bugs).
+-DCRC_ARMV8=On  - optional parameter. set it if you want to enable armv8 hardware CRC. It will be ignored if -DCRC_OPT=On.
 -DNEON_OPT=On - optional parameter. set it if you want to enable additional ARM NEON optimization (can cause additional bugs).
 -DNOHQ=On - build without realtime texture enhancer library (GLideNHQ).
 -DUSE_UNIFORMBLOCK=On - Use uniform blocks in shaders. May help to improve performance. Not supported by GLES2 hardware.
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -185,14 +185,20 @@ if(VEC4_OPT)
   )
 endif(VEC4_OPT)
 
-if(CRC_OPT)
-  list(APPEND GLideN64_SOURCES
-    CRC_OPT.cpp
-  )
+if(CRC_OPT OR CRC_ARMV8)
   list(REMOVE_ITEM GLideN64_SOURCES
     CRC.cpp
   )
-endif(CRC_OPT)
+  if(CRC_OPT)
+    list(APPEND GLideN64_SOURCES
+      CRC_OPT.cpp
+    )
+  elseif(CRC_ARMV8)
+    list(APPEND GLideN64_SOURCES
+      CRC_ARMV8.cpp
+    )
+  endif(CRC_OPT)
+endif(CRC_OPT OR CRC_ARMV8)
 
 if(NEON_OPT)
   add_definitions(
--- a/src/CRC_ARMV8.cpp
+++ b/src/CRC_ARMV8.cpp
@@ -1,0 +1,51 @@
+#include "CRC.h"
+#include <arm_acle.h>
+
+void CRC_BuildTable()
+{
+}
+
+u32 CRC_Calculate( u32 crc, const void * buffer, u32 count )
+{
+	u8 *p;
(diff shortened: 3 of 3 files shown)


src/CRC.cpp
```diff
@@ -37,8 +41,24 @@ u32 CRC_Calculate( u32 crc, const void * buffer, u32 count )
 	u32 orig = crc;
 
 	p = (u8*) buffer;
+
+#if defined(__ARM_FEATURE_CRC32)
+	// use four byte crc intrinsic __crc32w if count is high enough.
+	while (count >= 4) {
+		crc = __crc32w(crc, (p[0]  | p[1] << 8 | p[2] << 16 | p[3] << 24));
```
gonetz: why not just
__crc32w(crc, (u32*)p)
?

src/CRC.cpp
```diff
@@ -37,8 +41,24 @@ u32 CRC_Calculate( u32 crc, const void * buffer, u32 count )
...
+		p += 4;
+		count -= 4;
+	}
+	if (count == 2) {
+		crc = __crc32h(crc, (p[0]  | p[1] << 8));
```
gonetz: why not just
__crc32w(crc, (u16*)p)
?

src/CRC.cpp
gizmo98: Thank you very much. I have tried to use this but had done something wrong. I just looked how the linux kernel is doing it. It just shifts bytes. Your solution runs much better. I can see a speed gain of factor 9.3.

gizmo98:
@gonetz do you have any other requirements? Otherwise i would squash or repost this PR.

gonetz:
Well, I have no more questions to the code. The problem is how new code is introduced. You may notice that I don't like when several implementations of the same functionality present in the same source file separated by #ifdef/#else/#endif. Such code is hard to read and I'm trying to avoid it. Regarding crc calculation, the project already has optimized checksum functions, implemented in CRC_OPT.cpp This file included instead of CRC.cpp when corresponding cmake input parameter is used. You may do the same: move ARMv8 crc implementation to a separate file and add another cmake parameter for ARMv8 crc, which will include that file to the project instead of CRC.cpp and add corresponding compiler flags to GCC_CPP11_COMPILE_FLAGS list. Alternatively, you may try to ask cmake to detect support of ARMv8 crc, it should be able to probe for it somehow.

loganmc10:
@gizmo98 is there no way to check for this functionality at run time? It's kind of a bummer that is has to be set at compile time, especially for a platform like Android, where you can't really compile a different binary for every kind of CPU out in the ecosystem.

gizmo98:
@gonetz cmake does not seem to know this arm crc extension. I will add a new file and something like -DCRC_ARM_OPT=On. Do you have any naming conventions? Shall i add a file for SSE4.2 CRC32C as well -DCRC_SSE42_OPT)? 

@loganmc10 mupen64plus arm dynarec uses something like this:
https://github.com/mupen64plus/mupen64plus-core/blob/b2c24debe842eba9d4afafa2b32ebd9d05a7fd7a/src/r4300/new_dynarec/arm/assem_arm.c#L4794 
https://github.com/mupen64plus/mupen64plus-core/blob/b2c24debe842eba9d4afafa2b32ebd9d05a7fd7a/src/r4300/new_dynarec/arm/arm_cpu_features.c
I don't think it is easy to add something which reads /proc/cpuinfo and use this info.

gizmo98:
Ok. I moved everything into an own file and squashed everything.

src/CMakeLists.txt
```diff
@@ -194,6 +194,19 @@ if(CRC_OPT)
   )
 endif(CRC_OPT)
 
+if(CRC_ARMV8)
```
gonetz: What if user will use both CRC_OPT and CRC_ARMV8? Need some protection from this situation.

projects/cmake/Readme.txt
```diff
@@ -5,6 +5,7 @@ cmake [-DCMAKE_BUILD_TYPE=Debug] [-DVEC4_OPT=On] [-DCRC_OPT=On] [-DNEON_OPT=On]
 -DCMAKE_BUILD_TYPE=Debug - optional parameter, if you want debug build. Default buid type is Release
 -DVEC4_OPT=On  - optional parameter. set it if you want to enable additional VEC4 optimization (can cause additional bugs).
 -DCRC_OPT=On  - optional parameter. set it if you want to enable additional CRC optimization (can cause additional bugs).
+-DCRC_ARMV8=On  - optional parameter. set it if you want to enable armv8 hardware CRC.
```
gonetz: Please add a note that CRC_OPT and CRC_ARMV8 can't be used together.

gizmo98:
Ok. I removed -march=armv8-a+crc add all. The user can decide if he uses -march=native or something more specific. He just has to use a target which supports the crc extension. If we use -DNEON_OPT -mfpu=neon will be also not set.

gonetz:
Merged: a471b130e79

PR #1118, opened 2016-09-11
Remove old arm assembler

gizmo98:
GCC replaces (value << 8) | (value >> 8) with __builtin_bswap16 if -Os,
-O1 or higher optimization levels are used. __builtin_bswap16 is as fast as the assembler
declaration.


gonetz:
Merged. Thanks!

Issue #795, opened 2015-11-17
NBA Hangtime crash ingame


purplemarshmallow:
The emulator eats gigabytes of RAM. Then eventually it crashes. Most likely has to do something with all the failed assertions in this game #107

oddMLan:
Yep, there's a memory leak somewhere. Memory usage is constantly increasing.
But the game does not crash for me, it only slows down to a crawl. How long until it crashes? I have 8GB RAM so it might take a while to exhaust all the available RAM, which I assume it's what causes the crash.

gonetz:
It is not memory leak. Texture cache does not work properly, new textures constantly allocated.

gonetz:
Update: new textures constantly allocated because the game programmed that way. It seems that it has some large background image and game viewport constantly moves along it. Each frame it moves slightly, resulting in new set of slightly different textures. Texture cache helpless here.

I've added texture cache size limit to avoid game crash. Plugin clears whole texture cache when texture cache size limit reached.

BTW, texture cache in Glide64 also get overloaded almost instantly. Glide64 also clears cache when it overflowed. Since cache size is limited to 4k textures, cache clear works fast. GLideN64 cache size limit is much larger, so its clear can take time.

purplemarshmallow:
not crashing anymore

loganmc10:
I still have some people that have issues with slowdowns related to this.

For starters, shouldn't the 500MB cache size limit have stopped this? @purplemarshmallow @oddMLan mentioned GB's of memory.

Also, 128000 might be too high, @orbea had slowdowns and a crash playing Beetle Adventure Racing that was fixed by changing the number to 32000.

Finally, wiping the cache causes a "hiccup", since the cache needs to be rebuilt for the next frame obviously. Wouldn't it be better to just delete the oldest texture to make room for the new one like I did when I submitted a pull request for this issue on the Raspberry Pi? (https://github.com/gonetz/GLideN64/pull/1017/files)

PR #1126, opened 2016-09-15
Fix Texture Cache slow down and crashes

orbea:
I experienced an issue on Slackware64-current, nouveau, GLideN64 and the game Beetle Adventure Racing where after playing for a while the game would begin to lag progressively worse until it would clear the texture cache and potentially bring my entire system down.

The problem was that Beetle Adventure Racing is capable of producing infinite textures and that GLideN64 had a max cache limit of 128,000 textures. This is problematic because some video drivers such as nouveau convert every single one of these textures into a 128 KB "large" page which would require 16 GB of vram where I have 3 GB.  After discussing this with the nouveau developers at #nouveau @ freenode it was suggested to have this limit as 16,384 instead which would require 2 GB of vram and is much saner. 

This avoided the slow down and crashes, but introduced an annoying issue where the game would freeze (Hiccup) momentarily every time it hit that limit and cleared the cache. To combat this loganmc10 suggested to clear only the oldest textures instead of the entire cache. This works wonderfully and solves all issues encountered.

So in short, this will fix slow downs, crashes and hiccups with GLideN64 and Beetle Adventure Racing while using some linux video drivers.

Please read this issue report that this PR will fix for more information.
https://github.com/loganmc10/GLupeN64/issues/56

### DIFF
--- a/src/Textures.cpp
+++ b/src/Textures.cpp
@@ -526,12 +526,14 @@ void TextureCache::_checkCacheSize()
 #ifdef VC
 	const size_t maxCacheSize = 15000;
 #else
-	const size_t maxCacheSize = 128000;
+	const size_t maxCacheSize = 16384;
 #endif
-	// Clear cache if its size is too large.
 	if (m_textures.size() >= maxCacheSize) {
-		_clear();
-		return;
+		CachedTexture& clsTex = m_textures.back();
+		m_cachedBytes -= clsTex.textureBytes;
+		glDeleteTextures(1, &clsTex.glName);
+		m_lruTextureLocations.erase(clsTex.crc);
+		m_textures.pop_back();
 	}
 
 	if (m_cachedBytes <= m_maxBytes)


gonetz:
Thanks for well explained commit. I tried do something like this when worked on #795, but the game generates large number of new textures each frame, so removing them by one was too slow. It was better to clear cache by the price of performance hiccup. Good if this fix works well for BAR.

orbea:
I've tested with NBA Hangtime and seems to require a more complete solution for that game. With this change its slow, but if the entire cache is cleared instead it suffers from being both slow and hiccuping. At least BAR seems to work great now.

orbea:
I guess NBA Hangtime is much faster if the limit is even lower, but even if the limit is reduced to something like 1,000 it still often dips to 40 fps.

loganmc10:
You could always delete them in a batch. Like when it hits the 16384 limit, delete the 100 oldest textures, instead of wiping the whole cache or doing it one at a time.

orbea:
I think its more that the game creates textures far faster than nouveau can handle. I'm not sure about other video drivers, mupen64plus-libretro with glide64 is pretty slow too.

gonetz:
> You could always delete them in a batch. Like when it hits the 16384 limit, delete the 100 oldest textures, instead of wiping the whole cache or doing it one at a time.

Sounds like a good optimization trick.

PR #1128, opened 2016-09-20
Add 'static' to convert.h variables

loganmc10:
I am working on porting this to emscripten (to be able to play mupen64plus+GLideN64 in a browser), it compiles with clang+llvm, and I get these errors when it tries to link it all together:

```ERROR: Linking globals named 'Five2Eight': symbol multiply defined```

Adding 'static' solves the problem.

### DIFF
--- a/src/convert.h
+++ b/src/convert.h
@@ -3,7 +3,7 @@
 
 #include "Types.h"
 
-const volatile unsigned char Five2Eight[32] =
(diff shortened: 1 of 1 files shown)


gonetz:
I think it is better to have only one copy of these arrays .
My solution: 3db8c102fbcdd
Please check that it fixes your problem.

loganmc10:
Yep looks good, thanks

PR #1137, opened 2016-09-26
Fix Depth Buffer emulation

orbea:
This reverts part of commit https://github.com/gonetz/GLideN64/commit/8dad84cc39571c295d10a3b3c9326fa942478fcd to fix the depth buffer issue for Link's portrait in Orcarina of Time while still working with Toy Story 2.

%20%5b!%5d-160926-022058.png)

### DIFF
--- a/src/DepthBuffer.cpp
+++ b/src/DepthBuffer.cpp
@@ -387,7 +387,7 @@ void DepthBufferList::saveBuffer(u32 _address)
 		pDepthBuffer = nullptr;
 	}
 
-	if (pDepthBuffer == nullptr) {
+	if (m_pCurrent == nullptr) {
 		m_list.emplace_front();
 		DepthBuffer & buffer = m_list.front();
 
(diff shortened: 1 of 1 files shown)


SigmaVirus:
Brilliant, nicely done!

orbea:
Actually, don't merge this. It introduce more issues in Orcarina of Time. Hopefully it will provide clues for a proper solution

SigmaVirus:
OK, what kind of more issues are we talking about here?

orbea:
I refreshed the commit and seem to have fixed it now, I did not revert enough. The issue was similar to the original Toy Story 2 issue, lots of missing textures after leaving Link's house. Might be nice to test to make sure there isn't anything else broken as a result.

orbea:
I've tried various games without any new issues, I think this might be good now.

gonetz:
Thanks. It is a step in the right direction, but not quite correct one because it breaks logic of current depth buffer, which I tried to fix in the original commit. My fix: 3db8c102

orbea:
Thanks for the better solution, I'm not very experienced at programming, but I try anyways...

gonetz:
You did it almost right. Also, it helped me to find, where I made the mistake.

SigmaVirus:
Great work both of you nevertheless!

PR #1141, opened 2016-09-27
Set the program before updating uniforms

loganmc10:
See https://github.com/loganmc10/GLupeN64/issues/64 for a disucussion on this issue.

Basically the proper program wasn't being set prior to updating the uniform values.

@gonetz if there is a better place to put this go ahead, this just seemed like the right place to me.

### DIFF
--- a/src/GLUniforms/UniformSet.cpp
+++ b/src/GLUniforms/UniformSet.cpp
@@ -131,6 +131,8 @@ void UniformSet::updateUniforms(ShaderCombiner * _pCombiner, OGLRender::RENDER_S
 {
 	UniformSetLocation & location = m_uniforms.at(_pCombiner->getKey());
 
+	glUseProgram(_pCombiner->m_program);
+
 	_updateColorUniforms(location, false);
 
 	if ((_renderState == OGLRender::rsTriangle || _renderState == OGLRender::rsLine) && _pCombiner->usesTexture())


gonetz:
Ok, UniformSet::updateUniforms called from CombinerInfo::updateParameters

CombinerInfo::updateParameters always follows CombinerInfo::update()

CombinerInfo::update() always calls CombinerInfo::setCombine which always calls ShaderCombiner::update

ShaderCombiner::update always calls glUseProgram(m_program)

Could you explain, how "the proper program wasn't being set prior to updating the uniform values."?
This is the problem, which need to be found and fixed.

loganmc10:
I'll have to do some testing to figure out the exact code path that is followed that causes this issue, but could it be this line?

https://github.com/gonetz/GLideN64/blob/master/src/OpenGL.cpp#L475

It doesn't look like that is preceded by a call to CombinerInfo::update()

loganmc10:
Actually based on a little testing it seems like something is going wrong inside ```OGLRender::_updateStates``` but I'll test a bit tomorrow to narrow it down

gonetz:
This is OGLRender::TexrectDrawer, which is used when "draw native res texrect" option is enabled.
I did not see in  loganmc10/GLupeN64#64 discussion that that option is enabled.

TexrectDrawer assumes that all texrects, which form the picture are drawn with the same combiner. Thus, it is not necessary to update combiner each time.

If the problem appears only when "draw native res texrect" enabled, this is the place to dig.

orbea:
I leave "draw native res texrect" disabled and have tested now, I can reproduce the issue either way.

gonetz:
> I can reproduce the issue either way.

Well, probably current shader program changed somewhere in OGLRender::_updateStates on the way from cmbInfo.update() to cmbInfo.updateParameters.

loganmc10:
@gonetz here is where I am so far:

The current shader program is changed by this line:

https://github.com/gonetz/GLideN64/blob/master/src/OpenGL.cpp#L1165

```_updateTextures(_renderState)``` inside ```OGLRender::_updateStates```

Inside _updateTextures, it is changed by this loop:
```
for (u32 t = 0; t < 2; ++t) {
   if (pCurrentCombiner->usesTile(t))
      textureCache().update(t);
   else
      textureCache().activateDummy(t);
}
```

I haven't gotten further than that yet, there is a lot going on inside ```TextureCache::update```, but I assume somewhere in there the current shader program is being modified.

loganmc10:
Also, this only happens with Framebuffer Emulation enabled.

loganmc10:
If I had to guess I would assume the chain looks something like this:

textureCache().update(t) -> FrameBuffer_ActivateBufferTexture() -> pBuffer->getTexture(t) -> _getSubTexture(_t) -> OGLRender::copyTexturedRect(), which does a glUseProgram.

gonetz:
Thanks!

gonetz:
Hmm. It was very bad idea to use copyTexturedRect in _getSubTexture, because it alters lots of GL states, which were set prior it by OGLRender::_updateStates. In fact, we need to call _updateStates again after it.
I need to think what to do.

orbea:
This seems to have been fixed in the latest commits for me, thanks.

loganmc10:
thanks @gonetz

gonetz:
I just disabled _getSubTexture for GLES2. Some effects will work incorrect, e.g. reflections in bubbles in BT.

PR #1152, opened 2016-10-02
Modify german translation

gizmo98:


gonetz:
I added Crop and OSD features to master. I took your translation and updated it with new text lines for Crop and OSD : a5a432c4fd. Please translate them too.

PR #1161, opened 2016-10-07
Use full CRC calculation for CRC_OPT and GBI

loganmc10:
The optimized CRC calculation doesn't work with GBIInfo::loadMicrocode, when you try to launch Perfect Dark, you get a "[GLideN64]: error - unknown ucode!!!" message.

This change makes CRC_OPT use the regular CRC calculator for the ucode calculation

### DIFF
--- a/src/CRC.cpp
+++ b/src/CRC.cpp
@@ -31,7 +31,7 @@ void CRC_BuildTable()
 	}
 }
 
-u32 CRC_Calculate( u32 crc, const void * buffer, u32 count )
+u32 CRC_Calculate_Strict( u32 crc, const void * buffer, u32 count )
 {
 	u8 *p;
 	u32 orig = crc;
@@ -41,6 +41,11 @@ u32 CRC_Calculate( u32 crc, const void * buffer, u32 count )
 		crc = (crc >> 8) ^ CRCTable[(crc & 0xFF) ^ *p++];
 
 	return crc ^ orig;
+}
+
+u32 CRC_Calculate( u32 crc, const void * buffer, u32 count )
+{
+        return CRC_Calculate_Strict(crc, buffer, count);
 }
 
 u32 CRC_CalculatePalette(u32 crc, const void * buffer, u32 count )
--- a/src/CRC.h
+++ b/src/CRC.h
@@ -3,5 +3,6 @@
 void CRC_BuildTable();
 
 // CRC32
+u32 CRC_Calculate_Strict( u32 crc, const void *buffer, u32 count );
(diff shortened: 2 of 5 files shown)


gonetz:
1. CRC_Calculate_Full - bad name. CRC_Calculate_Strict is better.
2. Is optimized CRC really useful? May be remove it if it can't provide robust checksum?

loganmc10:
1. I agree, I couldn't think of a very good name for it.
2. It works very well, this change fixes a major issue with it. I don't see any graphical glitches with it. I was planning on enabling CRC_OPT for GLES devices with GLupeN64

gizmo98:
Isn't ARMV8_CRC faster than OPT_CRC on a rpi3? Wouldn't it be better to add OPT_CRC as a option like crc_calc = 0/"strict",1/"fast hash calc"?

loganmc10:
I'd bet that CRC_OPT is marginally faster but I haven't compared them. I was going to leave CRC_ARMV8 for the rpi3 and use CRC_OPT for other GLES devices like rpi2 and Android.

I'm not sure if I understand your second question, but if you're asking why CRC_OPT is in its own .cpp file instead of some kind of run-time option, that's how @gonetz wanted the code organized.

gizmo98:
All right. I did not know you had already discussed code organization.

gizmo98:
I compared crc armv8 with crc opt today and crc armv8  is always much faster.

gonetz:
@gizmo98 
> I compared crc armv8 with crc opt today and crc armv8 is always much faster.

How you tested it? Did you write a special test to compare both algorithms on the same data?

gizmo98:
I just used the following code snipped and added the hash function. 
https://gist.github.com/gizmo98/8a83aef820f1e6d15da4a4baf3b7e0dc

time_spent --> crc32_software > crc_opt(hash) > crc32_armv8

gonetz:
Ok, thanks

fzurita:
@loganmc10 how much of a performance boost is CRC_OPT. I don't have it enabled right now for Mupen64Plus-AE, but I'm wondering if I should.

loganmc10:
You can read a bit about the testing I did here:

https://github.com/gonetz/GLideN64/pull/1043#issuecomment-230388065

It's not life changing but it does lower the CPU usage by 5-10% on the Raspberry Pi. Obviously the improvements would be more drastic on a device with a weaker CPU

fzurita:
Ok, thanks.

gonetz:
Merged

PR #1162, opened 2016-10-08
Add getTextSize to TextDrawerStub

loganmc10:
Fixes compile when using TextDrawerStub.cpp instead of TextDrawer.cpp


gonetz:
Merged

PR #1164, opened 2016-10-09
Don't do performance counting if it's not enabled

loganmc10:
For some reason this causes a segfault in GLupeN64 (https://github.com/loganmc10/GLupeN64/issues/72)

It doesn't really need to be doing this if performance counting is disabled anyway

### DIFF
--- a/src/Performance.cpp
+++ b/src/Performance.cpp
@@ -1,5 +1,6 @@
 #include "VI.h"
 #include "Performance.h"
+#include "Config.h"
 
 Performance perf;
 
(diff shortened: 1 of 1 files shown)


loganmc10:
It probably doesn't matter either way if increaseFramesCount() happens, it's just increasing a counter

gonetz:
I need to know the reason of the segfault.  The code looks harmless. On which line the crash happens?

loganmc10:
You can read the original report from the user here: https://github.com/loganmc10/GLupeN64/issues/72. The segfault happened at an unrelated spot, but this commit fixed the issue for the user, and the issue started from this commit.

I assume it has something to do with std::clock since it happens after about 5 minutes for the user and the rest of it is just fairly simple math. I haven't had a chance to try and reproduce it, it could be a compiler or hardware specific issue, I'm not sure.

It doesn't need to be done exactly like this, but this code doesn't need to be executed if performance counting isn't being used.

gonetz:
Ok

gonetz:
793d1c4911

loganmc10:
Thanks

PR #1166, opened 2016-10-10
Fixes GLES 3.1 in Adreno devices

fzurita:
See https://github.com/gonetz/GLideN64/issues/895

I fixed this issue with this one specific driver by making any function (just readTex) that uses Sampler2D as argument into a macro instead.

For some reason the Adreno drivers don't like Sampler2D as an argument type. Using a macro forces the inlining of the code during pre-processing.

Tested in the following devices:
Nvidia Shield TV (Tegra X1) in Full GL and GLES 3.1 modes
Nexus Player (PowerVR)
Samsung Galaxy S6 Active (Mali)
ZTE Axon 7 (Adreno)

### DIFF
--- a/src/OGL3X/GLSLCombiner_ogl3x.cpp
+++ b/src/OGL3X/GLSLCombiner_ogl3x.cpp
@@ -28,7 +28,6 @@ static GLuint  g_calc_noise_shader_object;
 static GLuint  g_write_depth_shader_object;
 static GLuint  g_calc_depth_shader_object;
 static GLuint  g_render_depth_shader_object;
-static GLuint  g_readtex_shader_object;
 static GLuint  g_readtex_ms_shader_object;
 static GLuint  g_dither_shader_object;
 static GLuint  g_monochrome_image_program = 0;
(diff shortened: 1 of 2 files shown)


src/OGL3X/GLSLCombiner_ogl3x.cpp
```diff
@@ -361,24 +358,32 @@ ShaderCombiner::ShaderCombiner(Combiner & _color, Combiner & _alpha, const gDPCo
 		if (usesTile(0)) {
 			if (config.video.multisampling > 0) {
 				strFragmentShader.append("  lowp vec4 readtex0; \n");
-				strFragmentShader.append("  if (uMSTexEnabled[0] == 0) readtex0 = readTex(uTex0, vTexCoord0, uFbMonochrome[0], uFbFixedAlpha[0]); \n");
+				strFragmentShader.append("  if (uMSTexEnabled[0] == 0) READ_TEX(readtex0, uTex0, vTexCoord0, uFbMonochrome[0], uFbFixedAlpha[0]) \n");
 				strFragmentShader.append("  else readtex0 = readTexMS(uMSTex0, vTexCoord0, uFbMonochrome[0], uFbFixedAlpha[0]); \n");
```
gonetz: Why you did not change readTexMS the same way? It also uses sampler as input.

fzurita: For some reason Sampler2DMS type doesn't have this problem.

fzurita:
I just thought about this. The shader cache version should probably be incremented.

gonetz:
Ok, I'll up shader cache version.

gonetz:
f614c26e9

PR #1171, opened 2016-10-12
GLES3 fix for nvc0

orbea:
This fix suggested by a nouveau dev will resolve a blending / transparency issue with Ocarina of Time, GLES3, GLupeN64 and the nouveau nvc0 driver. Strangely enough this problem does not occur with the intel, llvmpipe or nv50 drivers nor does it occur with OpenGL or GLES2.

Please read issue https://github.com/loganmc10/GLupeN64/issues/74 for more discussion.

### DIFF
--- a/src/FBOTextureFormats.cpp
+++ b/src/FBOTextureFormats.cpp
@@ -43,9 +43,9 @@ void FBOTextureFormats::init()
 	monochromeType = GL_UNSIGNED_BYTE;
 	monochromeFormatBytes = 1;
 
-	depthInternalFormat = GL_DEPTH_COMPONENT32F;
+	depthInternalFormat = GL_DEPTH_COMPONENT24;
 	depthFormat = GL_DEPTH_COMPONENT;
-	depthType = GL_FLOAT;
+	depthType = GL_UNSIGNED_INT;
 	depthFormatBytes = 4;
 
 	depthImageInternalFormat = GL_RGBA32F;


fzurita:
I'm scared of changes like this. Will it break or slow down other drivers?

loganmc10:
gonetz will probably know better, but I believe the effect of this change will be less accurate depth calculations? Basically it will work like GLES2 instead of OpenGL (like it is now), but I could be completely wrong.

gonetz:
I can't predict all side effects of such modifications.
Common rule: device/drivers specific fixes must be enabled either via defines, as for VC
```
#ifndef VC
...
#else
...
#endif
```
or as renderer fixes, as it is done for Adreno.

imirkin:
(nouveau developer here)

FWIW my recommendation would actually be to set this to GL_DEPTH_COMPONENT (just like the GLES2 one is). For unsized internal formats, the effective internal format is determined by the format/type combo in GLES. In desktop GL, GL_DEPTH_COMPONENT will work out to a Z24X8 surface as well (the format/type is not used to determine effective internal format there).

I don't think this is a device-specific fix. The real issue is that you're using polygon offsets with float depth buffers. I doubt the intention was to use Z24 for GLES2 and desktop GL, but Z32F for GLES3. Polygon offsets units are well-defined for fixed point buffers (the units are whatever the smallest unit of depth buffer is, so 1/2^24 for a Z24 buffer), but for float buffers, it's not well defined. In practice, it tends to be in ULPs of the depth of one of the vertices of the polygon being rasterized. This is tough to rely on.

gonetz:
@imirkin I don't remember exactly why GL_DEPTH_COMPONENT32F selected as internal depth format for GLES3. I guess, it was selected because it works well on Android devices, which we have on hands. So, this change must be tested on all devices again to be sure that nothing is broken. I can push that change to a separate branch and ask @fzurita to test it. 

BTW, what "ULPs" stands for?

fzurita:
I'll go ahead and test it, no need to make a new branch. The change is simple enough. @gonetz anything specific that you could see breaking with this change?

gonetz:
@fzurita 
I need to know, how depth buffer emulation will work with these parameters: 
- depth buffer copy to RDRAM (coronas in Zeldas)
- render to depth buffer (shadows in Pilot wings, VS screen in Mario Tennis, TV in Quarterback club 98 menus)

fzurita:
Sure, I will test and report back.

gonetz:
Thanks in advance

fzurita:
Tested torches in majora's mask. They are not going through walls
Tested Pilotwings shadow and Mario Tennis VS screen. Both work correctly.

gonetz:
So, this fix does no harm. Good. Thanks!

fzurita:
Np!

imirkin:
@gonetz ULP = "Unit of Least Precision". It's the idea of what happens when you add "1" to the mantissa of a floating point number - depending on the exponent, it'll change the real representation by a different amount. So effectively the units (r) of polygon offset are dependent on the exponent of the floating point depth value for one of the vertices (might be highest, I don't remember). So you get a different amount of offset if polygon (rasterized) depth is 1.0 vs if it's 0.00001.

Issue #1091, opened 2016-08-22
Paper Mario slow async copy color buffer on Adreno


fzurita:
On Paper Mario, on Adreno devices, the async copy color buffer to RDRAM is slow. I suspect that it's a color format or texture size that the driver is not happy with.

fzurita:
Here are the texture parameters for paper mario. The texture data is identical to other games where async copies work fine.

Is there anything special about Paper Mario when it comes to the frame buffer texture?

fzurita:
This is still an issue, my fix breaks all of CopyColorToRDRAM.

fzurita:
I forgot to close this. It was fixed with buffer storage on Adreno 400 series GPUs. Adreno 500 series GPUs have a different problem where this is slow for all games.

fzurita:
P

PR #1172, opened 2016-10-13
Fixes slow PBOs in Paper Mario.

fzurita:
For some reason this only affected Paper Mario. The game now runs at a good speed with async copy color buffer to RDRAM on Adreno devices. I'm wondering if the same treatment needs to be applied to monochrome copies as well.

This issue: https://github.com/gonetz/GLideN64/issues/1091


gonetz:
Monochrome buffers not used as primary frame buffers and copied to RDRAM only when "Aux buffers copy" option enabled.

gonetz:
92b39b6b725

PR #1184, opened 2016-10-19
Zelda subscreen hack always enables copy color to RDRAM

fzurita:
Zelda subscreen hack always enables copy color to RDRAM. It's annoying to have to tell people to enable CopyColorToRDRAM to fix the subscreen delay for zelda games.


AmbientMalice:
Shouldn't copy colour to RDRAM be enabled by default, anyway? More games need it than not.

fzurita:
It's a big slowdown on some devices in Android.

Issue #1110, opened 2016-09-07
Mario Tennis pause menu causes Adreno devices to crash


fzurita:
The device driver seems to be crashing when bringing up the Mario Tennis in-game pause menu. I have yet to track down which call is causing the crash to happen since I can't seem to get a back trace.

fzurita:
Ok, I started trying to figure this one out. The one thing I notice when trying to check for GL errors in this game is that whenever we call ```glDrawArrays``` in OpenGL.cpp line 1931, I'm getting a GL error with status 0x506, meaning that the frame buffer is incomplete

When calling ```glCheckFramebufferStatus``` on ```GL_DRAW_FRAMEBUFFER```, I'm getting a status of ```GL_FRAMEBUFFER_INCOMPLETE_DIMENSIONS_EXT```. I looked up this error code online, and here is what it means:

GL_FRAMEBUFFER_INCOMPLETE_DIMENSIONS
Not all attached images have the same width and height.

fzurita:
Another GL error I'm seeing, glTexImage2D, for GLES 3.0/3.1 should use GL_R8 for the internal format.

gonetz:
AFAIR, only GLES2 has limitation that all attached images must have the same width and height.
N64 games often use small aux color buffers with the same full size depth buffer. Mario Tennis is among these games. I put special code to avoid GL_FRAMEBUFFER_INCOMPLETE_DIMENSIONS with GLES2. In FrameBufferList::attachDepthBuffer() :

```
#ifdef GLES2
	if (pDepthBuffer->m_pDepthBufferTexture->realWidth == m_pCurrent->m_pTexture->realWidth) {
#else
	if (pDepthBuffer->m_pDepthBufferTexture->realWidth >= m_pCurrent->m_pTexture->realWidth) {
#endif // GLES2

```

The depth buffer will not be attached to FBO if its width is not the same as width of color buffer texture in case of GLES2. Height should be tested too, but I decided that it is redundant.

fzurita:
Ok, I think GLES 3 has the same limitation. I'll have to check GLES 3.1

fzurita:
Ok, same problem in GLES 3.1. That means that I have to change the code to this
```
#ifdef GLESX
    if (pDepthBuffer->m_pDepthBufferTexture->realWidth == m_pCurrent->m_pTexture->realWidth) {
#else
[...]

PR #1193, opened 2016-10-28
Fix verious GL errors in GLES 3.1 mode and Mario tennis.

fzurita:
This fixes all the GL errors I encountered in GLES 3.1 and Mario Tennis. See https://github.com/gonetz/GLideN64/issues/1110.


fzurita:
Also, thanks to the Android GL Errors traceability, I was able to find why Multipsampling wasn't working, at least in Android. I added that to the pull request.

src/OGL3X/GLSLCombiner_ogl3x.cpp
```diff
@@ -142,18 +142,25 @@ void InitZlutTexture()
...
 	glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
+
+#ifndef GLESX
 	glTexImage2D(GL_TEXTURE_2D, 0, fboFormats.lutInternalFormat,
 		512, 512, 0, fboFormats.lutFormat, fboFormats.lutType,
 		zLUT);
-	glBindImageTexture(ZlutImageUnit, g_zlut_tex, 0, GL_FALSE, 0, GL_READ_ONLY, fboFormats.lutInternalFormat);
+#else
+	glTexStorage2D(	GL_TEXTURE_2D, 1, fboFormats.lutInternalFormat, 512, 512);
+#endif
```
gonetz: Please explain that part. Original code uses glTexImage2D with zLUT as texture data. Why you change that and when g_zlut_tex will be filled with data?

fzurita: glBindImageTexture requires an immutable texture object. See https://www.khronos.org/opengles/sdk/docs/man31/html/glBindImageTexture.xhtml

glTexStorage2D doesn't create immutable texture objects but glTexStorage2D does.

I think you did make me realize a mistake. glBindImageTexture takes texture data as an argument, glTexStorage2D doesn't. I need to find how to set texture data after.

fzurita: I think I can use glTexSubImage2D to update the data after.

fzurita:
I updated the pull request with a fix for the error you made me realize.

src/OGL3X/GLSLCombiner_ogl3x.cpp
```diff
@@ -142,18 +142,26 @@ void InitZlutTexture()
...
 	glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
+
+#ifndef GLESX
 	glTexImage2D(GL_TEXTURE_2D, 0, fboFormats.lutInternalFormat,
 		512, 512, 0, fboFormats.lutFormat, fboFormats.lutType,
 		zLUT);
-	glBindImageTexture(ZlutImageUnit, g_zlut_tex, 0, GL_FALSE, 0, GL_READ_ONLY, fboFormats.lutInternalFormat);
+#else
+	glTexStorage2D(	GL_TEXTURE_2D, 1, fboFormats.lutInternalFormat, 512, 512);
+	glTexSubImage2D(GL_TEXTURE_2D, 0, 0, 0, 512, 512, fboFormats.lutFormat, fboFormats.lutType, zLUT);
```
gonetz: This must work with GL as well, so let make that code common. Please replace the old code, without ifdef endif

fzurita: Ok, will do.

fzurita: Ok, done.

gonetz:
Thanks! Merged: 010c415d7186

gonetz:
OOPS! Assert in OpenGl.cpp line 418
assert(m_textureBoundsLoc >= 0);

fzurita:
Hmm, not good, I should probably start running debug soi that I will hit the asserts. Any idea what's causing that? The changes I made should had been benign.

gonetz:
The code looks correct. I don't see any problems, but glsl compiler removes the uniform now. I can't find how to fix it.

fzurita:
Yikes, it sounds like a GLSL compiler bug, it must be removing the variable because it thinks that it's not being used, but both versions of the macro is clearly using it.

You could make it easier for the GLSL compiler, in the main method of the shader, you could do:
```
mediump vec4 temp = uTextureBounds;
```

as the first call. Then, from then on, instead of using ```uTextureBounds```, use ```temp```.

gonetz:
Does not help;

fzurita:
Is the compilation of the shader succeeding, but m_textureBoundsLoc is not being located?

gonetz:
Yes

gonetz:
```
const char * strTexrectDrawerTexBilinearFilter =
MAIN_SHADER_VERSION
"#define TEX_OFFSET(off, tex, texCoord, texSize) texture(tex, texCoord - (off)/texSize)							\n"
"#define TEX_FILTER(name, tex, texCoord)															\\\n"
"{																												\\\n"
"  mediump vec2 texSize = vec2(textureSize(tex,0));																\\\n"
"  mediump vec2 texelSize = vec2(1.0) / texSize;																\\\n"
"  lowp vec4 c = texture(tex, texCoord);																		\\\n"
"  if (abs(texCoord.s - textureBounds[0]) < texelSize.x || abs(texCoord.s - textureBounds[2]) < texelSize.x){	\\\n"
"    name = c;												 													\\\n"
"  } else if (abs(texCoord.t - textureBounds[1]) < texelSize.y || abs(texCoord.t - textureBounds[3]) < texelSize.y){	\\\n"
"    name = c;												 													\\\n"
"  } else {													 													
...
"																												\n"
;

const char * strTexrectDrawerFragmentShaderTex =
"uniform sampler2D uTex0;																						\n"
"uniform lowp int uEnableAlphaTest;																				\n"
"uniform mediump vec4 uTextureBounds;																			\n"
"lowp vec4 uTestColor = vec4(4.0/255.0, 2.0/255.0, 1.0/255.0, 0.0);												\n"
"in mediump vec2 vTexCoord0;																					\n"
"out lowp vec4 fragColor;																						\n"
"void main()																									\n"
"{																												\n"
" mediump vec4 textureBounds = uTextureBounds;																			"  TEX_FILTER(fragColor, uTex0, vTexCoord0);													\n"
"  if (fragColor == uTestColor) discard;																		\n"
"  if (uEnableAlphaTest == 1 && !(fragColor.a > 0.0)) discard;													\n"
"}																												\n"
;

```

gonetz:
Definitely compiler error: when I replace "name = c;" by "discard;" the problem gone. Looks like the compiler can't detect that "name = c;" affect shader output.

gonetz:
You also did a mistake in this macro:
```
"    name = mix( pInterp_q0, pInterp_q1, interpolationFactor.y ); 												\\\n" // Interpolate in Y direction.
"    name = c;												 													\\\n"

```
I overlooked it.

fzurita:
Whoops, I missed that.

gonetz:
Yep, that was the reason! When I removed wrong line
`"    name = c;"`
compiler found the variable. It was not compiler error. uTextureBound is redundant if in all cases name = c.

fzurita:
Yep, a case of the compiler being too smart!

gonetz:
Just as smart as it should be. Otherwise we would not find that bug.

fzurita:
The Adreno compiler wasn't as smart unfortunately, otherwise I would have had the same issue.

gonetz:
3point version of the shader has compilation error
```
Fragment shader failed to compile with the following errors:
ERROR: 0:31: error(#112) Too few args in Macro: TEX_FILTER
ERROR: 0:31: error(#202) No matching overloaded function found: textureSize
ERROR: 0:31: error(#132) Syntax error: ")" parse error
ERROR: error(#273) 3 compilation errors.  No code generated

```

gonetz:
Sorry, it is probably my fault :(

gonetz:
Yes, I did not remove some debug modifications. It is ok now. I'll push the patch.

gonetz:
Done.

gonetz:
Re-done.

fzurita:
Cool, thanks. Sorry about that bad code in there.

PR #1195, opened 2016-10-29
Blit frame buffer workaround

fzurita:
Some older devices with Adreno GPUs with older drivers produce output that is upside down. I've been doing some reading and it seems to be due to glBlitFrameBuffer producing upside down output in those older drivers when the screen is rotated. This is a workaround for that.

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -859,8 +859,9 @@ void FrameBufferList::renderBuffer(u32 _address)
 		srcY0 = (GLint)(srcY0*yScale);
 		srcY1 = srcY0 + VI.real_height;
 	}
-
-	FrameBuffer * pFilteredBuffer = PostProcessor::get().doBlur(PostProcessor::get().doGammaCorrection(pBuffer));
+	PostProcessor & postProcessor = PostProcessor::get();
+	FrameBuffer * pFilteredBuffer = postProcessor.doBlur(postProcessor.doGammaCorrection(
+		postProcessor.doOrientationCorrection(pBuffer)));
 
 	const bool vi_fsaa = (*REG.VI_STATUS & 512) == 0;
 	const bool vi_divot = (*REG.VI_STATUS & 16) != 0;
@@ -931,7 +932,8 @@ void FrameBufferList::renderBuffer(u32 _address)
 		const u32 size = *REG.VI_STATUS & 3;
 		pBuffer = findBuffer(_address + (((*REG.VI_WIDTH)*VI.height)<<size>>1));
 		if (pBuffer != nullptr) {
-			pFilteredBuffer = PostProcessor::get().doBlur(PostProcessor::get().doGammaCorrection(pBuffer));
+			pFilteredBuffer = postProcessor.doBlur(postProcessor.doGammaCorrection(
+				postProcessor.doOrientationCorrection(pBuffer)));
 			srcY0 = 0;
 			srcY1 = srcPartHeight;
 			dstY0 = dstY1;
--- a/src/PostProcessor.cpp
+++ b/src/PostProcessor.cpp
@@ -242,6 +242,27 @@ FRAGMENT_SHADER_END
 "}																			\n"
 ;
 
+static const char* orientationCorrectionShader =
+SHADER_VERSION
+"#if (__VERSION__ > 120)													\n"
+"# define IN in																\n"
+"# define OUT out															\n"
+"# define texture2D texture													\n"
+"#else																		\n"
+"# define IN varying														\n"
+"# define OUT																\n"
+"#endif // __VERSION __														\n"
+"IN mediump vec2 vTexCoord;													\n"
+"uniform sampler2D Sample0;													\n"
+"OUT lowp vec4 fragColor;													\n"
+"																			\n"
+"void main()																\n"
+"{																			\n"
+"    fragColor = texture2D( Sample0, vec2(1.0 - vTexCoord.x, 1.0 - vTexCoord.y));       \n"
+FRAGMENT_SHADER_END
+"}																			\n"
+;
+
 static
 GLuint _createShaderProgram(const char * _strVertex, const char * _strFragment)
 {
@@ -321,6 +342,7 @@ PostProcessor::PostProcessor()
 	, m_glowProgram(0)
 	, m_bloomProgram(0)
 	, m_gammaCorrectionProgram(0)
+	, m_orientationCorrectionProgram(0)
 	, m_pResultBuffer(nullptr)
 	, m_FBO_glowMap(0)
 	, m_FBO_blur(0)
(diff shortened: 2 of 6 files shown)


gonetz:
I don't like the idea to add new common config parameter, which is useless for most of users. GLideN64 config panel in M64Py is already a mess. Please guard all these changes by #if GLESX define.

Another solution is to use alternative version of copyTexturedRect for devices, which can't do glBlitFramebuffer right.

loganmc10:
@fzurita i know that Dolphin Emulator ran into these issues with Adreno and glBlitFramebuffer. Maybe if you get on their IRC channel they can tell you how they detect when a fix needs to be applied for this.

fzurita:
That's a good point, I can look at their code.

fzurita:
@loganmc10 I don't see any workarounds in their code for glBlitFrameBuffer for older Adreno drivers. I think they gave up on older Adreno drivers since those older devices can't run Dolphin at all anymore since it requires 64bit. Their organization is nice, they keep all driver bug workarounds in one file, see here:
https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/VideoCommon/DriverDetails.h

@gonetz Yes, I could use that alternative code if I can make it into a run time configuration option. There are many glBlitFrameBuffer calls though in the code, and for the GLES 3.0 version, they are not all being replaced. So I'm not exactly sure using that alternative code path is going to work.

fzurita:
Pull request updated.

src/mupenplus/Config_mupenplus.cpp
```diff
@@ -3,6 +3,7 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <string.h>
+#include <Config.h>
```
gonetz: Why include Config.h twice ? (see next line)

src/PostProcessor.cpp
```diff
@@ -242,6 +242,27 @@ FRAGMENT_SHADER_END
...
+"# define IN varying														\n"
+"# define OUT																\n"
+"#endif // __VERSION __														\n"
+"IN mediump vec2 vTexCoord;													\n"
+"uniform sampler2D Sample0;													\n"
+"OUT lowp vec4 fragColor;													\n"
+"																			\n"
+"void main()																\n"
+"{																			\n"
+"    fragColor = texture2D( Sample0, vec2(1.0 - vTexCoord.x, 1.0 - vTexCoord.y));       \n"
```
gonetz: This option explained as "Enable to render everything upside down"
You invert X coord as well. Is it not only upside down but also inside out?

src/mupenplus/Config_mupenplus.cpp
fzurita: I'll fix that.

fzurita:
Ok, updated again.

src/PostProcessor.cpp
```diff
@@ -591,3 +632,23 @@ FrameBuffer * PostProcessor::doGammaCorrection(FrameBuffer * _pBuffer)
...
 	return m_pResultBuffer;
 }
+
+FrameBuffer * PostProcessor::doOrientationCorrection(FrameBuffer * _pBuffer)
+{
+	if (_pBuffer == nullptr)
+		return nullptr;
+
+	if (((*REG.VI_STATUS & 8) | config.generalEmulation.enableBlitScreenWorkaround) == 0)
+		return _pBuffer;
```
gonetz: If (*REG.VI_STATUS & 8) != 0 this function will do its work regardless of enableBlitScreenWorkaround value.
(*REG.VI_STATUS & 8) checks gamma correction bit. Why you use it?

src/FrameBuffer.cpp
```diff
@@ -931,7 +931,8 @@ void FrameBufferList::renderBuffer(u32 _address)
 		const u32 size = *REG.VI_STATUS & 3;
 		pBuffer = findBuffer(_address + (((*REG.VI_WIDTH)*VI.height)<<size>>1));
 		if (pBuffer != nullptr) {
-			pFilteredBuffer = PostProcessor::get().doBlur(PostProcessor::get().doGammaCorrection(pBuffer));
+			pFilteredBuffer = PostProcessor::get().doBlur(PostProcessor::get().doGammaCorrection(
+				PostProcessor::get().doOrientationCorrection()));
```
gonetz: Compilation error. Need doOrientationCorrection(pBuffer)

src/PostProcessor.cpp
fzurita: I was following the gamma correction example without fully understanding what that did. Thanks for explaining. I'll fix that.

fzurita:
Updated

gonetz:
This commit causes heavy glitches on my system. It breaks gamma correction and other things. I can't find what is going on because debugger is broken too! Very dark magic.

fzurita:
Hmm, great. It seemed to work ok on Android. All it's doing is mirroring the X and Y texture pixels. If you configure it to off, do there glitches go away?

gonetz:
My fault. You add new config option. GUI lib uses it and thus must be rebuilt. It is not automatic :(

I found two things to improve. I'll point on them.

src/PostProcessor.cpp
```diff
@@ -405,10 +427,21 @@ void PostProcessor::_initBlur()
...
+	assert(loc >= 0);
+	glUniform1i(loc, 0);
+	glUseProgram(0);
+}
+
 void PostProcessor::init()
 {
 	_initCommon();
 	_initGammaCorrection();
+	_initOrientationCorrection();
```
gonetz: let's make _initOrientationCorrection call optional, as for bloom.
why create resources, which will not be used?

fzurita:
Ok

src/PostProcessor.cpp
fzurita: Sure, that's no problem.

src/FrameBuffer.cpp
```diff
@@ -931,7 +931,8 @@ void FrameBufferList::renderBuffer(u32 _address)
 		const u32 size = *REG.VI_STATUS & 3;
 		pBuffer = findBuffer(_address + (((*REG.VI_WIDTH)*VI.height)<<size>>1));
 		if (pBuffer != nullptr) {
-			pFilteredBuffer = PostProcessor::get().doBlur(PostProcessor::get().doGammaCorrection(pBuffer));
+			pFilteredBuffer = PostProcessor::get().doBlur(PostProcessor::get().doGammaCorrection(
+				PostProcessor::get().doOrientationCorrection(pBuffer)));
```
gonetz: postProcessor.doBlur(postProcessor.doGamma(postProcessor.do...

src/FrameBuffer.cpp
```diff
@@ -860,8 +860,8 @@ void FrameBufferList::renderBuffer(u32 _address)
 		srcY1 = srcY0 + VI.real_height;
 	}
 
-	FrameBuffer * pFilteredBuffer = PostProcessor::get().doBlur(PostProcessor::get().doGammaCorrection(pBuffer));
-
+	FrameBuffer * pFilteredBuffer = PostProcessor::get().doBlur(PostProcessor::get().doGammaCorrection(
+		PostProcessor::get().doOrientationCorrection(pBuffer)));
```
fzurita: Sure, that will make it more concise.

fzurita:
Ok, updated again.

src/FrameBuffer.cpp
```diff
@@ -931,7 +932,9 @@ void FrameBufferList::renderBuffer(u32 _address)
 		const u32 size = *REG.VI_STATUS & 3;
 		pBuffer = findBuffer(_address + (((*REG.VI_WIDTH)*VI.height)<<size>>1));
 		if (pBuffer != nullptr) {
-			pFilteredBuffer = PostProcessor::get().doBlur(PostProcessor::get().doGammaCorrection(pBuffer));
+			PostProcessor & postProcessor = PostProcessor::get();
```
gonetz: Why? We already get it above.

fzurita:
Yikes, good point!

fzurita:
Ok, updated again.

gonetz:
Merged: 7ca4d0430b4
Phew!

fzurita:
Thanks! Sorry to make it difficult.

PR #1199, opened 2016-10-31
Minor GLSL optimization

loganmc10:
There is no point setting alphaValue and then potentially changing it, might as well wrap the alternative in an "else" statement


fzurita:
You would think that this kind of optimization the glsl compiler would do for you. Maybe they are not that smart yet.

loganmc10:
Maybe a very smart compiler, but I think like most things it would just execute it line by line, so it would potentially do the clamp and then overwrite the result with 0.125 in some scenarios

gonetz:
Merged: 2160f41451d554

PR #1201, opened 2016-11-01
2 column UI and simplified English

Aquatakat:
As more features have been added to GLideN64 I've noticed the UI becoming more cluttered and complex. I've made an attempt to clear up the UI and simplify the language used in the tooltips. Most of the changes I've made were based around Windows UI guidelines.

I'm pretty much useless with C/++ so there's still some work to be done to hook the UI up to the plugin. Also I think it will need to be re-localised.

I hope you like it.


gonetz:
You did a large work, thanks!

I'm not a specialist in GUI design. So, I'll not criticize or praise the new layout.  It is for users to decide. I have several comments about things, which either have errors or, as  I think, made wrong:
- Copy color and depth buffer. Replacement radio buttons by combo box is questionable. IMO, radio buttons are more convenient here. I see, that you added explanation for these options, so user can select necessary option without tooltips. Radio buttons are better for experienced users.
- The same can be said about "Frame buffer swap"
- "Emulate depth compare" - confusing option name. Depth compare is always emulated. Almost no game will work without it.  This option is "(Shader based) emulation of N64 depth compare". It requires Frame buffer emulation enabled and thus logically should be inside Frame buffer group box.
- "Remember color calculation" -  what is it? Shader storage? Very confusing.
- "Fix black lines between 2D elements" enabling this option does not affect speed at all. The difference between two modes is not speed, but graphical glitches they can remove (or add!).
- "Render 2D elements in N64 resolution" may cause performance drop, but this is not mentioned.
- Warning on top of Frame buffer group box is a very good idea. However: mip-mapping does not depend on frame buffer. Depth compare is basic GL functionality and it works without fb emulation. Only shader based emulation of N64 depth compare requires fb emulation.
- "Render entire frame buffer to output" - it does not look better or clearer after rename. 
- "Textures" tab is all about textures enhancement. "Initial filtering" group box should not be there, it is general emulation option. If you put this option on "Textures" tab, why you did not put other texture-related options there? Anisotropic filtering, mip-mapping, 2D elements. The name "Initial filtering" is very confusing too.
- "Compress texture cache", "Convert textures to 16 bpp" are not General settings. These options are for Enhanced/HD textures only.
- Cache size (MB) - cache for what? Explanation needed.
- Enhanced cache size (MB). Enhanced cache - confusing name. It is cache for enhanced textures. It uses conventional RAM, not video memory, as the main textures cache.
- In general: old design of "Textures" tab is more clear and convenient, imo.
- "Filters" tab is ok. "Selecting this option disables game specified gamma correction." I'd say overrides, not disables. "Selecting this option overrides game's gamma settings."
- "On screen display" - ok.

Sarkie:
Do you have a screenshot to show this for others?

gonetz:
You may use this build to test new GUI layout:
https://drive.google.com/file/d/0B0YqMPjGo3B2TG5palRnNnBnLU0/view?usp=sharing

Don't run games with it. I removed lots of code to make it compilable, so many controls do nothing there.

Aquatakat:
> - Copy color and depth buffer. Replacement radio buttons by combo box is questionable. IMO, radio buttons are more convenient here. I see, that you added explanation for these options, so user can select necessary option without tooltips. Radio buttons are better for experienced users.
> - The same can be said about "Frame buffer swap"

Here I disagree, mostly because the frame buffer emulation group box is very complex. By using a drop down here there are fewer visible elements which makes changing settings here less intimidating.

Alternatively we could move all the frame buffer settings to their own tab - similar to an "Advanced" tab in other programs. Then it would make more sense to use radio buttons.

> - "Emulate depth compare" - confusing option name. Depth compare is always emulated. Almost no game will work without it. This option is "(Shader based) emulation of N64 depth compare". It requires Frame buffer emulation enabled and thus logically should be inside Frame buffer group box.

The word "emulate" I think covers any confusion here. [The word implies perfection](http://www.merriam-webster.com/dictionary/emulate), and because every plugin offers some kind of baseline depth compare without notifying the user or giving them the option to turn it off I think that it could be expected that this option isn't too confusing and offers something different.

The reason I moved it out of the frame buffer group box is because it emulates a very specific effect that feels more in line with the other options inside General settings. So many options in GLideN64 require frame buffer emulation that it would be impossible to constrain them all to that group box. Because of that I've left only options that deal directly with managing the frame buffer there.

> - "Remember color calculation" - what is it? Shader storage? Very confusing.

I agree - it is confusing. I was thinking about this option when I was working on this. Is there any reason why this shouldn't be turned on all the time? Then we could remove the checkbox entirely.

> - "Fix black lines between 2D elements" enabling this option does not affect speed at all. The difference between two modes is not speed, but graphical glitches they can remove (or add!).
> - "Render 2D elements in N64 resolution" may cause performance drop, but this is not mentioned.

I see - I'll make some adjustments.

> - Warning on top of Frame buffer group box is a very good idea. However: mip-mapping does not depend on frame buffer. Depth compare is basic GL functionality and it works without fb emulation. Only shader based emulation of N64 depth compare requires fb emulation.

Thank you re. the warning. I thought most (or all?) of the shader based emulation was dependent on the frame buffer being enabled. Is that wrong?

> - "Render entire frame buffer to output" - it does not look better or clearer after rename. 

"Render frame buffer as texture" is ambiguous as to what will happen with the texture. By specifying "to output" it becomes clear that the frame buffer will be displayed.

> - "Textures" tab is all about textures enhancement. "Initial filtering" group box should not be there, it is general emulation option. If you put this option on "Textures" tab, why you did not put other texture-related options there? Anisotropic filtering, mip-mapping, 2D elements. The name "Initial filtering" is very confusing too.

I agree that "Initial filtering" is a confusing name. It was a difficult control to label. This might have been a bad change. Your original labels are more clear.

The reason I put these settings here is because they all are applicable before the textures are rendered to the screen. Anisotropic filtering in every UI I've ever seen always lives with anti-aliasing. Mip-mapping is a specific emulation effect and the 2D fixes resize texrects after they've been rendered - or at least they appear to.

> - "Compress texture cache", "Convert textures to 16 bpp" are not General settings. These options are for Enhanced/HD textures only.

Ah, I didn't know that. They were outside of the Texture enhancement group box so I thought they were for all textures. I'll adjust this.

> - Cache size (MB) - cache for what? Explanation needed.
> - Enhanced cache size (MB). Enhanced cache - confusing name. It is cache for enhanced textures. It uses conventional RAM, not video memory, as the main textures cache.

I think being on the Textures tab implies that these caches will be for textures. I don't think the user cares what type of RAM the cache will use.

> - "Filters" tab is ok. "Selecting this option disables game specified gamma correction." I'd say overrides, not disables. "Selecting this option overrides game's gamma settings."

Yes, I agree.

> - "On screen display" - ok.

Thank you. :)

Aquatakat:
Here are some screenshots:

About window:
image

Video tab:
image

Emulation tab (no FB emulator help):
image

Emulation tab (with FB emulator help):
image

Textures tab:
image

Filters tab:
image

On screen display tab:
image

oddMLan:
Could you try to fit more controls in less space? The empty space in some of the tabs is bothering me.
Maybe make the UI more like a wide rectangle instead of a semi-square?
Also IMO the vertical sliders look jarring. The vertical design fits more if they were volume controls IMO. Every single Antialias and Anisotropic slider you see in any GUI for any game, software or emulator has been horizontal as far as I know.

**Video**
- Language combobox needs to be slightly wider
- "Crop image" W and H boxes should be proportional to the W and H boxes of the Windowed resolution section
- Windowed resolution should be a editable combobox with preincorporated resolutions in the dropdown list
http://www.wpf-tutorial.com/chapters/list-controls/images/combobox_editable.png
that way you can do away with the W and H boxes
"Custom" option still should be in the list but selecting it focuses the editable combobox with the blinking cursor so the user knows that they can type something

- Maybe also the resolutions in the list should specify aspect ratio in parenthesis (eg. "640x480 (4:3), 1280x720 (16:9)"

**Emulation**
_Frame buffer_  -> **Framebuffer**
"_Clearing_ this option" should be or _"Unchecking"_ IMO (much more common usage)

**Textures**
- "General settings" and "Texture enhancements" should be placed backwards (general on the right and enhancements on the left)
- "Use texture pack" could fit beneath "Texture enhancements", with "General settings" being at the rightmost part, leaving "Use texture pack" beneath it (only if you follow my suggestion of making the UI wider)
- "Filter" and "Enhancement" comboboxes should be wider
- "Texture pack path" should include editable current path so one can copy & paste it

**Filters** LGTM - only get rid of that dreaded empty space with a more rectangular layout

**On screen display** - **_On-screen display_** or just OSD (a general term which is commonly known so abbreviating it won't hurt)
ea
or better yet, just dragging the OSD preview in a square should be possible but IDK how hard would that be to code

Aquatakat:
> Could you try to fit more controls in less space? The empty space in some of the tabs is bothering me.
Maybe make the UI more like a wide rectangle instead of a semi-square?

The size of the UI is mostly automatically generated by Qt based on the contents. The Emulation tab takes up most of the space. I was thinking we might move the frame buffer options out of that tab.

> Also IMO the vertical sliders look jarring. The vertical design fits more if they were volume controls IMO. Every single Antialias and Anisotropic slider you see in any GUI for any game, software or emulator has been horizontal as far as I know.

Actually I agree, but fitting horizontal sliders into that wider layout looked too wide. The vertical sliders just fit better into the UI. Maybe I could limit the horizontal sliders' width, but that would create some unnecessary negative space to the right.

> - Language combobox needs to be slightly wider

The language combobox resizes to fit its content.

image

Do you think it needs to be wider?

> - "Crop image" W and H boxes should be proportional to the W and H boxes of the Windowed resolution section
> - Windowed resolution should be a editable combobox with preincorporated resolutions in the dropdown list  
> http://www.wpf-tutorial.com/chapters/list-controls/images/combobox_editable.png
that way you can do away with the W and H boxes  
> "Custom" option still should be in the list but selecting it focuses the editable combobox with the > - blinking cursor so the user knows that they can type something
> - Maybe also the resolutions in the list should specify aspect ratio in parenthesis (eg. "640x480 (4:3), 1280x720 (16:9)"

I like these ideas a lot - making the windowed resolution an editable combobox. It would require some wrangling with C++ that's beyond me though.

I think if the aspect ratios are specified in the resolution lists they could be constrained to the most common ones: 4:3, 16:9, 16:10.

> - Frame buffer -> Framebuffer

I disagree. Framebuffer isn't a word. Being specific and separating the words frame and buffer doesn't affect the meaning.

> - "Clearing this option" should be or "Unchecking" IMO (much more common usage)

I agree. I was working based on [these guidelines](https://msdn.microsoft.com/en-us/library/windows/desktop/dn742401(v=vs.85).aspx#documentation) but it's probably better to use more common language.

> - "General settings" and "Texture enhancements" should be placed backwards (general on the right and enhancements on the left)

Yes, I agree.

> - "Use texture pack" could fit beneath "Texture enhancements", with "General settings" being at the rightmost part, leaving "Use texture pack" beneath it (only if you follow my suggestion of making the UI wider)

I would prefer to keep the Texture enhancements and Use texture pack group boxes together vertically. It implies a subtle connection between the two.

> - "Filter" and "Enhancement" comboboxes should be wider

See above re. the language combobox.

> - "Texture pack path" should include editable current path so one can copy & paste it

I like this idea, but again it's beyond my abilities. Gonetz?

> - Filters LGTM - only get rid of that dreaded empty space with a more rectangular layout

The only real way to get rid of that negative space would be to move these controls to another tab, but they don't fit anywhere else particularly well. Personally, I'm alright with a little blank space. Maybe we need more opinions.

> On screen display - On-screen display or just OSD (a general term which is commonly known so abbreviating it won't hurt)

I agree. OSD is a good name.

Given a little time I could probably manage something similar to your mock-up in Qt Designer. Dragging a preview is again beyond me. Gonetz might have some input.

gonetz:
General note. It is hard to make GUI design, which everybody will love. We all have different tastes. However, such discussions allows to find compromise solution. I suggest to turn that GLideNUI based project into stand-alone application, which will run the GUI, so everyone will be able to play with it. That prototype can be put on GitHub as separate project, so everyone with skills in design and with QT installed can modify the project, build it, test the result and pull-request updates.

gonetz:
> Alternatively we could move all the frame buffer settings to their own tab - similar to an "Advanced" tab in other programs. Then it would make more sense to use radio buttons.

Well, not a bad idea. At least worth to try.

> The word "emulate" I think covers any confusion here. The word implies perfection, and because every plugin offers some kind of baseline depth compare without notifying the user or giving them the option to turn it off I think that it could be expected that this option isn't too confusing and offers something different.

English is not my native language, I don't feel that subtle nuance of meaning. I use "emulate smth" as "enable emulation of smth". E..g we have option "emulate frame buffer". It does not implies perfection, it implies that with that option disabled no frame buffer effects will work.

> The reason I moved it out of the frame buffer group box is because it emulates a very specific effect

It emulates depth compare in general, not a specific effect. It is alternative implementation of depth compare, and it can be used for every game.

> I agree - it is confusing. I was thinking about this option when I was working on this. Is there any reason why this shouldn't be turned on all the time? Then we could remove the checkbox entirely.

Yes and no. This feature designed to be always enabled, but during development it is useful to be able to disable it.

>I thought most (or all?) of the shader based emulation was dependent on the frame buffer being enabled. Is that wrong?

It is wrong. Shaders and frame buffers are not related to each other. Shader based depth compare requires frame-depth emulation because it creates new depth buffer. Noise, mip-map, hw lighting, 3point filtering work regardless of fb emulation.

> The reason I put these settings here is because they all are applicable before the textures are rendered to the screen. Anisotropic filtering in every UI I've ever seen always lives with anti-aliasing. Mip-mapping is a specific emulation effect and the 2D fixes resize texrects after they've been rendered - or at least they appear to.

Anisotropic filtering applicable before the textures are rendered to the screen.
Mip-mapping is not a specific emulation effect - it has the same meaning as GL mip-maping, but implemented differently. It is designed to be always enabled, but some people don't like it.
2D fixes. "Render 2D in native resolution" renders block of textures into a buffer and then renders that buffer on its place to screen. "Fix black lines" works differently. It is a hack, which corrects coordinates of texture rectangles to eliminate splits between them.

> I think being on the Textures tab implies that these caches will be for textures. I don't think the user cares what type of RAM the cache will use.

They should care. Low size for GL buffer will affect performance of graphics card, too large value may lead to crash or performance problems. Low size of cache for enhanced textures will increase CPU load, high value will increase RAM and disk space usage.

Aquatakat:
> General note. It is hard to make GUI design, which everybody will love. We all have different tastes. However, such discussions allows us to find compromise solution. I suggest to turn that GLideNUI based project into stand-alone application, which will run the GUI, so everyone will be able to play with it. That prototype can be put on GitHub as separate project, so everyone with skills in design and with QT installed can modify the project, build it, test the result and pull-request updates.

I've actually never built a standalone Windows program or Qt program before so I have no idea what I'm doing in that respect. Either way people could probably pull Aquatakat/GlideN64 and open the GlideNUI.pro project in QT Creator or Designer and using the Alt+Shift+R preview which is what I've been doing.

Aquatakat:
I changed a bunch of the  language and layout and I'm interested to hear input.

image

image

image

image

image

image

image

gonetz:
That way it looks much better, imo. Good work! Comments:

* "Render 2D in native res" and "Fix black lines" options are mutually exclusive. GUI should reflect that fact somehow.

* Where "read in 4kb chunks" options for "Use emulator help" ?

* Texture enhancement tab. There are two modes: 
  1. filter original N64 textures
  2. use external texture pack.

Usually only one mode is used. It is possible to use both to filter N64 textures missing in texture packs, but I think it is better to make separate group box for texture enhancement options. Also, options "Convert to 16bpp" and "Compress cache" are common for both modes and must be placed out of group boxes.

* OSD. Position control looks cool, but gauges currently can be displayed only in corners. I'm not sure that placing them on center of a screen bound is convenient.

AmbientMalice:
"From VRAM" depth buffer has far more issues than "Software". Software is more compatible by a huge margin, and is also dramatically faster. So the UI should reflect that. I mean, I'm honestly not sure if there's any good reason for people to be using VRAM depth buffer on Desktop, at least.

Aquatakat:
> "Render 2D in native res" and "Fix black lines" options are mutually exclusive. GUI should reflect that fact somehow.

Like in the original layout, checking the first option disables the second. I think that's probably enough.

> Where "read in 4kb chunks" options for "Use emulator help" ?

These options appear/disappear based on whether or not the "Use emulator help" checkbox is checked:

image

> - Texture enhancement tab. There are two modes: 
> - - filter original N64 textures
> - - use external texture pack.
> Usually only one mode is used. It is possible to use both to filter N64 textures missing in texture packs, but I think it is better to make separate group box for texture enhancement options. Also, options "Convert to 16bpp" and "Compress cache" are common for both modes and must be placed out of group boxes.

I have an idea for this. I'll put some work into it.

> OSD. Position control looks cool, but gauges currently can be displayed only in corners. I'm not sure that placing them on center of a screen bound is convenient.

It's probably convenient for top centre and bottom centre because few games use those spots to place HUD elements. I only included left and right centre to complete the square. It's easy to disable buttons here, though.

***

> "From VRAM" depth buffer has far more issues than "Software". Software is more compatible by a huge margin, and is also dramatically faster. So the UI should reflect that. I mean, I'm honestly not sure if there's any good reason for people to be using VRAM depth buffer on Desktop, at least.

Could you provide an example? At 1080p I get about the same performance with it on VRAM or software in Perfect Dark (40-50 VI/s).

Aquatakat:
Here's what I came up with for the texture enhancements tab:

image

image

AmbientMalice:
>Could you provide an example? At 1080p I get about the same performance with it on VRAM or software in Perfect Dark (40-50 VI/s).

My PC gets 2x (or more) the framerate with some games using software depth buffer. Perfect Dark should not be running at 40-50VIs with no vsync or frame capping. This is with async FB, BTW.

This is a screenshot I took a while back.

copyfromvram_vs_softwaredepth

Aquatakat:
I adjusted a bunch of settings and I managed to get about 160-170 VIs in that same scene with the plugin set to VRAM or software. I'm guessing it's pretty hardware-dependent.

Here's the language I came up with:

image

gonetz:
@Aquatakat 
> These options appear/disappear based on whether or not the "Use emulator help" checkbox is checked:

This is a cool design trick, I like it! I myself use only disabling options, not hiding them.
One notice: "Render entire frame buffer to output" should disappear also when "Use emulator help" enabled, because emulator helps to write to screen only necessary pixels . Thus, it should be placed nearby to fb copy options.

And again: "Render entire frame buffer to output" is confusing. May be "Render entire N64 frame buffer to output". At least the option should inform that we take data from N64 memory, not from video card.
Also "Copy frame buffer to N64 memory:" -> "Copy GL frame buffer to N64 memory:"
Also "Copy depth buffer to N64 memory:" -> "Copy GL depth buffer to N64 memory:"

>It's probably convenient for top centre and bottom centre because few games use those spots to place HUD elements. I only included left and right centre to complete the square. It's easy to disable buttons here, though.

I agree that top centre and bottom centre can be useful. It will require some additional programming, but not that much. Left and right centre buttons should be removed or disabled.

> Here's what I came up with for the texture enhancements tab:

I like it. Simple and clear. Please make it.

Aquatakat:
> This is a cool design trick, I like it! I myself use only disabling options, not hiding them.
One notice: "Render entire frame buffer to output" should disappear also when "Use emulator help" enabled, because emulator helps to write to screen only necessary pixels . Thus, it should be placed nearby to fb copy options.
> 
> And again: "Render entire frame buffer to output" is confusing. May be "Render entire N64 frame buffer to output". At least the option should inform that we take data from N64 memory, not from video card.  
> Also "Copy frame buffer to N64 memory:" -> "Copy GL frame buffer to N64 memory:"  
> Also "Copy depth buffer to N64 memory:" -> "Copy GL depth buffer to N64 memory:"

I think GL might be a confusing term. This is what I came up with:

image

> I agree that top centre and bottom centre can be useful. It will require some additional programming, but not that much. Left and right centre buttons should be removed or disabled.

I'll disable the buttons. You could re-enable them if there's some demand for them to be included. I like maintaining the square of buttons.

AmbientMalice:
@Aquatakat Copying every depth buffer from VRAM to RAM is slower than rendering it in software. Software depth buffer is significantly faster unless there's a CPU speed issue, which is unlikely, and it is also significantly more accurate.  It's faster+more accurate by a huge margin. Vigilante 8 performs far better with software than From VRAM, as an example. I've never seen a game where it is slower.

Part of the reason the software depth buffer was ported from GLide64 is because "From VRAM" produces incorrect results in a range of games.

>I adjusted a bunch of settings and I managed to get about 160-170 VIs in that same scene with the plugin set to VRAM or software.

You didn't have "use per-game settings" enabled, did you? Because it will override that particular setting and set it to From VRAM regardless.

loganmc10:
@AmbientMalice While we're on this topic, do you think some of the entries in GLideN64.custom.ini could be changed from ```frameBufferEmulation\copyDepthToRDRAM=1``` (vram) to ```frameBufferEmulation\copyDepthToRDRAM=2``` (software)

Aquatakat:
@AmbientMalice 

Whoops, I feel a bit dumb. You're right - when set to software I got about 240 VI's. I adjusted the language to this:

image

AmbientMalice:
@loganmc10 IMO, those entries should probably be removed since "software depth" is now the default setting. The less bloat, the better. In some cases, depth buffer will break things. Shadow Man breaks, I believe. So that should be forced to 0, probably.

@Aquatakat All's well that ends well.

@Gonetz Does the desktop version of the plugin actually need From VRAM depth buffer emulation? I mean, I understand it being useful for devices with weaker CPUs, but the actual overheads of copying the buffers seem fairly counterproductive.

Another thing is the partial framebuffer copy stuff. That doesn't work in PJ64, partially works in mupen64plus, and kinda works in mupen64. I wonder whether it should be greyed out by default in the Zilmar spec version, with a "Enable experimental FB method - won't work on most emulators" checkbox? Confusion could arise to what those setting actually do, even though they're essentially placebos on PJ64.

gonetz:
@AmbientMalice 

Depth buffer in VRAM is more complete and correct than software one, it is obvious. Read from VRAM should be more correct. There are cases, where it is not so, e.g. Perfect Dark. Also, Read from VRAM can be faster when FBInfo is enabled. It is faster to read few chunks from VRAM when it is necessary than always do software rendering.

Ragarding FB info. It kinda works in mupen64 (JFG works great with it) and with 1964. I hope zilmar will add support to it someday. "Use emulator help" is disabled by default.

gonetz:
> I think GL might be a confusing term. This is what I came up with:

Yes, that's good.

AmbientMalice:
@gonetz Ah, I see what you mean. I did figure that From VRAM **should** be more accurate, but in its current form it isn't. I remember this sectioned from your blog:

>**"Incorrect result. Some pixels of a polygon can be discarded by alpha compare and thus not change color and depth buffers. Software rendering implemented in Glide64 renders only polygon’s depth without any care about its alpha. Thus, pixels which would normally be discarded by alpha compare are stored in the depth buffer. Again, it’s not a problem for coronas emulation, but very noticeable for depth buffer based fog. Run Beetle Adventure Racing with Glide64 – fog around trees is missing, because transparent parts of trees polygons were not discarded during software rendering. Software render have to do color rendering too to get alpha compare information. That is fully functional software rendering is necessary to get correct depth buffer. N64 software plugins/emulators don’t run full speed even on many modern CPU."**

And yea, if partial copies could be widely supported, it would be even faster than software. But unfortunately that's not likely to happen soon.

Aquatakat:
@gonetz I pushed the most recent version. I think we're pretty close to a final version.

Implementing it will require a little programming glue:

I liked @oddMLan 's suggestion for the `windowedResolutionComboBox`, but working with QT Creator/the XML file I can't create the correct behaviour.

The combobox should accept a value like (0-9 x 0-9) as an arbitrary resolution. Here's a regular expression, if it's useful:

    /^0*([0-9]+)\s*x\s*0*([0-9]+)$/

`\1` should contain width and `\2` height. When the text entered doesn't match the expression it should reset to the previous value.

Like @oddMLan suggested, the Custom option should still appear in the dropdown, but when it's selected it should clear and focus the combobox, implying you can type into it. Another option here would be to change the value to something like "Width x Height", but selecting it so when it's typed into it replaces it. I'm not sure how easy that would be to localise, though.

When I preview the form, Qt Creator doesn't show/hide or disable/enable controls based on signals initially. I'm not sure if that's a thing that's fixed when the form is actually compiled. If not, the new controls that send signals are `fbInfoEnableCheckBox`, `texturePackOffRadioButton` and possibly `nativeRes2D_checkBox`. I don't think I changed any of the other controls that send signals.

Other than that I think it's pretty simple, but I noticed in the most recent builds some options don't work properly because their `objectName`s have been renamed, which would need to be fixed.

gonetz:
Great, thanks!

I checked it briefly. Already looks good.
Some names still need polishing. For example: "Store complex N64 color calculations as shaders for performance (recommended)" Actually "Store compiled shader programs ..."

I'll try to build it tomorrow. I'll see, what can be done for "Custom window resolution" option.

oddMLan:
The newest version looks great!
But yeah I agree with Gonetz about the shader thing... that's a mouthful of a sentence and would sound scary to any end user
Are tooltips on option hover staying? Because some options might need clarification. I think 
"Store compiled N64 shaders (recommended)" should be enough. The tooltip for it should say something "Uncheck this option if you see wrong lights or wrong transparencies"
Tooltips are really useful to unclutter the UI and unintrusive. Dolphin has them.

Also just a nitpick but I think (recommended) should only be kept for options that might not sound obvious to the end user. OR in the tooltip for them (like Dolphin).
Options that might not need a (recommended) next to them might be Frame buffer (the warning makes the point already) and Compress textures. This also applies to options that have text next to them. Currently there's much text in the UI that it looks a bit cluttered and intimidating.
If that's too hard to decide on, you could always turn Frame buffer swap, Copy video frame buffer & Copy video depth buffer into comboboxes (current style) and leave the text next to each option as it currently is. That way only one option is visible at a time, making the UI less cluttered while ALSO being able to make the window less tall.

@gonetz  @AmbientMalice  thoughts?

Aquatakat:
@oddMLan 

> The newest version looks great!

Thank you :)

> But yeah I agree with Gonetz about the shader thing... that's a mouthful of a sentence and would sound scary to any end user
> Are tooltips on option hover staying? Because some options might need clarification. I think 
> "Store compiled N64 shaders (recommended)" should be enough. The tooltip for it should say something "Uncheck this option if you see wrong lights or wrong transparencies"
> Tooltips are unintrusive really useful to unclutter the UI. Dolphin uses tooltips.

This option is really contentious. I almost want to hide it away in a "debug" or "advanced" window or tab or something, but it seems a bit silly to create an entire area for one control. "Store compiled shaders for performance" is where I'm at right now with it.

I didn't remove the tooltips. I like them a lot. For controls that have both a label and a control like the comboboxes, etc., I removed them from the control, leaving them on the label, but that's mostly so if I edited one I wouldn't have to edit all of them.

> Also just a nitpick but I think (recommended) should only be kept for options that might not sound obvious to the end user. OR in the tooltip for them (like Dolphin).
> Options that might not need a (recommended) next to them might be Frame buffer (the warning makes the point already) and Compress textures. 

I agree for these.

> This also applies to options that have text next to them. Currently there's much text in the UI that it looks a bit cluttered and intimidating.
> If that's too hard to decide on, you could always turn Frame buffer swap, Copy video frame buffer & Copy video depth buffer into comboboxes (current style) and leave the text next to each option as it currently is. That way only one option is visible at a time, making the UI less cluttered while ALSO being able to make the window less tall. These options rarely need to be changed anyways. The GLideN64.ini already has the most optimal config most of the time.

Personally I like how the frame buffer tab turned out, even with the text. You're right that these options rarely need to be changed, so I'm not too terribly worried about how intimidating the tab appears. By limiting them to this tab they're essentially hidden away only for users that need to use them.

For advanced users the vertical orientation makes the window easy to scan, and for novice users the extra explanations can give the options some utility without having to check tooltips for each of them.

gonetz:
@Aquatakat 
I started to fix compilation errors. Please do not modify the layout until I finish, or your changes will be lost. I'll put the result in a new branch, and then we will be able to work async again.

gonetz:
@Aquatakat 
code is in new feature branch 'feature_new_gui": 7c5683e69d0b
It is incomplete. I fixed compilation issues, but some controls are not supported in code yet.

gonetz:
@Aquatakat 
I noticed another problem with current design.
"Use emulator help to read/write frame buffers" option hides buffer read/write options.
It is wrong behavior. FBInfo mode and "read/write each frame" are mutually exclusive. However, related config options are not. FBInfo is not supported by all emulators. Plugin can detect FBInfo usage in run time. It uses that info to enable/disable buffer read/write.
- UseFBInfo flag is set to false when "Use emulator help" is off.
- otherwise it is on only when emulator uses FBInfo. FBInfo can be disabled on emulator side as well (1964)

Plugin checks UseFBInfo flag to read/write buffers, not "Use emulator help" option. Thus, "Use emulator help" and "Buffer read - write" should be available at the same time, as with current design. The option can be renamed, for example as "Use emulator help to read/write frame buffers if supported by emulator", but it is too long and can be too much long on other languages.

Aquatakat:
@gonetz I changed it back from showing/hiding to the original design, where it disables the two checkboxes only.

gonetz:
Conflicts. I changed configDialog.ui to fix compilation errors. You need to pull my feature branch and modify the new configDialog.ui.

gonetz:
@Aquatakat another request:
Editable windowedResolutionComboBox is too complex to program and not very convenient. Besides, Crop option uses different style for Custom, with spin boxes. Let make custom windowedResolution via spin boxes, as it was.

gonetz:
I understood, that my explanation about texture enhancement options was not quite correct.
The situation is similar to FBInfo- FBRead/Write.
Texture enhancement modes supposed to be mutually exclusive, but options are not.
Texture packs exist only for few games. Suppose, I want to use texture pack when it is available, and when it is not, I want my N64 textures be enhanced by some filter. Plugin allows that, GUI does not.

gonetz:
I think this can be closed now.

PR #1202, opened 2016-11-01
Specify a specific CRC function for microcodes

loganmc10:
The CRC calculation in GBI.cpp is unique because it is comparing the results to a pre-created list of CRC's (specialMicrocodes):

https://github.com/gonetz/GLideN64/blob/master/src/GBI.cpp#L34-L56

GLupeN64 is using xxHash for the CRC calculations by intercepting the CRC_Calculate call and running xxHash instead of the regular function. The microcode check fails because that specific CRC needs to be run by the CRC32 calculation. This will allow me to implement xxHash for everything except the microcode calculation, which I will run with CRC32.

This won't change any functionality for the "regular" codepath.

### DIFF
--- a/src/CRC.h
+++ b/src/CRC.h
@@ -2,5 +2,6 @@
 
 void CRC_Init();
 
+u32 CRC_Calculate_Strict( u32 crc, const void *buffer, u32 count );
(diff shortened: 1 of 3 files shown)


gonetz:
merged: 617517737d4

PR #1204, opened 2016-11-03
Fix some issues with multisampling which were causing GL errors

fzurita:
Without these changes the orbs in Paper Mario oracle have corrupt graphics when multisampling is enabled.

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -364,7 +364,14 @@ CachedTexture * FrameBuffer::_getSubTexture(u32 _t)
 	if (y0 + copyHeight > m_pTexture->realHeight)
 		copyHeight = m_pTexture->realHeight - y0;
 
+#ifdef GLESX
+	if (m_pTexture->frameBufferTexture == CachedTexture::fbMultiSample){
+		resolveMultisampledTexture(true);
+		glBindFramebuffer(GL_READ_FRAMEBUFFER, m_resolveFBO);
+	}else
+#endif
 	glBindFramebuffer(GL_READ_FRAMEBUFFER, m_FBO);
+
 	glBindFramebuffer(GL_DRAW_FRAMEBUFFER, m_SubFBO);
 	glDisable(GL_SCISSOR_TEST);
 	glBlitFramebuffer(
--- a/src/DepthBuffer.cpp
+++ b/src/DepthBuffer.cpp
@@ -34,7 +34,7 @@ DepthBuffer::DepthBuffer(DepthBuffer && _other) :
 	m_depthImageFBO(_other.m_depthImageFBO), m_pDepthImageTexture(_other.m_pDepthImageTexture), m_pDepthBufferTexture(_other.m_pDepthBufferTexture),
 	m_depthRenderbuffer(_other.m_depthRenderbuffer), m_depthRenderbufferWidth(_other.m_depthRenderbufferWidth),
 	m_cleared(_other.m_cleared), m_pResolveDepthBufferTexture(_other.m_pResolveDepthBufferTexture), m_resolved(_other.m_resolved),
-	m_pDepthBufferCopyTexture(_other.m_pDepthBufferCopyTexture), m_copied(m_copied)
+	m_pDepthBufferCopyTexture(_other.m_pDepthBufferCopyTexture), m_copied(_other.m_copied)
 {
 	_other.m_depthImageFBO = 0;
 	_other.m_pDepthImageTexture = nullptr;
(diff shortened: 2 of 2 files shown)


src/FrameBuffer.cpp
```diff
@@ -31,8 +31,7 @@ FrameBuffer::FrameBuffer() :
 	m_startAddress(0), m_endAddress(0), m_size(0), m_width(0), m_height(0), m_validityChecked(0),
 	m_scaleX(0), m_scaleY(0),
 	m_copiedToRdram(false), m_fingerprint(false), m_cleared(false), m_changed(false), m_cfb(false),
-	m_isDepthBuffer(false), m_isPauseScreen(false), m_isOBScreen(false),
-	m_needHeightCorrection(false), m_readable(false),
+	m_isDepthBuffer(false), m_isPauseScreen(false), m_isOBScreen(false), m_needHeightCorrection(false),
```
gonetz: Why?

src/FrameBuffer.cpp
```diff
@@ -141,7 +140,9 @@ void FrameBuffer::init(u32 _address, u32 _endAddress, u16 _format, u16 _size, u1
 #ifdef GL_MULTISAMPLING_SUPPORT
 	if (config.video.multisampling != 0) {
 		glBindTexture(GL_TEXTURE_2D_MULTISAMPLE, m_pTexture->glName);
+
 #if defined(GLES3_1)
+
```
gonetz: redundant

src/FrameBuffer.cpp
```diff
@@ -364,9 +365,18 @@ CachedTexture * FrameBuffer::_getSubTexture(u32 _t)
 	if (y0 + copyHeight > m_pTexture->realHeight)
 		copyHeight = m_pTexture->realHeight - y0;
 
-	glBindFramebuffer(GL_READ_FRAMEBUFFER, m_FBO);
+	if (m_pTexture->frameBufferTexture == CachedTexture::fbMultiSample)
+	{
```
gonetz: formatting

src/FrameBuffer.cpp
```diff
@@ -557,7 +567,6 @@ void FrameBufferList::saveBuffer(u32 _address, u16 _format, u16 _size, u16 _widt
 
 	if (m_pCurrent != nullptr) {
 		bPrevIsDepth = m_pCurrent->m_isDepthBuffer;
-		m_pCurrent->m_readable = true;
```
fzurita: I'm using Android studio, it's doing some automatic fixing on its own. I need to pay closer attention to that. It seems to have removed m_readable due to it not being used.

src/FrameBuffer.cpp
```diff
@@ -364,9 +365,18 @@ CachedTexture * FrameBuffer::_getSubTexture(u32 _t)
...
-	glBindFramebuffer(GL_READ_FRAMEBUFFER, m_FBO);
+	if (m_pTexture->frameBufferTexture == CachedTexture::fbMultiSample)
+	{
+		resolveMultisampledTexture(true);
+		glBindFramebuffer(GL_READ_FRAMEBUFFER, m_resolveFBO);
+	}
+	else{
+		glBindFramebuffer(GL_READ_FRAMEBUFFER, m_FBO);
+	}
+
```
fzurita: It's possible, I'm getting GL errors when this is attempted. Once I started blitting the non-multisampled one, the issues went away.

fzurita:
Pull request has been updated.

src/FrameBuffer.cpp
gonetz: What kind of errors?

src/FrameBuffer.cpp
gonetz: It is used. CheckForFrameBufferTexture, line 383

src/FrameBuffer.cpp
fzurita: GL_INVALID_OPERATION

src/FrameBuffer.cpp
fzurita: Interesting, then I guessed it must had been a merge error. I'm not sure how it went away.

src/FrameBuffer.cpp
gonetz: Yes, it seems to be GLES specific.

GLES
GL_INVALID_OPERATION is generated if **GL_SAMPLE_BUFFERS for the read buffer is greater than zero** and the formats of draw and read buffers are not identical, or **the source and destination rectangles are not defined with the same (X0, Y0) and (X1, Y1) bounds**.

Corresponding condition for OpenGL is much less restrictive:
GL_INVALID_OPERATION is generated if the **value of GL_SAMPLE_BUFFERS for both read and draw buffers is greater than zero** and **the dimensions of the source and destination rectangles is not identical**.

Do you see the difference? We have multi sampled read buffer, the dimensions of the source and destination rectangles are identical, but (X0, Y0) and (X1, Y1) bounds are not the same. Conditions are ok for OpenGL, but not for GLES3.

Let's make it as this:
```
#ifdef GLESX
	if (m_pTexture->frameBufferTexture == CachedTexture::fbMultiSample){
		resolveMultisampledTexture(true);
		glBindFramebuffer(GL_READ_FRAMEBUFFER, m_resolveFBO);
	} else
#endif
	glBindFramebuffer(GL_READ_FRAMEBUFFER, m_FBO);

```

fzurita: Sure, I can updates it to that. I do see that GLES is more strict.

fzurita:
Updated again with #ifdef

gonetz:
merged: 478344ab5f570
Thanks! Depth buffers fixes are very useful. I overlooked these issues.

fzurita:
No problem!

PR #1206, opened 2016-11-04
Remove copyDepthToRDRAM from GLideN64.custom.ini

loganmc10:
Since software depth copy is the default these entries should be removed.

That is unless someone is aware of a game that requires VRAM copy?

### DIFF
--- a/ini/GLideN64.custom.ini
+++ b/ini/GLideN64.custom.ini
@@ -9,7 +9,6 @@ frameBufferEmulation\N64DepthCompare=1
 [1080%20SNOWBOARDING]
 Good_Name=1080 Snowboarding (E)(JU)
 frameBufferEmulation\copyToRDRAM=2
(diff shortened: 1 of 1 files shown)


AmbientMalice:
I think Gonetz prefers to leave the game names in the ini file, even if they're not in use.

loganmc10:
Don't they need something in the section though? I removed them because they don't even have a "Good_Name" entry. I don't use PJ64 so I'm not sure how it parses the ini file.

gonetz:
> I think Gonetz prefers to leave the game names in the ini file, even if they're not in use.

No. Why?

AmbientMalice:
@gonetz I seem to recall removing unused names once, and you said I shouldn't. But that might have been for Japanese games where the actual "name" isn't always obvious. If you're fine with it, no problem.

PR #1207, opened 2016-11-04
Ini Cleanup.

AmbientMalice:
Removed redundant async entries. (Async is default. Forcing it is counterproductive. It should only be forced to sync and OFF when needed,) Disabled depth buffer for Shadow Man, since that breaks pause menu.


loganmc10:
For some reason this is showing up as a binary file on GitHub, the diff isn't working.

AmbientMalice:
@loganmc10 I noticed that. Maybe there's something wrong with the web editor?

loganmc10:
I'm not sure, the pull request I just sent was edited on GitHub

AmbientMalice:
I'll try again.

PR #1208, opened 2016-11-04
INI cleanup.

AmbientMalice:
Removed redundant async FB copies, and disabled depth buffer copies on Shadow Man to fix pause menu.

### DIFF
--- a/ini/GLideN64.custom.ini
+++ b/ini/GLideN64.custom.ini
@@ -126,6 +126,7 @@ frameBufferEmulation\copyToRDRAM=1
 
 [F-ZERO%20X]
 Good_Name=F-Zero X (E)(J)(U)
+
 [18000458]
 Good_Name=Famista 64 (J)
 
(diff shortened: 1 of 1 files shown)


ini/GLideN64.custom.ini
```diff
@@ -136,8 +125,6 @@ frameBufferEmulation\copyToRDRAM=1
 
 [F-ZERO%20X]
 Good_Name=F-Zero X (E)(J)(U)
-frameBufferEmulation\copyToRDRAM=2
-
```
loganmc10: This will make the spacing a little ugly

AmbientMalice:
I considered removing sync FB copies for DK64, but I dunno. Basically, they have a minor HUD side effect, just like Perfect Dark and Quake II, but I figure most players are going to prefer performance.

gonetz:
Please remove all empty entries with no options set.

AmbientMalice:
Okay. What about DK64? Should I leave sync FB forced? Async is quite a bit faster, and the side effects are generally minor.

gonetz:
Ok, remove Dk64 as well.

gonetz:
merged: d061ca044de1
Thanks!

purplemarshmallow:
@AmbientMalice have you checked Donkey Kong 64 (E)?
This version crashes with async fb

PR #1212, opened 2016-11-05
Windowed spinboxes re-added, textures tab adjusted, FB tab adjusted

Aquatakat:


gonetz:
Thanks!
I add support in code. Everything is functional now.

I now see two issues:
* all frame buffer option must be grayed out when 'Emulate frame buffer' checkbox is off to show that these options have no result without 'Emulate frame buffer' on
* the same is for 'Use texture pack" checkbox

These options either should be placed to checkable groupbox or be disabled manually one by one.

PR #1214, opened 2016-11-06
Frame buffer and texture pack disables controls; new OSD arrows

Aquatakat:


gonetz:
Merged: 7c3b28cbb51

Problem 1. All arrows .ico files and Info.ico are broken. QT and IrfanView can't open them. Thus, they are not displayed. Info.png is correct, so I converted it to ico myself and it works. However, original file is in low resolution and does not look good after conversion. Png files for arrows are broken too.
Please pull current state of feature_new_gui branch, correct icons and make pull request over current state of the branch.

Problem 2. The idea with Info warning is good, but that warning always appears when I open config dialog, even when FB emulation is checked.

gonetz:
Problem 2 is fixed: 54ce47952

PR #1218, opened 2016-11-07
Icons reconverted; filters tab frame buffer message fixed

Aquatakat:
ffmpeg apparently does a poor job converting images. I used some silly online utility. They display in Irfanview.

The filters tab also has a message that might appear at the wrong time, so I duplicated/edited some C++ for those messages also.


gonetz:
Issues left:

* Texture pack options are enabled when you just open 'Texture pack' tab regardless of 'Use texure pack' checkbox state.
* 'Read buffer in 4KB chunks' options are not disabled when 'Emulate frame buffer' checkbox not checked.

Issue #1213, opened 2016-11-06
Feature: new GUI


gonetz:
Aquatakat designed new nice GUI layout. Code is in https://github.com/gonetz/GLideN64/tree/feature_new_gui 
feature branch.

WIP build:
https://drive.google.com/file/d/0B0YqMPjGo3B2TGpsenpZeTVsWUU/view?usp=sharing

We need your feedback.

LuigiBlood:
I actually like this GUI better overall. I do have one issue (not that I'll ever use it though):
![image](https://cloud.githubusercontent.com/assets/952602/20037625/52f0fd50-a423-11e6-8aba-e995ca946202.png)

oddMLan:
Looks good but I still think the Frame buffer copy/depth sections should use comboboxes so less text is occupying the interface in the Frame buffer tab. That would also allow us to get rid of the white space at the bottom of the window in other tabs providing a small, compact and functional UI, while making it look less cluttered with text.

![mockup](https://cloud.githubusercontent.com/assets/7278372/20040050/87376dbc-a40d-11e6-8b43-a62210709e58.png)

Seriously though this seems like the perfect window size.
![mockup0](https://cloud.githubusercontent.com/assets/7278372/20040079/3a7a3d96-a40e-11e6-8e2b-557c00122451.png)
![mockup1](https://cloud.githubusercontent.com/assets/7278372/20040107/8bed67f2-a40e-11e6-9022-f2382a36e976.png)

Aquatakat:
@oddMLan @gonetz 

About the comboboxes vs. the radio buttons I like both layouts. The reason I changed it back to radio buttons was at gonetz' suggestion. Maybe we just need a little more input on it.

@LuigiBlood 

It looks like you're using Windows 7? I'll have to find out which fonts preinstalled on Windows 7 include those arrows, or maybe just make some arrow graphics myself.

***

I'm getting ready to prepare another pull request with this added feature:

![image](https://cloud.githubusercontent.com/assets/9537912/20041213/16a1170e-a423-11e6-8609-4a9d8be2757a.png)

LuigiBlood:
Yeah I'm a Windows 7 user.

Aquatakat:
I used some skills I have with Paint.net to create some new arrows. They're a bit bold but they do the job.

![image](https://cloud.githubusercontent.com/assets/9537912/20042456/8978cbaa-a437-11e6-8ee4-74123856038a.png)

[...]

PR #1220, opened 2016-11-08
A possible fix for disabled/enabled controls

Aquatakat:
This fix is a bit dumb but I have a feeling it might work. There's probably a better way of doing this.

### DIFF
--- a/src/GLideNUI/ConfigDialog.cpp
+++ b/src/GLideNUI/ConfigDialog.cpp
@@ -151,10 +151,10 @@ void ConfigDialog::_init()
 		break;
 	}
 
-	ui->fbInfoEnableCheckBox->setChecked(config.frameBufferEmulation.fbInfoDisabled != 0);
+	ui->fbInfoEnableCheckBox->toggle();
 	ui->fbInfoEnableCheckBox->setChecked(config.frameBufferEmulation.fbInfoDisabled == 0);
 
-	ui->frameBufferCheckBox->setChecked(config.frameBufferEmulation.enable == 0);
+	ui->frameBufferCheckBox->toggle();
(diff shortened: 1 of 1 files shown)


gonetz:
Are you trying to send "toggle"signal that way? I'm sure there should be another way.

Aquatakat:
I'm sure there's another way too, but I don't know what it is.

gonetz:
Why you don't want to put texture pack settings to checkable group box? If you don't like borders, it can be made with no border:
http://www.qtforum.org/article/27045/remove-border-qgroupbox.html

Aquatakat:
The borders add unnecessary visual complexity.

When the borders are removed, it doesn't appear that you can adjust the position of the group box's checkbox:

image

Aquatakat:
Given the way UIs are designed I'm sure there's some kind of method like `Qt::processAllSignals` or something like that. It's weird that I haven't been able to find it yet.

gonetz:
Ok, another way is to call toggle before setChecked:
```
    ui->texturePackOnCheckBox->toggle();
    ui->texturePackOnCheckBox->setChecked(config.textureFilter.txHiresEnable != 0);

```
It is slightly better than two setChecked, and it works.

Aquatakat:
I've been looking and couldn't find a clean solution. I'll submit new code using `->toggle()`

gonetz:
Ok.

Aquatakat:
It really is weird that whatever object the `ui` variable is doesn't seem to have at Qt method of processing all of the signals before displaying the output.

gonetz:
Ok, it works. Thanks!
Now need to polish options names and tooltips.

Also, let's try to implement oddMLan suggestion:
https://github.com/gonetz/GLideN64/issues/1213#issuecomment-259053325
I'll put it to alternative branch, and give users to compare.

PR #1223, opened 2016-11-08
Fix debug build on Linux

loganmc10:
__debugbreak() is Windows specific


gonetz:
Ok. It is third-party code anyway.

PR #1226, opened 2016-11-10
Fix error in GLES2 implementation of LOD.

fzurita:
Looks like I missed this. I'm still seeing errors in Paper Mario, even with this fix, even though the peach painting in Mario 64 works ok now.

I must be missing something else somewhere.

Edit: I found the last problem, now Paper Mario works correctly.


src/GLES2/Shaders_gles2.h
```diff
@@ -317,7 +317,7 @@ static const char* fragment_shader_mipmap =
 "  if (magnify && (uTextureDetail == 1 || uTextureDetail == 3))			\n"
 "      lod_frac = 1.0 - lod_frac;						\n"
 "  if (uMaxTile == 0) {									\n"
-"    if (uEnableLod != 0 && (uTextureDetail == 2 || uTextureDetail == 3))	\n"
+"    if (uEnableLod != 0 && (uTextureDetail != 2 && uTextureDetail != 3))	\n"
```
gonetz: if (uEnableLod != 0 && uTextureDetail < 2)

fzurita: Heh, yeah, that's simpler.

fzurita:
Ok, updated.

PR #1227, opened 2016-11-11
N64 resolution factor moved to emulation tab

Aquatakat:
Some C++ glue is necessary, but I don't think too much.
I actually tried to do these within configDialog.ui but the signals sent there don't seem to be flexible enough.

Some psuedocode:

`factor0xRadioButton`: `resolutionFactorSlider->setNum(0)`
`factor1xRadioButton`: `resolutionFactorSlider->setNum(1)`

`resolutionFactorSlider` is set to 0: `factor0xRadioButton->setChecked(true)`
`resolutionFactorSlider` is set to 1: `factor1xRadioButton->setChecked(true)`

Regarding the language "Render 2D elements in N64 resolution," I couldn't find good language for this, so I just added "best quality" in the parentheses. My original thought was to move that checkbox in the "Fix black lines between 2D elements" radio button and redo all the language there, but I failed to figure out something that worked well.

### DIFF
--- a/src/GLideNUI/configDialog.ui
+++ b/src/GLideNUI/configDialog.ui
@@ -1085,6 +1085,9 @@
                  </property>
                  <item>
                   <widget class="QSlider" name="resolutionFactorSlider">
+                   <property name="minimum">
+                    <number>2</number>
+                   </property>
                    <property name="maximum">
                     <number>16</number>
                    </property>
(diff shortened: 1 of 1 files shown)


Aquatakat:
For some reason I really enjoy how Qt Creator decides for every commit that the buttonGroups need to be rearranged. :)

gonetz:
Issue:
If nativeResFactor is 0 or 1, the slider is disabled. It is right, but the label under it shows "0x". It still show "0x" when third option enabled until you change slider.

Aquatakat:
This is why I suggested these changes to the C++ form:

image

gonetz:
The label is tied with slider's value. Slider's minimal value is 2. Why label shows 0x?
It is ok to show 2x when slider is disabled.

Aquatakat:
Oh, I see what you're saying. I'm not sure I can change this within the XML file though.

Aquatakat:
This might be a good time to hide the slider control actually.

gonetz:
If native res factor is >= 2 and you open config panel, label shows correct value of the slider. I tried to force set slider to 2 when factor is < 2, but label shows 0x anyway.

It either should be fixed (I can set label's text in cpp code) or , as you said, slider can be hidden.

Aquatakat:
It probably needs more explicit modification in C++. Hiding it is a simpler and more elegant solution. It's nice for it not to appear when not needed. I'll prepare another pull request for you.

gonetz:
Ok, let's hide it, but I guess you need to send a signal to the label to get proper initial value.

PR #1228, opened 2016-11-11
Factor of N64 resolution slider is hidden rather than disabled

Aquatakat:


gonetz:
The slider becomes hidden only when you set option 3 and then one or two. And, as I said, initial value for slider's label is 0x.
Which signals should be send to make slider hidden and label set to slider minimum value?

Aquatakat:
I guess now that the slider's minimum is 2 you can't always grab the config from the slider.

Honestly I'm not sure how to deal with signals with a radio button.

Sending a `->toggle()` to the Xx slider before setting it to the right value might work.

The label's probably going to need an old fashioned switch statement I think.

I tried doing a `setNum(0)` and a `setNum(1)` in the XML, which didn't work. It might work in C++. If not, I don't know.

gonetz:
> I guess now that the slider's minimum is 2 you can't always grab the config from the slider.

Yes, 0 and 1 set from radio buttons.

gonetz:
> Honestly I'm not sure how to deal with signals with a radio button.

I made slider invisible when first two options selected. How?

Slider is not hidden when you open the dialog. I can hide it in code, but there are also 3 labels connected to it. Manual hiding looks ugly. I need to know, how you hide them.

Aquatakat:
They're all tied to `factorXxRadioButton`.

An ugly solution might be to reverse all the values of `factorXxRadioButton`, `factor0xRadioButton` and `factor1xRadioButton` then reverse them to the correct values.

gonetz:
Ok, I fixed that.

PR #1229, opened 2016-11-11
Aspect: "Adjust:" -> "Try to adjust"; "Per-pixel" -> "Phong shading"; fixing OSD autoselect

Aquatakat:
"Phong shading" actually isn't a bad label for the checkbox, but "per-pixel lighting" might be better for novice users.

### DIFF
--- a/src/GLideNUI/configDialog.ui
+++ b/src/GLideNUI/configDialog.ui
@@ -368,7 +368,7 @@
              <item>
               <widget class="QRadioButton" name="aspectAdjustRadioButton">
                <property name="text">
-                <string>Adjust game to fit</string>
+                <string>Try to adjust game to fit</string>
                </property>
                <attribute name="buttonGroup">
                 <string notr="true">aspectButtonGroup</string>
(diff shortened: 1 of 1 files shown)


gonetz:
Merged.

There is one issue left:
Read by chunks checkboxes disabled either when fb emulation disabled or when FBinfo disabled.
It works correct on start. Suppose, I disable FBInfo. Read chunks both disabled. Then I disable Fb emulation.  Read chunks still disabled. I enable fb emulation again and Read chunks are enabled too, while FBInfo is still off. Something is wrong with signal processing logic. Can you fix it? If not, I'll do it with code.

Aquatakat:
I see the issue, but it would require chaining signals in the XML in a way that doesn't seem possible.

It seems easy to fix in C++ by re-toggling `fbInfoEnableCheckBox` after `frameBufferCheckBox` has been enabled.

Aquatakat:
A similar issue happens with the initial resolution slider and its related labels that might be fixed at the same time.

PR #1233, opened 2016-11-14
Custom windowed mode and custom crop W/H labels are also disabled

Aquatakat:
This is just for consistency with the other controls.

Also there's a small bug: the crop image spinboxes are not disabled on init when they should be.


Aquatakat:
The same fix for [these issues](https://github.com/gonetz/GLideN64/pull/1229#issuecomment-260119343) seems to be necessary.

gonetz:
Merged: d69a78e9b26c

Issue: When Crop set to 'disable' or 'auto' related controls are grayed, but 'W:' label does not.

gonetz:
@Aquatakat please fix the issue with label "W:" , and I will add necessary code to update state of Crop controls.

PR #1234, opened 2016-11-14
Fix some differences between GLES2 and GLES3 code.

fzurita:
I found a few other differences between GLSLCombiner_gles2.cpp/GLSLCombiner_ogl3x.cpp and Shaders_gles2.h/Shaders_ogl3x.h. Also, I did a search for #ifndef GLES2 and found one thing that didn't seem right in FrameBuffer.cpp so I fixed that too.

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -788,8 +788,8 @@ void FrameBuffer_Destroy()
 	RDRAMtoColorBuffer::get().destroy();
 #ifndef GLES2
 	DepthBufferToRDRAM::get().destroy();
-	ColorBufferToRDRAM::get().destroy();
 #endif
+	ColorBufferToRDRAM::get().destroy();
 	frameBufferList().destroy();
 }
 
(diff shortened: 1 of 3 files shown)


loganmc10:
Sometimes they are different for a reason though. I don't think uDepthScale is used in GLES2 at all (it is used with fragment depth writing), so updating it in GLES2 is pointless

fzurita:
Yep, I understand that. That one slipped by me.

fzurita:
Ok, I fixed the issue you saw.

src/FrameBuffer.cpp
```diff
@@ -788,7 +788,6 @@ void FrameBuffer_Destroy()
 	RDRAMtoColorBuffer::get().destroy();
 #ifndef GLES2
 	DepthBufferToRDRAM::get().destroy();
-	ColorBufferToRDRAM::get().destroy();
```
fzurita: ok, fixed.

gonetz:
Merged: 16262e52ad4

PR #1237, opened 2016-11-18
Fixed crop image W: label not disabling

Aquatakat:
I tried my hand at fixing ConfigDialog.cpp for the initial display but there seems to be something wrong with lines 84 and 86 that I couldn't figure out.

### DIFF
--- a/src/GLideNUI/ConfigDialog.cpp
+++ b/src/GLideNUI/ConfigDialog.cpp
@@ -588,7 +588,7 @@ void ConfigDialog::on_nativeRes2D_checkBox_toggled(bool checked)
 void ConfigDialog::on_cropImageComboBox_currentIndexChanged(int index)
 {
 	const bool bCustom = index == Config::cmCustom;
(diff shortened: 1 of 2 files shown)


gonetz:
Issues:

* Crop image spinbox enable/disable works incorrect after fn emulation toggled. I'll try to fix it in code.
* OSD preview label not changed after new color/font selected. Font label is changed instead, as before. Probably, code support is needed.

Aquatakat:
The behaviour for the OSD preview label could possibly be how I intended (I can't compile so I haven't tested):

Font: [Font - Size] [Select]
Color: #hex [Pick]

[Preview]

With the preview displaying "Preview" but showing the correct background and foreground colour.

Aquatakat:
First issue I don't know how to fix - mostly because the init value is currently incorrect.

The second issue I'll commit in a moment.

gonetz:
I fixed both issues

Aquatakat:
Good.

PR #1238, opened 2016-11-18
Tooltip fixes

Aquatakat:


gonetz:
I dislike these changes.

My variant:

"This option set output render buffer size to a factor of the original N64 resolution. The buffer has size of output resolution by default, but you can output to the Original N64 resolution or a Multiple of N64 resolution."

"N64 calculates lighting per vertex. This option enables Phong shading, which provides smoother and more realistic lighting."

Aquatakat:
One thing to note:
Factor typically refers to a number divided by some number (1/2, 1/3, 1/4, 1/6), which makes little sense in the context of the UI

gonetz:
"This option set output render buffer size. The buffer has size of output resolution by default, but you can output to the Original N64 resolution or a Multiple of N64 resolution."

Aquatakat:
"This option sets the output render buffer size. By default, the render buffer is set to the Same **as output resolution**, but you can set it to the **Original N64 resolution** or a **Multiple of N64 resolution**."

??

Bold options imply selectable options in the dropdown.

gonetz:
Good

gonetz:
Please update the pull request.

Aquatakat:
Commit #4fc73ec looks OK to me. Am I missing something?

gonetz:
Yes, it is ok now, thanks!

gonetz:
General note: currently tooltip for group of elements attached only to the first element of the group.
For example, you may get tooltip for Aspect ratio only when you place cursor over "Aspect ratio:" label. IMO, it will be better to show that tooltip for any element of the group. Old GUI works that way.

PR #1244, opened 2016-11-20
Check for GL_ARB_shader_image_load_store

loganmc10:
Image Textures might be available on devices with OpenGL < 4.3 if they support GL_ARB_shader_image_load_store


gonetz:
merged: c5c7845294fc5

PR #1247, opened 2016-11-21
MAX_PATH broke Ubuntu 16.04 build

ericcurtin:
It's PATH_MAX on Ubuntu. I don't know all the platforms this is
supported on. Can build on Ubuntu 16.04 later if needs be. May have to
include linux/limits.h on Linux.

Logged issue here initially:

https://github.com/RetroPie/RetroPie-Setup/issues/1756

### DIFF
--- a/src/Log.cpp
+++ b/src/Log.cpp
@@ -5,6 +5,7 @@
 #include "Log.h"
 #include "PluginAPI.h"
 #include "wst.h"
+#include "Ext_TxFilter.h"
 
 void LOG(u16 type, const char * format, ...) {
 	if (type > LOG_LEVEL)


ericcurtin:
Do you have a number of build machines to test this on? I don't know all the supported platforms here, all I know is Ubuntu 16.04 is broken, which it supported by Retropie

loganmc10:
It's broken on Fedora as well

loganmc10:
PATH_MAX is what's used in Combiner.cpp as well.

However I need to:
```
#include <osal_files.h>
```
for PATH_MAX to be defined

ericcurtin:
Aha, I see what they did, I'm brand new to this project. Only tried an install yesterday. They don't use PATH_MAX which is the native Linux #define instead they have created their own define MAX_PATH in https://github.com/jquesnelle/mupen64plus-ui-console/blob/master/src/osal_preproc.h. This is what needs to be included. Can fix later after work if needs be. Just a question. How do people insure this stuff builds on all platforms? Pot luck, an educated decision? Wait for someone to find out it's broken like @loganmc10 and I? Or are there build systems set up somewhere to test all this stuff?

ericcurtin:
Might be able to set up travis ci, to test branches before they are merged, to ensure this doesn't happen in future. Haven't used it myself, but I see other open source projects use it to great effect! They provide free infrastructure for open source projects I believe https://travis-ci.org

loganmc10:
That works well for pull requests, but not for things committed directly, plus I don't think Travis supports Visual Studio (but there are other projects that do).

This plugin is quite popular on Windows and Linux, so things like this are generally caught pretty quick

ericcurtin:
I see you have a Travis build set up here, so the idea isn't new to you guys!

https://travis-ci.org/GLupeN64/GLupeN64

loganmc10:
Yes I use Travis to test GLupeN64 on Linux, Windows, and Android. It also serves as my buildbot, if the build is successful (and the branch is master), then it uploads the binary to Amazon S3 for people to download

loganmc10:
Setting up Travis can never hurt, and it would help catch stuff like this automatically even if it was committed directly to master. Committing everything through pull requests makes sense for large projects with multiple regular contributors (I think Dolphin Emulator works this way), the vast majority of the code for this project comes from gonetz, it would add more work for him with little benefit (just catching the odd bad commit before it hits master every now and then, they are rare)

ericcurtin:
This fixes the Retropie script, dunno if it fixes your fedora build @loganmc10:

`#include <osal_files.h>`

doesn't fix the Retropie script.

loganmc10:
Changing it to PATH_MAX and including osal_files.h would be the correct way to fix it I would think

ericcurtin:
MAX_PATH seems to be more commonly used in the repo, will we see what @gonetz decides so we can get this merged soon. Or do you have the power to do that also? @loganmc10

loganmc10:
Nope I am powerless

gonetz:
9491633b

gonetz:
Sorry for inconvenience.

PR #1249, opened 2016-11-21
Remove some GLES3_1 ifdefs

loganmc10:
Anytime that GL_IMAGE_TEXTURES_SUPPORT is defined.

This is just a small step towards getting rid of the need for the GLES3_1 define. Right now GL_IMAGE_TEXTURES_SUPPORT is only defined for GLES3_1 (not GLES2 or GLES3), so was redundant anyway.

With this change, you should be able to define GL_IMAGE_TEXTURES_SUPPORT for GLES3. The only problem is that you'll probably run into some compile time issues because of missing definitions for glBindImageTexure (and glMemoryBarrier).

That could get sorted out with eglGetProcAddress for those GLES calls. GLupeN64 handles all that in the background, so this change is all that is needed to get Image Texture support working in our GLES3 plugin. However I'm sure something would need to be done for @fzurita, but I'm not sure exactly how that would be handled.

### DIFF
--- a/src/DepthBuffer.cpp
+++ b/src/DepthBuffer.cpp
@@ -145,7 +145,7 @@ void DepthBuffer::_initDepthBufferTexture(FrameBuffer * _pBuffer, CachedTexture 
 #ifdef GL_MULTISAMPLING_SUPPORT
 	if (_multisample) {
 		glBindTexture(GL_TEXTURE_2D_MULTISAMPLE, _pTexture->glName);
-#if defined(GLES3_1)
+#if defined(GLESX)
 		glTexStorage2DMultisample(GL_TEXTURE_2D_MULTISAMPLE, config.video.multisampling, fboFormats.depthInternalFormat, _pTexture->realWidth, _pTexture->realHeight, false);
 #else
 		glTexImage2DMultisample(GL_TEXTURE_2D_MULTISAMPLE, config.video.multisampling, fboFormats.depthInternalFormat, _pTexture->realWidth, _pTexture->realHeight, false);
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -141,7 +141,7 @@ void FrameBuffer::init(u32 _address, u32 _endAddress, u16 _format, u16 _size, u1
 #ifdef GL_MULTISAMPLING_SUPPORT
 	if (config.video.multisampling != 0) {
 		glBindTexture(GL_TEXTURE_2D_MULTISAMPLE, m_pTexture->glName);
-#if defined(GLES3_1)
+#if defined(GLESX)
 		if (_size > G_IM_SIZ_8b)
 			glTexStorage2DMultisample(GL_TEXTURE_2D_MULTISAMPLE, config.video.multisampling, GL_RGBA8, m_pTexture->realWidth, m_pTexture->realHeight, false);
 		else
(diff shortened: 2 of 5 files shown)


loganmc10:
I suppose there is probably some code in the GLSL that requires ```#version 310 es``` so I'm not 100% how to deal with that yet...

loganmc10:
I guess it would be as simple as checking the GLES version and setting the GLSL header based on that instead of a hardcoded #ifdef

fzurita:
It would not be hard at all to merge GLES 3.0 and 3.1 variants. We can just use the gl31 header file and user eglGetProcAddress to get all the functions. The gl31 header file can also be used for GLES 2.0, but this would be harder to implement due to the number of #ifdefs that need to be replaced.

loganmc10:
Ya GLES2 merger would be pretty ambitious, GLES3_1 just needs a few ifdefs replaced with GLES and the GLSL headers needs to be determined by the GLES version instead of being hard coded

fzurita:
Also, we can't use multisampling with GLSL version 300, so if the version of GLES is older than 3.1, multisampling always needs to be turned off.

loganmc10:
GLideN64 checks for GL_MAX_SAMPLES, I can't remember where (on my phone), so that should already be covered

loganmc10:
Although maybe checking the value of GL_MAX_SAMPLES on a device that doesn't support multi sampling is a GL error, I'm not sure

loganmc10:
I added a few more instances where we can use GLESX instead of GLES3_1 since this stuff is already protected inside GL_IMAGE_TEXTURES_SUPPORT or GL_MULTISAMPLING_SUPPORT

gonetz:
I don't quite understand, what the purpose of these changes? Which problem this commit solves?
Define GL_IMAGE_TEXTURES_SUPPORT for GLES3?

gonetz:
> there will be no need for a separate GLES3.1 version of the plugin

I still don't see any benefits in removal of GLES3.1 version. As I understand, clumsy solution via defines will be replaced by clumsy solution via GL version check, which in turn will be separated with defines for GLES and OpenGL.

loganmc10:
GLES and OpenGL will still be separated by defines, but I wouldn't call a GL version check a "clumsy solution", that is the proper way to do it. Image Textures work this exact way with OpenGL, there is no ```#ifdef OPENGL_4_3```, it checks if the device has the right GL version.

I guess @fzurita could comment on if he sees any benefit, but I think it is a little clumsy to distribute a separate plugin for GLES3 and GLES3.1

fzurita:
There is another benefit. It removes the needs to distribute multiple binaries with the Mupen64Plus AE front end. This is good for reducing total APK size. Currently we distrubte 8 GLideN64 binaries in Android:

- armv7-GLES2
- armv7-GLES3
- armv7-GLES3.1
- armv7-FullOpenGL
- x86-GLES2
- x86-GLES3
- x86-GLES3.1
- x86-FullOpenGL

This will become 16 once we add 64 bit support. Anything to reduce the number of binaries would be appreciated.

gonetz:
> there is no #ifdef OPENGL_4_3, it checks if the device has the right GL version.

right GL version for OpenGL is 4.3, but for GLES 3.0 (or 3.1?). Thus GL version check will be different for OpenGL and GLES.

gonetz:
> Anything to reduce the number of binaries would be appreciated.

Now I see the benefit, thanks.

gonetz:
@fzurita could you check that this commit does not break Android build?

fzurita:
Android still builds fine with this, it also still runs fine.

gonetz:
Ok, thanks

gonetz:
merged: ad77e3db52e

PR #1251, opened 2016-11-22
Minor spacing changes; many Layouts changed to QFrames

Aquatakat:
All of the signals have been retooled so there's a bit of testing needed to make sure things behave right once compiled.


gonetz:
2D fix controls now work incorrect: they are disabled until I toggle Internal resolution buttons.
nativeRes2D_checkBox is not disabled in code, so you broke it somewhere.

PR #1252, opened 2016-11-22
Use Buffer storage extension for faster glReadPixels call

fzurita:
Use Buffer storage extension for faster glReadPixels call.

Also, I made several small changes along the way in different commits.

I have tested two devices so far and it fixes https://github.com/gonetz/GLideN64/issues/1091

I still need to test more devices to make sure that there are no more issues.

I was hoping you could make sure that it works OK on desktop as well.


loganmc10:
I may be missing something, but unless a PBO is bound (GL_PIXEL_UNPACK_BUFFER) a call to glTexSubImage2D with a null pointer as the data is essentially a "nothing" call. glTexStorage2D initilizes the storage, so I don't think the glTexSubImage2D call will do anything.

fzurita:
Yes, I agree with you there. I may have gone a little zealous in copying and pasting. I'll fix those.

fzurita:
Ok, fixed.

loganmc10:
In haven't tried to compile this yet but it looks like ColorBufferToRDRAM_BufferStorageExt.cpp will need to b added to the cmake and visual studio source lists?

Also, does this account for the "stub" version? The "GLES" version is a little deceiving since it only supports Android and not the Raspberry Pi

fzurita:
Yes, it needs to be added to cmake and visual studio, I'm not familiar with those, so I was hoping that gonetz could add them when it's ready.

The GLES version should really be renamed to Android and the GL version should be renamed to PBO in my opinion. This doesn't take into account the Stub version currently, but I can add it.

gonetz:
I fixed compilation on Windows. Unfortunately, you stepped on the same rakes with glTexStorage2D, as I did when tried to use glTexStorage2DMultisample : https://github.com/gonetz/GLideN64/commit/1fed809ac0c3a2b01db3911b89d03add5583345b
driver reports that necessary extension is supported, but the checkFBO fails. That is such modification is incompatible with old GL 3.3 cards on Windows.

fzurita:
Ok. I'll revert the change around glTexStorage2DMultisample in that case.

gonetz:
It will not help. glTexStorage2D must be reverted as well. FBO does not work when texture created with glTexStorage2D

fzurita:
Ok, I made check more refined. glTexStorage2D will only be used for GLES 3.0 and GLES 3.1.

gonetz:
It works. Now I have to check the code :)

src/FrameBuffer.cpp
```diff
@@ -87,10 +87,25 @@ void FrameBuffer::_initTexture(u16 _width, u16 _height, u16 _format, u16 _size,
...
+
+#ifndef GLES2
+		glTexStorage2D(GL_TEXTURE_2D, 1, fboFormats.colorInternalFormat, _pTexture->realWidth, _pTexture->realHeight);
+		glTexSubImage2D(GL_TEXTURE_2D, 0, 0, 0, _pTexture->realWidth, _pTexture->realHeight, fboFormats.colorFormat, fboFormats.colorType, nullptr);
+#else
+		glTexImage2D(GL_TEXTURE_2D, 0, fboFormats.colorInternalFormat, _pTexture->realWidth, _pTexture->realHeight, 0,
+					 fboFormats.colorFormat, fboFormats.colorType, nullptr);
+#endif
+
+	}else {
```
gonetz: } else {

src/FrameBuffer.cpp
```diff
@@ -336,10 +351,24 @@ bool FrameBuffer::_initSubTexture(u32 _t)
...
-		glTexImage2D(GL_TEXTURE_2D, 0, fboFormats.monochromeInternalFormat, m_pSubTexture->realWidth, m_pSubTexture->realHeight, 0, fboFormats.monochromeFormat, fboFormats.monochromeType, nullptr);
+	if (m_pSubTexture->size > G_IM_SIZ_8b) {
+#ifndef GLES2
+		glTexStorage2D(GL_TEXTURE_2D, 1, fboFormats.colorInternalFormat, m_pSubTexture->realWidth, m_pSubTexture->realHeight);
+		glTexSubImage2D(GL_TEXTURE_2D, 0, 0, 0, m_pSubTexture->realWidth, m_pSubTexture->realHeight, fboFormats.colorFormat, fboFormats.colorType, nullptr);
+#else
+		glTexImage2D(GL_TEXTURE_2D, 0, fboFormats.colorInternalFormat, m_pSubTexture->realWidth,
+					 m_pSubTexture->realHeight, 0, fboFormats.colorFormat, fboFormats.colorType, nullptr);
+#endif
+	}else {
```
gonetz: } else {

src/FrameBuffer.cpp
```diff
@@ -91,7 +91,6 @@ void FrameBuffer::_setAndAttachTexture(u16 _size, CachedTexture *_pTexture)
 
 #ifndef GLES2
 		glTexStorage2D(GL_TEXTURE_2D, 1, fboFormats.colorInternalFormat, _pTexture->realWidth, _pTexture->realHeight);
-		glTexSubImage2D(GL_TEXTURE_2D, 0, 0, 0, _pTexture->realWidth, _pTexture->realHeight, fboFormats.colorFormat, fboFormats.colorType, nullptr);
```
gonetz: All these changes goes into "Use glTexStorage for frame buffer textures." commit

src/BufferCopy/ColorBufferToRDRAM_BufferStorageExt.h
```diff
@@ -0,0 +1,30 @@
...
+	ColorBufferToRDRAM_BufferStorageExt();
+	~ColorBufferToRDRAM_BufferStorageExt() = default;
+
+private:
+	void _init() override;
+	void _destroy() override;
+	bool _readPixels(GLint _x0, GLint _y0, GLsizei _width, GLsizei _height, u32 _size, bool _sync)  override;
+	void _cleanUp()  override;
+	void _initBuffers(void) override;
+	static const int _numPBO = 10;
```
gonetz: Yes, 10 is overkill. Besides it will be bad to have lag in 9 frames. I think 2 should be used there and in ColorBufferToRDRAM_GL

src/BufferCopy/ColorBufferToRDRAM_BufferStorageExt.cpp
```diff
@@ -0,0 +1,99 @@
...
+	glBufferStorage = 0;
+	glBufferStorage = (PFNDOLBUFFERSTORAGEPROC)eglGetProcAddress("glBufferStorageEXT");
+#endif
+}
+
+void ColorBufferToRDRAM_BufferStorageExt::_init()
+{
+	// Generate Pixel Buffer Objects
+	glGenBuffers(_numPBO, m_PBO);
+	isGLError();
```
gonetz: Why?
may be assert(!isGLError()); ?

src/BufferCopy/ColorBufferToRDRAM_BufferStorageExt.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+void ColorBufferToRDRAM_BufferStorageExt::_destroy()
+{
+	for(int index = 0; index < _numPBO; ++index){
```
gonetz: for (int index = 0; index < _numPBO; ++index) {

src/BufferCopy/ColorBufferToRDRAM_BufferStorageExt.cpp
```diff
@@ -0,0 +1,99 @@
...
+		if(m_PBOData[index] != nullptr){
+			glBindBuffer(GL_PIXEL_PACK_BUFFER, m_PBO[index]);
+			glUnmapBuffer(GL_PIXEL_PACK_BUFFER);
+		}
+	}
+
+	glDeleteBuffers(_numPBO, m_PBO);
+
+	for(int index = 0; index < _numPBO; ++index){
```
gonetz: for (int index = 0; index < _numPBO; ++index) {

src/BufferCopy/ColorBufferToRDRAM_BufferStorageExt.cpp
```diff
@@ -0,0 +1,99 @@
...
+		m_PBO[index] = 0;
+	}
+}
+
+void ColorBufferToRDRAM_BufferStorageExt::_initBuffers(void)
+{
+	// Initialize Pixel Buffer Objects
+	for(int index = 0; index < _numPBO; ++index){
```
gonetz: for (int index = 0; index < _numPBO; ++index) {

src/BufferCopy/ColorBufferToRDRAM_BufferStorageExt.cpp
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	glBindBuffer(GL_PIXEL_PACK_BUFFER, m_PBO[m_curIndex]);
+	glReadPixels(_x0, _y0, m_pTexture->realWidth, _height, colorFormat, colorType, 0);
+
+	//Setup a fence sync object so that we know when glReadPixels completes
+	glMemoryBarrier(GL_CLIENT_MAPPED_BUFFER_BARRIER_BIT);
+	m_fence[m_curIndex] = glFenceSync(GL_SYNC_GPU_COMMANDS_COMPLETE, 0);
+
+	if(!_sync){
```
gonetz: if (!_sync) {

src/BufferCopy/ColorBufferToRDRAM_BufferStorageExt.cpp
```diff
@@ -0,0 +1,99 @@
...
+		m_curIndex = (m_curIndex+1)%_numPBO;
+	}
+
+	//Wait for glReadPixels to complete for the currently selected PBO
+	if(m_fence[m_curIndex] != 0){
```
gonetz: if (m_fence[m_curIndex] != 0) {

src/BufferCopy/ColorBufferToRDRAM_BufferStorageExt.cpp
```diff
@@ -0,0 +1,99 @@
...
+		glClientWaitSync(m_fence[m_curIndex], 0, 1e9);
```
gonetz: 1e9 what this? Wait forever?

src/BufferCopy/ColorBufferToRDRAM_BufferStorageExt.cpp
```diff
@@ -0,0 +1,99 @@
...
+		glDeleteSync(m_fence[m_curIndex]);
+	}
+
+	GLubyte* pixelData = reinterpret_cast<GLubyte*>(m_PBOData[m_curIndex]);
+	if (pixelData == nullptr)
+		return false;
+
+	int widthBytes = _width*colorFormatBytes;
```
gonetz: _width * colorFormatBytes;

src/BufferCopy/ColorBufferToRDRAM.cpp
```diff
@@ -260,6 +271,31 @@ void ColorBufferToRDRAM::copyChunkToRDRAM(u32 _address)
...
 
+
+ColorBufferToRDRAM & ColorBufferToRDRAM::get()
+{
+#ifndef GLES2
+
+	static bool supportsBufferStorage = OGLVideo::isExtensionSupported("GL_EXT_buffer_storage") ||
+		OGLVideo::isExtensionSupported("GL_ARB_buffer_storage");
+
+	if(supportsBufferStorage){
```
gonetz: if (supportsBufferStorage) {

src/BufferCopy/ColorBufferToRDRAM.cpp
```diff
@@ -260,6 +271,31 @@ void ColorBufferToRDRAM::copyChunkToRDRAM(u32 _address)
...
+		static ColorBufferToRDRAM_BufferStorageExt cbCopy;
+		return cbCopy;
+	}else{
```
gonetz: } else {

gonetz:
The code looks OK, but there are plenty of formatting issues and some commits need to be re-arranged.

src/BufferCopy/ColorBufferToRDRAM_BufferStorageExt.cpp
fzurita: That was part of some early debug. I'll just remove it.

src/BufferCopy/ColorBufferToRDRAM_BufferStorageExt.cpp
fzurita: I may be missing something, but this looks the same as what you typed better?

fzurita:
Ok, I'll make adjustments. Nothing you pointed out looks hard to adjust.

fzurita:
Ok, done.

gonetz:
Thanks! I'll check it asap.

fzurita:
@gonetz Do you see any performance difference with this?

fzurita:
oh, wait... I'm getting a crash in GLES 2.0 mode....

fzurita:
Updated again.

gonetz:
> Do you see any performance difference with this?

I can't answer such questions without precise measure in code. It works, that what I see for sure.

src/inc/ARB_buffer_storage.h
```diff
@@ -0,0 +1,48 @@
...
+#include <GL/glcorearb.h>
+#include "common/GLFunctions.h"
+#include <GL/glext.h>
+#endif
+
+
+
+typedef void (APIENTRYP PFNDOLBUFFERSTORAGEPROC)(GLenum target, GLsizeiptr size, const void* data,
+                                                GLbitfield flags);
+typedef void (APIENTRYP PFNDOLNAMEDBUFFERSTORAGEEXTPROC)(GLuint buffer, GLsizeiptr size,
```
gonetz: Never used

src/BufferCopy/ColorBufferToRDRAM_BufferStorageExt.h
```diff
@@ -0,0 +1,30 @@
...
+	bool _readPixels(GLint _x0, GLint _y0, GLsizei _width, GLsizei _height, u32 _size, bool _sync)  override;
+	void _cleanUp()  override;
+	void _initBuffers(void) override;
+	static const int _numPBO = 2;
+	GLuint m_PBO[_numPBO];
+	void* m_PBOData[_numPBO];
+	u32 m_curIndex;
+	GLsync m_fence[_numPBO];
+#ifndef EGL
+	PFNDOLBUFFERSTORAGEPROC glBufferStorage;
```
gonetz: It is necessary for GLESX only, right?
Why #ifndef EGL ?

gonetz:
I pushed it to fzurita-buffer_storage_ext branch along with my fixes for desktop version.
My old GeForce does not support "GL_EXT_buffer_storage" and "GL_ARB_buffer_storage" extensions, so I did not check yet, how it works on desktops.

gonetz:
My fix does not work on Linux :(

gonetz:
Fixed Linux compilation too.

gonetz:
I tested ColorBufferToRDRAM_BufferStorageExt on Linux - it works fine. I did not measured performance though.

src/inc/ARB_buffer_storage.h
```diff
@@ -0,0 +1,48 @@
...
+#define APIENTRYP GL_APIENTRY*
+#else
+#include <GL/glcorearb.h>
+#include "common/GLFunctions.h"
+#include <GL/glext.h>
+#endif
+
+
+
+typedef void (APIENTRYP PFNDOLBUFFERSTORAGEPROC)(GLenum target, GLsizeiptr size, const void* data,
```
fzurita: We don't use glext.h for GLES?

src/BufferCopy/ColorBufferToRDRAM_BufferStorageExt.h
fzurita: Yeah, good point.

gonetz:
@fzurita @loganmc10 please build from 012deec5dee2c and check that compilation and functionality not broken.

fzurita:
Checking

fzurita:
ok, a couple of issues:
https://github.com/gonetz/GLideN64/commit/012deec5dee2c0c61061135bdf7a75c9c0b233a1#diff-d9cae051f1525bb0d40029b7d7efd39aR5

That should be || instead of &&

Typo in here:
https://github.com/gonetz/GLideN64/blob/6832a14add86d96e225d482a250d5617dcbbbc45/src/inc/ARB_buffer_storage.h#L38
and here
https://github.com/gonetz/GLideN64/blob/6832a14add86d96e225d482a250d5617dcbbbc45/src/BufferCopy/ColorBufferToRDRAM_BufferStorageExt.cpp#L17

it should be: PFNGLBUFFERSTORAGEPROC

gonetz:
Ok, new fix uploaded.

fzurita:
Ok, it builds now.

gonetz:
Does it work properly, as before?

fzurita:
I will check

fzurita:
It still works fine on my GLES 3.1 device. I will have to check the shield TV in a sec for the full GL version.

fzurita:
Shield TV still works as before.

gonetz:
Great! I'll merge it to master then.

gonetz:
I have one question: ColorBufferToRDRAMStub.cpp what it is for now?

gonetz:
I removed ColorBufferToRDRAMStub.cpp
I need confirmation from @loganmc10  that build not broken on his platforms.

loganmc10:
ColorBufferToRDRAMStub.cpp was for GLES2 devices that don't run Android since ColorBufferToRDRAM_GLES.cpp has Android specific code. That includes the Raspberry Pi and probably other boards like the odroid. I'm not sure if it's still valid after all these changes, I'll try compiling on a Raspberry Pi tomorrow

gonetz:
ColorBufferToRDRAMStub.cpp code moved to  ColorBufferToRDRAMStub.h, so it should work as before. I'll wait your confirmation.

loganmc10:
I just tested compiling on the Raspberry Pi and it works fine, I don't have mupen64plus setup on there right now and I've got to head to bed so I can't test running it right now, but I assume since it compiles we are probably ok

fzurita:
A version could be made that works in Raspberry pi. It wouldn't wouldn't work with async mode though.

@loganmc10 you should investigate if it's possible to do the same thing that was done in Android for GLES 2.0. See this extension:
https://www.khronos.org/registry/gles/extensions/OES/OES_EGL_image_external.txt

The only problem with this extension is the implementation is platform specific, that's why there is platform specific code.

gonetz:
Ok, merged. Thanks!

loganmc10:
@fzurita Yes I looked into it a while ago, Broadcom has a function called eglCreateGlobalImageBRCM and some other definitions that I believe are for this purpose, but there is absolutely no documentation on how it works. Plus having code in there for 1 specific device would be a little odd, at least the Android one works across multiple devices.

Anyway, I'm sure that you're aware but even the Android solution is a little "unofficial", given that the code to make it work isn't part of the SDK or NDK.

loganmc10:
@fzurita is this line required?
https://github.com/gonetz/GLideN64/blob/master/src/inc/ARB_buffer_storage.h#L24

GLupeN64 (Android GLES3) compiles fine without it. When building with the Android standalone toolchain for GLES3 (make_standalone_toolchain.py --api 18 --arch arm), it doesn't have gl31.h

I don't think the GLES3 build should depend on gl31.h being present.

fzurita:
Just checked, it doesn't compile. Let me see if I can find a way to get rid of it.

fzurita:
Ok, buffer storage is actually only supported for GLES 3.1. It should probably be removed from the GLES 3.0 build of the plugin.

loganmc10:
It's true that it's only support for GLES3.1, but if the goal is to eventually merge the GLES3/3.1 versions then there would need to be another solution. GL_EXT_buffer_storage will only show as available on a device that supports it, so it's not like there will be a runtime issue, you just need a way to compile it.

fzurita:
You can manually pull the gl31.h header file into your build. gl3.h is a subset of gl31.h. I think it may be the same with gl2.h.

Actually, it may not be a bad idea to make gl31.h be part of the source code of GLideN64 in the fame way that the full GL version of that file is party of it.

loganmc10:
Fair enough, I'll just add gl31.h to my include path. I don't think there is a need to include it with GLideN64 but that's just me

gonetz:
> the full GL version of that file is part of it.

For Windows only.

PR #1254, opened 2016-11-22
Possible fix for 2D UI controls

Aquatakat:
I'm not sure why once compiled the 2D fix controls aren't working properly. It works properly in the preview. This is an attempt at a fix.

### DIFF
--- a/src/GLideNUI/ConfigDialog.cpp
+++ b/src/GLideNUI/ConfigDialog.cpp
@@ -143,10 +143,9 @@ void ConfigDialog::_init()
 	ui->frameBufferCheckBox->toggle();
 	const bool fbEmulationEnabled = config.frameBufferEmulation.enable != 0;
 	ui->frameBufferCheckBox->setChecked(fbEmulationEnabled);
-	ui->frameBufferInfoLabel->setVisible(!fbEmulationEnabled);
-	ui->frameBufferInfoIcon->setVisible(!fbEmulationEnabled);
-	ui->frameBufferInfoLabel2->setVisible(!fbEmulationEnabled);
-	ui->frameBufferInfoIcon2->setVisible(!fbEmulationEnabled);
+    ui->frameBufferInfoFrame->setVisible(!fbEmulationEnabled);
+    ui->frameBufferInfoFrame2->setVisible(!fbEmulationEnabled);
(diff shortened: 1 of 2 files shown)


Aquatakat:
From what I've read `setChecked()` doesn't fire signals but `toggled()` does, so hopefully this fixes the problem.

Because the frame buffer emulation checkbox now enables/disables QFrames, rather than controls directly, any `setChecked()` or `toggled()` where `fbEmulationEnabled &&` was added is probably redundant now.

gonetz:
No changes in behavior.

Aquatakat:
Could I have a build to test with? It might help to find exactly where the problem is occurring.

gonetz:
A build from your top commit?

Aquatakat:
It could include all the commits, the other ones are minor changes that shouldn't affect behaviour.

gonetz:
Here:
https://drive.google.com/file/d/0B0YqMPjGo3B2aml2SnNXMWtydlk/view?usp=sharing

gonetz:
Built from Revert "Another possible fix for 2D UI controls"

Aquatakat:
Thanks.

I've noticed that behaviour is correct as long as "Multiple of N64 resolution:" is selected. I'll try to figure out why.

Aquatakat:
It's hard to debug without being able to build the plugin but I wonder if this is the issue.

gonetz:
"Disabled 2D fixes controls on dialog open" problem is fixed.
One problem remains: Set fb emulation off - go to Emulation panel - 2D fixes disabled. It is wrong, these settings can work without FB emulation.
Enable fb emulation again, go to Emulation panel: 2D fixes disabled if Internal resolution was set any first two options. If Multiple of N64 resolution was selected, 2D will be enabled.

gonetz:
Build from  A third fix for 2D UI controls
https://drive.google.com/file/d/0B0YqMPjGo3B2RERfdG42M21zY1E/view?usp=sharing

Aquatakat:
I commented out the same line as "A third fix..." but in `ConfigDialog::on_frameBufferCheckBox_toggled`.

gonetz:
It work way better now. I found only one logical error:
If internal resolution is set to Original and then we disable FB emulation, 2D fixes remained disabled.

PR #1255, opened 2016-11-23
Get GLSL version dynamically

loganmc10:
Instead of hardcoding the ```#version XXX core``` header into the GLSL sources, this will use GL_MAJOR_VERSION and GL_MINOR_VERSION to automatically generate the header.

For instance, if the user has a GFX card that reports OpenGL 4.2, the header will become ```#version 420 core```

The purpose here is to get rid of the separate GLES3.1 version of the plugin. With this change, even when using the GLES3 version, the GLSL header will be ```#version 310 es``` if the user's phone supports it. This will allow GL_MULTISAMPLING_SUPPORT and GL_IMAGE_TEXTURE_SUPPORT to work in the GLES3 version of the plugin if the user's phone supports GLES3.1.

### DIFF
--- a/src/ShaderUtils.cpp
+++ b/src/ShaderUtils.cpp
@@ -1,5 +1,6 @@
 #include <assert.h>
 #include <stdio.h>
+#include <sstream>
 #include "ShaderUtils.h"
 #include "Config.h"
 #include "Log.h"
@@ -56,10 +57,46 @@ void logErrorShader(GLenum _shaderType, const std::string & _strShader)
 	}
 }
 
+const std::string &  getHeader()
+{
+	static std::string strHeader;
+	if (!strHeader.empty())
+		return strHeader;
+
+	OGLVideo & ogl = video();
+	OGLRender & render = ogl.getRender();
+	std::stringstream headerStream;
+#if defined(GLES2)
+	headerStream << "#version 100 " << std::endl;
+	if (config.generalEmulation.enableLOD > 0) {
+		headerStream << "#extension GL_EXT_shader_texture_lod : enable " << std::endl \
+			<< "#extension GL_OES_standard_derivatives : enable " << std::endl;
+	}
+#elif defined(GLESX)
+	headerStream << "#version " << render.majorVersion << render.minorVersion << "0 es " << std::endl;
+#else
+	headerStream << "#version " << render.majorVersion << render.minorVersion << "0 core " << std::endl;
+	if (video().getRender().isImageTexturesSupported() && ((render.majorVersion * 10) + render.minorVersion) < 42)
+		headerStream << "#extension GL_ARB_shader_image_load_store : enable " << std::endl \
+			<< "#extension GL_ARB_shading_language_420pack : enable " << std::endl;
+#endif
+	strHeader = headerStream.str();
+	return strHeader;
+}
+
+const char* addGLSLVersion(const char* shaderString)
+{
+	static std::string strResult;
+	strResult = getHeader();
+	strResult.append(shaderString);
+	return strResult.c_str();
+}
+
 GLuint createShaderProgram(const char * _strVertex, const char * _strFragment)
 {
 	GLuint vertex_shader_object = glCreateShader(GL_VERTEX_SHADER);
-	glShaderSource(vertex_shader_object, 1, &_strVertex, nullptr);
+	const char* vertex_shader_string = addGLSLVersion(_strVertex);
+	glShaderSource(vertex_shader_object, 1, &vertex_shader_string, nullptr);
 	glCompileShader(vertex_shader_object);
 	assert(checkShaderCompileStatus(vertex_shader_object));
 
@@ -67,7 +104,8 @@ GLuint createShaderProgram(const char * _strVertex, const char * _strFragment)
 		logErrorShader(GL_VERTEX_SHADER, _strVertex);
 
 	GLuint fragment_shader_object = glCreateShader(GL_FRAGMENT_SHADER);
-	glShaderSource(fragment_shader_object, 1, &_strFragment, nullptr);
+	const char* fragment_shader_string = addGLSLVersion(_strFragment);
+	glShaderSource(fragment_shader_object, 1, &fragment_shader_string, nullptr);
 	glCompileShader(fragment_shader_object);
 	assert(checkShaderCompileStatus(fragment_shader_object));
 
--- a/src/ShaderUtils.h
+++ b/src/ShaderUtils.h
@@ -4,6 +4,7 @@
 #include "OpenGL.h"
 #include "Combiner.h"
 
+const char* addGLSLVersion(const char* shaderString);
 GLuint createShaderProgram(const char * _strVertex, const char * _strFragment);
 bool checkShaderCompileStatus(GLuint obj);
 bool checkProgramLinkStatus(GLuint obj);
--- a/src/GLES2/GLSLCombiner_gles2.cpp
+++ b/src/GLES2/GLSLCombiner_gles2.cpp
@@ -100,7 +100,8 @@ static
 GLuint _createShader(GLenum _type, const char * _strShader)
 {
 	GLuint shader_object = glCreateShader(_type);
-	glShaderSource(shader_object, 1, &_strShader, nullptr);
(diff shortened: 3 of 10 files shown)


src/ShaderUtils.h
```diff
@@ -4,6 +4,8 @@
 #include "OpenGL.h"
 #include "Combiner.h"
 
+static std::string finalString;
```
gonetz: 1. finalString is a bad name. Very bad for global name.
2. You don't need to declare it there. Move declaration to addGLSLVersion

src/ShaderUtils.cpp
```diff
@@ -56,18 +56,42 @@ void logErrorShader(GLenum _shaderType, const std::string & _strShader)
...
+	finalString = "#version "+ GLversion + "0 es \n";
+#else
+	finalString = "#version "+ GLversion + "0 core \n";
+	if (video().getRender().isImageTexturesSupported() && stoi(GLversion) < 42)
+		finalString += "#extension GL_ARB_shader_image_load_store : enable \n";
+#endif
+	finalString.append(shaderString);
+	return finalString.c_str();
+}
+
```
gonetz: I don't like that the strings calculated each time addGLSLVersion called.

I suggest:

+const std::string &  getHeader()
 +{
       static std::string strHeader;
       if (!strHeader().empty())
          return strHeader;


 +	OGLVideo & ogl = video();
 +	OGLRender & render = ogl.getRender();
 +	std::string GLversion = std::to_string(render.majorVersion) + std::to_string(render.minorVersion);
 +#if defined(GLES2)
 +	strHeader= "#version 100 \n";
 +	if (config.generalEmulation.enableLOD > 0) {
 +		strHeader+= "#extension GL_EXT_shader_texture_lod : enable \n" \
 +			"#extension GL_OES_standard_derivatives : enable \n";
 +	}
 +#elif defined(GLESX)
 +	strHeader= "#version "+ GLversion + "0 es \n";
 +#else
 +	strHeader= "#version "+ GLversion + "0 core \n";
 +	if (video().getRender().isImageTexturesSupported() && stoi(GLversion) < 42)
 +		strHeader+= "#extension GL_ARB_shader_image_load_store : enable \n";
 +#endif
 +	return strHeader;
 +}

const char* addGLSLVersion(const char* shaderString)
 +{
       static std::string strResult;
        strResult = getHeader();
 +	strResult .append(shaderString);
 +	return strResult .c_str();
 +}

src/ShaderUtils.h
loganmc10: I agree it is a bad name. glShaderSource takes a pointer to the string as an argument. If the string is declared inside the function then it is destroyed when the function returns and the pointer becomes invalid. If there is some other way to do this I'm open to it, I'm not a C++ expert

gonetz:
>If the string is declared inside the function then it is destroyed when the function returns

Not if variable declared as static.

loganmc10:
Thanks for the tips, I've updated the pull request, I'll do a bit of testing on different platforms to make sure there are no issues

loganmc10:
I had to modify it from using std::to_string to using sprintf() since Android doesn't support std::to_string

loganmc10:
So I've tested this on Windows and Linux, including Image Texture support on a machine with GL < 4.2 but that supports the required extensions.

It all seems to work except I am getting a crash on Android, I forgot my USB cable today so I'm not able to debug it right now, I'll let you know when I've sorted it out.

fzurita:
spring is a C style call stringstream is the C++ version

loganmc10:
Ok I updated the code to use stringstream, I still get a crash on Android, it works on Linux using GLES (it correctly appends ```#version 300 es``` for instance when I use GLES3 mode.

On Android I get:
```
11-28 10:15:26.056: A/libc(22243): GLideN64/src/OGL3X/GLSLCombiner_ogl3x.cpp:220: GLuint _createShader(GLenum, const char*): assertion "checkShaderCompileStatus(shader_object)" failed
11-28 10:15:26.057: A/libc(22243): Fatal signal 6 (SIGABRT), code -6 in tid 22259 (Thread-29714)
```

I'm not sure how to view the shaders on Android to see what is wrong, @fzurita perhaps you could give this pull request a try?

fzurita:
Looks like the shader is failing to compile. I'll give it a shot tonight at some point if I can get a chance.

fzurita:
Shader compilation is not failing for me. Please replace the assert in the ```_createShader``` method with this:

```
	if (!checkShaderCompileStatus(shader_object))
		logErrorShader(_type, shader_string);
```

Then check the logs for the shader compilation error.

loganmc10:
It turns out the issue was because I use glsl-optimizer in GLupeN64, I guess it doesn't support GLSL 3.10.

Anyway I'll just have to get rid of that, it's unmaintained nowadays anyway, so I think this is okay to be merged.

loganmc10:
By the way, for some reason multisampling doesn't work on my device (Xperia XA). I tried mupen64plus-fz from the app store and GLupeN64 (with these changes), the same result, its verrry slow and all the textures are messed up. When I compile a debug build I get:

```
11-29 08:49:05.363: A/libc(30393): GLideN64/src/OGL3X/GLSLCombiner_ogl3x.cpp:509: ShaderCombiner::ShaderCombiner(Combiner&, Combiner&, const gDPCombine&): assertion "checkProgramLinkStatus(m_program)" failed
```

This is the same device that didn't properly detect GLES3.1 support in mupen64plus-fz using the old method of detection, maybe it has some kind of broken ES3.1 implementation.

Anyway that is unrelated to this pull request, I just found it strange, I haven't figured out what the cause of the linking error is yet

fzurita:
I just tried multi-sampling again just to make sure and it still seems to work with Mupen64plus-fz on my device.

It does look like the Xperia XA driver has issues compiling the Multisampling version of the shader.

If you enable developer options and enable "Backtrace on glErrors", you can check if there is something we are doing that it's complaining about. This option prints out a backtrace every time there is a glError in the logcat.

loganmc10:
I don't see any GL errors with that option. This is what Smash Bros looks like with multi-sampling on in mupen64plus-fz:
screenshot_20161129-120712

Shouldn't the linker error be printed to the logcat though? checkProgramLinkStatus should be printing the error message to LOG:

https://github.com/gonetz/GLideN64/blob/master/src/ShaderUtils.cpp#L31-L36

But I don't see anything in the logcat with a tag of "GLideN64" like it seems there should be:

https://github.com/gonetz/GLideN64/blob/master/src/Log.h#L20

loganmc10:
I got it to print out the shader error log by using this change: https://github.com/gonetz/GLideN64/pull/1273, but it just prints out the contents of the shader, it doesn't point to what is wrong.

I tested Image Textures using the GLES3.1 plugin in mupen64plus-fz (using the fog in Beetle Adventure Racing as a test), and it doesn't work either, so I think my device just has a broken implementation of ES3.1, there's probably a reason they weren't advertising it via reqGlEsVersion...

loganmc10:
Anyway that was all off topic, I've tested this commit with the Nvidia driver on desktop (supports GLES 3.2) and it all looks good.

By the way, with GLupeN64, this allows me to set:
```
#define GL_IMAGE_TEXTURES_SUPPORT
#define GL_MULTISAMPLING_SUPPORT
```
And it all works with the GLES3 version of the plugin, although the shader storage cache version should probably be incremented if this is merged (and again if GL_IMAGE_TEXTURES_SUPPORT or GL_MULTISAMPLING_SUPPORT is set for GLES3)

GLupeN64 handles all the eglGetProcAddress stuff in the background so that's why it works, mupen64plus-ae would just need a way to expose the GLES3.1 functions like glMemoryBarrier, glBindImageTexture, and glTexStorage2DMultisample and then there should be no need for the GLES3.1 version anymore.

fzurita:
Yeah, this is doable. I'll get to that next. I'll probably stop including the GLESv2 library and only include EGL and user eglGetProcAddress to get all the functions. This I think can be done in GLFunctions.h/cpp.

After that, we can get rid of the two defines that you mentioned and not allow them to be enabled if the GLES version is not high enough.

gonetz:
I pulled it to loganmc10-glslversion branch.
I'll not merge it with master until new public release.

PR #1256, opened 2016-11-24
2D fixes UI controls stay enabled when FB emulation disabled

Aquatakat:
The original issue,

> If internal resolution is set to Original and then we disable FB emulation, 2D fixes remained disabled.

exists on the plugin before `QFrame`s were introduced. I hoped the new signal logic would solve this problem without additional code, but it seems like a small bit is needed.


gonetz:
Yes some logic is still needed.
1. Set Internal to Original
2. Disable FB
3. Set any Fix 2D option
4. Enable FB.

Internal set to original, but 2Dfix enabled.

gonetz:
Seems to be fixed

PR #1260, opened 2016-11-26
Different approach to re-initializing buffers

fzurita:
Here is what I would do to re-initialize the mappings.

### DIFF
--- a/src/BufferCopy/ColorBufferToRDRAM.cpp
+++ b/src/BufferCopy/ColorBufferToRDRAM.cpp
@@ -98,6 +98,8 @@ void ColorBufferToRDRAM::_destroyFBTexure(void)
 	if (m_pTexture != nullptr) {
 		textureCache().removeFrameBufferTexture(m_pTexture);
 		m_pTexture = nullptr;
+
+		_destroyBuffers();
 	}
 }
 
--- a/src/BufferCopy/ColorBufferToRDRAM.h
+++ b/src/BufferCopy/ColorBufferToRDRAM.h
@@ -28,9 +28,10 @@ protected:
 	std::vector<GLubyte> m_pixelData;
 
 private:
-	virtual void _init() = 0;
-	virtual void _destroy() = 0;
+	virtual void _init() {};
+	virtual void _destroy() {};
 	virtual void _initBuffers(void) = 0;
+	virtual void _destroyBuffers(void){};
 	virtual bool _readPixels(GLint _x0, GLint _y0, GLsizei _width, GLsizei _height, u32 _size, bool _sync) = 0;
 	virtual void _cleanUp() = 0;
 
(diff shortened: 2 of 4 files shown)


gonetz:
The idea is right, implementation is wrong. It is not enough to call glUnmapBuffer for buffer, initialized with persistent storage. glBufferStorage will cause GL error anyway.
Buffer must be deleted, and new buffer created. That is _destroy, _init

fzurita:
Ok, I can move the PBO initialization and cleanup to ```_initBuffers``` and ```_destroyBuffers``` as well.

fzurita:
Ok, updated.

loganmc10:
I'm not sure if this pull request fixes this issue, but right now GLupeN64 users are getting a GL error because of these lines of code:

https://github.com/gonetz/GLideN64/blob/master/src/BufferCopy/ColorBufferToRDRAM_BufferStorageExt.cpp#L32-L33

By the time they execute, the buffers no longer exist, and I get these GL errors:

```
GL_INVALID_OPERATION in glBindBuffer(non-gen name)
GL_INVALID_OPERATION in glUnmapBuffer(no buffer bound)
GL_INVALID_OPERATION in glBindBuffer(non-gen name)
GL_INVALID_OPERATION in glUnmapBuffer(no buffer bound)
```

I'm not sure of the exact reason that is happening, but with persistent mapped buffers there is really no need to unmap the buffer, so those lines could probably be removed altogether. Once you delete the buffer the pointer just becomes invalid, unmapping the buffer doesn't really do anything, it was only really useful with non-persistent buffers where you have to unmap the buffer before using it.

fzurita:
Yes, this will most likely fix it. Also, Gonetz already did a fix that should work as well, but I think I like this version better.

loganmc10:
Well this issue I'm having is with the current master (except the latest "Neon version of InverseTransformVectorNormalize" commit)

fzurita:
The ```_destroy()``` method you pointed out will only get called on a shutdown. Is that when your users are seeing the issue?

loganmc10:
The issue is happening during a full screen toggle, which triggers OGLVideo::changeWindow(). Closing the application doesn't trigger any GL errors, just the full screen/windowed toggle.

I realise now that the issue is probably because RetroArch deletes and recreates the GL context before changeWindow has a chance to run, so when it runs the _destroy() method the buffer has already been deleted. I'm not sure if any other emulators like Project64 behave this way. I can work around it in my code, but the point still stands that those glBindBuffer/glUnmapBuffer calls don't really do anything since the buffer is deleted in the next line anyway (which will unmap it).

gonetz:
@fzurita I took your commit and rewrote some part of it: 6735db132

fzurita:
I see what you did, thanks.

gonetz:
Thanks to you :)

PR #1261, opened 2016-11-26
Neon version of InverseTransformVectorNormalize

fzurita:
I'm not sure if this is the best way to do this using neon, but it works. Verified by running neon and normal versions side by side and making sure the results were the same.


gizmo98:
I have rewritten your function. I use matrix multiplication of TransformVectorNormalize and replaced vld1 with vld4 to get array elements in the right order. 
https://gist.github.com/gizmo98/1255930cbc103ca78144e58ca4da315c

fzurita:
Cool, you should make a pull request. I'm by no means an expert in writing NEON code, I just knew enough to get by in this case, so I'm sure your version would be more efficient than mine.

PR #1269, opened 2016-11-28
Use Five2Eight and One2Eight from convert.h

loganmc10:
When this is all built as one project (instead of building GLideNHQ as a static library), some compilers complain about the multiple definitions of these variables.


gonetz:
merged: 5a66cfc897

Issue #975, opened 2016-04-27
Random texture corruption, possible instability.


AmbientMalice:
I can't reliably reproduce this problem. Sometimes it happens after settings are changed, sometimes it happens after booting a game for the first time. It could perhaps be a weird shader cache bug, but it occurs even with the cache disabled.

![gliden64_star_wars_ep1_racer_001](https://cloud.githubusercontent.com/assets/10704701/14838083/3b9b6848-0c61-11e6-9559-024bd885eac9.jpg)

![gliden64_turok_3 _shadow_of_o_000](https://cloud.githubusercontent.com/assets/10704701/14838095/47deb966-0c61-11e6-9073-eeb3acaf7e21.jpg)

![gliden64_super_mario_64_000](https://cloud.githubusercontent.com/assets/10704701/14838100/56b43290-0c61-11e6-8f4b-cb335be62164.jpg)

Also, the plugin seems to have gotten strangely unstable. When this bug occurs, changing settings or closing the emulator (PJ64 in this case) can cause the emulator to crash. I'm not 100% sure whether the emulator might be a factor.

olivieryuyu:
have the same with the new blender plugin

oddMLan:
Hmm.. that does ring a bell... could this be the same issue I faced in DK64?
https://u.pomf.is/zelokv.webm

At first I thought it happened because I forgot to delete the old shader cache but @gonetz said that he changed the shader cache version so the new blender version shouldn't have loaded anyways.

AmbientMalice:
@oddMLan Yes, that looks like the issue. It's very hard to reproduce, and seemingly impossible to reproduce 100% reliably. Common factor seems to be transparent textures, however. Text, bubbles in water, smoke, that sort of thing. Looks to me like the melon you pick up is a 2D HUD icon that rapidly pans.

I initially thought it might be related to https://github.com/gonetz/GLideN64/issues/882 where copying Aux buffers to RDRAM gives you garbage, but I think this occurs regardless of whether aux buffers are enabled.

gonetz:
It may happen when you run several games without closing the emu. Something is not cleared between sessions.

oddMLan:
For what it's worth, I've been experiencing this issue more often. I came to the conclusion that deleting the shader cache does not help or alleviate the issue; it still happens (and it's recurrent) with shader storage disabled. It's even recurrent between emulator restarts.

[...]

PR #1270, opened 2016-11-29
Fix issues when re-initializing Combiner class after destruction.

fzurita:
This seems to fix https://github.com/gonetz/GLideN64/issues/975 at least in Android.

I'm not sure which of the changes in this file fixes the issue for me. The changes in the destroy method seem to be correct to me.

The updates in the setCombine method probably fixed it, but I don't understand fully why yet.

### DIFF
--- a/src/GLSLCombiner.h
+++ b/src/GLSLCombiner.h
@@ -50,7 +50,7 @@ private:
 
 	struct iUniform	{
 		GLint loc = -1;
-		int val;
+		int val = -999;
 		void set(int _val, bool _force) {
 			if (loc >= 0 && (_force || val != _val)) {
 				val = _val;
@@ -61,7 +61,7 @@ private:
 
 	struct fUniform {
 		GLint loc = -1;
-		float val;
+		float val = -9999.9f;
 		void set(float _val, bool _force) {
 			if (loc >= 0 && (_force || val != _val)) {
 				val = _val;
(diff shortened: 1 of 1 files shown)


src/Combiner.cpp
```diff
@@ -264,13 +268,13 @@ void CombinerInfo::setCombine(u64 _mux )
 	const u64 key = getCombinerKey(_mux);
 	if (m_pCurrent != nullptr && m_pCurrent->getKey() == key) {
 		m_bChanged = false;
-		m_pCurrent->update(false);
+		m_pCurrent->update(true);
```
gonetz: It is bad approach to mask problem instead of investigation, what is wrong.

If force update really helps, we must know, why.

fzurita: Yeah, I agree. It should at least get us started into a good solution.

fzurita:
When I have time tonight, I can try doing a more refined forced update inside the update method. It should tell me which set of uniforms are not being updated correctly.

fzurita:
A little progress, in GLSLCombiner_ogl3x.cpp ```ShaderCombiner::updateAlphaTestInfo``` always setting the force parameter of ```uEnableAlphaTest``` to true seems to fix the issue. I don't quite understand why yet.

fzurita:
Pull request updated.

gonetz:
> Not all parameters are always updated in ShaderCombiner::updateAlphaTestInfo on the first time around the shader is executed which is when _bForce is set to true.

Great discovery! Thanks!

gonetz:
merged: 4ef15e7e2da

PR #1272, opened 2016-11-29
Get Pokemon Snap playable

purplemarshmallow:


fzurita:
Can this work with frameBufferEmulation\copyToRDRAM=2?

purplemarshmallow:
No, it need to be in sync mode here

AmbientMalice:
Would it be possible to get a Zilmar test build of this? I imagine Mupen64 (not 64plus) will probably be able to emulate the red dot+everything else via partial copies.

purplemarshmallow:
>I imagine Mupen64 (not 64plus) will probably be able to emulate the red dot+everything else via partial copies.

I only tested mupen64plus

src/FrameBuffer.cpp
```diff
@@ -571,7 +571,7 @@ FrameBuffer * FrameBufferList::findTmpBuffer(u32 _address)
 
 void FrameBufferList::saveBuffer(u32 _address, u16 _format, u16 _size, u16 _width, u16 _height, bool _cfb)
 {
-	if (m_pCurrent != nullptr && config.frameBufferEmulation.copyAuxToRDRAM != 0) {
```
gonetz: Ok, leave it as is, if it helps with other games.

gonetz:
Merged: 36fb0d180

Red dot works perfectly with Mupen64, but game hangs when course ended and you returned to lab.

weinerschnitzel:
What about 1964?

tony971:
Using Glupen64 with default settings, the red dot is now gone. However, the constant flickering and freeze at Oak's evaluation are also gone.

no pokemon

PR #1273, opened 2016-11-29
Have android LOG function use __android_log_vprint

loganmc10:
This changes the Android LOG function from a Macro in Log.h to integrating it into Log.cpp

It also changes it from using ```__android_log_print``` to using ```__android_log_vprint```, since that is how the log messages are formatted.

### DIFF
--- a/src/Log.cpp
+++ b/src/Log.cpp
@@ -1,5 +1,3 @@
-#ifndef ANDROID
-
 #include <stdarg.h>
 #include <stdio.h>
 #include "Log.h"
@@ -21,7 +19,7 @@ void LOG(u16 type, const char * format, ...) {
 	char cbuf[bufSize];
 	wcstombs(cbuf, logPath, bufSize);
 	FILE *dumpFile = fopen(cbuf, "a+");
-#endif
+#endif //OS_WINDOWS
 
 	if (dumpFile == nullptr)
 		return;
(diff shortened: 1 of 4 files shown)


gonetz:
The code is unreadable now.
Please move all android code to Log_android.cpp and include it in mupen64plus-video-gliden64.mk

loganmc10:
Most the code is reused which is why I thought it best to use the #ifdef's. Technically I don't think there needs to be a specific Android version of the logging at all anymore since it uses GetUserDataPath to determine where to write gliden64.log (I don't know where that would be in mupen64plus-ae).

I'd defer to @fzurita as to what he'd rather (leave it like it is now, create Log_android.cpp, or use the regular version of logging that would write a gliden64.log file to GetUserDataPath). Personally I find reading gliden64.log easier than parsing through logcat output, but I don't know if GetUserDataPath is supported in mupen64plus-ae

gonetz:
> Most the code is reused which is why I thought it best to use the #ifdef's. 

You use __android_log_vprint, so most of code is useless for android version.

GetUserDataPath is supported in mupen64plus-ae, but I prefer logcat for debug.

loganmc10:
Ok I've made that change

gonetz:
Thanks! Merged: 4706bb52

PR #1277, opened 2016-12-01
New about dialog with more credits and layout changes

Aquatakat:
image


Aquatakat:
I may need to make some adjustments if the tab box height changes when you switch tabs - let me know.

gonetz:
Looks great and works correct. Thanks!

PR #1306, opened 2016-12-21
UI: Bloom sliders converted to grid and word wrap added to labels

Aquatakat:
This change is designed to prevent the window from becoming too wide:

image

Compared to:

image


gonetz:
Looks good, thanks!

PR #1320, opened 2017-01-05
Fix runtime library loading errors with GLES 3.0.

fzurita:
Some devices with GLES 3.0 support only are not tolerant of GLES 3.1
specific calls.


loganmc10:
I still think it would be better not to have a GLES3.1 plugin. It would be better to figure out a way to make the GLES3 version work with 3.0/3.1/3.2 features

fzurita:
Yeah, I completely agree with that, but that would require a lot more effort. Maybe now it's the time to merge the two versions since release 2.0 is complete.

PR #1324, opened 2017-01-09
Add "Animal Forest" to RSP list

loganmc10:
Fan translations of Dōbutsu no Mori exist where the RSP.romname=ANIMAL FOREST

EDIT: Also for some reason on some versions of the ROM the ucode wasn't being detected properly, so I added the CRC to the special microcode list, and this allows the ROM to boot

### DIFF
--- a/src/GBI.cpp
+++ b/src/GBI.cpp
@@ -54,7 +54,8 @@ SpecialMicrocodeInfo specialMicrocodes[] =
 	{ F3DJFG,		false,	0xbde9d1fb, "Jet Force Gemini" },
 	{ F3DPD,		true,	0x1c4f7869, "Perfect Dark" },
 	{ Turbo3D,		false,	0x2bdcfc8a, "Turbo3D" },
-	{ F3DEX2CBFD,	true,	0x1b4ace88, "Conker's Bad Fur Day" }
+	{ F3DEX2CBFD,	true,	0x1b4ace88, "Conker's Bad Fur Day" },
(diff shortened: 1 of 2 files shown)


Enker:
I'm getting an unknown ucode error with the translations. The original rom works, so the ucode detection might need an update.

loganmc10:
I found 2 version, one was 16MB and it worked, one was 32MB and didn't work but I never looked into it. The one I got working apparently only has some translation done, but both report the ROM name as Animal Forest

Enker:
I tested the 90% & 2010-12-02 translations and they both are 32MB and don't work. Seems odd that expanding the rom would cause this error, so maybe they change something else as well.

loganmc10:
Yes it seems that the ucode needed to be added to the list, I pushed that changed to this PR, and now the newer translation works for me

Enker:
Thanks! I tested both with glupen64 and they are working now.

loganmc10:
I get a seizure inducing flashing when any menu is open (when the guy asks your name on the train for instance). I need to look into whether that happens with the Japanese rom and mupen64plus+GLideN64

loganmc10:
It seems like the flashing only happens with the English translations and not Dōbutsu no Mori...

Enker:
I noticed that glN64 also has this bug with the translations. GLideN64's LLE mode works for (J) & translations, but the menu background is broken.

loganmc10:
Ya LLE mode looks like this:
image

Better than the flashing I think, maybe I'll play in LLE mode

gonetz:
Merged: 3ad872cff4ef1e78a

Issue #1335, opened 2017-01-28
Major code rafactor: remove direct use of graphics API


gonetz:
GLideN64 uses OpenGL graphics API. There are several major desktop versions of OpenGL with different capabilities, there are 3 supported versions of mobile OpenGL. That causes issues, described in #1276

I refactored GLideN64 code to totally remove direct calls to graphics API from main code. All core  GLideN64 classes use proxy class **Context** from namespace **graphics** to manipulate with textures , shaders, buffers and so on,  and to draw objects. Context passes calls to backend class. Currently there is one backend, which uses OpenGL. If somebody wants to add Vulkan or DirectX support, it can be made much easier now: just write new backend.

Of course, use of proxy class instead of direct call to GL API causes additional load on CPU. From the other side, CPU usually is not a bottleneck for plugin's performance. Number of API calls should be the same or less than now. Use of more advanced GL functions also can give some boost.

OpenGL backend designed as dynamically adoptable for available GL version. It may use different functions for the same task. For example, if available GL supports glTexStorage2D the new texture will be initialized with glTexStorage2D and with glTexImage2D otherwise.
BTW, I tried to use new functions, available in OpenGL 4.5, but they do not work yet. That code presents, but currently not used.

The amount of refactoring work was very large because OpenGL calls were everywhere. Now this work is mostly done: checkout refactor_graphics branch:
https://github.com/gonetz/GLideN64/tree/refactor_graphics
The new code works on Windows and Linux. I'm still working on polishing it, but it already can be used and tested.
Currently the main task is to port it on Android. I moved most of old GLES specific code to the new backend and corrected list of source files in mupen64plus-video-gliden64.mk project file. This is not enough, and I count on @fzurita help. I hope, that this refactoring will allow us to have one version of the plugin for all GL versions available on Android devices.

fzurita:
Thanks gonetz. This is great work, I can see you are quite busy. I'll start on this today as soon as I have a chance. I'm sure @loganmc10 will be interested as well.

gonetz:
Yes, I hope that @loganmc10 will help me to polish the backend and tune it for max performance. Currently the tuning is very simple, and based only on GL version.

gonetz:
If you want to help me to find regressions, here the current binaries for Windows:
https://drive.google.com/file/d/0B0YqMPjGo3B2NE8zSm1IbnJNWTg/view?usp=sharing

Note: text drawer and blur not refactored yet and thus do not work.

[...]

PR #1339, opened 2017-01-31
Fix some GL Errors

loganmc10:
As I mentioned here:
https://github.com/gonetz/GLideN64/issues/1335#issuecomment-276447209

Having the function pointers use the exact same name as the actual OpenGL function is a problem. It breaks apitrace, which is a very useful tool in GL debugging. There might be other issues as well, I'm not sure.

This change prepends ```m_``` to the function pointers. I tested this on Linux, but I haven't tested EGL or Windows.

The 2nd commit fixes a couple extension issues. You can't rely on checking if a function pointer is NULL, sometimes an implementation will return a valid pointer even if the function is unsupported. I also added a check for ARB_compute_shader for Image Textures, since it provides glMemoryBarrier and that is required.


fzurita:
Instead of a blanket search and replace, I would consider using macros to redefine those strings. For example:
```
#define glEnable m_glEnable
```

loganmc10:
ok ill look into that

loganmc10:
Thanks for the tip, that works as well and it's a lot cleaner. I've updated the pull request.

loganmc10:
I've also pushed a commit to fix Image Texture support on devices with GL < 4.2. They need a couple lines in the GLSL header to work.

loganmc10:
I've also added a commit to fix the EGL build on non-Android devices. I've tested it on Linux and it works good, I haven't tried the Raspberry Pi.

gonetz:
'm_ ' prefix is for class members. It is confusing to use it for global functions.

I use 'g_' for global objects. 

If you don't like 'g_gl' you may use just '_gl' or '_' postfix, e.g. 'glBlendColor_'

gonetz:
Other commits look good.

loganmc10:
I'm fine with g_, I just did m_ because I don't know much about coding conventions, I'll correct it when I get a chance

loganmc10:
Ok should be good now

PR #1340, opened 2017-02-01
Fix compile on Fedora

loganmc10:
Compilation fails on Fedora with ```-static-libstdc++``` in the Makefile. I've tested this on Ubuntu 16.04 and Fedora and everything works correctly with this change.


gonetz:
merged:  bf1e748d7

PR #1342, opened 2017-02-01
A few extension cleanups

loganmc10:
Adds support for EXT_buffer_storage, and ARB/EXT_texture_storage

### DIFF
--- a/src/Graphics/OpenGLContext/GLFunctions.cpp
+++ b/src/Graphics/OpenGLContext/GLFunctions.cpp
@@ -253,6 +253,10 @@ void initGLFunctions()
 	GL_GET_PROC_ADR(PFNGLGETSTRINGIPROC, glGetStringi);
 	GL_GET_PROC_ADR(PFNGLINVALIDATEFRAMEBUFFERPROC, glInvalidateFramebuffer);
 	GL_GET_PROC_ADR(PFNGLBUFFERSTORAGEPROC, glBufferStorage);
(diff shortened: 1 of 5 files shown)


src/Graphics/OpenGLContext/GLFunctions.cpp
```diff
@@ -280,4 +280,10 @@ void initGLFunctions()
 	GL_GET_PROC_ADR(PFNGLCREATEBUFFERSPROC, glCreateBuffers);
 	GL_GET_PROC_ADR(PFNGLCREATEFRAMEBUFFERSPROC, glCreateFramebuffers);
 	GL_GET_PROC_ADR(PFNGLNAMEDFRAMEBUFFERTEXTUREPROC, glNamedFramebufferTexture);
+#ifdef EGL
+	if (g_glBufferStorage == nullptr)
+		g_glBufferStorage = (PFNGLBUFFERSTORAGEPROC) eglGetProcAddress("glBufferStorageEXT");
+	if (g_glTexStorage2D == nullptr)
+		g_glTexStorage2D = (PFNGLTEXSTORAGE2DPROC) eglGetProcAddress("glTexStorage2DEXT");
+#endif
```
loganmc10: I know this is a little ugly, but with OpenGL ES they append "EXT" onto functions that come from an extension, so this is the best way I could think of to support ARB_buffer_storage and EXT_buffer_storage.

Issue #1313, opened 2016-12-28
[Mupen64Plus] No video in Mupen64Plus 2.5.0


gourdcaptain:
When using GLideN64 2.0 with Mupen64Plus 2.5.0 on Arch Linux 64-bit (Intel HD Graphics 515, Mesa 13.0.2) the screen shows no visuals and a black screen, while the audio plays for the game on all games I have access to. Another Arch Linux user is having the same issue with the plugin as noted here on the AUR package: https://aur.archlinux.org/packages/mupen64plus-video-gliden64/

This occurs when using Mupen64plus standalone as well as in m64py. A log from Mupen64Plus is below:

|  \/  |_   _ _ __   ___ _ __  / /_ | || | |  _ \| |_   _ ___ 
| |\/| | | | | '_ \ / _ \ '_ \| '_ \| || |_| |_) | | | | / __|  
| |  | | |_| | |_) |  __/ | | | (_) |__   _|  __/| | |_| \__ \  
|_|  |_|\__,_| .__/ \___|_| |_|\___/   |_| |_|   |_|\__,_|___/  
             |_|         http://code.google.com/p/mupen64plus/  
Mupen64Plus Console User-Interface Version 2.5.0

UI-Console: attached to core library 'Mupen64Plus Core' version 2.5.0
UI-Console:             Includes support for Dynamic Recompiler.
Core: Goodname: Banjo-Kazooie (U) (V1.0) [!]
Core: Name: Banjo-Kazooie       
Core: MD5: B29599651A13F681C9923D69354BF4A3
Core: CRC: A4BF9306 BF0CDFD1
Core: Imagetype: .z64 (native)
Core: Rom size: 16777216 bytes (or 16 Mb or 128 Megabits)
Core: Version: 1448
Core: Manufacturer: Nintendo
Core: Country: USA
UI-Console Status: Cheat codes disabled.
UI-Console: using Video plugin: 'GLideN64' v2.0.0
UI-Console: using Audio plugin: 'Mupen64Plus SDL Audio Plugin' v2.5.0
UI-Console: using Input plugin: 'Mupen64Plus SDL Input Plugin' v2.5.0
UI-Console: using RSP plugin: 'Hacktarux/Azimer High-Level Emulation RSP Plugin' v2.5.0
Core Warning: Fallback for Video plugin API (02.02.00) < 2.2.0. Resizable video will not work
Input: 0 SDL joysticks were found.
[...]

Issue #454, opened 2015-05-01
GLideN64 won't work on Linux with an Intel GPU (Mesa driver)


Mastergatto:
But other video plugins will actually work on Mupen64plus. I have also tried to build the source, but (aside from few compilation errors that I have had to manually fix) GLideN64 still doesn't work. 
In your blog you have said that GLideN64 now supports OpenGL 3.3, which is the higher version supported by the Mesa driver.

> **\* Error in `./mupen64plus': free(): invalid pointer: 0x00007f25b0d48340 ***
> [1]    9340 abort      

My system:
Debian GNU/Linux Jessie 8.0
Intel i7-4770K with HD Graphics 4600

sergiobenrocha2:
Minimum OpenGL is 4.2, where did you read that?

Mastergatto:
http://gliden64.blogspot.it/2015/05/public-release-1.html?showComment=1430498152518#c7901910148167356341
http://gliden64.blogspot.it/2015/05/public-release-1.html?showComment=1430501984030#c6990029942897199087

Predator82Germany:
I have only OpenGL 3.3 & plugin works

sergiobenrocha2:
I tried here with my Intel HD Graphics 4600:

```
python3: /home/sergio/Projects/Mupen64Plus/GLideN64/GLideN64_Public_Release_1/src/GLSLCombiner.cpp:265: GLuint _createShader(GLenum, const char*): Assertion `checkShaderCompileStatus(shader_object)' failed.
Aborted (core dumped)
```

With the binary from the release:
```
Core: Starting R4300 emulator: Dynamic Recompiler
Core: R4300: starting 64-bit dynamic recompiler at: 0x7f1e6a5214e0
OpenGL Error: invalid operation
```

sergiobenrocha2:
glxinfo | grep OpenGL:
```
OpenGL vendor string: Intel Open Source Technology Center
OpenGL renderer string: Mesa DRI Intel(R) Haswell Mobile 
OpenGL core profile version string: 3.3 (Core Profile) Mesa 10.1.3
OpenGL core profile shading language version string: 3.30
OpenGL core profile context flags: (none)
OpenGL core profile profile mask: core profile
OpenGL core profile extensions:
OpenGL version string: 3.0 Mesa 10.1.3
OpenGL shading language version string: 1.30
OpenGL context flags: (none)
OpenGL extensions:
```

Mastergatto:
[...]

PR #1343, opened 2017-02-01
VBO Implementation

loganmc10:
This commit adds support for using VBO's. It fixes the long standing "black screen with Mesa/Linux" issue (https://github.com/gonetz/GLideN64/issues/1313 https://github.com/gonetz/GLideN64/issues/454).

VBO's are mandatory in core OpenGL. This implementation has 2 different methods for dealing with them. If the device supports ARB_buffer_storage, it uses that. Essentially there is a permanent mapping into the buffer and a simple memcpy is used to transfer the data.

If the device doesn't support ARB_buffer_storage, it uses glMapBufferRange + GL_MAP_UNSYNCHRONIZED_BIT. This method is not as fast, since glMapBufferRange takes some time, but it's still quite fast, and like I mentioned it is mandatory to do with in core OpenGL, and it is the fastest approach outside of ARB_buffer_storage. Most computers nowadays support ARB_buffer_storage.

OpenGL ES never removed support for client side vertex arrays. For ES devices, VBO's are enabled if the device supports EXT_buffer_storage and GLES3.2 (required for glDrawElementsBaseVertex). If EXT_buffer_storage is not supported, it uses client side arrays, I haven't found any VBO implementation besides EXT_buffer_storage that is as fast as client side arrays.

This setup has been used for some time with GLupeN64 (a RetroArch mupen64plus + GLideN64 core, which is used by a number of people on Windows, Linux, and Android) and results have been promising.


src/Graphics/OpenGLContext/opengl_UnbufferedDrawer.cpp
```diff
@@ -21,63 +24,123 @@ UnbufferedDrawer::UnbufferedDrawer(const GLInfo & _glinfo, CachedVertexAttribArr
 	m_cachedAttribArray->enableVertexAttribArray(rectAttrib::texcoord0, false);
 	m_cachedAttribArray->enableVertexAttribArray(rectAttrib::texcoord1, false);
 
-	m_attribsData.fill(nullptr);
+	int fake;
+	m_attribsData.fill(&fake);
```
loganmc10: I know this looks a little weird, but nullptr is a valid value when using VBO's, so if they default to nullptr, the caching fails because it thinks the proper value is already set.

If there is a better way to initialize these to a "useless value that is not NULL" I'm open to it.

src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
```diff
@@ -59,7 +59,9 @@ void DisplayWindowMupen64plus::_setAttributes()
 	CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_MAJOR_VERSION, 3);
 	CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_MINOR_VERSION, 2);
 #else
-	// Do nothing
+	CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_PROFILE_MASK, M64P_GL_CONTEXT_PROFILE_CORE);
+	CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_MAJOR_VERSION, 3);
+	CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_MINOR_VERSION, 3);
```
gonetz: These changes should work for OS_MAC_OS_X, so please just remove OS_MAC_OS_X section above.

src/Graphics/OpenGLContext/opengl_UnbufferedDrawer.h
```diff
@@ -19,9 +20,25 @@ namespace opengl {
 		void drawLine(f32 _width, SPVertex * _vertices) override;
 
 	private:
-		bool _updateAttribPointer(u32 _index, const void * _ptr);
+		bool _updateAttribPointer(u32 _type, u32 _index, const void * _ptr);
+		void _updateBuffer(u32 _type, u32 _length, const void * _data);
 
+		CachedBindBuffer * m_bindBuffer;
```
gonetz: UnbufferedDrawer got this name because it implements drawing without buffers.
There must be another implementation of GraphicsDrawer interface for VAO/VBO.
ContextImpl will select suitable implementation based on available GL capabilities.

UnbufferedDrawer made first because it works for all GL version.
I planned to do VBO implementation when code with UnbufferedDrawer will start to work on all platforms.
I made preliminary work for it in my experimantal_vbo branch.

gonetz: We probably need two VBO implementation classes: one for buffer_storage and another for glMapBufferRange

loganmc10:
Ok I seperated this implementation into it's own Class. I've updated the cmake and Android makefiles, but I didn't update the Visual Studio makefiles since I don't have VS, and I didn't want to try to edit those files by hand.

Right now the glBufferStorage and glMapBufferRange implementations are both in opengl_BufferedDrawer.cpp. There is very little difference between the 2 (you can just look for the ```if (m_glInfo.bufferStorage)``` blocks to see the differences).

gonetz:
Hi! I just started to port my VBO branch to refactor_branch, and ops - you finished with your version.
I checked it briefly, it mostly ok. I have few question to implementation, but that later.
I want to discuss general implementation strategy with you.
As you may noticed, I separated vertex attributes for Rects and Triangles.
My idea is to have separate VAO for rects and triangles. VAO keeps attribute settings, so when we need to render rect after triangle, we can just switch VAO and don't call VertexAttribArray enable/disable. I suppose that it should be faster. What do you think?

src/Graphics/OpenGLContext/opengl_BufferedDrawer.cpp
```diff
@@ -0,0 +1,218 @@
...
+	m_cachedAttribArray->enableVertexAttribArray(triangleAttrib::numlights, false);
+	m_cachedAttribArray->enableVertexAttribArray(triangleAttrib::modify, false);
+
+	m_cachedAttribArray->enableVertexAttribArray(rectAttrib::position, false);
+	m_cachedAttribArray->enableVertexAttribArray(rectAttrib::color, false);
+	m_cachedAttribArray->enableVertexAttribArray(rectAttrib::texcoord0, false);
+	m_cachedAttribArray->enableVertexAttribArray(rectAttrib::texcoord1, false);
+
+	int fake;
+	m_attribsData.fill(&fake);
```
gonetz: Why?

loganmc10:
Yes using 2 VAO's would be the "proper" way to do it, and yes then you would avoid enabling and disabling attributes. From what I've read the performance difference isn't much, but it certainly wouldn't hurt.

VAO's aren't available on GLES2 unfortunately though, so you'd have to deal with that.

I don't mind if you don't use this implementation and just use it for reference or something, doesn't matter to me.

src/Graphics/OpenGLContext/opengl_BufferedDrawer.cpp
loganmc10: Because nullptr is a valid pointer value for VBO's (because it is referring to an offset in the buffer). If they get initialized to nullptr, then the caching doesn't work, because it thinks it already has the correct value, so it never executes glVertexAttribPointer.

There may be a better way to setup these values, I was just trying to say "uselsss value that isn't nullptr"

gonetz:
Thanks!

> VAO's aren't available on GLES2 unfortunately though, so you'd have to deal with that.

As you said, VBO does not bring performance boost to GLES2, so it will live with UnbufferedDrawer

> I don't mind if you don't use this implementation and just use it for reference or something, doesn't matter to me.

Ok, I'll finish with porting my version then. Your commit is clear now, and is good reference for filling buffers in the right way.

src/Graphics/OpenGLContext/opengl_BufferedDrawer.cpp
gonetz: Ok, I'll keep it in mind.

src/Graphics/OpenGLContext/opengl_BufferedDrawer.cpp
```diff
@@ -0,0 +1,218 @@
...
+	m_bufMaxSize = 4194304;
```
gonetz: Why 4194304? Is not it too much 4mb per buffer? Especially for indices.

loganmc10: It probably is overkill, it's the same reason I did ```m_size * 32``` in the PBO persistent mapping code.

When the buffer is full, it resets the offset back to 0 and starts over at the beginning of the buffer. There is nothing preventing you from overwriting something in the buffer that hasn't been used yet, which would result in corruption/crash. Basically I made the buffer large enough that I can be pretty sure by the time I've "made my way around the buffer", the contents aren't neede anymore.

The safer way to do it would be to use a sync object to make sure it is safe to overwrite a portion of the buffer, but I've never ran into any problems doing it this way (it's been like this in GLupeN64 for a while now and I've never had a single issue). I figured using 12MB of memory (3 buffers) is pretty small and then I don't have to have the overhead of the sync objects.

gonetz: I had circa 24mb for all buffers and textures (including HD ones) on my Voodoo5. Now I have gigs of video memory, but still can't used to it. 4MB just to be sure that it is enough! Such buffer can hold circa 95000 vertex elements btw. That is, at least for 100 frames. Of course, it is overkill.
I left that limit in my implementation for now, but it is matter for future optimizations.

gonetz:
@loganmc10 
VBO implementation: a62522532

It is mix of my experimental_vbo branch and your commit.
Please check it, may be I did something not optimal.

Issue #710, opened 2015-09-07
Star Wars Episode I - Racer bad performance


purplemarshmallow:
The CPU spends the most time in tidtable.c. I'm not sure what's wrong here.

purplemarshmallow:
I found the problem, it's NoiseTexture::update(). #711 fixes the problem but only if noise emulation is disabled. It's this line:
GLubyte* ptr = (GLubyte*)glMapBufferRange(GL_PIXEL_UNPACK_BUFFER, 0, dataSize, GL_MAP_WRITE_BIT);

mudlord:
Would be more helpful to implement then noise emulation as a pixel shader. Considering that Shader Model 4.0 seems to be the baseline for this plugin, would make sense to offload work as a small pixel shader.

purplemarshmallow:
Noise causes some slowdown in every game but in other games it's not so bad. This game uses noise excessively and it can use noise almost everywhere. Noise is already emulated using shaders (fragment_shader_noise). The slowdown seems to come from the RNG this seems to be done by the CPU. @mudlord you think the solution here is to generate random numbers on the GPU using shaders?

AmbientMalice:
@purplemarshmallow Before the recent PR, was GLideN64 generating noise even if noise was disabled?

purplemarshmallow:
One performance intensive part of noise emulation was not disabled if you disable noise before my PR.

purplemarshmallow:
I'm not sure if the slowdown is caused by the RNG. Here's a screenshot. The CPU also spends time in tidtable.c. If I skip NoiseTexture::update() all performance problems are gone and there's also no bottleneck anymore in tidtable.c.
![unbenannt](https://cloud.githubusercontent.com/assets/7283660/9799156/cbc3db32-5805-11e5-96e4-f65e8171ad0b.jpg)

mudlord:
In that case a shader for RNG sounds reasonable then. Would offload that performance bottleneck, tho I don't know whats so special about glMapBufferRange that causes such a performance hit.

purplemarshmallow:
It only causes a performance hit in Star Wars Episode I - Racer. In other games there's only a small impact. Would be nice to find out why this happens. dataSize is always VI.width*VI.height. At this point it shouldn't matter how much noise a game uses. Here's Ocarina of Time. In both cases noise is updated once per frame.
[...]

PR #1344, opened 2017-02-01
Faster noise emulation

loganmc10:
Sorry for the onslaught of PR's

Right now, noise textures are generated "on-the-fly". Some games do this every single frame (GoldenEye, Star Wars Ep I Racer, probably many others). This results is glTexSubImage2D (texture upload) happening every frame.

This has resulted in some performance problems like this:
https://github.com/gonetz/GLideN64/issues/710

This change generates 30 random noise textures at launch, and cycles through them randomly. I looked at the character selection screen in Super Smash Bros and the result looks the same to me, although someone with a keener eye could verify. The "30" number could always be increased to increase the randomness.

This should result in a pretty big performance improvement for noise emulation.

### DIFF
--- a/src/NoiseTexture.cpp
+++ b/src/NoiseTexture.cpp
@@ -14,8 +14,7 @@ using namespace graphics;
 NoiseTexture g_noiseTexture;
 
 NoiseTexture::NoiseTexture()
-	: m_pTexture(nullptr)
-	, m_DList(0)
+	: m_DList(0)
 {
 }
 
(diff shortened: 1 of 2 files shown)


gonetz:
> Sorry for the onslaught of PR's

Nothing. Any code improvement is welcome. 
I can't promise that I'll approve them all quickly, but I'll try to do it asap.

src/NoiseTexture.cpp
```diff
@@ -14,90 +14,89 @@ using namespace graphics;
...
+	for (u32 i = 0; i < NOISE_TEX_NUM; ++i) {
+		textureCache().removeFrameBufferTexture(m_pTexture[i]);
+		m_pTexture[i] = nullptr;
+	}
 }
 
 void NoiseTexture::update()
 {
-	if (!m_pbuf || m_pTexture == nullptr)
+	int tex_num = rand() % NOISE_TEX_NUM;
```
gonetz: Put this code after if (m_DList == dwnd().getBuffersSwapCount() ...
No need to call rand() if texture will not be used.

gonetz: Also, current tex_num need to be class member. If new rand index is the same as current one, call rand again.

gonetz: ... or just increase/decrease the current index.

loganmc10: I thought about just increasing the index but then you still have a pattern, I figured randomizing the texture would help with the "randomness"

gonetz: I don't think that rand will often return the same value, but when it does, you must not use the same texture twice.

loganmc10:
Ok I've updated the PR

gonetz:
Thanks!
Merged: cd17543447d

PR #1345, opened 2017-02-02
Remove need for USE_DEPTH_RENDERBUFFER

loganmc10:
In GLES2, depth textures require GL_OES_depth_texture (https://www.khronos.org/registry/OpenGL/extensions/OES/OES_depth_texture.txt).

This extension is almost universal on Android, but it is missing on the Raspberry Pi, which is why renderbuffers were needed.

This change now has GLideN64 check for GL_OES_depth_texture instead of using the hardcoded USE_DEPTH_RENDERBUFFER


gonetz:
Good fix, thanks. Merged: 4456d43d

PR #1351, opened 2017-02-04
Improve PBO writer performance

loganmc10:
See https://github.com/gonetz/GLideN64/pull/1350 for my comments on why I added ```GL_MAP_UNSYNCHRONIZED_BIT | GL_MAP_INVALIDATE_BUFFER_BIT``` to the glMapBufferRange call.

This adds a new PixelWriteBuffer implementation that uses ARB/EXT_buffer_storage. It should offer a pretty good performance improvement, especially on mobile devices.

I tested this with Dr. Mario since I know it's a game that uses CopyColorFromRDRAM for the pills and it worked well. My laptop doesn't support image textures so I was unable to test that.


gonetz:
Merged: d5052c0c70c7

PR #1354, opened 2017-02-05
[Refactor] Get GLES 2.0 working again

fzurita:
Ok, I got the GLES 2.0 version working again. I'm not getting any more GL errors and graphics appear to show correctly at least.

Also, I keep getting this in logcat:

E/libEGL: called unimplemented OpenGL ES API 
Now to track down which call is doing this will be difficult.

Another thing
3-point filtering seems to be broken with GLES 2.0

### DIFF
--- a/src/Textures.cpp
+++ b/src/Textures.cpp
@@ -1015,7 +1015,7 @@ void TextureCache::_loadBackground(CachedTexture *pTexture)
 		params.width = pTexture->realWidth;
 		params.height = pTexture->realHeight;
 		params.format = colorFormat::RGBA;
-		params.internalFormat = glInternalFormat;
+		params.internalFormat = gfxContext.convertGHQTextureFormat(u32(glInternalFormat));
 		params.dataType = glType;
 		params.data = pDest;
 		gfxContext.init2DTexture(params);
@@ -1338,7 +1338,7 @@ void TextureCache::_load(u32 _tile, CachedTexture *_pTexture)
 			params.msaaLevel = 0;
 			params.width = tmptex.realWidth;
 			params.height = tmptex.realHeight;
-			params.internalFormat = glInternalFormat;
+			params.internalFormat = gfxContext.convertGHQTextureFormat(u32(glInternalFormat));
 			params.format = colorFormat::RGBA;
 			params.dataType = glType;
 			params.data = pDest;
--- a/src/Graphics/OpenGLContext/opengl_ContextImpl.cpp
+++ b/src/Graphics/OpenGLContext/opengl_ContextImpl.cpp
@@ -408,17 +408,7 @@ bool ContextImpl::isSupported(graphics::SpecialFeatures _feature) const
 	case graphics::SpecialFeatures::ImageTextures:
 		return m_glInfo.imageTextures;
 	case graphics::SpecialFeatures::ShaderProgramBinary:
-	{
-	   GLint numBinaryFormats = 0;
-#ifdef GL_NUM_PROGRAM_BINARY_FORMATS
-		glGetIntegerv(GL_NUM_PROGRAM_BINARY_FORMATS, &numBinaryFormats);
-#endif
-
-		if (m_glInfo.isGLESX)
-		   return numBinaryFormats != 0 && Utils::isExtensionSupported(m_glInfo, "GL_OES_get_program_binary");
-
-	   return numBinaryFormats != 0 && Utils::isExtensionSupported(m_glInfo, "GL_ARB_get_program_binary");
-	}
+		return m_glInfo.shaderStorage;
 	case graphics::SpecialFeatures::DepthFramebufferTextures:
 		if (!m_glInfo.isGLES2 || Utils::isExtensionSupported(m_glInfo, "GL_OES_depth_texture"))
 			return true;


gonetz:
Wow, great work!
Some solutions are questionable though, I'll see how make them better.

How far we are from having just one EGL version for Android?

fzurita:
That's my next focus.

gonetz:
> E/libEGL: called unimplemented OpenGL ES API

I see this code in ContextImpl::isSupported :
```
#ifdef GL_NUM_PROGRAM_BINARY_FORMATS
		glGetIntegerv(GL_NUM_PROGRAM_BINARY_FORMATS, &numBinaryFormats);
#endif

```
I guess, GL_NUM_PROGRAM_BINARY_FORMATS defined for EGL, but GLES2 driver may not know about it. In that case call of glGetIntegerv with unknown parameter will at least cause GL error, is not it?

fzurita:
Yes, that was causing a GL error, that was the only reason I did that.

gonetz:
Can GLES2 support shader storage? May be just return false for it?

fzurita:
I'm not exactly sure what you mean. GLES 2 can't do shader storage. Although, the EGL function calls can be kept around. Is that what you mean?

gonetz:
I mean, that part of code 
```
bool ContextImpl::isSupported(graphics::SpecialFeatures _feature) const
{
...
	case graphics::SpecialFeatures::ShaderProgramBinary:
	{
...
	}

```
can just return false for GLES2.

gonetz:
I merged your fixes. I polished a bit few of them and made another implementation for 'Fix textures.cpp for GLES 2' and 'Fix wrong texture internal format for GLES2' commits. Should be equivalent.

fzurita:
Yes, I think so, I don't think that there is an extension that allows it in GLES2.

fzurita:
Never mind, I'm wrong: https://www.khronos.org/registry/OpenGL/extensions/OES/OES_get_program_binary.txt

gonetz:
Ok, let it be as it is.

fzurita:
Although, we shouldn't use this macro to check for support: ```GL_NUM_PROGRAM_BINARY_FORMATS```

fzurita:
I just realized, you can just do this:
```
	case graphics::SpecialFeatures::ShaderProgramBinary:
		return m_glInfo.shaderStorage;
```

You already have the right check at m_glInfo.

fzurita:
A few other thing. glFragDepth can be supported by GLES 2 using this extension:
https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_frag_depth.txt

Also, GL_OES_depth_texture can be supported can be supported by GLES 2 as well.

loganmc10:
> Also, GL_OES_depth_texture can be supported by GLES 2 as well.

Yes, that's the only time it's useful, higher GL versions support that natively

fzurita:
And one last thing that @loganmc10 pointed out to me as well. We should not be checking for extensions if the extension is already a standard part of a given GL version. The extension may not necessarily report as supported even though it is.

fzurita:
I added a couple of commits.

src/Textures.cpp
```diff
@@ -1015,7 +1015,7 @@ void TextureCache::_loadBackground(CachedTexture *pTexture)
 		params.width = pTexture->realWidth;
 		params.height = pTexture->realHeight;
 		params.format = colorFormat::RGBA;
-		params.internalFormat = glInternalFormat;
+		params.internalFormat = gfxContext.convertGHQTextureFormat(u32(glInternalFormat));
```
gonetz: How it works when glInternalFormat is RGB5_A1 or RGBA4?
DataType information is enough to load such textures?

gonetz: Also, if GLES2 does not support sized internal formats, they all can be replaced by RGBA in ContextImpl::init(), as I already did for RGBA8. That will guarantee that GLES2 will not met unsupported types in other places of code.

fzurita: I'll check on that.

fzurita: Those two seem ok, but this contradicts:
https://www.khronos.org/registry/OpenGL-Refpages/es2.0/xhtml/glTexImage2D.xml

That seems to suggest that it can only be GL_ALPHA, GL_LUMINANCE, GL_LUMINANCE_ALPHA, GL_RGB, GL_RGBA in GLES 2.0. I wonder if there is an extension allowing it.

fzurita: Ok, it seems like the extension OES_rgb8_rgba8 is allowing the additional color formats, but it's still not fully compatible with GLES 3.0.

gonetz: > Those two seem ok, but this contradicts:

You misunderstood me. Your fix replaces converts any internal formats, including GL_RGB5_A1, GL_RGBA4 to GL_RGBA. However, internal structure of texture data does not changed. It still will be 5551 or 4444. I mean that GLES2 works with these textures correctly because it has DataType information, as GL_UNSIGNED_SHORT_5_5_5_1 and it is enough.

Currently, we have the following internal color formats, which GLES2 code can met:
```
		InternalColorFormatParam RGB8(GL_RGB);
		InternalColorFormatParam RGBA8(GL_RGBA8);
		InternalColorFormatParam RGBA4(GL_RGBA4);
		InternalColorFormatParam RGB5_A1(GL_RGB5_A1);
		InternalColorFormatParam LUMINANCE(0x1909);

```
I suggest, instead of 
`params.internalFormat = gfxContext.convertGHQTextureFormat(u32(glInternalFormat));`
which is contradicts with semantic of gfxContext.convertGHQTextureFormat function,
replace internal RGB format by GL_RGB and RGBA formats by GL_RGBA in ContextImpl::init() for GLES2.

fzurita: Ok

gonetz:
0c6564f
3f855ff

PR #1355, opened 2017-02-05
A few GL fixes

loganmc10:
This makes sure it only uses glBufferStorageEXT if glBufferStorage isn't supportd.

It also removes support for EXT_texture_storage (glTexStorage2D in GLES2),since glTexStorage2D requires sized internal formats (ie GL_RGBA8, not GL_RGBA), and we aren't setup to support that.

It also changes the check for using GL_NUM_EXTENSIONS to ```_glinfo.majorVersion >= 3```, which basically excludes everything except GLES2.

@fzurita perhaps you could review this to make sure it's all ok


gonetz:
Merged : 5d5124225241

PR #1357, opened 2017-02-06
UI: Mild N64 textures tab adjustment; less XML on spacers

Aquatakat:
The vertical label/controls on this tab were bothering me. Once compiled it'll look something like this:

image

I removed the size hint from all the spacers because the default is 0, so the XML can be a little cleaner.


gonetz:
Yes, it looks better that way .

PR #1358, opened 2017-02-06
A few corrections to BufferedDrawer

loganmc10:
2 things:
Disabled vertices caching if buffer_storage is in use (there might be a cleaner way to do this?)

GL_ARRAY_BUFFER was hardcoded in _updateBuffer, I changed that to _buffer.type (this would have only affected computer without ARB_buffer_storage)


gonetz:
7e360c613
ca7da541

PR #1360, opened 2017-02-06
use glCopyImageSubData when possible

loganmc10:
This is not ready to be merged yet, it should be complete but I need to do more testing. I'm just putting it here so it can be reviewed/receive suggestions.

Testing the opening to Mario Tennis (the screen with the Blimp), this reduces glBindFramebuffer calls from 23 to 10 for a single frame. It is accomplished in 2 ways:

First: It only binds the framebuffer when it needs to, so a call to gfxContext.bindFramebuffer() does nothing more than record the desired framebuffer, and the framebuffer is actually bound when it is needed. This avoids any unnecessary "flip-flopping" (switching back and forth between framebuffers needlessly).

Second: It uses glCopyImageSubData instead of glBlitFramebuffer when it can. glCopyImageSubData is like glBlitFramebuffer, but it can only copy images of the same size, and the same number of samples (can't copy from a multisampled texture into a regular one).

The benefit of glCopyImageSubData can be seen the the documentation:

> It should be considered to operate in a manner similar to a CPU memcpy

It avoids the need to bind GL_DRAW_FRAMEBUFFER or GL_READ_FRAMEBUFFER, and simply copies one texture to another.


src/Graphics/OpenGLContext/opengl_BufferManipulationObjectFactory.cpp
```diff
@@ -81,6 +81,7 @@ class AddFramebufferTexture2D : public AddFramebufferRenderTarget
 
 	void addFrameBufferRenderTarget(const graphics::Context::FrameBufferRenderTarget & _params) override
 	{
+		gfxContext.bindFramebuffer(_params.bufferTarget, _params.bufferHandle);
```
loganmc10: This is necessary to update the m_boundDrawFBO/m_boundReadFBO variables, there may be a better way?

The other option would be to not count on addFrameBufferRenderTarget to bind the FBO, and call gfx.bindFramebuffer() in the frontend before calling addFrameBufferRenderTarget. The same goes for blitFramebuffers

gonetz:
I suggest to split this commit on two parts:
* definitely handful use of glCopyImageSubData instead of glBlitFramebuffer when it is possible
* questionable reduce of glBindFramebuffer calls. This commit in its current state introduces non-evident code flow, while practical bonus in performance is not proved. Mario Tennis is very special case, other games do not switch buffers as often. As I read in your discussion with Francisco, even in Mario Tennis performance boost is not evident.

src/Graphics/OpenGLContext/opengl_BufferManipulationObjectFactory.cpp
```diff
@@ -357,34 +363,44 @@ class BlitFramebuffersImpl : public BlitFramebuffers
...
-		m_enableScissor->enable(false);
-		glBlitFramebuffer(
-			_params.srcX0, _params.srcY0, _params.srcX1, _params.srcY1,
-			adrenoCoordFix +_params.dstX0, _params.dstY0, _params.dstX1, _params.dstY1,
-			GLbitfield(_params.mask), GLenum(_params.filter)
-		);
-		m_enableScissor->enable(true);
-
+		if (_params.dstX1 - _params.dstX0 == _params.srcX1 - _params.srcX0 &&_params.dstY1 - _params.dstY0 == _params.srcY1 - _params.srcY0 &&
+		_params.readBufferAttachment != graphics::ObjectHandle() && _params.drawBufferAttachment != graphics::ObjectHandle() && m_copyImage) {
```
gonetz: _params.readBufferAttachment != graphics::ObjectHandle()
Use ObjectHandle::isNotNull() instead: _params.readBufferAttachment.isNotNull()

gonetz: m_copyImage need to be checked first: why check parameters if m_copyImage is not supported?

loganmc10:
Good points.

Probably the best way to properly reduce redundant glBindFramebuffer calls would be to get rid of gfxContext.bindFramebuffer altogether. Then you just pass the desired framebuffer to the other API calls, for instance you add a parameter to gfxContext.drawTriangles that includes the framebuffer you want to draw to. I looked at doing this, but it was a little complex for me to figure out how to accomplish that in the front-end

gonetz:
>Probably the best way to properly reduce redundant glBindFramebuffer calls would be to get rid of gfxContext.bindFramebuffer altogether.

Well, sounds reasonable. In most cases current DRAW FBO is the one, which bound with current Framebuffer object. READ FBO is used for ReadPixels and for BlitFramebuffer. I think that we can get rid of bindFramebuffer with careful code investigation. This is potentially dangerous refactoring. I suggest to  open a ticket for that task, but postpone its solution until other works in that branch completed and it will go to master.

loganmc10:
After doing some more testing, even the glCopyImageSubData change is a little dangerous, the internal formats also need to match, the srcX/srcY can't be negative.. I think there are too many chances for error.

I'll close this for now, I may look at it at a later date, but I don't think it will be good to possibly introduce new issues at this time, when there are still many existing issues to fix.

PR #1365, opened 2017-02-07
Force disable fragment depth writing in GLES2

loganmc10:
There were a few spots that weren't properly checking for enableFragmentDepthWrite support.


gonetz:
merged: acff19e66c

PR #1367, opened 2017-02-08
Remove macros checking for GL version and use runtime checks instead

fzurita:
This allows one build to work across multiple GL versions.

I'm not too sure if what I did in the mupen64plus Displaywindow class is the correct thing. @loganmc10 can you check that it works in your environment for mupen64plus?

I did that way because it doesn't seem like other video plugins ever set the GL version and I can't check the GL version until the GL context has been created.


src/GBI.cpp
```diff
@@ -190,7 +190,8 @@ void GBIInfo::_makeCurrent(MicrocodeInfo * _pCurrent)
 			case F3DEX2MM:	F3DEX2MM_Init();	break;
 		}
 
-#ifndef GLESX
+		if(gfxContext.isSupported(graphics::SpecialFeatures::NearPlaneClipping))
+		{
 		if (m_pCurrent->NoN) {
```
loganmc10: The indentation needs to be moved over here

src/GBI.cpp
```diff
@@ -200,9 +201,10 @@ void GBIInfo::_makeCurrent(MicrocodeInfo * _pCurrent)
 			gfxContext.enable(graphics::enable::DEPTH_CLAMP, false);
 			gfxContext.enable(graphics::enable::CLIP_DISTANCE0, false);
 		}
-#endif
+		}
 	} else if (m_pCurrent->NoN != _pCurrent->NoN) {
-#ifndef GLESX
+		if(gfxContext.isSupported(graphics::SpecialFeatures::NearPlaneClipping))
+		{
 		if (_pCurrent->NoN) {
```
loganmc10: Same here, indentation needs to be shifted right

src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
```diff
@@ -193,10 +172,12 @@ void DisplayWindowMupen64plus::_readScreen2(void * _dest, int * _width, int * _h
 	if (_dest == nullptr)
 		return;
 
+	m_glInfo.init();
```
loganmc10: is this right? m_glInfo already has a .init() in ContextImpl::init()

src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
```diff
@@ -193,10 +172,12 @@ void DisplayWindowMupen64plus::_readScreen2(void * _dest, int * _width, int * _h
...
+
 	u8 *pBufferData = (u8*)malloc((*_width)*(*_height) * 4);
 	u8 *pDest = (u8*)_dest;
 
-#ifndef GLES2
+	if(!m_glInfo.isGLES2) {
 	GLint oldMode;
```
loganmc10: Indentation needs to be shifted here

src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
```diff
@@ -205,9 +186,9 @@ void DisplayWindowMupen64plus::_readScreen2(void * _dest, int * _width, int * _h
 		glReadBuffer(GL_BACK);
 	glReadPixels(0, m_heightOffset, m_screenWidth, m_screenHeight, GL_RGBA, GL_UNSIGNED_BYTE, pBufferData);
 	glReadBuffer(oldMode);
-#else
+	} else {
 	glReadPixels(0, m_heightOffset, m_screenWidth, m_screenHeight, GL_RGBA, GL_UNSIGNED_BYTE, pBufferData);
```
loganmc10: Indentation needs to be shifted here

fzurita:
Whoops, I used WinMerge to copy my changes to this branch from my other branch.... It apparently doesn't pick up indentation. 

I'll fix it...

src/GBI.cpp
gonetz: space after if 
```
if () {
} else {
}
```

src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
fzurita: It's not a global or a single ton. It has to be re-created here. It would had been a pain to use the same instance of GLInfo in both places.

gonetz: ContextImpl can be initialized after we got valid GL context. But why call init in DisplayWindowMupen64plus::_readScreen2 ?

fzurita: When start is called here, there is no valid gl context. I should probably just make it be a local variable in this function.

src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
```diff
@@ -205,9 +186,9 @@ void DisplayWindowMupen64plus::_readScreen2(void * _dest, int * _width, int * _h
...
-#endif
+	}
```
gonetz: BTW, is it possible that that function is called with _front != 0?
If _front is always zero, all that code is redundant and we can left only glReadPixels

loganmc10:
With regards to mupen64plus Displaywindow:

You need to keep:
```
CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_PROFILE_MASK, M64P_GL_CONTEXT_PROFILE_CORE);
CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_MAJOR_VERSION, 3);
CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_MINOR_VERSION, 3);
```
Everything else can go. GLideN64 requires an OpenGL core profile, and mupen64plus defaults to compatibility, so it needs to be specifically requested.

As for the version numbers, the purpose of those numbers isn't to try and guess the user's GL level, it's to specify the minimum required GL version for the application to run. GLideN64 requires at least GL 3.3, so that is specified as the minimum, the actual version of the GL context created will be the highest backwards compatible GL version available (so if the machine supports OpenGL 4.5, and that is backwards compatible with 3.3, then a 4.5 context will be created).

Fortunately, when mupen64plus-core is compiled with USE_GLES, it seems to ignore the M64P_GL_CONTEXT_PROFILE_CORE request and it properly creates a GLES context.

gonetz:
> GLideN64 requires an OpenGL core profile, and mupen64plus defaults to compatibility, so it needs to be specifically requested.

I'm afraid GLES2 will not work with 3.3 core profile. However, GLES2 should be available only on pure GLES2 hardware, which probably can't create 3.3 core profile.

> GLideN64 requires at least GL 3.3, so that is specified as the minimum, the actual version of the GL context created will be the highest backwards compatible GL version available

It is in theory. On practice, when I ask 3.3 core, my Radeon driver returns me 3.3 core. I asked minor and major version, and got 3 and 3. When I asked for 4.5 core I got 4.5.

src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
fzurita: I'm not sure about that. I wouldn't mind just ignoring the _front parameter. I mean, how does the core know that is should read either the front or the back buffer?

loganmc10:
> I'm afraid GLES2 will not work with 3.3 core profile. However, GLES2 should be available only on pure GLES2 hardware, which probably can't create 3.3 core profile.

Ya I tested mupen64plus and requesting Core 3.3 has no effect when used in GLES mode, it just requests a GLES context, so it works fine.

> It is in theory. On practice, when I ask 3.3 core, my Radeon driver returns me 3.3 core. I asked minor and major version, and got 3 and 3. When I asked for 4.5 core I got 4.5.

I've noticed this as well. We can try just asking for the major version, like:
```
CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_PROFILE_MASK, M64P_GL_CONTEXT_PROFILE_CORE);
CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_MAJOR_VERSION, 3);
```
That works for me (I get a 3.3 context, my max), I don't have an OpenGL 4+ machine to test, but that may properly request a higher GL context.

gonetz:
On windows, core profile creation is two pass. First, you create some profile. Then, you get necessary extension functions using that profile and wglGetProcAddress. Only then you can create core profile. When I get first profile, I ask it for minor-major version numbers, and use these numbers in request for core profile.

fzurita:
Mupen64Plus delegates the context creation to SDL it seems, from my look at the code. In Android, SDL then delegates the context creation to the app itself. In mupen64plus-ae, I can safely ignore the requested context version and use max GLES or OpenGL if the hardware allows.

loganmc10:
I think the best way to do it for now would just be:

```
CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_PROFILE_MASK, M64P_GL_CONTEXT_PROFILE_CORE);
CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_MAJOR_VERSION, 3);
```

That will work with GLES, and it properly request a core profile. I don't know of any better way to do it in mupen64plus

gonetz:
@fzurita Can you check in code, when ReadScreen2 API function is called with _front != 0 ?

fzurita:
Yeah, I'll check tonight. I'm at my job right now.

loganmc10:
Actually it seems like if you just request ```CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_MAJOR_VERSION, 3)```, then SDL requests 3.1.

So I would say
```
CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_PROFILE_MASK, M64P_GL_CONTEXT_PROFILE_CORE);
CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_MAJOR_VERSION, 3);
CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_MINOR_VERSION, 3);
```

Is the best way to do it. As long as we properly check for extensions we should be ok. If a machine supports OpenGL 4.5, but returns a 3.3, it should still report support for extensions like GL_ARB_shader_image_load_store and GL_ARB_compute_shader, so Image Textures should still work.

loganmc10:
As far as the ReadScreen2 front/back buffer issue:

The mupen64plus API contains M64CMD_READ_SCREEN. It allows the frontend to choose:

> 1 to copy the buffer that is currently displayed (front buffer), 0 to copy the buffer that is being drawn (back buffer).

Info is here:
https://github.com/mupen64plus/mupen64plus-core/blob/master/doc/emuwiki-api-doc/Mupen64Plus-v2.0-Core-Front-End.mediawiki#command-functions

In practice, it looks like neither official frontend (mupen64plus-ui-console, m64py) actually implement this feature. They just support taking regular screenshots.

loganmc10:
Ok I've done more testing with the context creation:
On Linux (Mesa), if I request 3.1, 3.2, or 3.3, it gives me 3.3, which is the best. If I request 3.0, I get 3.0

On Windows (Nvidia), if I request 3.1, 3.2, or 3.3, that's what I get. If I request 3.0, I get 4.5, which is the best.

So it's almost like they behave exactly opposite of each other. so I would definitely do:

```
CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_PROFILE_MASK, M64P_GL_CONTEXT_PROFILE_CORE);
CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_MAJOR_VERSION, 3);
CoreVideo_GL_SetAttribute(M64P_GL_CONTEXT_MINOR_VERSION, 3);
```

Which will ensure the highest compatibility, unless someone else has a different idea.

gonetz:
>In practice, it looks like neither official frontend (mupen64plus-ui-console, m64py) actually implement this feature. They just support taking regular screenshots.

Well, in that case we can safely replace 
```
void DisplayWindowMupen64plus::_readScreen2(void * _dest, int * _width, int * _height, int _front)
{
#ifndef GLES2

```
by 
```
void DisplayWindowMupen64plus::_readScreen2(void * _dest, int * _width, int * _height, int _front)
{
#ifndef ANDROID

```
and not use glInfo.

BTW, can we replace ANDROID by OS_ANDROID? OS_LINUX, OS_WINDOWS, OS_ANDROID

loganmc10:
I believe ANDROID is an automatic system define but I could be wrong

gonetz:
Cite mupen64plus-video-gliden64.mk:

```
MY_LOCAL_CFLAGS :=      \
    $(COMMON_CFLAGS)    \
    -g                  \
    -DTXFILTER_LIB      \
    -DANDROID           \
...
```

loganmc10:
I'm pretty sure ANDROID would still be defined if that wasn't there, the toolchain defines it, but of course it's still possible to change that to OS_ANDROID

fzurita:
Yes, loganmc10 is right about the Android define. I could still replace it with OS_ANDROID, it wouldn't hurt anything.

gonetz:
Ok. It is not urgent of course.

fzurita:
Pull request updated.

Also, a lot of the wstring for Android may not be needed any more.

gonetz:
> Also, a lot of the wstring for Android may not be needed any more.

Which wstring(s) ?

gonetz:
Merged.

PR #1369, opened 2017-02-08
Disable N64 Depth Compare if Image Textures are unsupported

loganmc10:
Instead of checking for the config setting and checking for Image Texture support every time we need it, this disables the config setting in GLInfo.

The same goes for Fragment based depth writing and GLES2

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
@@ -762,7 +762,7 @@ public:
 				"uniform lowp int uRenderTarget;	\n"
 				"uniform mediump vec2 uDepthScale;	\n"
 				;
-			if (_glinfo.imageTextures && config.frameBufferEmulation.N64DepthCompare != 0) {
+			if (config.frameBufferEmulation.N64DepthCompare != 0) {
 				m_part +=
 					"uniform lowp int uEnableDepthCompare;	\n"
 					;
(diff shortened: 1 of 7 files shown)


src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -1128,7 +1128,7 @@ class ShaderFragmentCallN64Depth : public ShaderPart
 public:
 	ShaderFragmentCallN64Depth(const opengl::GLInfo & _glinfo)
 	{
-		if (!_glinfo.isGLES2 && _glinfo.imageTextures && config.frameBufferEmulation.N64DepthCompare != 0) {
+		if (!_glinfo.isGLES2 && config.frameBufferEmulation.N64DepthCompare != 0) {
```
gonetz: !_glinfo.isGLES2 check looks redundant after these modifications.

src/Graphics/OpenGLContext/opengl_GLInfo.cpp
```diff
@@ -41,6 +41,12 @@ void GLInfo::init() {
 				Utils::isExtensionSupported(*this, "GL_ARB_compute_shader"))) && (glBindImageTexture != nullptr);
 		msaa = true;
 	}
+	if (!imageTextures && config.frameBufferEmulation.N64DepthCompare !=0) {
```
gonetz: missing space after !=

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
loganmc10: yes you are right

loganmc10: yes you are right

loganmc10:
Ok I've updated the PR with those changes

gonetz:
Thanks! Merged: 1f74b9

PR #1370, opened 2017-02-08
Make cmake work with mingw

loganmc10:
Again, unrelated to refactor_graphics, but I don't want to cause merge conflicts.

This allows GLideN64 to be built on Windows using cmake/MinGW. It also removes the GLideNHQ dependencies on the included libz.a and libpng.a, and has it use the system's libpng and zlib. This was required since they don't work with MinGW. It's also good practice to use the system libraries, since they receive updates.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -166,6 +166,12 @@ if(UNIX)
   endif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
 endif(UNIX)
 
+if(MINGW)
+  add_definitions(
+    -DMINGW
+  )
+endif(MINGW)
+
 if(WIN32)
   list(APPEND GLideN64_SOURCES ${GLideN64_SOURCES_WIN})
   add_definitions(
--- a/src/GLideNHQ/CMakeLists.txt
+++ b/src/GLideNHQ/CMakeLists.txt
@@ -20,11 +20,11 @@ set(GLideNHQ_SOURCES
   TxUtil.cpp
 )
 
-if(PANDORA OR BCMHOST)
+if(MINGW OR PANDORA OR BCMHOST)
   include_directories( ${CMAKE_CURRENT_SOURCE_DIR}/../osal )
-else(PANDORA OR BCMHOST)
+else(MINGW OR PANDORA OR BCMHOST)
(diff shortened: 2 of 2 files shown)


gonetz:
> it's also good practice to use the system libraries, since they receive updates.

It is arguable. Your code can suddenly stop working, because you used deprecated function, which dissappear with new update.
I use prebuilt png library because libpng in official Ubuntu 14 repository is too old. May be it is updated in Ubuntu 16. Anyway, if you don' want surprises caused by third party libraries, best practice is to keep suitable version of their source code in your repo, and build them from sources. I think about it. This requires additional work in cmake, I don't have time for it now.

loganmc10:
It's really quite common in Linux to depend on libraries like zlib and libpng, and I think in the absence of built-in source code like you mentioned, linking to the system libraries should be the preferred method.

I've tested this in Ubuntu 16.04, Fedora 25, and MinGW on Windows. I'll test it on Ubuntu 14.04 to be sure, since that is the version by buildbot uses and it is still supported, but I think if it works in 14.04 it should probably be merged

gonetz:
I tested new project file on my Ubuntu 14.04 It is installed in VirtualBox and has all recent updates. Log:
```
Looking for bcm_host.h
bcm_host.h not found
-- Found ZLIB: /usr/lib/x86_64-linux-gnu/libz.so (found version "1.2.8") 
-- Found PNG: /usr/lib/x86_64-linux-gnu/libpng.so (found version "1.2.50") 
-- Configuring done

```

make finished successfully:
```
Linking CXX shared library plugin/debug/mupen64plus-video-GLideN64.so
[100%] Built target mupen64plus-video-GLideN64

```

However, it will not work properly. Minimal required version of libpng is 1.4.xx

gonetz:
You may add new cmake option, like
-DUSE_SYSTEM_LIBPNG

loganmc10:
Fair enough, I won't need to do that, I'll just add MINGW to the if statement along side BCMHOST, that will solve the problem with MinGW, all the other platforms work fine currently

loganmc10:
Ok I've updated it just to fix the MINGW build

gonetz:
Merged

PR #1372, opened 2017-02-11
Implement generic copy color to RDRAM for GLES2 devices

fzurita:
Async copy is not supported for GLES2 devices unless using EGL Image extension which has platform independent implementations.

I will add the EGL Image extension for Android next.

### DIFF
--- a/src/Graphics/OpenGLContext/opengl_ColorBufferReaderWithReadPixels.cpp
+++ b/src/Graphics/OpenGLContext/opengl_ColorBufferReaderWithReadPixels.cpp
@@ -1,0 +1,43 @@
+#include <Graphics/Context.h>
+#include "opengl_ColorBufferReaderWithReadPixels.h"
+#include <algorithm>
+
+using namespace graphics;
+using namespace opengl;
+
+ColorBufferReaderWithReadPixels::ColorBufferReaderWithReadPixels(CachedTexture *_pTexture)
+	: ColorBufferReader(_pTexture)
+{
+
+}
+
+u8 * ColorBufferReaderWithReadPixels::readPixels(s32 _x0, s32 _y0, u32 _width, u32 _height, u32 _size, bool _sync)
+{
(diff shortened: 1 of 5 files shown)


src/Graphics/OpenGLContext/opengl_ColorBufferReaderWithReadPixels.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+	for (unsigned int lnIndex = 0; lnIndex < _height; ++lnIndex) {
+		std::copy_n(pixelData.get() + (lnIndex*strideBytes), widthBytes, m_pixelData.data() + lnIndex*widthBytes);
+	}
+	return m_pixelData.data();
+}
+
+void ColorBufferReaderWithReadPixels::cleanUp()
+{
+	glUnmapBuffer(GL_PIXEL_PACK_BUFFER);
```
loganmc10: I believe this should be removed

loganmc10:
Where does it check for _sync? Won't this run even is the user has Asynx copies selected? It should probably only run if Sync is selected, and be disabled otherwise, or there will be performance impacts

fzurita:
It doesn't check for it. My reasoning was that if a user enables copy color buffer to RDRAM, even in sync mode, that they would rather have the option do something rather than nothing at all.

src/Graphics/OpenGLContext/opengl_ColorBufferReaderWithReadPixels.cpp
fzurita: Whoops, missed that.

loganmc10:
Ya but Async mode is the default, so by default this is going to cause poor performance, and the user won't know why.

fzurita:
Ok, we could make the default be "off" for GLES 2 devices.

gonetz:
Merged.

PR #1376, opened 2017-02-14
Fix GLES 2.0 texture initialization errors

fzurita:
It seems that not all GL errors were addressed. After making this change, I'm not seeing any more.

### DIFF
--- a/src/Textures.cpp
+++ b/src/Textures.cpp
@@ -626,7 +626,7 @@ void TextureCache::init()
 	params.width = m_pDummy->realWidth;
 	params.height = m_pDummy->realHeight;
 	params.format = colorFormat::RGBA;
-	params.internalFormat = internalcolorFormat::RGBA8;
+	params.internalFormat = gfxContext.convertInternalTextureFormat(u32(internalcolorFormat::RGBA8));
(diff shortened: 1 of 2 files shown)


src/Textures.cpp
```diff
@@ -906,7 +906,7 @@ bool TextureCache::_loadHiresBackground(CachedTexture *_pTexture)
 		params.width = ghqTexInfo.width;
 		params.height = ghqTexInfo.height;
 		params.format = ColorFormatParam(ghqTexInfo.texture_format);
-		params.internalFormat = InternalColorFormatParam(ghqTexInfo.format);
+		params.internalFormat = gfxContext.convertInternalTextureFormat(u32(InternalColorFormatParam(ghqTexInfo.format)));
```
gonetz: Too much conversions.
gfxContext.convertInternalTextureFormat(ghqTexInfo.format));
This and below.

fzurita:
Ok, fixed.

src/Textures.cpp
gonetz: In fact it is redundant code. I already did 
ghqTexInfo.format = gfxContext.convertInternalTextureFormat(ghqTexInfo.format);
few lines above.

PR #1377, opened 2017-02-14
Implemented Async buffer copies for GLES2 using EGLImage in Android

fzurita:
Run time loading of the library used this implementation:

https://github.com/Rakashazi/emu-ex-plus-alpha/tree/e5f5d8fa1c9c04338932c0282ea1dbe916e2d6af/imagine/src/base/android/privateApi

Please let me know if this doesn't compile in Windows or Linux.

Also, I left the code I pulled in almost as is, with a few minor modifications to fit our code. Let me know if you want me to adjust it to fit our coding style guidelines.

Also, does this crash some GLES 2.0 devices, so it will only be used when Async copies are enabled.


gonetz:
Changes in code, which Windows or Linux use are minimal. No problems.

PR #1378, opened 2017-02-14
UI supports AA in powers of 2; fixed spaces to tabs

Aquatakat:
I'm still pretty useless with C++ but hopefully this compiles.

I noticed some things I added earlier used four spaces where you prefer tabs so I changed them to tabs.


gonetz:
I fixed compilation errors and made it as one commit: 02b10ac946fd

PR #1386, opened 2017-02-19
German translation 1/2

gizmo98:
It’s not complete but it is a start.


gonetz:
Thanks! I put it to master branch: b34e9df3fe

refactor_graphics branch is for refactoring of rendering code .

PR #1387, opened 2017-02-19
German translation part 2

gizmo98:


gonetz:
@gizmo98 
I merged the translation without check. It actually does not work. There are plenty of such errors:
lrelease error: Parse error at gliden64_de.ts:186:235: Entity 'auml' not declared.
lrelease error: Parse error at gliden64_de.ts:317:113: Entity 'uuml' not declared.
lrelease error: Parse error at gliden64_de.ts:317:128: Entity 'ouml' not declared.

xml can't take strings like this "Aufl&ouml" "Fensterr&auml"

gizmo98:
Ok. Sorry. I will replace them.

gizmo98:
I thought this would be the right way to write ä, ö and ü in html.

gizmo98:
xml is different:
http://www.devtrain.de/artikel_296.aspx
&auml; --> &#228;
&uuml; --> &#246;
&ouml; --> &#246;

PR #1392, opened 2017-02-20
Disable async color buffer copies on non-android GLES2

loganmc10:
The implementation only supports synchronous copies, so this disables it altogether if they have async chosen (right now if they choose async it will give them sync)


gonetz:
I fixed compilation and merged: c42c1a4a4

loganmc10:
Thanks, sorry it didn't compile, I couldn't compile because of that noise texture thing so I was just guessing

PR #1393, opened 2017-02-21
german translation: fix special signs

gizmo98:
Replace html special sign notation with xml notation.

@gonetz 
Please take a look if there is no error now.

XML validator says there is no error anymore.


gonetz:
You should use QT Linguist tool to avoid such problems.
I open the translation file with QT Linguist, so the problem with xml is fixed, file can be opened.
However, there are plenty of "Translation does not end with the same punctuation as the source text" errors. For example:  "Correction-level:" => "Korrekturwert", colon is missing.
There are many missing translations for tooltips. All these errors are easy to find and fix with QT Linguist.

gizmo98:
Ok. I will use Qt Linguist Tool and update the PR.

gizmo98:
All "ends with different punctuation" and "missing translation" points are fixed now.

gonetz:
Cool! Looks great, thanks!
Merged.

PR #1394, opened 2017-02-22
Allow clang build to work in Android.

fzurita:
This doesn't fix 3DMathNeon.

### DIFF
--- a/src/mupen64plus-video-gliden64.mk
+++ b/src/mupen64plus-video-gliden64.mk
@@ -128,7 +128,7 @@ ifeq ($(TARGET_ARCH_ABI), armeabi-v7a)
     MY_LOCAL_SRC_FILES += $(SRCDIR)/3DMathNeon.cpp.neon
     MY_LOCAL_SRC_FILES += $(SRCDIR)/gSPNeon.cpp.neon
     MY_LOCAL_CFLAGS += -D__NEON_OPT
-    MY_LOCAL_CFLAGS += -D__VEC4_OPT -mfpu=neon -mfloat-abi=softfp -ftree-vectorize -mvectorize-with-neon-quad -ftree-vectorizer-verbose=2 -funsafe-math-optimizations -fno-finite-math-only
+    MY_LOCAL_CFLAGS += -D__VEC4_OPT -mfpu=neon -mfloat-abi=softfp -ftree-vectorize -funsafe-math-optimizations -fno-finite-math-only
 
 else ifeq ($(TARGET_ARCH_ABI), x86)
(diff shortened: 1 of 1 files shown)


fzurita:
I just tried to build the newest master with clang and it's not building, so probably more fixes incoming.

gonetz:
Ok. Can I help?

gonetz:
I see error in Textures.cpp. I'll fix it. VS compiler also can't compile that code.

gonetz:
Fixed: bfe5b6078e7

gonetz:
I tested clang build on Note3. Frame buffer effects do not work, plugin crashed on exit.

fzurita:
Interesting, I didn't have that problem. Maybe there's is some underlying bug in the code that the clang compiler is exposing.

gonetz:
I wish soonest recovery for your family. All my family also were sick with the flu this winter. It was nasty. Anti-cough syrup was consumed by liters.

fzurita:
Thanks!

fzurita:
So this file name is too long for some reason: opengl_ColorBufferReaderWithBufferStorage.h/cpp. It doesn't compile with clang unless I reduce it by a few characters.

gonetz:
Windows limitation I guess.

fzurita:
I made a pull request to address it and another easy fix. 3DMathNeon fix should be coming soon.

gonetz:
@fzurita I can be wrong, but  I suppose you need to exclude neon asm code from the project for clang build because assembler inlines contradict with clang way of code generation.

fzurita:
Possibly, I want to try to convince myself of that first.

fzurita:
Also, we should probably bump ShaderStorageFormatVersion. It's still the same version as before the refactor which may hide issues if they are present.

gonetz:
clang works atop of LLVM. Ready asm code will prevent LLVM from doing its optimizations. At least I'd first try to build without neon asm and see how it works.

I'll bump ShaderStorageFormatVersion.

fzurita:
I'm running the latest master with clang. Everything seems to work ok so far. I'm not sure what issues you were running into. Try building my latest version: https://github.com/mupen64plus-ae/mupen64plus-ae

Also, the issue with corrupt lighting when HW Lighting was enabled is now gone.

fzurita:
Conker's bad fur day is in pretty bad shape. Getting into the water in the starting area causes severe slow downs that are not present before the refactor.

gonetz:
> Try building my latest version

The latest version works ok. At least it does not crash.

PR #1407, opened 2017-03-05
Cleanup Android make files

fzurita:
Cleaning up Android make files as well as moving from static libraries to shared libraries.

Also, fix circular dependency of GLideNHQ to GLideN64 in TxQuantize


src/GLideNHQ/TxQuantize.h
```diff
@@ -32,6 +32,48 @@ class TxQuantize
...
+    197, // 11000 = 11000101
+    206, // 11001 = 11001110
+    214, // 11010 = 11010110
+    222, // 11011 = 11011110
+    230, // 11100 = 11100110
+    239, // 11101 = 11101111
+    247, // 11110 = 11110111
+    255  // 11111 = 11111111
+  };
+
```
gonetz: Why add them to header?

Would not it better to make them as static variables in TxQuantize.cpp ?

fzurita:
Is there much of a difference now a days? Having it in the CPP I guess would allow you to change this without having a bunch of files having to be recompiled. I don't think these values are ever going to change though.

If you prefer though, I can move them to the CPP file.

fzurita:
I updated the pull request to put those arrays in the CPP file.

gonetz:
> Is there much of a difference now a days? Having it in the CPP I guess would allow you to change this without having a bunch of files having to be recompiled.

No, I care about clean code. Headers are for interfaces. Implementation details must be hidden as much as possible.

src/GLideNHQ/TxQuantize.h
```diff
@@ -32,6 +32,9 @@ class TxQuantize
 private:
   int _numcore;
 
+  static const unsigned char One2Eight[2];
+  static const unsigned char Five2Eight[32];
+
```
gonetz: No need to make these arrays as class members.

src/GLideNHQ/TxQuantize.cpp
```diff
@@ -33,6 +33,48 @@
 #include <convert.h>
 #include "TxQuantize.h"
 
+const unsigned char TxQuantize::One2Eight[2] =
```
gonetz: static const unsigned char One2Eight[2] =

src/GLideNHQ/TxQuantize.cpp
```diff
@@ -33,6 +33,48 @@
 #include <convert.h>
```
gonetz: convert.h is not needed if you copy arrays declared in it to there.

src/GLideNHQ/TxQuantize.cpp
```diff
@@ -33,6 +33,48 @@
...
 #include "TxQuantize.h"
 
+const unsigned char TxQuantize::One2Eight[2] =
+{
+	0, // 0 = 00000000
+	255, // 1 = 11111111
+};
+
+const unsigned char TxQuantize::Five2Eight[32] =
```
gonetz: static const unsigned char Five2Eight[2] =

fzurita:
Ok, got it. That's a good stance to have. I updated the pull request.

fzurita:
Ok, pull request updated again.

gonetz:
Thanks!

fzurita:
Yep, np!

Issue #1390, opened 2017-02-20
[Feature Request] Configurable hotkey for enabling/disabling Vsync


DonelBueno:
It would be nice being able to disable Vsync when you disable the framelimiter (F4 on P64), so you can go past 60 fps when you want to speed up the gameplay, but also to enable it back to avoid screen tearing when you are on 60 fps (framelimiter enabled).

Since GLide can't have control over this, the idea is to have a configurable hotkey to toggle Vsync on the fly.

Discussed in issue #1364

DonelBueno:
@gonetz, any idea on when you will be able to implement this?

gonetz:
Implemented in #1497.
I close this task.
Please report in #1497 if something does not work.

PR #1409, opened 2017-03-07
UI: Custom crop and windowed resolution shows/hides rather than enables/disables

Aquatakat:
For a more clean look on the Video tab, the width/height spinboxes might be hidden rather than disabled when they're not in use. This will also make the left column less tall compared to the rest of the form.

image

The C++ changes are pretty simple so they should compile... Is there a guide somewhere I could use to figure out how to compile GLideN64? Then I could make sure any C++ changes I make actually work.

### DIFF
--- a/src/GLideNUI/ConfigDialog.cpp
+++ b/src/GLideNUI/ConfigDialog.cpp
@@ -106,11 +106,11 @@ void ConfigDialog::_init()
 	}
 	ui->windowedResolutionComboBox->insertItems(0, windowedModesList);
 	ui->windowedResolutionComboBox->setCurrentIndex(windowedModesCurrent);
+
 	ui->cropImageComboBox->setCurrentIndex(config.video.cropMode);
 	ui->cropImageWidthSpinBox->setValue(config.video.cropWidth);
-	ui->cropImageWidthSpinBox->setEnabled(config.video.cropMode == Config::cmCustom);
 	ui->cropImageHeightSpinBox->setValue(config.video.cropHeight);
-	ui->cropImageHeightSpinBox->setEnabled(config.video.cropMode == Config::cmCustom);
+	ui->cropImageCustomFrame->setVisible(config.video.cropMode == Config::cmCustom);
 
 	QStringList fullscreenModesList, fullscreenRatesList;
 	int fullscreenMode, fullscreenRate;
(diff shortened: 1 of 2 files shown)


Aquatakat:
Also I have a couple small questions:

In ConfigDialog.cpp there are arrays of the texture filter and texture enhancements. I could probably move those arrays to the XML so they can be localised. It doesn't seem too difficult, unless there's some other reason not to.

Also the Custom option in the list of windowed modes could be moved to the XML to be localised, but the C++ changes are a bit out of my wheelhouse. I was thinking of possibly introducing a hidden QtLabel where the text could be copied over.

gonetz:
I wrote a guide how to build for Windows:
https://github.com/gonetz/GLideN64/wiki/Build-From-Source-(Windows)
but it is outdated. I'm planning to switch to new version of MSVS and new version of QT. May be I'll update the guide after that.

List of texture enhancements hardly can be localized; i think HQ4X is the same on any language.
List of texture filters can be moved to XML.

Custom option also can be localized, but I have no time for it now.

gonetz:
Merged: 27cc8099e

Aquatakat:
@gonetz I sat down and tried to figure out how to build based on your guide but I haven't had any luck. Visual Studio and Qt 5.4 and 5.6 resulted in a bunch of errors compiling GLideN64.sln on this one line:

image

If you know an easy way to fix this that would be nice. If not, don't worry about it.

Once I can compile GLideN64 I could probably figure out how to localise that custom option in windowed modes. When I mentioned texture enhancements I was talking about "None" and "Store." The others can be listed in the XML with the translatable tag turned off.

Aquatakat:
I could probably also figure out a UI for options like #1390. I've been thinking about this since they've been mentioned.

gonetz:
These errors have nothing with QT. It can be issue with Visual Studio 2013, but I use 2013 too and this code compiled for me.

First, try to build the project for mupen64plus: select Debug_mupenplus or Release_mupenplus  configuration and build. If it is not compiled, something wrong with the Studio.

Second: I use static build of QT only for Release build, because I need everything linked statically before sending dll to users. Debug build uses QT dlls, and 'Debug' configuration links with non-static QT. If you don't need to share your builds, you also may use standard non-static QT build. Just create new QT configuration in QT plugin for Studio, which points to QT location and set that configuration for GLideNGUI solution.

Aquatakat:
Based on [this](https://msdn.microsoft.com/en-us/library/dn793970.aspx) it seems like Microsoft identified an error in the compiler for update 3+ and "fixed" the problem by throwing an exception. AFAIK this isn't an issue in VC2015+, but I had other various issues trying those versions. I found a copy of the RTM release of VC2013 which compiles just fine.

As far as the static Qt/non static Qt I kind of want to use the static one just because I waited so long for it to compile. >_>

PR #1410, opened 2017-03-08
Shortened path to help Android Studio build in windows

fzurita:
I'm having more issues building in windows unfortunately. The path name is too long now when I'm attempting to build a release build. This was only a problem with the debug build with only 1 file.


gonetz:
Oh-ho.
May be it is possible to increase path length limit?

fzurita:
I'm not sure of a way. It may be a Windows problem...

gonetz:
https://msdn.microsoft.com/en-us/library/aa365247%28VS.85%29.aspx?f=255&MSPPError=-2147217396#maxpath

fzurita:
That's interesting. A registry key allows overriding the behavior of 256 charterers, but only for Windows 10.... I would hate to make that mandatory to compile in Windows for Android.

I'm not sure why Windows has those silly path length restrictions.

Also, I do have my project located a few folders down in my filesystem. Maybe the right solution is to place it closer to root. I could try that.

gonetz:
http://stackoverflow.com/questions/1880321/why-does-the-260-character-path-length-limit-exist-in-windows

gonetz:
> Also, I do have my project located a few folders down in my filesystem. Maybe the right solution is to place it closer to root. I could try that.

Or make a symlink.

gonetz:
One way to cope with the path limit is to shorten path entries with symbolic links.

For example:

create a C:\p directory to keep short links to long paths
mklink /J C:\p\foo C:\Some\Crazy\Long\Path\foo
add C:\p\foo to your path instead of the long path

fzurita:
Yeah, I'll probably just make a folder in my C five for Android projects. A symbolic link would clutter my hard drive I feel.

gonetz:
> I'll probably just make a folder in my C five for Android projects. 

It should help. I don't have that issue with building the project on Windows with MSVS.

fzurita:
This is not an issue after moving my workspace to the root of my c drive.

gonetz:
Good. May be 58ddf8822c can be reverted as well?

fzurita:
Yep

PR #1413, opened 2017-03-11
UI: Texture filter/enhancement dropdowns can now be localised

Aquatakat:
I was able to confirm this compiles properly and the settings are set properly. Some old commits are floating around though and I'm not sure why.


gonetz:
> Some old commits are floating around though and I'm not sure why.

I think it is because you did not pull current master to your fork, or did not switch to it.

gonetz:
Merged: d72adc6b

PR #1415, opened 2017-03-11
UI: Custom windowed resolution option is translatable

Aquatakat:
Turns out figuring out how to make that label translatable was pretty simple. Qt really makes this sort of thing easy, I can see why you chose to use it. Even I can figure it out.

I also figured out why my last pull request had all those other commits. Sometimes cygwin on my computer stops working properly which made git behave weirdly. When I restart my PC it goes back to normal.


gonetz:
It works. Cool!
Merged: 2fa86eee4

PR #1421, opened 2017-03-14
Add missing GLSL extension

loganmc10:
Sorry, I missed this one. This is required if the GL context is less than 4.3, but compute shaders/image textures are supported.

The error message I was getting on Windows/Nvidia when using N64 Depth Compare:

```
shader_compile error: 0(139) : error C7532: global function memoryBarrierImage requires "#version 430" or later
0(139) : error C0000: ... or #extension GL_ARB_compute_shader : enable
```


loganmc10:
Sorry hold off a bit I forgot to also do the fragment shader

loganmc10:
Ok it's been fixed

gonetz:
Thanks!

Issue #1418, opened 2017-03-13
Various regressions since PR2.0 with N64 depth compare enabled [AMD Issue?]


oddMLan:
![imagen](https://cloud.githubusercontent.com/assets/7278372/23850277/f2c1bbae-079b-11e7-91f4-57455c133371.png)

Just start a new game and watch the intro.
Using latest hotfix from #1364 

This is a recent regression, it doesnt happen with build e8c47c0 (public release 2.0)

gonetz:
I can't reproduce it. The game works for me as usual. May be you use non-standard RSP plugin?

oddMLan:
Tested with both the RSP that comes with the emulator and with cxd4's RSP. The problem doesn't seem to go away. Only tested with PJ64 2.3.x though, could it be a core regression? Public Release 2.0 doesn't seem to have any problems.

oddMLan:
It doesn't happen with Mupen64plus build. Closing

oddMLan:
Oddly, it doesn't happen in LLE with Project64... hmm

gonetz:
> I'm thinking one of the functions refactored to use OpenGL freely might be having trouble with my graphics card. AMD Radeon R7 250 series

Hardly. Polygons on your screen shot looks very incorrect. Either ucode detected incorrect (wrong rom version) or problem with rsp.

oddMLan:
Nice catch. Culprit was the PJ64's rsp after all. I'll make an issue for them.
I got confused since it seemed to be wrong with cxd4's rsp too? what even. now it's not the wrong result anymore

oddMLan:
Problem is N64 depth compare. Reopening

oddMLan:
N64 depth compare works properly in Public Release 2.0. Confirming this is a regression in N64 depth compare

gonetz:
Please check mupen64plus build.

oddMLan:
Of what? Public release 2.0?

I tested the latest build of the VI refactor branch with mupen64plus above and this time with N64 depth compare enabled (when I said it was correct before, I didn't have it enabled because I didn't have the custom ini with the setting) and it gave the same wrong result.

gonetz:
> Of what? Public release 2.0?

Latest master. I don't see problems with N64 depth compare in that game.

gonetz:
Also, please check for gliden64.log file. It may contain records about problems with shaders compilation.

oddMLan:
[...]

PR #1422, opened 2017-03-15
Fix GL_ARB_compute_shader statement

loganmc10:
Sorry I got the last commit wrong, you can't have ```#extension GL_ARB_compute_shader : enable``` at the beginning of normal shaders, it gives errors about having user-defined input variables. I somehow dodged this problem when I tested before, perhaps it was shader storage.

Anyway I've tested this with shader storage disabled on my Nvidia machine, and this works (tested with TWINE). However, I do still notice the issue in https://github.com/gonetz/GLideN64/issues/1418. This commit just fixes shader compilation, I was getting a completely black screen before this fix.


gonetz:
Merged: c0d4644a

Issue #1297, opened 2016-12-06
SM64 screen flicker


CornN64:
I compiled the latest version from source on raspberry pi3 (retropie) and the screen start to flicker once the super mario head comes up. I tested a few other games like perfect dark and wave racer which had no problems and was rendering fine...
Looks like some screen buffer issue is causing it.

gonetz:
Could you find the regression commit?

loganmc10:
This started happening at this commit:

313741d8276f0e018ba0c137892d67794468cc28

Turning off Async color buffer copys (or disabling framebuffer emulation) fixes the problem

fzurita:
Interesting. I didn't think the raspberry pi 3 supported any form of Async color buffer read, so it must be falling back to the stub version which shouldn't be doing anything.

loganmc10:
Ya it would be using the stub version, I'm not really sure what caused it, it happens on Linux too if you compile the GLES2 version of GLideN64, which also uses the stub version

fzurita:
If there is no support for sync buffer copies, they should be getting turned off all together in the configuration in my opinion.

loganmc10:
Probably true, I read that Android 7 breaks the GraphicBuffer thing as well (no more access to those "internal" libraries), but I'm not 100% sure about that

fzurita:
It's indeed true, it doesn't work any more, I can tell you from personal experience.

HpNoTiQ56:
Hi,
I've compiled from source today the new NGLIDE64(Raspberry Pi3) and I'm experiencing the flickering issue in SM64!
I've tried mariokart which works fine but can't find a way to make the flickering out of SM64.
I've tried to :
EnableCopyColorToRDRAM : 0 or 1
EnableN64DepthCompare : False/True
EnableFBEmulation : False/True
Here is my config in retropie (Glide section) :
[Video-GLideN64]

# Settings version. Don't touch it.
configVersion = 16
# Crop resulted image (0=disable, 1=auto crop, 2=user defined crop)
CropMode = 0
# Crop width pixels from left and right of resulted image (in native resolution)
CropWidth = 0
# Crop height pixels from top and bottom of resulted image (in native resolution)
CropHeight = 0
# Enable/Disable MultiSampling (0=off, 2,4,8,16=quality)
MultiSampling = 0
# Screen aspect ratio (0=stretch, 1=force 4:3, 2=force 16:9, 3=adjust)
AspectRatio = 1
[...]

PR #1423, opened 2017-03-15
Neon asm optimization

gizmo98:
-rewrite InverseTransformVectorNormalize. Load array deinterleaved (lvd4) so we can reuse TransformVectorNormalize. Fixes whatever reason texture flickering in SM64 on a PI as well. https://github.com/gonetz/GLideN64/issues/1297. Pi problems with enabled framebuffer persist. 
-reorder instructions to get best throughput
-use vld1.32 and vst1.32 to load and store four d registers at once
-add MultMatrix2 neon version without memcpy

### DIFF
--- a/src/3DMath.cpp
+++ b/src/3DMath.cpp
@@ -13,6 +13,13 @@ void MultMatrix(float m0[4][4], float m1[4][4], float dest[4][4])
 	}
 }
 
+void MultMatrix2(float m0[4][4], float m1[4][4])
+{
+	float dst[4][4];
+	MultMatrix(m0, m1, dst);
+	memcpy( m0, dst, sizeof(float) * 16 );
+}
+
 void TransformVectorNormalize(float vec[3], float mtx[4][4])
 {
 	float vres[3];
--- a/src/3DMath.h
+++ b/src/3DMath.h
@@ -3,18 +3,12 @@
 #include <memory.h>
 #include <string.h>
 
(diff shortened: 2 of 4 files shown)


loganmc10:
@gizmo98 this is totally off topic, but what issues are you having with framebuffer emulation on the Pi? Does it still lose the depth after a while/randomly? Or does it happen right away?

gizmo98:
I build my branch and master today. Zelda OoT and MM are broken if i load a game. Beetle adventure racing is also broken. Looks always like the old depth problem. But it happens right after a saved game is loaded. On the other hand Mario Tennis and Mario Golf have less depth issues, but Mario Golf is crashing after a short time now. Don't know if regression happened during refactor graphics.

loganmc10:
hm ok, I'll try and take a look sometime. I tried to re-implement the (imperfect) fix the Raspberry Pi had before refactor graphics here:

https://github.com/gonetz/GLideN64/blob/master/src/Graphics/OpenGLContext/opengl_ContextImpl.cpp#L264-L268

If you remove those lines you should notice that it gets a lot worse, but I'm still not sure how to completely fix the problem. I also never experienced any crashing, I'll have to try out Mario Golf sometime.

gonetz:
Is it ready for pull?

gizmo98:
Yep. This one is ready to pull. We discussed off topic.

gonetz:
Ok

gonetz:
Merged.

Issue #1428, opened 2017-03-18
Raspberry Pi: multiple regressions


gizmo98:
After VI refactoring some games are not aligned correctly. SM64, Super Smash Brothers, Perfect Dark, Paper Mario, Yoshis Story, Mario Kart. FBEmu is off. 
![img_3821](https://cloud.githubusercontent.com/assets/6412699/24076210/5fb4b23c-0c2b-11e7-8a9b-5d9216d09c87.JPG)
![img_3819](https://cloud.githubusercontent.com/assets/6412699/24076216/768b5d8a-0c2b-11e7-97b0-7912826e31de.JPG)
![img_3812](https://cloud.githubusercontent.com/assets/6412699/24076218/9889ccfa-0c2b-11e7-8a8f-492ab0f7b406.JPG)

After graphics refactor zelda games had series issues. The new subscreen fix helps a little bit. After game startup the picture is messed up. if the menu was opened once picture is ok. FBEmu is on.
1. Startup:
![img_3832](https://cloud.githubusercontent.com/assets/6412699/24076233/f6bb012c-0c2b-11e7-99b1-d48600fb9088.JPG)
2. Menu:
![img_3835](https://cloud.githubusercontent.com/assets/6412699/24076238/0fb03436-0c2c-11e7-88ae-1bffd1972232.JPG)
3. Everything ok:
![img_3836](https://cloud.githubusercontent.com/assets/6412699/24076241/2db2bae4-0c2c-11e7-8c35-697c960b9f70.JPG)

Beetle adventure racing has black textures now. FBEmu is on:
![img_3840](https://cloud.githubusercontent.com/assets/6412699/24076249/4bb48da6-0c2c-11e7-88f5-fbd1213bcf15.JPG)

Killer Instinct and Mario Golf crash after a short time.

loganmc10:
I'll look into some of these issues when I get a chance. I'm almost tempted to just force disable Framebuffer emulation on the Raspberry Pi, the current ugly hack to get it sort of working is not ideal, and probably causes other issues. @gizmo98 what do you think? Until someone can figure out what will actually fix it properly it seems like it might be the best solution

loganmc10:
I couldn't reproduce any of the mis-aligned games with FBE on or off, they all look fine to me in that respect.

The issues with Zelda and Beetle Adventure Racing are both related to broken depth buffer on the Raspberry Pi as referenced here: https://github.com/gonetz/GLideN64/issues/1084

[...]

PR #1429, opened 2017-03-19
Disable LOD emulation on GLES2 devices that can't support it

loganmc10:
LOD emulation on GLES2 depends on GL_EXT_shader_texture_lod and GL_OES_standard_derivatives. 

Unfortunately not every device supports these extensions, like the Raspberry Pi.

This disables LOD if either of those extensions are not supported. @fzurita perhaps you could review this. It should be pretty harmless.

Fixes the issue with GoldenEye on the Raspberry Pi noted in https://github.com/gonetz/GLideN64/issues/1428


fzurita:
Looks ok to me.

PR #1432, opened 2017-03-19
add neon optimized RSP and gSP functions

gizmo98:
-add arm neon version of RSP_LoadMatrix.
-add arm neon version of gSPTransformVertex ported from https://github.com/ricrpi/mupen64plus-video-gles2n64/blob/master/src/gSPNeon.cpp#L409.
-use neon intrinsics instead of arm asm.
-tested cmake with and without NEON_OPT and VEC4_OPT.

@fzurita is the android *.mk file ok? *.neon extension on all neon cpp files looks suspicious.
@gonetz i have just added RSP_LoadMatrix.cpp to the visual studio project file. I have no visual studio  to test.


fzurita:
The .neon extension I only got from following examples from other Android makefiles that had neon assembly, I'm not entirely sure what it did or if it works if you remove the .neon extension. I may give it a shot later.

I should also probably enable the X86_ASM when building GLideN64 for Android X86 CPUs.

gonetz:
@gizmo98 Commits need to be rearranged:
* gSPTransformVertex_NEON and related changes into one commit.
* RSP_LoadMatrix and related changes into another one. 'neon asm: remove RSP_LoadMatrix' and 'neon asm: add comments' also should be in that commit. So, two commits.

Then, your decision to move out only Neon version of RSP_LoadMatrix and leave C++, x86 and x64 versions in the same .cpp does not look consistent. Either create separate .cpp for each implementation, or put everything to one .cpp. Solution with separate files will make code more readable, but harder to support in project files.

src/mupen64plus-video-gliden64.mk
```diff
@@ -128,13 +128,15 @@ ifeq ($(TARGET_ARCH_ABI), armeabi-v7a)
     # Use for ARM7a:
     MY_LOCAL_SRC_FILES += $(SRCDIR)/3DMathNeon.cpp.neon
     MY_LOCAL_SRC_FILES += $(SRCDIR)/gSPNeon.cpp.neon
+    MY_LOCAL_SRC_FILES += $(SRCDIR)/RSP_LoadMatrixNeon.cpp.neon
```
gonetz: How can it work? File name is RSP_LoadMatrixNeon.cpp

gizmo98:
Ok. I will rearange commits. As i can see there is only x86 asm for WIN32 and gcc/clang which is separatd with #ifdef. So there should be only one RSP_LoadMatrixX86.cpp i suppose.

src/mupen64plus-video-gliden64.mk
gizmo98: I don't know. But 3DMathNeon.cpp.neon and gSPNeon.cpp.neon should be also wrong then. As i asked fzurita these lines look suspicious.

gizmo98:
If i add a separate file for x86 asm i need a new "X86_OPT" option for cmake or check for x86. In android make file i can put it under TARGET_ARCH_ABI x86. But is visual studio always using WIN32_ASM? If target is arm64 i suppose not. Is there a way to select files target dependant?

src/mupen64plus-video-gliden64.mk
gonetz: I'm afraid that .neon extension will exclude these files from build. Sure, need to check it.

gonetz:
> So there should be only one RSP_LoadMatrixX86.cpp i suppose.

Yes. I thought that it is asm for x86-x64, but it is indeed just x86 for Visual and g++.

> If i add a separate file for x86 asm i need a new "X86_OPT" option for cmake or check for x86. 

Yes, that is what I meant by "harder to support in project files". X86_OPT option should add X86_ASM define.

Windows build currently is 32bit only, so visual studio projects always using WIN32_ASM.

gizmo98:
I will close this. A cleaner PR is here:
https://github.com/gonetz/GLideN64/pull/1438

PR #1445, opened 2017-03-26
arm neon: add multi vector versions of InverseTransformVectorNormalize()

gizmo98:
Arm neon performance is much better if more data can be loaded and
processed at the same time.

Four vectors. Opt level -O3
—————————————-
runtime 100% - C function
runtime 99% - Neon function
runtime 56% - Neon 2x function
runtime 36% - Neon 4x function

Four vectors. Opt level -O2
—————————————-
runtime 100% - C function
runtime 71% - Neon function
runtime 43% - Neon 2x function
runtime 30% - Neon 4x function

--> Best result can be achieved if four vectors are calculated as long as possible. gSPUpdateLightVectors() uses 4x function as long as possible now.

### DIFF
--- a/src/3DMath.cpp
+++ b/src/3DMath.cpp
@@ -87,3 +87,17 @@ End:
 	}
 #endif // WIN32_ASM
 }
+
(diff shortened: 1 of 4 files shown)


gonetz:
Merged: 79dca65a5d135c2

Issue #1326, opened 2017-01-11
Animal Forest broken menu


loganmc10:
Animal Forest is a fan translation of Dōbutsu no Mori.

Normally I wouldn't bring up ROM hacks, but this one works perfectly in angrylion, so I presume it should work properly.

GLideN64 HLE flashes between these 2 frames:

![screenshot from 2017-01-09 22-03-55](https://cloud.githubusercontent.com/assets/848146/21834262/103122ca-d771-11e6-9896-b2b8e1043199.png)

![screenshot from 2017-01-09 22-03-56](https://cloud.githubusercontent.com/assets/848146/21834269/1aa6b1ca-d771-11e6-8934-0d298ca2d8f6.png)

GLideN64 LLE looks like this:

![b5b2c27c-d73e-11e6-9a7a-b4472590de6a](https://cloud.githubusercontent.com/assets/848146/21834274/2ccabf36-d771-11e6-9921-f36cd452d961.png)

Angrylion looks like this:

![screenshot from 2017-01-10 20-10-25](https://cloud.githubusercontent.com/assets/848146/21834281/36439132-d771-11e6-9ea0-216c374ee061.png)

loganmc10:
This is what is looks like with Framebuffer Emulation disabled (doesn't flicker, just has this image):

![image](https://cloud.githubusercontent.com/assets/848146/21854764/3cb4296a-d7d9-11e6-81e8-8092e6d04d39.png)

loganmc10:
Still happening with current master

AmbientMalice:
This looks fairly similar to Resident Evil 2's background issue, IMO.

loganmc10:
The weird thing is that the issue is only present in the fan translation, the original Japanese version works fine, but both versions work fine with angrylion

gonetz:
```
	else if (strstr(RSP.romname, (const char *)"THE LEGEND OF ZELDA") != nullptr ||
			 strstr(RSP.romname, (const char *)"ZELDA MASTER QUEST") != nullptr ||
			 strstr(RSP.romname, (const char *)"DOUBUTSUNOMORI") != nullptr)
		config.generalEmulation.hacks |= hack_subscreen;

```
may be subscreen hack has a side effect?

loganmc10:
The current RSP.cpp has Animal Forest in the subscreen hack as well, I've tried it both ways. It's not a high priority obviously, just found it strange

loganmc10:
[...]

PR #1448, opened 2017-03-28
Add missing ucode for Animal Forest

loganmc10:
Fixes https://github.com/gonetz/GLideN64/issues/1326


loganmc10:
For some reason Animal Forest can't figure out the ucodes, while the Japanese ROM (Doubutsu no Mori) can.

I figured out this ucode by running the Japanese ROM and printing the string for the ucode, it came back as:

```
RSP Gfx ucode S2DEX       fifo 2.08  Yoshitaka Yasumoto 1999 Nintendo.
```

So I assume this is S2DEX2 with NoN set to "false". I tested this and it is working.

Issue #1312, opened 2016-12-25
Performance regression with Copy To RDRAM


Enker:
I found some games (Mario Party 1-3/Paper Mario) that run at 10 to 20 VI/s in recent builds. The first build with the issue is GLideN64_236902b (Release Candidate 1) from Nov 27, 2016.

Disabling "Copy video card frame buffer to N64 memory" makes them run at normal speed.

Tested with intel HD Graphics 4600.

**Edit**: I rechecked this and it affects all games if the setting is "Synchronous".

LegendOfDragoon:
yup i got this issue on Intel HD 530..

fzurita:
It sounds like Intel has problems with this change: 313741d8276f0e018ba0c137892d67794468cc28

Is async mode any faster?

Enker:
Async mode is ~5 to 10 VI/s faster than Sync with the games I listed.

LegendOfDragoon:
>Is async mode any faster?

On my end, async mode is fine.

fzurita:
I wonder if Async mode suffered a similar slow down.

Enker:
I did a quick test with two games and Async mode runs at the same VI/s as older builds when you unlimit the speed.

loganmc10:
So it just affects sync mode?

Enker:
No, Mario Party & Paper Mario run slow with Async mode.
I noticed that Doubutsu no Mori runs fine with Sync mode, so not all games are affected.

LegendOfDragoon:
>So it just affects sync mode?

On my end, yes. I guess Enker's issue is different.

loganmc10:
TBH I'm not surprised that Intel has issues. Intel OpenGL drivers for Windows are terrible. I have a user that gets the wrong colours in Mario 64 with GLupeN64 because it uses ARB_buffer_storage for VBO's

https://github.com/GLupeN64/GLupeN64/issues/140

It could be that Intel's ARB_buffer_storage implementation is a bit shaky

weinerschnitzel:
I mostly use my Intel lappy but I test GlideN64 on my AMD HTPC. If I'm not mistaken gonetz has been only focusing on AMD and Nvidia. Im not expecting any support for Intel as of yet. If there is strong interest in fixing up Intel, I will start submitting issues for it.

LegendOfDragoon:
I think a good solution might be to allow the user to choose which method to use. I also wonder if the GLES2 version is faster on my Intel HD 530.

[...]

PR #1449, opened 2017-03-29
ColorBufferReader Performance Improvements

loganmc10:
I tested Sync mode with both implementations (ColorBufferReaderWithBufferStorage and ColorBufferReaderWithPixelBuffer), and I got 170 VI/S in Luigi's Raceway with both methods. This was on Windows with an Nvidia card.

Given that some desktop users have experienced performance regressions with ColorBufferReaderWithBufferStorage (see https://github.com/gonetz/GLideN64/issues/1312), I think it's probably best to just use that method for GLES devices.

### DIFF
--- a/src/Graphics/OpenGLContext/opengl_ColorBufferReaderWithBufferStorage.cpp
+++ b/src/Graphics/OpenGLContext/opengl_ColorBufferReaderWithBufferStorage.cpp
@@ -26,8 +26,8 @@ void ColorBufferReaderWithBufferStorage::_initBuffers()
 	for (int index = 0; index < _numPBO; ++index) {
 		m_bindBuffer->bind(Parameter(GL_PIXEL_PACK_BUFFER), ObjectHandle(m_PBO[index]));
 		m_fence[index] = 0;
-		glBufferStorage(GL_PIXEL_PACK_BUFFER, m_pTexture->textureBytes, nullptr, GL_MAP_READ_BIT | GL_MAP_PERSISTENT_BIT);
-		m_PBOData[index] = glMapBufferRange(GL_PIXEL_PACK_BUFFER, 0, m_pTexture->textureBytes, GL_MAP_READ_BIT | GL_MAP_PERSISTENT_BIT);
+		glBufferStorage(GL_PIXEL_PACK_BUFFER, m_pTexture->textureBytes, nullptr, GL_MAP_READ_BIT | GL_MAP_PERSISTENT_BIT | GL_MAP_COHERENT_BIT);
+		m_PBOData[index] = glMapBufferRange(GL_PIXEL_PACK_BUFFER, 0, m_pTexture->textureBytes, GL_MAP_READ_BIT | GL_MAP_PERSISTENT_BIT | GL_MAP_COHERENT_BIT);
 	}
 
 	m_bindBuffer->bind(Parameter(GL_PIXEL_PACK_BUFFER), ObjectHandle::null);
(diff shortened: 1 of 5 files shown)


loganmc10:
By the way @fzurita , in my reading I noticed that instead of doing ```glMemoryBarrier(GL_CLIENT_MAPPED_BUFFER_BARRIER_BIT)```, you should be able to just call glBufferStorage and glMapBufferRange with ```GL_MAP_COHERENT_BIT```. I don't know if that will improve performance, it didn't make a difference for me.

```
If GL_MAP_COHERENT_BIT is not set and the server performs a write, the application must call glMemoryBarrier with the GL_CLIENT_MAPPED_BUFFER_BARRIER_BIT set and then call glFenceSync with GL_SYNC_GPU_COMMANDS_COMPLETE (or glFinish). Then the CPU will see the writes after the sync is complete.
```
```
If GL_MAP_COHERENT_BIT is set and the server does a write, the app must call glFenceSync with GL_SYNC_GPU_COMMANDS_COMPLETE (or glFinish). Then the CPU will see the writes after the sync is complete.
```

That might also fix the performance issues (just a random guess), but I can't reproduce the performance issues so it's hard for me to say.

fzurita:
Maybe we should add a check for Intel. We already have a check for Adreno.

loganmc10:
Yeah that would be fine as well, probably a better solution. I'm still curious if GL_MAP_COHERENT_BIT offers any better performance over glMemoryBarrier.

I have an Intel laptop with Linux I'll test tonight, but I don't have any Windows machine to test

loganmc10:
Ok I've updated the PR to exclude Intel GPU's. I still need to test the GL_MAP_COHERENT_BIT thing.

The thing that makes me nervous about ```glMemoryBarrier(GL_CLIENT_MAPPED_BUFFER_BARRIER_BIT)``` is that we have other persistently mapped buffers (the VBO's), and I'm worried that they will get synced as well when that is executed. The OpenGL wiki also has the note:

> Note that this may cause additional synchronization operations.

I don't know why there isn't an option to specify which buffer you want to use with the barrier. Anyway I'll try to see if there is any difference on my Intel laptop tonight.

fzurita:
When I wrote that code, I was using this as a guide:

https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_buffer_storage.txt

In there, specifically:

>         - If MAP_COHERENT_BIT_EXT is not set and the server performs a write, the application must call MemoryBarrier with the CLIENT_MAPPED_BUFFER_BARRIER_BIT_EXT set and then call FenceSync with SYNC_GPU_COMMANDS_COMPLETE (or Finish). Then the CPU will see the writes after the sync is complete.

fzurita:
We could, theoretically speaking, live dangerously and do no syncing at all. That could cause the copies from memory to have data from multiple frames in one read though.

loganmc10:
> In there, specifically:

Yes but the line right under that one says you can use MAP_COHERENT_BIT_EXT without MemoryBarrier, they just don't make clear the pros and cons of each approach.

> We could, theoretically speaking, live dangerously and do no syncing at all. That could cause the copies from memory to have data from multiple frames in one read though.

Yes I've tried that (removing all the Fence Sync stuff), you still need either MAP_COHERENT_BIT or MemoryBarrier so that the information is visible, but you can get away without the Fence Sync.

I wouldn't really call it dangerous, I don't think it would lead to any instability. Just like you say, if the computer can't keep up, it may be displaying data from old frames, and it may be mix-and-matching frames. It's really not a terrible approach since the whole point of async mode is that it's not perfectly accurate, you can still have it wait in sync mode.

loganmc10:
Although now that I think about it, it may cause instability, what if the program was trying to read from ```m_PBOData[m_curIndex]``` at the same moment that the GPU was trying to write to it? That would be possible if you didn't synchronize it, and I assume that would cause big issues.

fzurita:
I don't think that would be an issue, it's just a block of memory. Multiple threads can read/write from a block of memory without causing issues.

loganmc10:
Removing the Fence Sync makes me kind of nervous, especially since it kind of violates the API

```the app must call FenceSync with SYNC_GPU_COMMANDS_COMPLETE (or Finish)```

If we went that route it would need more extensive testing.

As this PR sits right now I've:

- Disabled the BufferStorage path for Intel GPU's
- Switched glBindBuffer to the cached version in ColorBufferReaderWithPixelBuffer
- In sync mode, it uses glFinish instead of glFenceSync, since it is equivalent to calling glFenceSync and glClientWaitSync back to back
- Switch from ```glMemoryBarrier(GL_CLIENT_MAPPED_BUFFER_BARRIER_BIT)``` to ```GL_MAP_COHERENT_BIT```, I still need to do some performance testing tonight to determine which is better.

loganmc10:
I just tested Intel + Linux and I saw no difference between PixelBuffer and BufferStorage. I also saw no difference between glMemoryBarrier and MAP_COHERENT_BIT.

I assume since the Intel GPU shares its memory with the CPU that memory copies like glReadPixels are actually rather efficient.

So I assume the performance issues are some kind of Intel Windows driver bug, unsurprising.

Anyway I'm happy with how this PR is now. I'm personally more partial to MAP_COHERENT_BIT over glMemoryBarrier, but it seems that either way would be fine.

gonetz:
Looks good, thanks!

Issue #1443, opened 2017-03-24
Resource leak


fzurita:
I haven't had time to determine what is leaking, but something is. Here is a save state for conker's bad Fur day:
[starting area.zip](https://github.com/gonetz/GLideN64/files/869334/starting.area.zip)

Jump in the water that Conker is facing above the water fall. Keep on jumping there for like a minute. Frame rate of the game should go down significantly. If you exit the game and come back, frame rate all go back to normal.

loganmc10:
Have you tried increasing the size of the texture cache? It's possible that that area in the game is generating a ton of textures (some games seem to generate non-stop unique textures somehow). It's possible that the texture cache is getting full, and so it has to delete a texture every time to add the new one.

I've ran into stuff like this a number of times in the past:
https://github.com/gonetz/GLideN64/issues/1015
https://github.com/gonetz/GLideN64/pull/1126
https://github.com/gonetz/GLideN64/issues/795
https://github.com/gonetz/GLideN64/issues/997

I suspect it might have something to do with that.

loganmc10:
Increasing the size of the texture cache obviously isn't a permanent solution, but if you double the size of the texture cache, and the problem takes twice as long to manifest itself, then you've got a pretty good indicator where the problem is.

fzurita:
Good idea, I'll give that a shot.

fzurita:
I gave it 800Mbytes of cache, it didn't seem to help.

gonetz:
> starting area.zip

I can't load that state, emulator crashes. Could you post a screen shot from where the problem happens?

fzurita:
Ok, right here:

![conker_bfd-001](https://cloud.githubusercontent.com/assets/11581687/24401842/013a765a-1384-11e7-9194-078a2324fa82.png)

loganmc10:
I've got to assume it has something to do with the texture cache, there aren't really many other things that get generated. I know that Mario Tennis seems to generate endless FBO's though, but I checked Conkers and that doesn't seem to be the case.

Conkers does (like many other games) generate a seemingly endless number of textures though. There are probably ways to optimize the texture cache, for instance, maybe we don't cache small textures at all? Maybe the overhead isn't worth it. Or if there was some way to identify these "one time use" textures that get generated in these games, and don't add them to the cache, that would probably also help.

loganmc10:
[...]

Issue #590, opened 2015-06-25
SD Hiryu


theboy181:
Menus run slow in HLE but are working fine in LLE

LegendOfDragoon:
Interestingly, this particular issue seems to be an issue in every other HW gfx plugin, for me. GlideN64 LLE is the only HW plugin that doesn't run the menu at a ridiculously low speed.

When I profiled the menu with 1964 Video, a ridiculous amount of time spent was in CreateTexture. So it looks like most hardware plugins are constantly creating textures, because I notice the RAM usage of the emulator skyrockets in these slow menus.

I'd like to see how this kind of problem is fixed.

gonetz:
What about z64? Is it slow too?

LegendOfDragoon:
z64gl is at least 2x as slow as GlideN64 LLE, in the menu. That's still much better than any HLE plugin, where it's like 3 VI/s and the RAM usage gets abnormally high.

Z64gl's problem must be something else since the RAM usage doesn't skyrocket.

My guess is that something with the HLE code is causing it to constantly recreate textures.

LegendOfDragoon:
@gonetz ,  I found a hack that fixes the performance problems for the menus, but idk what kind of possible side effects exist.

https://github.com/gonetz/GLideN64/blob/master/src/Textures.cpp#L1196

dividing numBytes by 32, significantly improves the performance for HLE in that game.

Hopefully this will help lead towards a proper fix. What I do know is that it's spending way too much time in CRC_Calculate().

purplemarshmallow:
reminds me of #329

theboy181:
![capture](https://cloud.githubusercontent.com/assets/9299743/18419273/c8bb971e-780b-11e6-8164-6d0c4f4f1983.PNG)


Forgot How slow HLE really is.

Do we know why?

theboy181:
@LegendOfDragoon anymore progress on this? should I close this ?

olivieryuyu:
@gonetz 

I guess this one can be closed, no?

gonetz:
Yes

PR #1454, opened 2017-04-04
Add xxHash as CRC_OPT

loganmc10:
xxHash is a very fast and accurate algorithm. This adds it to the code as "CRC_OPT". It also makes it the default on Android.

I haven't seen any regressions using this method. As a hashing algorithm it has received a fair bit of testing.

In my tests it is significantly faster than CRC32. I haven't yet compared it to the ARMv8 version.

### DIFF
--- a/src/CRC_OPT.cpp
+++ b/src/CRC_OPT.cpp
@@ -1,0 +1,69 @@
+#include "CRC.h"
+#include "xxHash/xxhash.h"
+
+#define CRC32_POLYNOMIAL     0x04C11DB7
+
+unsigned int CRCTable[ 256 ];
+
+static
+u32 Reflect( u32 ref, char ch )
+{
+	 u32 value = 0;
+
+	 // Swap bit 0 for bit 7
+	 // bit 1 for bit 6, etc.
+	 for (int i = 1; i < (ch + 1); ++i) {
+		  if(ref & 1)
+			value |= 1 << (ch - i);
+		  ref >>= 1;
+	 }
+	 return value;
+}
+
+void CRC_Init()
+{
+	u32 crc;
+
+	for (int i = 0; i < 256; ++i) {
+		crc = Reflect( i, 8 ) << 24;
+		for (int j = 0; j < 8; ++j)
+			crc = (crc << 1) ^ (crc & (1 << 31) ? CRC32_POLYNOMIAL : 0);
+
+		CRCTable[i] = Reflect( crc, 32 );
+	}
+}
+
+u32 CRC_Calculate_Strict( u32 crc, const void * buffer, u32 count )
+{
+	u8 *p;
+	u32 orig = crc;
+
+	p = (u8*) buffer;
+	while (count--)
+		crc = (crc >> 8) ^ CRCTable[(crc & 0xFF) ^ *p++];
+
+	return crc ^ orig;
+}
+
+u32 CRC_Calculate( u32 crc, const void * buffer, u32 count )
+{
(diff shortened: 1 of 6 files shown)


fzurita:
Maybe this should be a compile time flag and could be applied to all CPUs that are not x86.

loganmc10:
Since it hasn't been tested in this application I figured it would have to do it's time in CRC_OPT before being promoted.

If it does work well it would benefit x86 as much as arm. Even on an x86 machine you can be CPU limited if you are trying to play a game like Rogue Squadron in LLE or something.

src/CRC_OPT.cpp
```diff
@@ -0,0 +1,64 @@
...
+	return XXH64(buffer, count, crc);
+#else
+	return XXH32(buffer, count, crc);
+#endif
+}
+
+u32 CRC_CalculatePalette(u32 crc, const void * buffer, u32 count )
+{
+#if defined(__x86_64__) || defined(_M_X64)
+	return XXH64(buffer, count * 8, crc);
```
gonetz: Why count * 8 ?
CRC_CalculatePalette is used to calculate CRC for disperse data. Take 2 bytes, skip 6. Check CRC_CalculatePalette for CRC32.
If you will calculate checksum for the whole data, starting from the buffer address, you may get different checksum for the same palette, because data to skip may be different. To use xxHash for palette checksum, palette load need to be changed first in gDPLoadTLUT. It is not hard to do.

loganmc10: Thanks, I've updated the PR so that it works that way.

AmbientMalice:
I know it's not a proper solution, but how much effect does this alternate algorithm have on https://github.com/gonetz/GLideN64/issues/590 ?

loganmc10:
@AmbientMalice Without this change I get 5-6 VI/S in the menus of SD Hiryuu no Ken Densetsu. With this change, along with some other changes to the texture cache we discussed in https://github.com/gonetz/GLideN64/issues/1443, I get 110 VI/S :)

loganmc10:
On Nintama Rantarou 64 Game Gallery, without this change I get 10 VI/S on the "Press Start" screen, with this change and the other improvements to the cache, 80 VI/S

AmbientMalice:
Very promising. So this isn't enabled in the x86 version currently, correct? I only have Visual Studio 2013, but it looks like the flag is for CMAKE.

loganmc10:
> So this isn't enabled in the x86 version currently, correct?

You mean the Visual Studio version. I use cmake to build GLideN64 on Windows and Linux.

I believe gonetz just changed the VS build to use it: https://github.com/gonetz/GLideN64/commit/a118a2e3dce9c8c5ab7de2c4c70ba0e0c6b384a9

AmbientMalice:
@loganmc10 Coolbeans. This is good news. I mean, it is still extremely demanding, but those games are a common source of complaints.

loganmc10:
There is a 64-bit and 32-bit version of xxHash. In their tests the 64-bit version ran at 13.8 GB/s and the 32-bit version ran at 1.9 GB/s.

I did my tests using the 64-bit version with mupen64plus. I don't think Project64 supports 64-bit and I believe the mupen64plus builds that gonetz distributes are 32-bit, so you probably won't experience the same performance benefits as I did, but it should still be much better.

gonetz:
@loganmc10 > With this change, along with some other changes to the texture cache we discussed in #1443 (using an unordered_map and reducing the cache size to 8000), I get 110 VI/S :)

Please do a PR with these changes. I still can't find time to make them.

PR #1455, opened 2017-04-04
Texture Cache performance improvements

loganmc10:


src/Textures.cpp
```diff
@@ -551,11 +551,7 @@ void TextureCache::destroy()
 
 void TextureCache::_checkCacheSize()
 {
-#ifdef VC
-	const size_t maxCacheSize = 15000;
-#else
-	const size_t maxCacheSize = 16384;
-#endif
+	const size_t maxCacheSize = 8000;
```
fzurita: Why not 5000?

loganmc10: I haven't done very thorough testing, I've tested CBFD and a few other games, which led me to believe that 4000 would be a pretty good number.

I'm not sure if games exist that might benefit from a number higher than 4000, if there are games that use more than 4000 (or 5000) cached textures (unlikely but possible), then setting the limit lower than that would cause performance regressions.

I figured that 8000 would be a very safe number, that is still half of the current value, but yes, it is basically a guess.

gonetz:
Thanks!

Issue #1026, opened 2016-06-16
Paper Mario - Pause Menu Background Incorrectly Emulated


RibShark:
Here is what it should look like (using the Angrylion Graphics Plugin): 
![nmqe0000](https://cloud.githubusercontent.com/assets/1957489/16121650/fcd61e4e-33db-11e6-9e41-700ac15f1cf2.png)

As you can see everything including the scenery, background and characters is shown as the background to the pause menu.

Here is what it looks like under GLideN64 with "Copy Depth Buffer to RDRAM" set to either "Software" or "Disable":
![gliden64_paper_mario_000](https://cloud.githubusercontent.com/assets/1957489/16121772/71777f4a-33dc-11e6-8f9f-5b8ed4c09293.png)

All the scenery and characters are gone, leaving just the background. (With "Copy Depth Buffer to RDRAM set to "from VRAM", no graphical elements are visible leaving just a black background.)

SigmaVirus:
I can confirm this issue on newer versions and the last working version of GLideN64 that showed the pause menu correctly for me was dated 2015-09-28.
However it's not game breaking, but would be nice to have displayed correctly.

standard-two-simplex:
For the record, with the current build (rev.93be54a) the background in pause menu is black.

SigmaVirus:
Sometimes it's grey too, it's really weird.

ichee:
In latest version, set "Copy video card depth buffer to N64 memory" to never. Menu background then renders correctly.

ichee:
Fixed in latest master with copyToRDRAM is set to synchronous.

gonetz:
Good to know, thanks!

PR #1458, opened 2017-04-05
Update Paper Mario/Mario Story

AmbientMalice:
According to https://github.com/gonetz/GLideN64/issues/1026#issuecomment-291628122 Paper Mario needs synchronous RDRAM copies.

### DIFF
--- a/ini/GLideN64.custom.ini
+++ b/ini/GLideN64.custom.ini
@@ -79,6 +79,14 @@ frameBufferEmulation\N64DepthCompare=1
 Good_Name=Mario Golf (E)(J)(U)
 frameBufferEmulation\copyDepthToRDRAM=0
 
+[MARIO%20STORY]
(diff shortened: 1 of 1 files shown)


loganmc10:
I also see:

> In latest version, set "Copy video card depth buffer to N64 memory" to never. Menu background then renders correctly.

Does that also work then? Synchronous copies can be very demanding

AmbientMalice:
Someone more familiar with the game can probably comment on whether it needs RDRAM copies, but my personal assumption is that accurate emulation of its effects probably does.

gonetz:
> Does that also work then? Synchronous copies can be very demanding

Sync color buffer copy is required for backgrounds to work. Currently it works regardless of  depth buffer copy mode.

gonetz:
Merged: 0ad54a8887

PR #1462, opened 2017-04-06
UI: New font selection

Aquatakat:
The old font selection only works with a handful of fonts. I made a new one that isn't perfect but it works with any TTF font on the user's system.

image

I'm still pretty useless with C++ so most of the code I wrote was guesswork. I think it's probably all OK but you may want to have a quick go-over.


gonetz:
Looks good, thanks!

I put all your changes in one commit and corrected C++ code a bit: 41f40ed1409

PR #1466, opened 2017-04-16
Floating point color buffer reader support

fzurita:
This is much faster when using devices with Adreno 500 series GPUs.

I still need to add support in RDRAMtoColorBuffer and TexrectDrawer. But it's good to get it out there early just in case you notice something wrong.

Edit: Nothing to do in TextrectDrawer. Nothing is copied from/to GPU memory directly.

### DIFF
--- a/src/BufferCopy/RDRAMtoColorBuffer.cpp
+++ b/src/BufferCopy/RDRAMtoColorBuffer.cpp
@@ -11,6 +11,7 @@
 #include <Graphics/Context.h>
 #include <Graphics/Parameters.h>
 #include <DisplayWindow.h>
+#include <algorithm>
 
 using namespace graphics;
 
@@ -27,6 +28,8 @@ RDRAMtoColorBuffer & RDRAMtoColorBuffer::get()
 
 void RDRAMtoColorBuffer::init()
 {
+	const FramebufferTextureFormats & fbTexFormats = gfxContext.getFramebufferTextureFormats();
+
 	m_pTexture = textureCache().addFrameBufferTexture(false);
 	m_pTexture->format = G_IM_FMT_RGBA;
 	m_pTexture->clampS = 1;
@@ -38,10 +41,9 @@ void RDRAMtoColorBuffer::init()
 	m_pTexture->mirrorT = 0;
 	m_pTexture->realWidth = 640;
 	m_pTexture->realHeight = 580;
-	m_pTexture->textureBytes = m_pTexture->realWidth * m_pTexture->realHeight * 4;
+	m_pTexture->textureBytes = m_pTexture->realWidth * m_pTexture->realHeight * fbTexFormats.colorFormatBytes;
 	textureCache().addFrameBufferTextureSize(m_pTexture->textureBytes);
 
-	const FramebufferTextureFormats & fbTexFormats = gfxContext.getFramebufferTextureFormats();
 	Context::InitTextureParams initParams;
 	initParams.handle = m_pTexture->name;
 	initParams.width = m_pTexture->realWidth;
@@ -197,16 +199,29 @@ void RDRAMtoColorBuffer::copyFromRDRAM(u32 _address, bool _bCFB)
 
 	const bool bUseAlpha = !_bCFB && m_pCurBuffer->m_changed;
 
+	const FramebufferTextureFormats & fbTexFormats = gfxContext.getFramebufferTextureFormats();
+
 	m_pTexture->width = width;
 	m_pTexture->height = height;
-	const u32 dataSize = width*height * 4;
+	const u32 gpuDataSize = width*height * fbTexFormats.colorFormatBytes;
 
 	PixelBufferBinder<PixelWriteBuffer> binder(m_pbuf.get());
-	u8* ptr = (u8*)m_pbuf->getWriteBuffer(dataSize);
+	u8* ptr = (u8*)m_pbuf->getWriteBuffer(gpuDataSize);
 	if (ptr == nullptr)
 		return;
 
-	u32 * dst = (u32*)ptr;
+	u32 * dst = nullptr;
+	std::unique_ptr<u8[]> dstData;
+
+	//If not using float, the initial coversion will already be correct
+	if (fbTexFormats.colorType == datatype::FLOAT) {
+		const u32 initialDataSize = width*height * 4;
+		dstData = std::unique_ptr<u8[]>(new u8[initialDataSize]);
+		dst = reinterpret_cast<u32*>(dstData.get());
+	} else {
+		dst = reinterpret_cast<u32*>(ptr);
+	}
+
 	bool bCopy;
 	if (m_vecAddress.empty()) {
 		if (m_pCurBuffer->m_size == G_IM_SIZ_16b)
@@ -219,6 +234,20 @@ void RDRAMtoColorBuffer::copyFromRDRAM(u32 _address, bool _bCFB)
 			bCopy = _copyPixelsFromRdram<u16>(address, m_vecAddress, dst, RGBA16ToABGR32, 1, width, height, _bCFB);
 		else
 			bCopy = _copyPixelsFromRdram<u32>(address, m_vecAddress, dst, RGBA32ToABGR32, 0, width, height, _bCFB);
+	}
+
+	//Convert integer format to float
+	if (fbTexFormats.colorType == datatype::FLOAT) {
+		f32* floatData = reinterpret_cast<f32*>(ptr);
+		u8* byteData = dstData.get();
+		const u32 widthPixels = width*4;
+		for (unsigned int heightIndex = 0; heightIndex < height; ++heightIndex) {
+			for (unsigned int widthIndex = 0; widthIndex < widthPixels; ++widthIndex) {
+				u8& src = *(byteData + heightIndex*widthPixels + widthIndex);
+				float& dst = *(floatData + heightIndex*widthPixels + widthIndex);
+				dst = src/255.0f;
+			}
+		}
 	}
 
 	if (bUseAlpha) {
@@ -238,8 +267,6 @@ void RDRAMtoColorBuffer::copyFromRDRAM(u32 _address, bool _bCFB)
 	CombinerInfo::get().setPolygonMode(DrawingState::TexRect);
 	CombinerInfo::get().update();
 	gDP.otherMode.cycleType = cycleType;
-
-	const FramebufferTextureFormats & fbTexFormats = gfxContext.getFramebufferTextureFormats();
 
 	Context::UpdateTextureDataParams updateParams;
 	updateParams.handle = m_pTexture->name;
--- a/src/Graphics/ColorBufferReader.cpp
+++ b/src/Graphics/ColorBufferReader.cpp
@@ -1,0 +1,89 @@
+
+#include "ColorBufferReader.h"
+#include "FramebufferTextureFormats.h"
+#include "Context.h"
+#include "Parameters.h"
+#include <algorithm>
+#include <cstring>
+
+namespace graphics {
+
+	ColorBufferReader::ColorBufferReader(CachedTexture * _pTexture)
+		: m_pTexture(_pTexture)
+	{
+		m_pixelData.resize(m_pTexture->textureBytes);
+		m_tempPixelData.resize(m_pTexture->textureBytes);
+	}
+
+	const u8* ColorBufferReader::_convertFloatTextureBuffer(const u8* _gpuData, u32 _width, u32 _height,
+		u32 _heightOffset, u32 _stride)
+	{
+		std::copy_n(_gpuData, m_tempPixelData.size(), m_tempPixelData.data());
+		u8* pixelDataAlloc = m_pixelData.data();
+		float* pixelData = reinterpret_cast<float*>(m_tempPixelData.data());
+		const u32 colorsPerPixel = 4;
+		const u32 widthPixels = _width * colorsPerPixel;
+		const u32 stridePixels = _stride * colorsPerPixel;
+
+		for (u32 index = 0; index < _height; ++index) {
+			for (u32 widthIndex = 0; widthIndex < widthPixels; ++widthIndex) {
+				u8& dest = *(pixelDataAlloc + index*widthPixels + widthIndex);
+				float& src = *(pixelData + (index+_heightOffset)*stridePixels + widthIndex);
+				dest = static_cast<u8>(src*255.0);
+			}
+		}
+
+		return pixelDataAlloc;
+	}
+
+	const u8* ColorBufferReader::_convertIntegerTextureBuffer(const u8* _gpuData, u32 _width, u32 _height,
+		u32 _heightOffset, u32 _stride)
+	{
+		const u32 colorsPerPixel = 4;
+		const u32 widthBytes = _width * colorsPerPixel;
+		const u32 strideBytes = _stride * colorsPerPixel;
+
+		u8* pixelDataAlloc = m_pixelData.data();
+		for (u32 lnIndex = 0; lnIndex < _height; ++lnIndex) {
+			memcpy(pixelDataAlloc + lnIndex*widthBytes, _gpuData + ((lnIndex+_heightOffset)*strideBytes), widthBytes);
+		}
+
+		return pixelDataAlloc;
+	}
+
+
+	const u8 * ColorBufferReader::readPixels(s32 _x0, s32 _y0, u32 _width, u32 _height, u32 _size, bool _sync)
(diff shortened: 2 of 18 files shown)


fzurita:
Ok, done. Also, when RDRAM to color buffer is enabled, it's not honoring the BufferSwapMode, which makes it much slower in Android.

fzurita:
This could be faster on other devices, like Intel or AMD. I don't have an easy way of testing this though, so I left this as Adreno 500 only.

loganmc10:
This seems very strange to me, is there any documentation from Adreno saying this is the case? Also, ColorBufferReader involves glReadPixels and RDRAM to color buffer does not, so does this affect reads and writes? Aren't most the textures in RGBA8 format and wouldn't they be affected as well?

fzurita:
It's hard to argue with numbers. I'm getting 3x the frame rate on that specific GPU when using FP buffers and making async glReadPixels calls.

This doesn't affect textures, only FBOs that we read and write from.

loganmc10:
Huh ok, I'm on vacation this week but I can try this out on an Intel and Nvidia GPU next week sometime. I guess I just assumed it would be easier for the GPU to deal with integers, but maybe it's converting them to floats anyway.

src/Graphics/ColorBufferReader.cpp
```diff
@@ -0,0 +1,89 @@
...
+		m_tempPixelData.resize(m_pTexture->textureBytes);
+	}
+
+	const u8* ColorBufferReader::_convertFloatTextureBuffer(const u8* _gpuData, u32 _width, u32 _height,
+		u32 _heightOffset, u32 _stride)
+	{
+		std::copy_n(_gpuData, m_tempPixelData.size(), m_tempPixelData.data());
+		u8* pixelDataAlloc = m_pixelData.data();
+		float* pixelData = reinterpret_cast<float*>(m_tempPixelData.data());
+		int colorsPerPixel = 4;
```
gonetz: const u32 colorsPerPixel = 4;

src/Graphics/ColorBufferReader.cpp
```diff
@@ -0,0 +1,89 @@
...
+		int widthPixels = _width * colorsPerPixel;
```
gonetz: const u32 int widthPixels =

src/Graphics/ColorBufferReader.cpp
```diff
@@ -0,0 +1,89 @@
...
+		int stridePixels = _stride * colorsPerPixel;
```
gonetz: const u32 stridePixels =

src/Graphics/ColorBufferReader.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+		for (unsigned int index = 0; index < _height; ++index) {
```
gonetz: for (u32 index = 0;...

src/Graphics/ColorBufferReader.cpp
```diff
@@ -0,0 +1,89 @@
...
+			}
+		}
+
+		return pixelDataAlloc;
+	}
+
+	const u8* ColorBufferReader::_convertIntegerTextureBuffer(const u8* _gpuData, u32 _width, u32 _height,
+		u32 _heightOffset, u32 _stride)
+	{
+		int colorsPerPixel = 4;
```
gonetz: const u32 colorsPerPixel = 4;

src/Graphics/ColorBufferReader.cpp
```diff
@@ -0,0 +1,89 @@
...
+		size_t widthBytes = _width * colorsPerPixel;
```
gonetz: const size_t widthBytes =

src/Graphics/ColorBufferReader.cpp
```diff
@@ -0,0 +1,89 @@
...
+		int strideBytes = _stride * colorsPerPixel;
```
gonetz: const u32 strideBytes =

src/Graphics/ColorBufferReader.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+		u8* pixelDataAlloc = m_pixelData.data();
+		for (unsigned int lnIndex = 0; lnIndex < _height; ++lnIndex) {
```
gonetz: for (u32 lnIndex = 0; ...

src/Graphics/ColorBufferReader.cpp
```diff
@@ -0,0 +1,89 @@
...
+		params.width = _width;
+		params.height = _height;
+		params.sync = _sync;
+
+		if (_size > G_IM_SIZ_8b) {
+			params.colorFormat = fbTexFormat.colorFormat;
+			params.colorType = fbTexFormat.colorType;
+			params.colorFormatBytes = fbTexFormat.colorFormatBytes;
+		}
+		else {
```
gonetz: } else {

src/Graphics/ColorBufferReader.h
```diff
@@ -7,19 +7,50 @@ namespace graphics {
...
+private:
+
+	//Faster float to integer conversion. Taken from here:
+	//http://stackoverflow.com/questions/429632/how-to-speed-up-floating-point-to-integer-number-conversion
+	inline int _float2int( double d )
+	{
+		union Cast
+		{
+			double d;
+			long l;
```
fzurita: It seemed to produce the correct output. But you are probably right, I did try with a regular cast and I didn't see any performance difference.

src/Graphics/ColorBufferReader.cpp
```diff
@@ -0,0 +1,89 @@
...
+	const u8* ColorBufferReader::_convertIntegerTextureBuffer(const u8* _gpuData, u32 _width, u32 _height,
+		u32 _heightOffset, u32 _stride)
+	{
+		int colorsPerPixel = 4;
+		size_t widthBytes = _width * colorsPerPixel;
+		int strideBytes = _stride * colorsPerPixel;
+
+		u8* pixelDataAlloc = m_pixelData.data();
+		for (unsigned int lnIndex = 0; lnIndex < _height; ++lnIndex) {
+			memcpy(pixelDataAlloc + lnIndex*widthBytes, _gpuData + ((lnIndex+_heightOffset)*strideBytes), widthBytes);
```
gonetz: error: ‘memcpy’ was not declared in this scope

src/Graphics/ColorBufferReader.cpp
```diff
@@ -0,0 +1,89 @@
...
+		}
+		else {
+			params.colorFormat = fbTexFormat.monochromeFormat;
+			params.colorType = fbTexFormat.monochromeType;
+			params.colorFormatBytes = fbTexFormat.monochromeFormatBytes;
+		}
+
+		u32 heightOffset = 0;
+		u32 stride = 0;
+		const u8* pixelData = _readPixels(params, heightOffset, stride);
```
fzurita: Whoops, they are supposed to be the 4th and 5th argument to ```_convertFloatTextureBuffer``` and ```_convertIntegerTextureBuffer```

src/Graphics/OpenGLContext/opengl_GLInfo.cpp
```diff
@@ -19,19 +20,22 @@ void GLInfo::init() {
 		glGetIntegerv(GL_MAJOR_VERSION, &majorVersion);
 		glGetIntegerv(GL_MINOR_VERSION, &minorVersion);
 	}
-	LOG(LOG_VERBOSE, "%s major version: %d\n", isGLESX ? "OpenGL ES" : "OpenGL", majorVersion);
-	LOG(LOG_VERBOSE, "%s minor version: %d\n", isGLESX ? "OpenGL ES" : "OpenGL", minorVersion);
+	LOG(LOG_MINIMAL, "%s major version: %d\n", isGLESX ? "OpenGL ES" : "OpenGL", majorVersion);
```
fzurita: I'll revert this change.

src/Graphics/ColorBufferReader.h
```diff
@@ -7,19 +7,50 @@ namespace graphics {
...
-		m_pixelData.resize(m_pTexture->textureBytes);
-	}
-	virtual ~ColorBufferReader() {}
+	ColorBufferReader(CachedTexture * _pTexture);
+	virtual ~ColorBufferReader() = default;
 
-	virtual u8 * readPixels(s32 _x0, s32 _y0, u32 _width, u32 _height, u32 _size, bool _sync) = 0;
+	virtual const u8 * readPixels(s32 _x0, s32 _y0, u32 _width, u32 _height, u32 _size, bool _sync);
 	virtual void cleanUp() = 0;
+	const u8* _convertFloatTextureBuffer(const u8* _gpuData, u32 _width, u32 _height, u32 _heightOffset, u32 _stride);
```
fzurita: This was intended to be private. I'll move to private section.

fzurita:
Ok, pull request has been updated.

src/BufferCopy/RDRAMtoColorBuffer.cpp
```diff
@@ -197,16 +199,31 @@ void RDRAMtoColorBuffer::copyFromRDRAM(u32 _address, bool _bCFB)
...
 
+	const FramebufferTextureFormats & fbTexFormats = gfxContext.getFramebufferTextureFormats();
+
 	m_pTexture->width = width;
 	m_pTexture->height = height;
-	const u32 dataSize = width*height * 4;
+	const u32 gpuDataSize = width*height * fbTexFormats.colorFormatBytes;
+	const u32 initialDataSize = width*height * 4;
+
+	std::unique_ptr<u8[]> dstData = std::unique_ptr<u8[]>(new u8[initialDataSize]);
```
gonetz: std::unique_ptr<u8[]> dstData(new u8[initialDataSize]);

src/BufferCopy/RDRAMtoColorBuffer.cpp
```diff
@@ -221,6 +238,22 @@ void RDRAMtoColorBuffer::copyFromRDRAM(u32 _address, bool _bCFB)
...
+		for (unsigned int heightIndex = 0; heightIndex < height; ++heightIndex) {
+			for (unsigned int widthIndex = 0; widthIndex < widthPixels; ++widthIndex) {
+				u8& src = *(byteData + heightIndex*widthPixels + widthIndex);
+				float& dst = *(floatData + heightIndex*widthPixels + widthIndex);
+				dst = src/255.0f;
+			}
+		}
+	}
+
+	std::copy_n(convertedDataPtr, gpuDataSize, ptr);
```
gonetz: I don't like that we do redundant data copy in case fbTexFormats.colorType != datatype::FLOAT

src/Graphics/ColorBufferReader.cpp
```diff
@@ -0,0 +1,89 @@
...
+		m_tempPixelData.resize(m_pTexture->textureBytes);
+	}
+
+	const u8* ColorBufferReader::_convertFloatTextureBuffer(const u8* _gpuData, u32 _width, u32 _height,
+		u32 _heightOffset, u32 _stride)
+	{
+		std::copy_n(_gpuData, m_tempPixelData.size(), m_tempPixelData.data());
+		u8* pixelDataAlloc = m_pixelData.data();
+		float* pixelData = reinterpret_cast<float*>(m_tempPixelData.data());
+		const int colorsPerPixel = 4;
```
gonetz: why you use signed integer for obviously unsigned values?  colorsPerPixel , widthPixels, stridePixels can't be negative

src/Graphics/ColorBufferReader.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+		return pixelDataAlloc;
+	}
+
+	const u8* ColorBufferReader::_convertIntegerTextureBuffer(const u8* _gpuData, u32 _width, u32 _height,
+		u32 _heightOffset, u32 _stride)
+	{
+		const int colorsPerPixel = 4;
+		const size_t widthBytes = _width * colorsPerPixel;
+		const int strideBytes = _stride * colorsPerPixel;
```
gonetz: why you use signed integer for obviously unsigned values? colorsPerPixel , widthBytes , strideBytes can't be negative

src/Graphics/ColorBufferReader.cpp
gonetz: man memcpy says:` #include <string.h>. `
`<cstring>` seems to work also

src/Graphics/ColorBufferReader.cpp
fzurita: I wasn't thinking about it. It's fixable.

src/Graphics/ColorBufferReader.cpp
fzurita: What I'm finding, at least for mobile, that it seems faster to do one big copy instead of many small copies from GPU memory. I didn't test the performance difference for writing to GPU memory, but I assumed that it would be similar.

What do you think?

src/BufferCopy/RDRAMtoColorBuffer.cpp
fzurita: What I'm finding, at least for mobile, is that it seems faster to do one big copy instead of many small copies from GPU memory. I didn't test the performance difference for writing to GPU memory, but I assumed that it would be similar. That's why I did that here.

What do you think?

fzurita:
Pull request updated. I left the extra copying of the buffer there for now.

fzurita:
I did a quick test:
* With floating point format, doing the extra copy and then doing one big copy to GPU memory was was about 1-2% faster in the same spot in Mario 64.
* With integer format, doing the extra copy and then doing one big copy to GPU memory was really about the same performance. I couldn't see a performance difference.

Given that, I think I will remove the extra copies.

fzurita:
Ok, pull request updated. Extra copy has been removing for floating point and integer versions.

On a side note, async color buffer to RDRAM seems to cause random corruption when using RDRAM to color buffer.

loganmc10:
I just tested this on my Nvidia GPU laptop and I didn't notice any performance difference (for worse or for better). I tested by changing:
```
colorInternalFormat = GL_RGBA8;
colorFormat = GL_RGBA;
colorType = GL_UNSIGNED_BYTE;
colorFormatBytes = 4;
```

To:
```
colorInternalFormat = GL_RGBA32F;
colorFormat = GL_RGBA;
colorType = GL_FLOAT;
colorFormatBytes = 16;
```

in FramebufferTextureFormatsOpenGL.

I tested Mario Kart, Luigi's Raceway, with the speed limiter disabled

fzurita:
Did you with Async color buffer to RDRAM? I didn't see any performance differences in sync mode, only async mode.

loganmc10:
I tested with async, I'll test with Intel graphics (Linux) tonight, I don't have a Windows + Intel machine to test.

Everything I read seems to point to the fact that OpenGL converts the values to floating point, so it's not unreasonable to assume that this might be worthwhile to have as a default setting.

It seems that GL_RGBA32F isn't available on GLES2, but is available in GLES3.0+ and OpenGL 3.0+, but it's hard to tell from the documentation when a color format becomes part of the standard.

fzurita:
I tried a few other GLES 3.x devices and they don't seem to work with this setting. The only ones that worked correctly was a device based on Adreno 530 and the Shield TV.

Nexus player and a device based on Adreno 330 didn't work with this setting due to various GL errors. I have a device based on Adreno 430 to test on. I wonder if that one will work with this. I also have a device with Mali graphics.

loganmc10:
I assume they probably require this extension:

https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_color_buffer_float.txt

It seems that that extension became core in GLES 3.2, so the device would either need EXT_color_buffer_float or GLES 3.2+ (if I understand the extension correctly).

I'm not sure about OpenGL, like I said, it's somewhat difficult to figure out when these core formats enter into the standard.

loganmc10:
My Googling tells me (http://opengles.gpuinfo.org/gles_listreports.php?extension=GL_EXT_color_buffer_float) that the Adreno 330 should support that extension though, so maybe something else is up

fzurita:
It would probably require investigating what these GL errors are. Maybe there is something wrong in the code somewhere.

gonetz:
Merged: b31bb1fe9e04d37

loganmc10:
I just tested this on Intel + Linux and the current master I get ~230 VI/S in Mario Kart, with floating point format I got ~200 VI/S, so I guess it's the kind of thing that would need to be tested on each GPU to determine if it benefits or not.

fzurita:
I would be really interested to know the performance with Intel drivers instead of Mesa drivers. They seem to behave differently with glReadPixels.

PR #1482, opened 2017-05-14
UI: Fullscreen resolution list displays aspect ratio

Aquatakat:
This is a feature requested a long time ago by @oddMLan. I'll let you decide if it's a good addition or not.

image

### DIFF
--- a/src/GLideNUI/fullscreenresolutions_windows.cpp
+++ b/src/GLideNUI/fullscreenresolutions_windows.cpp
@@ -69,6 +69,21 @@ void fillFullscreenResolutionsList(QStringList & _listResolutions, int & _resolu
 	fullscreen.numRefreshRates = 0;
 	_resolutionIdx = 0;
 
+	static
+	struct
+	{
+		unsigned short x, y;
+		const char *description;
+	} ratios[] = {
+		{ 3,  2, "3:2" },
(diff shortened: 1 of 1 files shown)


gonetz:
Looks nice, thanks!
Merged: a4d0796e737

Aquatakat:
@gonetz She also suggested a while ago turning the windowed drop down list into a combo box where one could simply type the resolution if they don't see the one they want in the list. I'm comfortable enough with the code and Qt to make that work if you think it's not a bad idea.

oddMLan:
>She

I'm a cis dude though :)
(not that I mind , I understand the confusion due to my name). But Kimberly is also a name given to males.

Aquatakat:
Whoops, sorry. I've never seen Kimberly as a male name, but there's always a big world out there.

gonetz:
Regarding "turning the windowed drop down list into a combo box where one could simply type the resolution if they don't see the one they want in the list" : if it will be more convenient than current solution for custom windowed resolutions, I don't mind.

gonetz:
>But Kimberly is also a name given to males.

Reminds me of Jayne Cobb :)
http://www.imdb.com/character/ch0003807/mediaviewer/rm650156288

Aquatakat:
A show to watch again.

PR #1483, opened 2017-05-15
UI: Custom windowed resolution set via combo box

Aquatakat:
Another request from @oddMLan. Again, I'll let you decide if this is a good change. You can type directly in the combo box values from 300x200 to 7999x3999.

image


oddMLan:
Can't compile this yet, but is the aspect ratio supposed to show here akin #1482 or not?
Btw nice job!

Aquatakat:
No, I didn't add that feature, mostly because I didn't see an easy signal that Qt fires when this control loses focus. (Also all the resolutions Gonetz put in that list are 4:3 anyway).

oddMLan:
D'oh. Right. I thought there were some different that 4:3 because of the last two resolutions in the list. They can't both be 4:3

Aquatakat:
Oh yeah. ... Why is 1600 x 1024 in there in the first place? I should probably get rid of it.

tony971:
As someone with a 3000x2000 monitor, I appreciate this.

gonetz:
Good idea overall. I see two issues:
* 'custom' element of the list becomes useless
* cursor appears in edit area after I select a resolution from the list. It is confusing, as if the control still waits something from me.

Aquatakat:
> 'custom' element of the list becomes useless
 When that option is selected the combo box clears indicating something can be typed.

> cursor appears in edit area after I select a resolution from the list. It is confusing, as if the control still waits something from me.
Maybe I need to do some more work.

Aquatakat:
> cursor appears in edit area after I select a resolution from the list. It is confusing, as if the control still waits something from me.

Alright, I think I fixed it.

gonetz:
Yes, cursor problem is fixed, thanks!
I found another one. Scenario:
* Open combo-box list, select 'Custom', enter custom resolution in **empty** edit box, press tab or click to other control. Ok.
* Open combo-box list again, select 'Custom': edit box contains 'Custom'. You need to erase it before enter new resolution.

Could you fix it?

Aquatakat:
> Could you fix it?

Yes, I changed the way it clears to check for the Custom text now. It works in the other languages also.

gonetz:
Thanks! Merged: 716bcee8962a0b7

PR #1493, opened 2017-05-25
arm neon: merge InverseTransformVectorNormalize functions and add gSPLightVertex functions 

gizmo98:
-merge InverseTransformVectorNormalize functions into one N element function:
-   -O3: 7 lights --> 0,23x time of C function 
-   -O2: 7 lights --> 0,19x time of C function 

-add gSPLightVertex and gSPLightVertex4 neon versions.
-   -O3: 0,71x time of C function 
-   -O2: 0,36x time of C function

### DIFF
--- a/src/3DMath.cpp
+++ b/src/3DMath.cpp
@@ -88,16 +88,10 @@ End:
 #endif // WIN32_ASM
 }
 
-void InverseTransformVectorNormalize2x(float src[2][3], float dst[2][3], float mtx[4][4] )
+void InverseTransformVectorNormalizeN(float src[][3], float dst[][3], float mtx[4][4], s32 count)
 {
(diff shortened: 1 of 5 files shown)


gonetz:
Overall looks good. I have minor remark:
You declared parameter 'count' for InverseTransformVectorNormalizeN as s32.
First, why use signed type for parameter, which can't be negative?
Then, count is **s32** in 3DMath.cpp, but **int** in 3DMathNeon.cpp. I know, s32 and int are the same, but such lack of uniformity is not good. I suggest to use u32 for count everywhere.

gonetz:
Merged

gizmo98:
Oh. Already merged! I just wanted to squash my last commit first. Sorry for late reply.

Issue #1498, opened 2017-05-29
Resident Evil 2: Missing video on Windows x64 Debug build


loganmc10:
This is a really random issue.

I am using CMake to compile builds on Windows and Linux. On Linux, everything works normally. On Windows, if I compile using ```-DCMAKE_BUILD_TYPE=Debug```, on x86 (32 bit), it works fine. On x64, I get a blank screen starting when it should say "Capcom".

The only similar issue I can find is this:
https://github.com/gonetz/GLideN64/issues/1266

Can anyone/@gonetz think of why this may be? What could be the difference with the debug builds that would affect the video output?

One other similar issue (although this happens on the Release build as well): Pokemon Puzzle League, on Windows x64 it flickers, but x86 works fine, and Linux x86/x64 works fine.

This was all tested with mupen64plus.

loganmc10:
The flickering in Pokemon Puzzle League happens on Android (mupen64plus-ae) as well

loganmc10:
Actually this seems like it may be related to the RSP plugin (cxd4), if I play around with the compile flags for that plugin, it brings back the video output (although I can't get it perfect yet), so I'll close this for now

loganmc10:
I just tested Glide64 on Android, and the video in Pokemon Puzzle League and Resident Evil 2 is correct, so there is some issue with GLideN64

loganmc10:
I am really starting to believe this is happening because of ```EnableCopyColorFromRDRAM```. I have the same problem in Dr. Mario

Here is the 32-bit build:
![1](https://cloud.githubusercontent.com/assets/848146/26590742/a9259abe-4518-11e7-972c-0eef6ba5a74a.png)

Here is the 64-bit build:
![2](https://cloud.githubusercontent.com/assets/848146/26590750/af9e7082-4518-11e7-8938-b73dbadfaebc.png)

loganmc10:
Indeed, disabled ```EnableCopyColorFromRDRAM``` fixed the distorted image on Dr. Mario 64 with the 64 bit build (but the pills are gone of course)

loganmc10:
Compiling with Clang solves this problem, everything works properly. I'll try and see why this is also happening on Android (which uses Clang, but probably has a CFLAG set that is interfering with this)

Still strange though, there is likely something wrong with the code somewhere

loganmc10:
I'm running ubsan on my Raspberry Pi and posting the findings here so I can look at them tomorrow:
```
[...]

Issue #555, opened 2015-05-25
CopyFromRDRAM loses colors


purplemarshmallow:
Sometimes colors are missing. The screecshot is from Namco Museum 64 but it can happen in every game that uses cfb. It seems to happen randomly. Switching window/fullscreen can make colors disappear but it can also bring colors back.
edit: happens with Project64 and mupen64plus
![gliden64_namcomuseum64_000](https://cloud.githubusercontent.com/assets/7283660/7796061/b7ac5aa4-02db-11e5-9442-a6e044982ce8.jpg)

purplemarshmallow:
There's a bug in CopyFromRDRAM, everything that relies on CopyFromRDRAM can have missing colors (pills in Dr. Mario, scoreboards in Mario Tennis, ...)

gonetz:
Is it still actual? I can't reproduce it.

purplemarshmallow:
Yes it happens with this week's master build. But it's hard to reproduce because it's completely random and hard to debug :/

weinerschnitzel:
I cant reproduce the colors disappearing in the pills in Dr. Mario using the current build and Project64 2.3. However there are some backgrounds that do not persist when switching fullscreen and back. Is this the behavior of the same bug?

gonetz:
> However there are some backgrounds that do not persist when switching fullscreen and back. Is this the behavior of the same bug?

No. Backgrounds in Dr. Mario stored in frame buffer objects, which destroyed on window mode change.
This bug is about rare case, when format of image in RDRAM wrongly detected as monochrome one.

fzurita:
I wonder if this is fixed now.

purplemarshmallow:
fixed

PR #1503, opened 2017-05-31
zero fill gDPTile in RDRAMtoColorBuffer

loganmc10:
Fixes https://github.com/gonetz/GLideN64/issues/1498

Creating ```gDPTile tile0;``` without zeroing it out allows the structure to be filled with garbage. It seems that some compilers/platforms zero it out by default, but you can't expect that.


loganmc10:
This may also fix https://github.com/gonetz/GLideN64/issues/555, I suspect it was probably the same issue

loganmc10:
I'm not 100% sure that ```gDPTile tile0 = {0};``` is valid in every compiler. I tested GCC and it worked, but there is a potential that some compilers don't like ```= {0}`` (from what I read, I could be wrong, I assume if it works in MSVC then we are good)

gonetz:
It compiles with MSVS, thanks!
Merged: 63e438f22c8c4

PR #1508, opened 2017-06-14
UI: Language drop down displays languages in foreign languages

Aquatakat:
Having them display always in English doesn't make sense for someone looking for their language. A simple change:

image


paradadf:
I'd take out the "de España" and "do Brasil". Those aren't really necessary.

Aquatakat:
The names are automatically generated by Qt based on the names of the filenames of the translation files, so "do Brasil" comes from "gliden64_pt_BR.ts". I'm not sure where "de España" comes from. I could probably strip off everything after the first space but if there's legitimate future dialects of the same language, like English (United Kingdom), that could cause a problem.

gonetz:
Merged: 08366f5c0962

PR #1514, opened 2017-06-25
UI: Debug tab cleanup

Aquatakat:
Small changes so text is consistent with other parts of the UI. I've also prevented a future bug where the tab wouldn't be removed if the tab text has been localised.


gonetz:
Thanks!

PR #1522, opened 2017-07-07
Addendum to readme

Aquatakat:
The front page of this Github should probably mention the WIP builds because those builds have far better game compatibility. Also I noticed a few issues submitted fixed in the newer builds.


gonetz:
Thanks! Merged: d4971b48b

PR #1530, opened 2017-07-20
CMake: fix 'UNIX build requires MUPENPLUSAPI' message

ligfx:


loganmc10:
Won't this break no Unix machines that want to set MUPENPLUSAPI? You're changing the initial value of the option to the value of "UNIX". What if I'm on Windows?

loganmc10:
You just need to build like ```cmake -DMUPENPLUSAPI=On /source/dir```

ligfx:
It already defaults to `MUPENPLUSAPI=OFF` on all platforms, right? So on Windows, there will be no behavior change.

gonetz:
Logan is right: you must use -DMUPENPLUSAPI=On parameter. Readme.txt explains it.

cmake was planned to use for generation of any kind of project files, including Visual Studio ones. It was not completed, and currently cmake used only for mupen64plus builds. Nevertheless, your solution is wrong. If you want to make mupen64plus build default, you need to remove MUPENPLUSAPI cmake parameter and add new optional ZILMARSPEC one, which will disable MUPENPLUSAPI define. I would not do it: 
- MUPENPLUSAPI  may already be used in automatic scripts. 
- it is not hard to add this parameter to cmake call.

ligfx:
I'm not sure I understand what you guys are saying. Is the message fix incorrect? Can UNIX build with MUPENPLUSAPI=OFF? I tried it on macOS, and it failed trying to pull in some Windows-only headers. This change would not prevent specifying MUPENPLUSAPI explicitly on any platform, and only default to ON on Unix to make building a little easier.

gonetz:
> Can UNIX build with MUPENPLUSAPI=OFF?

No, currently it can't

>  default to ON on Unix to make building a little easier

and little more confusing on Windows.  You need to call cmake only once, so it is not a problem to add one more parameter. Currently cmake -DMUPENPLUSAPI=On /source/dir can be used for any platform. I don't see much sense in unix-specific optimization.

ligfx:
Okay, I see. I've removed the commit that changes the default on Unix, leaving only the message fix. Does that look correct?

src/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 2.6)
 
 option(EGL "Set to ON if targeting an EGL device" ${EGL})
 option(PANDORA "Set to ON if targeting an OpenPandora" ${PANDORA})
-option(MUPENPLUSAPI "Set to ON for Mupen64Plus plugin" ${MUPENPLUSAPI})
+option(MUPENPLUSAPI "Set to ON for Mupen64Plus plugin" ${UNIX})
```
gonetz: Why? It was correct.

ligfx: Huh, GitHub doesn't seem to have updated what's it's showing here, even though the branch was updated. Can you re-open the pull request, or should I make another one?

gonetz: I don't see reopen button. Please make new PR

Issue #1539, opened 2017-07-28
Paper Mario: Red Station Crashing


FraganciaGH:
GlideN64 2.0 crashes when you enter the pitch black room leading up to the chapter boss fight and the room where the Big Lantern ghost is located at the Red Station inside the Shy Guys Toy Box.

This also applies to Chapter 5 when you have to rescue the Yoshi children where one Yoshi is located in pitch black area only accessible through a green pipe, the game will instantly crash as soon you enter the pipe.

![pm_rs_crash_p1](https://user-images.githubusercontent.com/14110425/28721893-102a052a-73b2-11e7-849f-6baa3ffc4997.png)
![pm_rs_crash_p2](https://user-images.githubusercontent.com/14110425/28721902-13af8f76-73b2-11e7-9004-5f754b485251.png)

Glide64 for PJ64 2.1.2.202 handles these pitch black areas pretty well with no noticeable problems.

Youtube video showing Glide64 in action.
https://www.youtube.com/watch?v=qkVEGuvSfvU

loganmc10:
You should test/post bug reports using the latest WIP build

Pop000100:
https://github.com/gonetz/GLideN64/issues/1364 here is the thread for above builds

FraganciaGH:
@loganmc10 
Using the latest build didn't make any difference, it still crashes.
Used Revision 5691fcb.

loganmc10:
Can you post a save game or save state so that I can get to the problem quickly? I can try to figure out why it's crashing

FraganciaGH:
@loganmc10 
Here you go.
[Paper Mario (U).pj.zip](https://github.com/gonetz/GLideN64/files/1184457/Paper.Mario.U.pj.zip)

loganmc10:
Is this also with the multiplayer ROM hack?

FraganciaGH:
@loganmc10 
Happens on both ROMs, I've tried it multiple times with GlideN64 2.0 and the WIP build
It crashed everytime.

loganmc10:
You're gonna have to help me out, I don't know Paper Mario too well. That save file puts me in front of a tree, where do I need to go from there to make the crash happen?

FraganciaGH:
@loganmc10 
Here's a youtube video showing you where to go
https://www.youtube.com/watch?v=ymjKiyM4TN8

loganmc10:
[...]

Issue #1250, opened 2016-11-21
[Project64] Banjo-Tooie Groggy Crash


vgf89:
It seems my issue is identical to this guy's: https://github.com/gonetz/GLideN64/issues/968, except that I'm using Project 64 2.3. I've tested with the 1.2 public release of GLideN64 and the latest WIP build from https://github.com/gonetz/GLideN64/issues/885.

This crash happens regardless of recompiler or interpreter setting in PJ64. This crash doesn't happen with Glide64.

EDIT: I've attached the save state. Just walk forward and behold the crash
[Banjo-Tooie (U).pj.zip](https://github.com/gonetz/GLideN64/files/604945/Banjo-Tooie.U.pj.zip)

Frank-74:
Using latest Project64 github master, and latest GLideN64 WIP.  I've loaded that save state 5 times and gone into the next area with no crash.

I renamed the save to "Banjo-Tooie (U).pj.zip", the link adds a "." after Tooie instead of a space.

gonetz:
Be sure that you don't have any cheats enabled.

olivieryuyu:
can it be closed?

I have zero issue as well.

gonetz:
Yes, we added a workaround for this issue.

PR #1547, opened 2017-08-12
Fix memory corruption in Banjo-Tooie

fzurita:
This fixes memory corruption in Banjo-Tooie in the case of the attached save state for mupen64plus

[grogged.zip](https://github.com/gonetz/GLideN64/files/1219533/grogged.zip)

Just walk forward to reproduce the issue.

### DIFF
--- a/src/Graphics/ColorBufferReader.cpp
+++ b/src/Graphics/ColorBufferReader.cpp
@@ -18,17 +18,21 @@ namespace graphics {
 	const u8* ColorBufferReader::_convertFloatTextureBuffer(const u8* _gpuData, u32 _width, u32 _height,
 		u32 _heightOffset, u32 _stride)
 	{
-		std::copy_n(_gpuData, m_tempPixelData.size(), m_tempPixelData.data());
+		int bytesToCopy = m_pTexture->realWidth * _height * 16;
+		std::copy_n(_gpuData, bytesToCopy, m_tempPixelData.data());
 		u8* pixelDataAlloc = m_pixelData.data();
 		float* pixelData = reinterpret_cast<float*>(m_tempPixelData.data());
 		const u32 colorsPerPixel = 4;
 		const u32 widthPixels = _width * colorsPerPixel;
 		const u32 stridePixels = _stride * colorsPerPixel;
+		
+		if (_height * widthPixels  > m_pixelData.size())
+			_height = m_pixelData.size() / widthPixels;
 
-		for (u32 index = 0; index < _height; ++index) {
+		for (u32 heightIndex = 0; heightIndex < _height; ++heightIndex) {
 			for (u32 widthIndex = 0; widthIndex < widthPixels; ++widthIndex) {
-				u8& dest = *(pixelDataAlloc + index*widthPixels + widthIndex);
-				float& src = *(pixelData + (index+_heightOffset)*stridePixels + widthIndex);
+				u8& dest = *(pixelDataAlloc + heightIndex*widthPixels + widthIndex);
+				float& src = *(pixelData + (heightIndex+_heightOffset)*stridePixels + widthIndex);
 				dest = static_cast<u8>(src*255.0);
 			}
 		}
@@ -37,15 +41,18 @@ namespace graphics {
 	}
 
 	const u8* ColorBufferReader::_convertIntegerTextureBuffer(const u8* _gpuData, u32 _width, u32 _height,
-		u32 _heightOffset, u32 _stride)
+		u32 _heightOffset, u32 _stride, u32 _colorsPerPixel)
 	{
-		const u32 colorsPerPixel = 4;
-		const u32 widthBytes = _width * colorsPerPixel;
-		const u32 strideBytes = _stride * colorsPerPixel;
+		const u32 widthBytes = _width * _colorsPerPixel;
+		const u32 strideBytes = _stride * _colorsPerPixel;
 
 		u8* pixelDataAlloc = m_pixelData.data();
-		for (u32 lnIndex = 0; lnIndex < _height; ++lnIndex) {
-			memcpy(pixelDataAlloc + lnIndex*widthBytes, _gpuData + ((lnIndex+_heightOffset)*strideBytes), widthBytes);
+		
+		if (_height * widthBytes  > m_pixelData.size())
+			_height = m_pixelData.size() / widthBytes;
+
+		for (u32 index = 0; index < _height; ++index) {
+			memcpy(pixelDataAlloc + index * widthBytes, _gpuData + ((index + _heightOffset) * strideBytes), widthBytes);
 		}
 
 		return pixelDataAlloc;
@@ -80,10 +87,11 @@ namespace graphics {
 		if (pixelData == nullptr)
 			return nullptr;
 
-		if(params.colorType == datatype::FLOAT) {
+		if(params.colorType == datatype::FLOAT && _size > G_IM_SIZ_8b) {
 			return _convertFloatTextureBuffer(pixelData, params.width, params.height, heightOffset, stride);
 		} else {
-			return _convertIntegerTextureBuffer(pixelData, params.width, params.height, heightOffset, stride);
+			return _convertIntegerTextureBuffer(pixelData, params.width, params.height, heightOffset, stride,
+												params.colorFormatBytes);
 		}
 	}
 }
--- a/src/gDP.cpp
+++ b/src/gDP.cpp
@@ -626,10 +626,14 @@ void gDPLoadTLUT( u32 tile, u32 uls, u32 ult, u32 lrs, u32 lrt )
 		DebugMsg(DEBUG_NORMAL | DEBUG_ERROR, "gDPLoadTLUT wrong tile tmem addr: tile[%d].tmem=%04x;\n", tile, gDP.tiles[tile].tmem);
 		return;
 	}
-	const u16 count = (u16)((gDP.tiles[tile].lrs - gDP.tiles[tile].uls + 1) * (gDP.tiles[tile].lrt - gDP.tiles[tile].ult + 1));
+	u16 count = (u16)((gDP.tiles[tile].lrs - gDP.tiles[tile].uls + 1) * (gDP.tiles[tile].lrt - gDP.tiles[tile].ult + 1));
 	u32 address = gDP.textureImage.address + gDP.tiles[tile].ult * gDP.textureImage.bpl + (gDP.tiles[tile].uls << gDP.textureImage.size >> 1);
 	u16 pal = (u16)((gDP.tiles[tile].tmem - 256) >> 4);
 	u16 *dest = (u16*)&TMEM[gDP.tiles[tile].tmem];
+	
+	// Workaround for possible game/emulator bug (see bug #1250)
+	if (pal != 0)
+		count = 16;
 
 	int i = 0;
 	while (i < count) {
--- a/src/BufferCopy/ColorBufferToRDRAM.cpp
+++ b/src/BufferCopy/ColorBufferToRDRAM.cpp
@@ -152,7 +152,7 @@ bool ColorBufferToRDRAM::_prepareCopy(u32 _startAddress)
 
 	if(m_pTexture == nullptr ||
 		m_pTexture->realWidth != _getRealWidth(pBuffer->m_width) ||
-		m_pTexture->realHeight != VI_GetMaxBufferHeight(m_lastBufferWidth))
+		m_pTexture->realHeight != VI_GetMaxBufferHeight(_getRealWidth(pBuffer->m_width)))
 	{
 		_destroyFBTexure();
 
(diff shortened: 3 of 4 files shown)


fzurita:
I also included a fix for possible memory corruption when reading G_IM_SIZ_8b images during buffer copies.

src/gDP.cpp
```diff
@@ -632,7 +632,7 @@ void gDPLoadTLUT( u32 tile, u32 uls, u32 ult, u32 lrs, u32 lrt )
 	u16 *dest = (u16*)&TMEM[gDP.tiles[tile].tmem];
 
 	int i = 0;
-	while (i < count) {
+	while (i < count && pal < 16) {
```
fzurita: Also, I'm not entirely sure what correct behavior should be. Should we wrap around instead of exiting the loop?

gonetz: This is impossible situation. pal can't be >= 16.
Palette has either 16 or 256 elements (count).
For 256 elements palette initial pal index **must be** 0.
Or if pal is not zero, count **must be** 16
If it is not so, something bad happened before. This fix masks the real problem. Speaking of which: I have no problems with your save on Windows. I tried to load it on Ubuntu and first got that impossible situation and then, after memory corruption, got assert in F3DEX2.cpp, line 68. I suppose this is an issue in mupen64plus core: such problem can happen when at some moment gDPSetTile is called, but corresponding gDPSetTileSize does not, thus palette count calculated incorrectly.

I can't check, does the older mupen64plus core has this issue, because your save can't be loaded with older core. Most likely we have to add a workaround for  this problem on plugin's side. Your fix works, but it is wrong, because it spoils palettes with index above pal. I suggest to add clearly visible workaround, like this:
```
	u16 count = (u16)((gDP.tiles[tile].lrs - gDP.tiles[tile].uls + 1) * (gDP.tiles[tile].lrt - gDP.tiles[tile].ult + 1));
	u32 address = gDP.textureImage.address + gDP.tiles[tile].ult * gDP.textureImage.bpl + (gDP.tiles[tile].uls << gDP.textureImage.size >> 1);
	u16 pal = (u16)((gDP.tiles[tile].tmem - 256) >> 4);
	u16 *dest = (u16*)&TMEM[gDP.tiles[tile].tmem];
#if defined(MUPENPLUSAPI) && !defined(OS_WINDOWS)
    // Workaround for mupen64plus error
    if (pal != 0)
        count = 16;
#endif
```

src/Graphics/ColorBufferReader.cpp
```diff
@@ -18,17 +18,18 @@ namespace graphics {
...
+		int bytesToCopy = m_pTexture->realWidth * _height * 16;
+		std::copy_n(_gpuData, bytesToCopy, m_tempPixelData.data());
 		u8* pixelDataAlloc = m_pixelData.data();
 		float* pixelData = reinterpret_cast<float*>(m_tempPixelData.data());
 		const u32 colorsPerPixel = 4;
 		const u32 widthPixels = _width * colorsPerPixel;
 		const u32 stridePixels = _stride * colorsPerPixel;
 
-		for (u32 index = 0; index < _height; ++index) {
+		for (u32 heightIndex = 0; heightIndex < _height && (heightIndex * widthPixels + widthPixels < m_pixelData.size()); ++heightIndex) {
```
gonetz: If  (heightIndex * widthPixels + widthPixels >= m_pixelData.size()) really possible, it is better to reduce _height instead of that ugly check:
```
if (_height * widthPixels  > m_pixelData.size())
  _height = m_pixelData.size() / widthPixels;
```

src/Graphics/ColorBufferReader.cpp
```diff
@@ -37,15 +38,15 @@ namespace graphics {
...
-		const u32 widthBytes = _width * colorsPerPixel;
-		const u32 strideBytes = _stride * colorsPerPixel;
+		const u32 widthBytes = _width * _colorsPerPixel;
+		const u32 strideBytes = _stride * _colorsPerPixel;
 
 		u8* pixelDataAlloc = m_pixelData.data();
-		for (u32 lnIndex = 0; lnIndex < _height; ++lnIndex) {
-			memcpy(pixelDataAlloc + lnIndex*widthBytes, _gpuData + ((lnIndex+_heightOffset)*strideBytes), widthBytes);
+
+		for (u32 index = 0; index < _height && (index * widthBytes + widthBytes < m_pixelData.size()); ++index) {
```
gonetz: Same as above: reduce _height to avoid out of bounds writes.

src/gDP.cpp
loganmc10: @fzurita did you create that save state? Didn't retroben first report this issue? Is it possible that some cheat has corrupted the memory?

fzurita: It was created by retroben.  I don't think a cheat should cause corruption in the display list itself. I guess @retrobenny could always create an in-game save and get to the same spot from the beginning to if the issue happens again.

loganmc10: Whatever the issue might be, I don't think it's specific to mupen64plus, another person reported the same issue with PJ64: https://github.com/gonetz/GLideN64/issues/1250

fzurita: @gonetz Also, my fix is intended to protect memory. If pal can really be greater than 16 then paletteCRC16 must be enlarged to allow for a bigger size.

gonetz: > If pal can really be greater than 16

It **must not ** be greater than 16. By N64 design.

gonetz: > another person reported the same issue with PJ64

another person could use cheats as well.

For me the problem is platform-specific.

loganmc10: > For me the problem is platform-specific.

Just because it didn't crash doesn't mean the problem doesn't exist. When dealing with out-of-bounds memory issues each computer might behave a bit differently. You might have gotten lucky on your Windows box, but someone else might not. There isn't really any platform specific emulation code in mupen64plus or GLideN64

loganmc10: Take this issue for instance (Paper Mario crash):
https://github.com/gonetz/GLideN64/issues/1539

Same thing (bogus pointer this time), it crashed in Linux, but not Windows.

Anyway I'm not saying there is necessarily something wrong with GLideN64, it could be a bad savestate corrupted by a cheat, I just don't think the issue is specific to mupen64plus or Linux.

fzurita: @gonetz Maybe instead of performing this check ```pal < 16``` we can do this instead?

```u16 pal = (u16)(((gDP.tiles[tile].tmem - 256) >> 4) & 0xF);```

I just added a 0xF to mask the value generated from the tmem parameter.

Comment (gonetz) [reply src/gDP.cpp]: You did not read carefully my explanation of the problem. The problem is not with the pallette index pal, but with tile size, which defines pallette size  (count). If pal is not zero, count **must be** 16. My fix insures that. I agree that it is too strict to apply that fix only for mupen64plus and Linux. This condition must always be satisfied. Comment, why it is added is necessary anyway.

File: src/Graphics/ColorBufferReader.cpp (hunk 3 above)
Comment (retrobenny) [reply src/Graphics/ColorBufferReader.cpp]: If it matters still,I went through a fresh file before on mupen Android and still encountered the issue with the entire process being without any cheats. I thought that I mentioned this already to @fzurita though? It was a native save too,no save states loaded in that lengthy test. I went out of my own way to replay a cleanly made save up to that point just to prove its a normal issue of the emulation or whatever part is actually triggering it.

Comment (fzurita) [reply src/Graphics/ColorBufferReader.cpp]: I think you did mention it, but I forgot.

Comment (gonetz) [reply src/Graphics/ColorBufferReader.cpp]: >> I went through a fresh file before on mupen Android and still encountered the issue with the entire process being without any cheats

So, it is not caused by cheats. Bad news. @fzurita @loganmc10  could you check, is this bug reproduced with other graphics plugins? I mean not crash in that point, but bad set of tile parameters, which leads to crash in load tlut: tile.tmem > 256 and tile.lrs == 255. It can be checked with debugger or with logging.
If it is reproduced, it is core issue. Otherwise, the plugin somehow breaks things.

Comment (fzurita) [reply src/Graphics/ColorBufferReader.cpp]: I have not experienced any crashing with Glide64mk2, but I have not checked that we are getting bad tile parameters. I'll try to check that when I get a chance, I have been busy with other things last couple of weeks.

Comment (loganmc10) [reply src/Graphics/ColorBufferReader.cpp]: So just to confirm a few things, @gonetz's fix does prevent the crash. I tested Glide64mk2 and it doesn't crash. I can test Glide64mk2 for the values:

Can someone point out to me where in https://github.com/mupen64plus/mupen64plus-video-glide64mk2/ I should be looking for these values? I'm not familiar with the source code at all

Comment (loganmc10) [reply src/Graphics/ColorBufferReader.cpp]: I think I found it here:
https://github.com/mupen64plus/mupen64plus-video-glide64mk2/blob/master/src/Glide64/rdp.cpp#L1693-L1696

I added:
```
if (start != 0 && count != 16)
      printf("start: %u count: %u\n", start, count);
```
Right where that crash happens it prints out a few:
```
start: 64 count: 256
```
It seems to be the only place it happens. So yes, it does seem that the same issue happens in Glide64, just no crash.

gonetz: @loganmc10 Yes, start: 64 count: 256 is violation of "if palette index != 0 palette size is 16" invariant.  Thanks! So, this is most likely a core issue. May be playing with counter factor can help. IIRC, Banjo-Tooie is very sensitive to counter factor and timing issues.

fzurita: Ok, it seems like we have established that bad tile parameters are a real possibility in mupen64plus and probably project64 as well. Could a custom microcode cause this issue?

gonetz: No, gDPSetTile and gDPSetTileSize are RDP commands. I did not check that gDPSetTileSize is actually missing, May be it presents, but sets wrong data. In both cases it is an error in display list, emulator's fail. Btw, have you checked that save in LLE mode? I can't check it now as I don't have Linux box on hands.  And the crash does not happen on windows for me.

fzurita:
I have not checked in LLE, I should try.

loganmc10:
Just tested LLE in Linux, it still crashes

fzurita:
Just checked as well. I tried interpreter and LLE. The rules out the RSP and dynarec.

fzurita:
I also tried several CountPerScanline and CountPerOp settings work no difference.

So what's left: Either the game is really doing that or there is some other timing issue in the core somewhere unrelated to what I checked.

gonetz:
> Just tested LLE in Linux, it still crashes

Thanks! It is expected to crash as it is not microcode issue.

loganmc10:
> So what's left: Either the game is really doing that or there is some other timing issue in the core somewhere unrelated to what I checked.

It could still be a timing issue, either one that is so precise that CountPerOp isn't accurate enough to solve, or (more likely), that the savestate was taken when it was already "too late" and the damage had already been done.

fzurita:
I believe retrobenny got there without using a save state and it still crashed. I'm not very familiar with that game, so I don't know how to get there from a fresh start. We would have to do that with multiple timing settings switched around to see if it makes a difference.

loganmc10:
> I believe retrobenny got there without using a save state and it still crashed. I'm not very familiar with that game, so I don't know how to get there from a fresh start. We would have to do that with multiple timing settings switched around to see if it makes a difference.

Yeah or at least take a save state at ~3-5 minutes before it happens and test from there. The save state we have now it like a second before the crash, so I assume something in the code might already be messed up.

fzurita:
Ok, I stumbled my way through the game starting from the in-game save. Here is a save state:
[earlier.zip](https://github.com/gonetz/GLideN64/files/1288326/earlier.zip)

Select "Pine grove", walk straight to the gate, after the gate walk straight to the swirly purple thing on the ground and push B, select "Space Zone". Once you are warped, walk to the right to get to the same area as the original save state.

fzurita:
I repeated the previous test without using a save state at all and only the in-game save. It still crashed. It was interpreter in LLE mode.

fzurita:
At this point, I believe that the invalid tile parameters must be happening in Windows as well, it's just not leading to a crash there.

I believe this because I was running with code that should be identical between Windows and arm/Linux builds.

We should really consider the possibility that the corrupt display list is coming from the game itself.

AmbientMalice:
Could it be a flaw in the emulator's implementation of Tooie's anti-piracy protection by any chance?

gonetz:
> At this point, I believe that the invalid tile parameters must be happening in Windows as well, it's just not leading to a crash there.

You're right: it happens on windows too.

> Could it be a flaw in the emulator's implementation of Tooie's anti-piracy protection by any chance?

Hardly. Rare's anti-piracy protection usually smarter.

retrobenny:
But isn't Banjo-Tooie's protection the same as Zelda Ocarina,or is it a heavily modified version of it which made it harder to cra-I mean bypass?

When I tried an old plugin with microcode selection for curiosity,the Zelda microcode option was the one that made it run instead continuing to hit gfx plugin microcode errors.

gonetz:
> We should really consider the possibility that the corrupt display list is coming from the game itself.

That is game's bug. Ok, let's consider that possibility:

    // Workaround for possible game/emulator bug (see bug #1250)
    if (pal != 0)
        count = 16;

fzurita:
This should be all set.

gonetz:
Thanks! Merged: 16d1aa6761034d84

PR #1551, opened 2017-08-19
Have mupen64plus maintain resolution in fullscreen mode

loganmc10:
This is mostly a suggestion/hope for some discussion.

Right now the default is to start at 640x480, windowed mode.

I get complaints from people sometimes that "performance is bad", and what is happening is that they put it into fullscreen, and thus it starts rendering at their monitor resolution, which might be quite high.

I think this would be a better default for most people, making the graphics look decent by default, while also allowing people to use fullscreen without having to mess with any settings to get decent performance.

### DIFF
--- a/src/BufferCopy/ColorBufferToRDRAM.cpp
+++ b/src/BufferCopy/ColorBufferToRDRAM.cpp
@@ -177,25 +177,14 @@ bool ColorBufferToRDRAM::_prepareCopy(u32 _startAddress)
 	}
 
 	if (m_pCurFrameBuffer->m_scale != 1.0f) {
-		u32 x0 = 0;
-		u32 width;
-		if (config.frameBufferEmulation.nativeResFactor == 0) {
-			const u32 screenWidth = wnd.getWidth();
-			width = screenWidth;
-			if (wnd.isAdjustScreen()) {
-				width = static_cast<u32>(screenWidth*wnd.getAdjustScale());
-				x0 = (screenWidth - width) / 2;
-			}
-		} else {
-			width = m_pCurFrameBuffer->m_pTexture->realWidth;
-		}
+		u32 width = m_pCurFrameBuffer->m_pTexture->realWidth;
 		u32 height = (u32)(bufferHeight * m_pCurFrameBuffer->m_scale);
 
 		CachedTexture * pInputTexture = m_pCurFrameBuffer->m_pTexture;
 		GraphicsDrawer::BlitOrCopyRectParams blitParams;
-		blitParams.srcX0 = x0;
+		blitParams.srcX0 = 0;
 		blitParams.srcY0 = 0;
-		blitParams.srcX1 = x0 + width;
+		blitParams.srcX1 = width;
 		blitParams.srcY1 = height;
 		blitParams.srcWidth = pInputTexture->realWidth;
 		blitParams.srcHeight = pInputTexture->realHeight;
--- a/src/DisplayWindow.cpp
+++ b/src/DisplayWindow.cpp
@@ -87,8 +87,21 @@ void DisplayWindow::updateScale()
 {
 	if (VI.width == 0 || VI.height == 0)
 		return;
-	m_scaleX = m_width / (float)VI.width;
-	m_scaleY = m_height / (float)VI.height;
+	u32 width, height;
+#ifdef MUPENPLUSAPI
+	if (m_bFullscreen)
+	{
+		width = config.video.windowedWidth;
+		height = config.video.windowedHeight;
+	}
+	else
+#endif
+	{
+		width = m_width;
+		height = m_height;
+	}
+	m_scaleX = width / (float)VI.width;
+	m_scaleY = height / (float)VI.height;
 }
 
 void DisplayWindow::_setBufferSize()
--- a/src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
+++ b/src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
@@ -78,6 +78,8 @@ bool DisplayWindowMupen64plus::_start()
 	m_screenHeight = config.video.windowedHeight;
 	_getDisplaySize();
 	_setBufferSize();
+	config.video.windowedWidth = m_width;
+	config.video.windowedHeight = m_height;
 
 	printf("(II) Setting video mode %dx%d...\n", m_screenWidth, m_screenHeight);
 	const m64p_video_flags flags = M64VIDEOFLAG_SUPPORT_RESIZING;
@@ -128,7 +130,6 @@ bool DisplayWindowMupen64plus::_resizeWindow()
 {
 	_setAttributes();
 
-	m_bFullscreen = false;
 	m_width = m_screenWidth = m_resizeWidth;
 	m_height = m_screenHeight = m_resizeHeight;
 	if (CoreVideo_ResizeWindow(m_screenWidth, m_screenHeight) != M64ERR_SUCCESS) {
@@ -145,6 +146,7 @@ bool DisplayWindowMupen64plus::_resizeWindow()
 
 void DisplayWindowMupen64plus::_changeWindow()
 {
+	m_bFullscreen = !m_bFullscreen;
 	CoreVideo_ToggleFullScreen();
 }
 


AmbientMalice:
I personally think native should be the default, since it has the highest level of accuracy. But 2x would be a good compromise. There is a bottleneck rendering at higher resolutions that needs dealing with, though. You get texture corruption and massive slowdown past around 7x.

loganmc10:
Yeah I also play at 1x, but people seem to complain when things look "blurry and pixelated". 2x should make most people happy, while still keeping performance at a decent place

SigmaVirus:
I also think native should be the default setting, it's Nintendo 64 not Nintendo Wii U

Things should look "blurry and pixelated" and if people don't like authentic looking then they can always change it to something else.

Accuracy first, eye-candy second!

gonetz:
Regardless of discussion about accuracy vs eye-candy, this can't be considered as general solution. Zilmar spec plugin has separate resolution settings for windowed and full screen modes. Both are 640x480 by default. Thus, zilmar-spec users will run full screen in 640x480 by default. Most of desktop users prefer full screen resolution equal to monitor resolution. The first thing I do when config version changed - set full screen resolution  to maximum. I would be confused to see low-res graphics in full screen after I set it to HD resolution. It is unexpected behavior for desktop graphics plugin. Android build has its own settings for rendering resolution, and as Android user I would be surprised to get low-res picture when I set rendering resolution to maximum. Thus, default nativeResFactor=2 is suitable only for platforms with low-performance graphics, as RPI.

loganmc10:
Fair enough

> Both are 640x480 by default. Thus, zilmar-spec users will run full screen in 640x480 by default

Can I ask, how is this achieved? Is it something done in this graphics plugin? I'm sure this is possible in mupen64plus, there is nothing in the spec preventing it, I just don't know much about scaling images and such

AmbientMalice:
Full screen output resolution should match actual screen for the simple reason that this allows 4:3 pillarboxing to work correctly.

>I would be confused to see low-res graphics in full screen after I set it to HD resolution.
>It is unexpected behavior for desktop graphics plugin.

Is it, though? Dolphin behaves that way. PCSX2 behaves that way. It's pretty much the norm to behave this way. Mind you, PCSX2 and Dolphin do it partially because non-standard resolutions cause memory corruption and/or graphics errors. Running emulators with a resolution that isn't in some way a multiple of the native one is pretty bizarre behavior that only really became common due to arguably poorly designed N64 and PS1 emulators.

An increasing number of modern videogames work that way in the sense that they decouple internal resolution from output resolution and use a percentage slider. When you set the screen resolution in the latest and greatest Call of Duty or Resident Evil or Bethesda game to 1920x1080, it is accepted and understood that the REAL resolution setting is the separate Internal Resolution setting.

>Android build has its own settings for rendering resolution, and as Android user I would be surprised to get low-res picture when I set rendering resolution to maximum.

Android users shouldn't be setting resolution to maximum, though. Especially when it causes performance to crash into the floor. And again, contemporary emulators don't work this way.

loganmc10:
I'll play around with this a bit tomorrow, but right now it's a little confusing, the config description for ScreenWidth/Screen height in mupen64plus says:

> Width of output window or fullscreen width

But if you resize the window, or put it into fullscreen, it changes the internal/rendering resolution to match. I think that the internal/rendering resolution should always match the ScreenWidth/Height setting, and if the user resizes the window, or uses fullscreen, it should scale the image to match. Does this make sense @gonetz ? You could always add another value to Video-General like "render fullscreen image at maximum resolution" or something

gonetz:
>Can I ask, how is this achieved? Is it something done in this graphics plugin? I'm sure this is possible in mupen64plus, there is nothing in the spec preventing it,

zilmar specs plugin handles all its options. Also, plugin creates graphics content. It's up to plugin to have separate settings for windowed and full screen modes. For example, early versions of Glide64 had only fullscreen settings, because Glide API worked only in full screen. Emulator sends notification to plugin about window mode change, plugin changes graphics content according to settings. GLideN64 defaults for fullscreen and window are 640x480. Yes, it is possible to select max available monitor resolution for full screen automatically, but it is not implemented yet.

mupen64plus core creates graphics context by itself and passes it to graphics plugin. Settings for graphics window size are part of [Video-General] section of mupen64plus config. This section is used by all graphics plugins, GLideN64 does not set it. GLideN64 reads these settings from config and use them to create graphics context. There are no separate settings for full screen and windowed modes. I don't know, why mupen64plus uses display size resolution for full screen in your case. mupen64plus for windows sets resolution, defined in [Video-General] for both window and full screen. 

> You could always add another value to Video-General like "render fullscreen image at maximum resolution" or something

I don't know, is it allowed for graphics plugin to change this section. Also, I don't know portable way to know maximum resolution. By mupen64plus ideology, plugin must not care about system dependent stuff.

gonetz:
> Is it, though? Dolphin behaves that way. PCSX2 behaves that way. It's pretty much the norm to behave this way

N64 emulation is older than Dolphin  and PCSX2 and it has its traditions.

> CSX2 and Dolphin do it partially because non-standard resolutions cause memory corruption and/or graphics errors.

Yes, but for N64 situation is different. Any resolution different from native may cause glitches with 2D graphics. Native N64 resolution is too low. If low-res graphics does not bother you, you may use software plugin and forget about graphics issues. Users select hardware plugins for their ability to produce hi-res graphics. And for speed of course.

> Android users shouldn't be setting resolution to maximum, though.

Why not? I want to run Zelda on my Shield with widescreen and max resolution. Why should I remeber, that I must reset nativeResFactor setting first?

loganmc10:
@gonetz I tested m64py in Windows, fullscreen toggling doesn't seem to work there either, it still updates the resolution to the monitor's full resolution in fullscreen mode, instead of 640x480 like it should be.

There are a few issues with the mupen64plus API implementation:
In DisplayWindowMupen64plus::_changeWindow(), m_bFullscreen is never modified, changeWindow() is called when fullscreen is toggled, so that function should have something like ```m_bFullscreen = !m_bFullscreen```

In mupen64plus, when _changeWindow() is called, the frontend will resize the window, and trigger a ResizeVideoOutput() call. Therefore, after _changeWindow() is called, _resizeWindow() will be called, with the values of m_resizeWidth and m_resizeHeight equaling the fullscreen resolution when m_bFullscreen is true (assuming m_bFullscreen was properly set in _changeWindow()).

Right now when _resizeWindow() is called, it forces ```m_bFullscreen = false``` and sets m_width and m_screenWidth to the value of m_resizeWidth (the fullscreen resolution).

mupen64plus frontends don't have the ability to scale the window like in Windows, therefore, if m_bFullscreen is true, it should be using the value of config.video.windowedWidth/config.video.windowedHeight for the rendering resolution, and stretching it to fix m_screenWidth/m_screenHeight

Does this make sense @gonetz ?

loganmc10:
Actually in mupen64plus-ui-console (the CLI mupen64plus frontend). It will actually try to switch the desktop's monitor resolution to match the window resolution using SDL_SetWindowFullscreen() (instead of calling ResizeVideoOutput).

This works if your monitor happens to support the resolution of the window, for instance, it worked when I had the resolution set to 640x480, but failed when I had to set to 320x240, since my monitor doesn't support that resolution. It's also a very ugly way to do it, especially on 2 monitor setups like I have, since it changes the resolution of the 2nd monitor. We should just pretend they never tried that approach....

In any event m64py and mupen64plus-gui call ResizeVideoOutput after ChangeWindow with the fullscreen resolution values, so we should be using those values to upscale the image, using the internal resolution values from config.video.windowedWidth/Height

loganmc10:
I have updated this PR with my proposed solution. Basically in ```DisplayWindow::updateScale()```, when using mupen64plus, if ```m_bFullscreen``` is true, then use the values from config.video.windowedWidth/Height to calculate the scaling, instead of m_width and m_height, which actually represent the monitor resolution.

loganmc10:
The only game that seems to have an issue with my changes is Dr. Mario:

dr mario_64-000

I assume it is the CopyFromRDRAM setting? I'm not sure why it doesn't like the scaling done this way.

loganmc10:
Honestly though even if I get that working, it's still a mess. What I would really love to see:

Set NativeResFactor to 2 by default. Have Framebuffer Emulation always on. Get rid of PJ64's full screen resolution options, just make the resolution the native resolution and use NativeResFactor to set the graphics quality. That makes everything so simple. Just as many people are confused as to why performance gets worse in full screen as would be confused why the quality might look poor

loganmc10:
I fixed the issue in Dr. Mario with https://github.com/gonetz/GLideN64/pull/1551/commits/1a41b2074cd88905f6f88dce3cd780f439a52bf9 I don't know for sure if that is correct. It seems strange to me to have one set of calculations if NativeResFactor = 0, and a different calculation for everything else, when the resolution could be the same in both cases (NativeResFactor=0 and res set to 640x480 is the same as NativeResFactor=2 in a lot of cases)

Using ```m_pTexture->realWidth``` when NativeResFactor=0 seems to work the same for me.

gonetz:
>  I just realized that PJ64 works the same way, it modifies the monitor resolution. That is a real ugly way to do it

This behavior is from CRT monitor times. CRT monitor does not have  "monitor resolution". Maximal resolution supported by CRT monitor usually not comfortable for your eyes. I agree that for modern monitors it is better to scale internal resolution to monitor one.

I need time to work through your propositions. I'll give my answer later.

src/BufferCopy/ColorBufferToRDRAM.cpp
```diff
@@ -177,25 +177,14 @@ bool ColorBufferToRDRAM::_prepareCopy(u32 _startAddress)
...
-			const u32 screenWidth = wnd.getWidth();
-			width = screenWidth;
-			if (wnd.isAdjustScreen()) {
-				width = static_cast<u32>(screenWidth*wnd.getAdjustScale());
-				x0 = (screenWidth - width) / 2;
-			}
-		} else {
-			width = m_pCurFrameBuffer->m_pTexture->realWidth;
-		}
+		u32 width = m_pCurFrameBuffer->m_pTexture->realWidth;
```
gonetz: You killed read from frame buffer for widescreen-hack mode.
We can't just read from frame buffer while in widescreen mode because our wide buffer does not fit its place in RDRAM.
We need to crop part of the image, which corresponds to original 4:3 frame.
The code you removed does that.

loganmc10: Ok, I figured there was some scenerio I was missing, I still need to think of some way to do this properly

gonetz:
@loganmc10 

> Set NativeResFactor to 2 by default.

As I said, I don't like to have that on desktops.

> Have Framebuffer Emulation always on.

Yes, it is not recommended to set Framebuffer Emulation off. However, I use that option sometimes and it is not hard for me to maintain it.

> Get rid of PJ64's full screen resolution options, just make the resolution the native resolution and use NativeResFactor to set the graphics quality.

I thought about it. It would be right idea if native res factor > 1 produced less glitches than arbitrary resolution. However, 3x native res has the same glitches as 1024x768 for example. Thus, why limit users with NativeResFactor? Besides, widescreen-hack should be used only with NativeResFactor=0
PJ64's full screen resolution options can be removed because I can detect available video modes on Windows and select maximal one. Hardly anyone uses CRT monitor today. Alternatively, I can add "auto-detect full screen resolution" option.

> Just as many people are confused as to why performance gets worse in full screen as would be confused why the quality might look poor

As I said, I don't like to have that on desktops. However, if you have 4K monitor, you may easily have bad performance in 4K resolution even with desktop GPU. May be NativeResFactor = 2 by default is not bad compromise.

AmbientMalice:
>As I said, I don't like to have that on desktops. However, if you have 4K monitor, you may easily have bad performance in 4K resolution even with desktop GPU. May be NativeResFactor = 2 by default is not bad compromise.

On that note, there is an issue with high resolutions that needs addressing, where performance can drop to single digits and textures start to corrupt. If that were resolved, the high resolutions wouldn't be as much of an issue.

loganmc10:
In thinking about this I don't think my solution is any better, with what I am doing, in mupen64plus it'll default to 640x480 in fullscreen mode, even if they have something like NativeResFactor=5, so that's no good. I guess I'll close this, having it run at the fullscreen resolution is what most people probably want anyway.

I'll just have to create a wiki or something to explain how to use NativeResFactor to control the internal resolution, although I still think that defaulting it to 2 wouldn't be a bad option

gonetz:
> I'll just have to create a wiki or something to explain how to use NativeResFactor to control the internal resolution

Any update of available documentation is welcome.

PR #1552, opened 2017-08-19
Add Google Cardboard VR support on Android

justinmichaud:
See the related change on mupen64plus-ae: https://github.com/mupen64plus-ae/mupen64plus-ae/pull/649

screenshot_20170817-214125

I tried to keep the changes confined to the android build as much as possible. Would you be willing to review this?

Thanks!

### DIFF
--- a/src/GraphicsDrawer.cpp
+++ b/src/GraphicsDrawer.cpp
@@ -775,14 +775,15 @@ void GraphicsDrawer::_drawStereo(void (GraphicsDrawer::*callback)(bool, void*), 
         (this->*callback)(i==1, data);
 
         gDP.scissor = orig_scissor;
-        gSP.changed |= CHANGED_SCISSOR;
-		updateScissor(frameBufferList().getCurrent());
     }
 
     // Hack so that we can avoid clipping next frame
     // See gSPCombineMatrices
     VR_CURRENTLY_RENDERING = false;
     gSPCombineMatrices();
+
+    gSP.changed |= CHANGED_SCISSOR;
+    updateScissor(frameBufferList().getCurrent());
 }
 
 void GraphicsDrawer::drawTriangles()
@@ -1345,8 +1346,10 @@ void GraphicsDrawer::_drawTexturedRectOneEye(bool finish, void *data)
 			m_rect[i].x *= scale;
 	}
 
-    for (int i=0; i<4; i++) {
-        m_rect[i].y *= 0.5;
+    if (config.vr.enable) {
+        for (int i=0; i<4; i++) {
+            m_rect[i].y *= 0.5;
+        }
     }
 
 	if (bUseTexrectDrawer)
--- a/src/GraphicsDrawer.cpp
+++ b/src/GraphicsDrawer.cpp
@@ -662,8 +662,36 @@ bool GraphicsDrawer::_canDraw() const
 	return config.frameBufferEmulation.enable == 0 || frameBufferList().getCurrent() != nullptr;
 }
 
-void GraphicsDrawer::_drawTrianglesUnmodified()
+void GraphicsDrawer::_drawTrianglesOneEye(bool finish)
 {
+    if (triangles.num == 0 || !_canDraw()) {
+        triangles.num = 0;
+        triangles.maxElement = 0;
+        return;
+    }
+
+    std::array<SPVertex, 256U> old_verts = triangles.vertices;
+
+    for (unsigned int j=0; j<triangles.vertices.size(); j++) {
+        SPVertex &vtx = triangles.vertices.at(j);
+        vtx.x = vtx.orig_x;
+        vtx.y = vtx.orig_y;
+        vtx.z = vtx.orig_z;
+        vtx.w = vtx.orig_w;
+    }
+
+#ifdef OS_ANDROID
+// ^ fixes a linker error, since gSPProcessVertex4 is
+//  not always included
+    unsigned int j=0;
+    for (;j<triangles.vertices.size(); j+=4) {
+        gSPProcessVertex4(j);
+    }
+    for (;j<triangles.vertices.size(); j++) {
+        gSPProcessVertex(j);
+    }
+#endif
+
 	_prepareDrawTriangle();
 
 	Context::DrawTriangleParameters triParams;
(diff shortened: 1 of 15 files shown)


fzurita:
Ok, I finally gave this a shot. I think it would be much better and produce a lot less graphical issues if you don't enable head tracking. I tried a few games and there were too many graphical issues for them to be playable.

If you did a simple split screen without giving each eye a different view point, I think it would fix those issues. If you tried to give each eye a different view point but disabled head tracking, I think the issues would remain.

A more viable approach in the end would be the way that dolphin did it, see this blog post:
https://dolphin-emu.org/blog/2015/05/13/a-second-perspective/

justinmichaud:
Hmm. What kind of graphical issues did you experience? I only tried it out on Super Mario 64 so far.

I think that applying the vr transformation to drawDMATriangles and drawScreenSpaceTriangle would hopefully fix some of the random graphical glitches and the hud. Unfortunately, glitches like the black sky in the screenshot above would probably require rom hacks to fix (they show up in Dolphin too).

I think that both eyes definitely need a different view point, otherwise people would get headaches. As for the head tracking, I think it might be useful as a config option to enable/disable, but I worry that removing it entirely would defeat the purpose of using a vr headset.

I took a look at the blog post, and I think this approach is as close as we can get, since AFAIK geometry shaders aren't available on most devices. Speaking of dolphin, I think that https://www.youtube.com/watch?v=pjMxt7Gx4F4 is a good example of what I was hoping to implement.

fzurita:
Aside from the black sky, one problem I saw in Mario 64 was that a lot of things out of the screen were not being drawn. The problem with head tracking is that any game that doesn't draw beyond the normal view port will have this problem.

Also, I tried Banjo-Tooie and attempting to move caused the camera to not really track the phone's movements. I think a lot of games need to be tried to make sure this works in general for most games.

Did you investigate why the sky is black? Maybe we can focus on one thing a time.

justinmichaud:
I will take a look at the drawDMATriangles and drawScreenSpaceTriangle today. That should make the hud look better, and fix a few small glitches.

For objects not appearing, it is almost certainly a problem with the original game code culling. I know that dolphin vr has cheat codes to fix issues like that, and so once the hud issues are taken care of, it might be worth looking into some game-specific hacks. In general, I don't think this is much of a problem if the hud is visible in-game, because then players will be looking through the hud most of the time.

For the sky, it is distracting, but I don't think there will be an easy fix. The same goes for one eye of the camera clipping through walls.

justinmichaud:
Hmm. There seems to be a weird glitch with the text (and I presume this is also happening with the sky texture). I commented out the HUD drawing code to try to isolate where the draw call is coming from, and it seems to be coming from drawTriangles(). I have no idea why this is happening:

screenshot_20170824-095947

fzurita:
Interesting glitch, not sure where that's coming from. Probably related to hearing the HUD twice.

justinmichaud:
I commented out the code that draws the HUD in the screenshot above though. Very confusing indeed.

src/GraphicsDrawer.cpp
```diff
@@ -735,6 +735,7 @@ void GraphicsDrawer::_drawStereo(void (GraphicsDrawer::*callback)(bool, void*),
         // Hack to work around lack of clearing in-game, exposed
         //  by VR viewport
         gfxContext.clearColorBuffer(0.0f, 0.0f, 0.0f, 0.0f);
+        gfxContext.clearDepthBuffer();
```
fzurita: You should only enable this hack if VR mode is enabled.

justinmichaud:
I fixed the issue above by only applying the VR transform if the current projection matrix is a perspective projection. Now, I am running into a new issue with textures getting cut off, and I was wondering if you have run into it before.

In the screenshot below, I reverted all of my changes. I checked out bbc7131 and added the following to drawTriangles:
```
void GraphicsDrawer::drawTriangles()
{
	if (triangles.num == 0 || !_canDraw()) {
		triangles.num = 0;
		triangles.maxElement = 0;
		return;
	}

    const u32 bufferWidth = VI.width;
    const u32 bufferHeight = VI_GetMaxBufferHeight(bufferWidth);
    const f32 viewportScale = DisplayWindow::get().getScaleX();

    const s32 size = bufferWidth / 2;
    const s32 start = 110;
    gfxContext.setViewport((s32) (start * viewportScale), 0, (s32) (size * viewportScale), (s32) (bufferHeight * viewportScale));
    gSP.changed &= ~CHANGED_VIEWPORT;

	_prepareDrawTriangle();

	Context::DrawTriangleParameters triParams;
	triParams.mode = drawmode::TRIANGLES;
	triParams.flatColors = m_bFlatColors;
[SNIP]
```
It seems that as I move the viewport right, the texture gets cut off. I tried everything I could think of to try to fix it, and now I am stuck. Any ideas?

screenshot_20170829-121326

fzurita:
I have not seen this problem before. It seems like an offset is not being applied correctly. I couldn't tell you where though.

theboy181:
This looks interesting, its too bad I don't have an android to full grasp the issues being reported.

I do have a question about the clipping, and scissoring hacks though. Would either of these help out the widescreen hacks built into GLideN64? I have been struggling to get around the culling that happens at the edges of the screen beyond the 4:3 ratio in some games. Good examples are SSB64, and Star Wars RS.

justinmichaud:
I think I have fixed most of the glitches that can be fixed without modifications to the game. Super Mario 64 seems to be playable, except for slowdown in a few areas. I was wondering if you could take a look and:
- Take screenshots of any remaining showstopping bugs that would need to be fixed before merging
- Take a look at the code to see what could be cleaner/less hacky

In addition, I was wondering if you had any ideas for improving performance.

Thanks again for your time and help. I think this will very cool once merged.

fzurita:
@gonetz What do you think about this?

gonetz:
I think that I have to find time and look at it.
Does it work? How to test it?

fzurita:
Set "enableVR" to TRUE in the configuration file. You can also force it on the code.

SpinAttack:
Sorry, new to GitHub is there any way I could try this?

BenMcLean:
Was this closed because it works now, or what?

justinmichaud:
@BenMcLean I closed it because it hasn't been active for a while. I would be happy to take a crack at making it work, although it was pretty glitchy the last time I worked on it.

BenMcLean:
I'd very much appreciate getting to play N64 games on Android with stereoscopic 3D for my Gear VR or Daydream headset. IMO headtracking isn't the really important thing: the stereoscopic 3D is.

SpinAttack:
Just add First Person Mario 64😁

BenMcLean:
That seems like it would be a nice game to try with it, sure, but of course first person and 3D are two different things. :)

SpinAttack:
Yeah I know, just saying.

justinmichaud:
I am breaking this up into smaller PRs, starting with stereo: https://github.com/gonetz/GLideN64/pull/1779

Issue #1553, opened 2017-08-19
Resetting once breaks rendering via transparency of many textures.


retrobenny:
This happens to me on a myriad of revisions for GLideN64 and its very frustrating because I have to re-open the emulator to fix it and soft/hard reset is completely useless as a result.
It appears to be both transparency and/or alpha that gets broken but some stuff that fades still look okay like the Dolby logo seen in one of the screenshots below.

![image](https://user-images.githubusercontent.com/16418716/29490411-77357a06-8500-11e7-8db7-383a860d43c4.png)

![image](https://user-images.githubusercontent.com/16418716/29490417-9bd849b0-8500-11e7-858a-9810d840bb68.png)

![image](https://user-images.githubusercontent.com/16418716/29490421-b751cea0-8500-11e7-9776-2fe5103cc3d3.png)

I have seen glitched rendering in another fashion after restarts before,but it wasn't specifically just entirely breaking transparency.
Zelda Ocarina looks much worse with solar flares happening on every single screen transition instead of normal fading.
I can't use savestates because loading them closes the memory viewer search windows (crash inducing) usually or triggers crashes anyway when trying to take advantage of debugger/breakpoints for code searches as they would've been convenient to undo any game freeze.
Sad thing is GLideN64 is actually faster than Project64 Video/Glide64-wxwidgets by a few frames or so with the right settings.

retrobenny:
This issue is not present on Public Release 2.0 so at least there is some info that can help for figuring out the issue.

loganmc10:
Lately I've run into a similar issue in Linux. It only happens if I am playing a game, stop the game, and then start a new game. If I close mupen64plus and then re-open it to play a new game, the issue doesn't occur. I thought it was an issue with a recent Mesa update or something, since it doesn't happen for me on Windows.

It could be some kind of Mesa issue, just thought I'd mentioned it since it looks like a similar issue.

![smash_brothers-000](https://user-images.githubusercontent.com/848146/29491472-3e08d4de-851a-11e7-9634-a1db65cc4878.png)

AmbientMalice:
This only happens with the new combiner, right?

AmbientMalice:
I've done a bit of testing, and I only seem to be able to trigger this in Project 64. (Run Mario Kart 64, quick save, load quick save.) Mupen64plus seems fine.

Aquatakat:
[...]

Issue #982, opened 2016-05-09
Error in latest git version when compling on rpi3


loganmc10:
I get this error when I run mupen64plus:

shader_link error: ERROR:CUSTOM-14 (fragment shader, line 73) Support for indexing array/vector/matrix with a non-constant is not mandated in the fragment shader

gonetz:
Looks like hardware-specific. What is your GPU?

loganmc10:
This is a Raspberry Pi 3, so it is a Broadcom Videocore IV I believe (GLES2)

gonetz:
Bad. Actually, GLES2 does not have that limitation. This code works on Android devices.
I may rewrite the code with only constant indexing, but it will look ugly.

loganmc10:
Ya when I Google that error, I only see references to the Raspberry Pi, it would be nice if this plugin worked with the Pi though

gonetz:
As I understand, the problem is with code like this:
`muxB[0] = 1.0 - muxA[uBlendMux1[1]];`
It is prohibited to access vector via non-constant index. Thus, it must be rewritten something like this:
```
lowp float muxa;
if (uBlendMux1[1] == 0)
  muxa = muxA[0];
else if (uBlendMux1[1] == 1)
  muxa = muxA[1];
else if (uBlendMux1[1] == 2)
  muxa = muxA[2];
else if (uBlendMux1[1] == 3)
  muxa = muxA[3];
muxB[0] = 1.0 - muxa;

```
And this is only beginning. Now we need to rewrite next equation:
`lowp vec4 blend1 = (muxPM[uBlendMux1[0]] * muxA[uBlendMux1[1]]) + (muxPM[uBlendMux1[2]] * muxB[uBlendMux1[3]]);
`
That is that clumsy set of ifs must be repeated for each mux source. I left it to you as an exercise :)

loganmc10:
So is it fair to say that 0ad53ad0e8a551e89563b03c279e1fb0ac4d9b86 caused this problem? I was wondering why it just stopped working recently

gonetz:
Yes, new blending code causes this problem.

loganmc10:
I'm not seeing this issue anymore, I'll re-open if it comes up again, I'm not sure why it went away.

loganmc10:
I'm willing to write all those if statements and test it, but can you confirm the possible values for uBlendMux1[0-3], are the possible values only 0, 1, 2, or 3?

gonetz:
[...]

PR #1566, opened 2017-09-06
Fix for #1553

loganmc10:
The problem currently is that these shaders are created when CombinerProgramBuilder is created, but g_cycleType isn't updated until later, however the shader text has an if statement that depends on the value of g_cycleType.

I moved the creation of the shader text to an "update()" function that is called right before the shader is written. This fixed the issue for me.

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
@@ -464,148 +464,121 @@ public:
 class ShaderBlender1 : public ShaderPart
 {
 public:
-	ShaderBlender1(const opengl::GLInfo & _glinfo)
+	void update()
 	{
-		if (_glinfo.isGLES2) {
-			m_part =
-				"  if (uForceBlendCycle1 != 0) {	\n"
-				"    muxPM[0] = clampedColor;		\n"
-			;
-			if (g_cycleType == G_CYC_2CYCLE) {
-				m_part +=
-					"    muxPM[1] = clampedColor;	\n"
-					;
-			}
-			m_part +=
-				"    muxA[0] = clampedColor.a;		\n"
-				"    lowp float muxa;				\n"
-				"    if (uBlendMux1[1] == 0)		\n"
-				"      muxa = muxA[0];				\n"
-				"    else if (uBlendMux1[1] == 1)	\n"
-				"      muxa = muxA[1];				\n"
-				"    else if (uBlendMux1[1] == 2)	\n"
-				"      muxa = muxA[2];				\n"
-				"    else if (uBlendMux1[1] == 3)	\n"
-				"      muxa = muxA[3];				\n"
-				"    muxB[0] = 1.0 - muxa;			\n"
-				"    lowp vec4 muxpm0;				\n"
-				"    if (uBlendMux1[0] == 0)		\n"
-				"      muxpm0 = muxPM[0];			\n"
-				"    else if (uBlendMux1[0] == 1)	\n"
-				"      muxpm0 = muxPM[1];			\n"
-				"    else if (uBlendMux1[0] == 2)	\n"
-				"      muxpm0 = muxPM[2];			\n"
-				"    else if (uBlendMux1[0] == 3)	\n"
-				"      muxpm0 = muxPM[3];			\n"
-				"    lowp vec4 muxpm2;				\n"
-				"    if (uBlendMux1[2] == 0)		\n"
-				"      muxpm2 = muxPM[0];			\n"
-				"    else if (uBlendMux1[2] == 1)	\n"
-				"      muxpm2 = muxPM[1];			\n"
-				"    else if (uBlendMux1[2] == 2)	\n"
-				"      muxpm2 = muxPM[2];			\n"
-				"    else if (uBlendMux1[2] == 3)	\n"
-				"      muxpm2 = muxPM[3];			\n"
-				"    lowp float muxb;				\n"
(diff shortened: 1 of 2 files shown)


gonetz:
It was the first task in my todo list. Great that you found the problem, thanks! I have few notes about implementation though.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -539,14 +544,21 @@ class ShaderBlender1 : public ShaderPart
...
+	opengl::GLInfo m_glinfo;
 };
 
 class ShaderBlender2 : public ShaderPart
 {
 public:
 	ShaderBlender2(const opengl::GLInfo & _glinfo)
 	{
-		if (_glinfo.isGLES2) {
+		m_glinfo = _glinfo;
```
gonetz: ShaderBlender2 does not use g_cycleType. These changes are redundant.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -539,14 +544,21 @@ class ShaderBlender1 : public ShaderPart
 
 		}
 	}
+private:
+	opengl::GLInfo m_glinfo;
```
gonetz: You need only _glinfo.isGLES2.
So,
`bool m_isGLES2;`

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -466,7 +466,12 @@ class ShaderBlender1 : public ShaderPart
 public:
 	ShaderBlender1(const opengl::GLInfo & _glinfo)
 	{
-		if (_glinfo.isGLES2) {
+		m_glinfo = _glinfo;
```
gonetz: `m_isGLES2 = _glinfo.isGLES2;`
see below for m_isGLES2

gonetz: Also, I would keep current value of g_cycleType:
m_cycleType = g_cycleType;

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -466,7 +466,12 @@ class ShaderBlender1 : public ShaderPart
...
+	}
+
+	void update()
+	{
+		if (m_glinfo.isGLES2) {
```
gonetz: This will update m_part each time new shader is created.
I suggest to add this check:
```
if (m_cycleType == g_cycleType)
  return;
```

gonetz: if (m_isGLES2)

gonetz:
Another idea: check both blenders for isGLESX, not for isGLES2. retrobenny said that Android devices have performance issues with new blending at high resolutions, so this optimization can be useful for all mobile platforms.

gonetz:
Another idea: add another class member
`std::string m_part2;`
and create two versions of shader: for 1cycle and 2cycle modes. Make ShaderPart::write virtual and override it for ShaderBlender1. It will be faster than repeatedly rewrite m_part. 
```
void ShaderBlender1::write(std::stringstream & shader) const override
{
  if (g_cycleType == G_CYC_1CYCLE)
	shader << m_part;
  else
	shader << m_part2;
}

```

gonetz:
Also, please make more detailed commit description.

loganmc10:
You can commit a fix for this if you'd like and just close this PR, the problem is known now, I don't mind if you just fix it as you see best.

> Another idea: check both blenders for isGLESX, not for isGLES2. retrobenny said that Android devices have performance issues with new blending at high resolutions, so this optimization can be useful for all mobile platforms.

retrobenny normally uses a Shield TV, which doesn't use GLES anyway, but either way, I wouldn't really call it an "optimization". It was put in place to work around this issue: https://github.com/gonetz/GLideN64/issues/982 . It's just 2 ways of writing the same thing, but I would think it's best to prefer the shorter version if it works and let the GPU's GLSL compiler optimize it as best it can

gonetz:
Ok.

gonetz:
> I wouldn't really call it an "optimization".

There was a discussion about this matter. I don't remember details, but main thought is that GPUs do not like variable as index for vertex/matrix even if they support such thing. Some shader compilers produce low performance warnings. So, your code is kind of optimization.

loganmc10:
> There was a discussion about this matter. I don't remember details, but main thought is that GPUs do not like variable as index for vertex/matrix even if they support such thing. Some shader compilers produce low performance warnings. So, your code is kind of optimization.

It's possible, I'm sure it wouldn't really hurt to enable it for all GLES devices

loganmc10:
@gonetz I've updated this PR with the changes you suggested. I also removed the original blending implementation and made the GLES2/Raspberry Pi version the default. It produces the same result, and from retrobenny's testing it sounds like it does perform better.

gonetz:
@loganmc10 I already re-implemented your fix. It affects the same places of code, which I touched during my work on F-Zero, thus I put it on top to avoid conflicts. It will be released within days. Sorry, I did not informed you about it.

Regarding original blending implementation : yes, retrobenny says that your optimized version works almost as fast as legacy blending. However I will not completely remove it from code. I'll disable it by #ifdef and keep as reference, because it has much more simple and clear form.

loganmc10:
Oh ok no problem, I'll close this then.

PR #1570, opened 2017-09-12
Only use XXH32 in CRC_OPT

loganmc10:
I realized the return type in XXH64 is an unsigned long long, since the output is stored in a u32 (unsigned int), this could create problems. Better to just use XXH32, which returns a u32/unsigned int


gonetz:
Merged: f567186e014a9d

PR #1571, opened 2017-09-13
Init mupen64plus config at PluginStartup

loganmc10:
Right now, the configuration isn't populated until RomOpen. This means that the configuration isn't populated until the first game is played. Most mupen64plus plugins (all that I know of), set the config defaults at PluginStartup. This way, in a GUI frontend, the configuration can be populated before a user plays their first game.


gonetz:
It works, thanks!
Merged: 528744de19cee

Issue #1555, opened 2017-08-25
Please add support for legacy blending on zilmar spec GLideN64.


retrobenny:
It would help greatly for Interpreter performance and would run even faster than Glide64/Project64 Video thanks to the Async and threading optimizations.

It already proved to me that it could outperform Glide64mk2 on the other version of it within BizHawk when set to 1x Native and the fastest settings quite a long while back thanks to enabling legacy blending for the speed boost.

For codes,I don't care about sacrificing some visual accuracy because I need Interpreter for real-time ASM changes currently until zilmar finds a solution for writing RDRAM ASM to cache when changes are made and Shygoo adds recompiler support for the Commands window and setting breakpoints.

retrobenny:
Can someone please point out if I could just use an ini parameter to enable legacy blending?
I won't need the change to UI if I can just exploit that to access much faster performance.

AmbientMalice:
The Zilmar version passes all configuration through the UI.

retrobenny:
Don't know how to deal with that. This probably means it lacks the setting parameter within PJ64 source code files as well.

gonetz:
@retrobenny I don't want to add legacy blending option to desktop builds. If you really need it, I can make a build specially for you with legacy blending always enabled.

retrobenny:
Cool,I can only hope its able to outperform everything else I have tried (it should at least be faster than Glide64) like it did comparing legacy mode GLideN64 to Glide64mk2 on BizHawk regardless of GPU choice.
It beat out Glide64mk2 in performance with legacy blending checked so it should really help for use with Interpreter to debug with BPs in order to make interesting new codes and to improve basic older codes if I can reduce freezing and crashing due to "mods" (corrupted ASM bytes) breaking things too much.

gonetz:
Current master with forced legacy blending:
https://drive.google.com/file/d/0B0YqMPjGo3B2bEVHYmRvaWxqQ1E/view?usp=sharing

retrobenny:
Thank you!

I do wish there was someone that would try to make legacy blending render more stuff without impacting on performance like at least the lighting effects in SSB64.
Even a version of Jabo can do that and still have great performance,but that is a D3D closed-source plugin so its hard to compare it to anything else. Having legacy blending optimized for functionality and sustained performance gains would make the plugin the best of both worlds. (e.g. legacy blending on Android would look better with working Smash Bros. lighting since its the only decent performing mode for that port)

[...]

Issue #1534, opened 2017-07-23
Severe performance impact at higher resolutions.


AmbientMalice:
This is a general issue, but I've noticed that the plugin performance falls apart the higher you push the resolution (over 10x multiplier, for example), and I'm wondering if there's a bottleneck somewhere that needs addressing.

fzurita:
@retrobenny has brought this up a few times.
https://github.com/gonetz/GLideN64/issues/1036
https://github.com/gonetz/GLideN64/issues/1052
https://github.com/gonetz/GLideN64/issues/1120

It seems to be related to the new blending method introduced a while back. Mupen64Plus has an option to use the legacy method which improves performance by a good amount. The option seems to be missing in GLIdeNUI.

fzurita:
It almost looks like we are applying the new blending shader to something that we should not be upscaling with resolution.

gonetz:
> The option seems to be missing in GLIdeNUI.

Desktop GPU must not have performance issues with blending shader, so this option disabled for zilmar specs. It is easy to check, is the blender shader a bottleneck on high resolutions with mupen64plus build, because it has "legacy blending" option.

olivieryuyu:
@AmbientMalice 

with the new blending shader, is the performance ok now?

AmbientMalice:
Although I haven't done extensive testing, performance does seem to be better. The problem isn't entirely fixed because performance collapses around/above 10x native on some games with default settings, resulting in single digit framerates. MSAA makes it worse.

![perfect_dark-012](https://user-images.githubusercontent.com/10704701/31017449-d490981c-a56b-11e7-9985-f983b3ed5321.png)

One factor is the cache setting, which by default is 500MB. This is insufficient for higher resolutions, and should probably be automated.

olivieryuyu:
10x native? it is a huge resolution! Why would need this?

AmbientMalice:
>10x native? it is a huge resolution! Why would need this?

I wouldn't personally recommend running games above 4x native, but some people want to run at very high resolutions.

320x240 times 10 is 3200x2400.

Some people are running 4K monitors, which natively render at 3840x2160.

Games like Perfect Dark are problematic because their high resolution mode is 640x222, which translates to 6400x2220 at 10 times native.

olivieryuyu:
well then honestly no wonder why it is getting slow. Is there any program won't be slow at this resolution? I cannot believe we have a performance issue.

AmbientMalice:
[...]

PR #1581, opened 2017-09-20
Optimize blender shaders

loganmc10:
Based on discussion in https://github.com/gonetz/GLideN64/issues/1555

I've copied the new code below so it's easier to look at, the git diff is a little messy.

I haven't tested this on any GLES devices, but it gets rid of all the if statements as well as any non-constant array indexing.

Thanks to @standard-two-simplex for the help

```
class ShaderBlender1 : public ShaderPart
{
public:
	ShaderBlender1()
	{
		m_part =
			"  if (uForceBlendCycle1 != 0) {	\n"
			"    muxPM[0] = clampedColor;		\n"
		;
		if (g_cycleType == G_CYC_2CYCLE) {
			m_part +=
				"    muxPM[1] = clampedColor;	\n"
				;
		}
		m_part +=
			"    muxA[0] = clampedColor.a;		\n"
			"    lowp float muxa = dot(muxA,vec4(equal(vec4(uBlendMux1[1]),vec4(0,1,2,3))));	\n"
			"    muxB[0] = 1.0 - muxa;			\n"
			"    lowp vec4 muxpm0 = muxPM*vec4(equal(vec4(uBlendMux1[0]), vec4(0,1,2,3)));	\n"
			"    lowp vec4 muxpm2 = muxPM*vec4(equal(vec4(uBlendMux1[2]), vec4(0,1,2,3)));	\n"
			"    lowp float muxb = dot(muxB,vec4(equal(vec4(uBlendMux1[3]),vec4(0,1,2,3))));	\n"
			"    lowp vec4 blend1 = (muxpm0 * muxa) + (muxpm2 * muxb);	\n"
			"    clampedColor.rgb = clamp(blend1.rgb, 0.0, 1.0);		\n"
			"  }								\n"
			;
	}
};

class ShaderBlender2 : public ShaderPart
{
public:
	ShaderBlender2()
	{
		m_part =
			"  if (uForceBlendCycle2 != 0) {	\n"
			"    muxPM[0] = clampedColor;		\n"
			"    muxPM[1] = vec4(0.0);			\n"
			"    muxA[0] = clampedColor.a;		\n"
			"    lowp float muxa = dot(muxA,vec4(equal(vec4(uBlendMux2[1]),vec4(0,1,2,3)))); \n"
			"    muxB[0] = 1.0 - muxa;			\n"
			"    lowp vec4 muxpm0 = muxPM*vec4(equal(vec4(uBlendMux2[0]), vec4(0,1,2,3)));  \n"
			"    lowp vec4 muxpm2 = muxPM*vec4(equal(vec4(uBlendMux2[2]), vec4(0,1,2,3)));  \n"
			"    lowp float muxb = dot(muxB,vec4(equal(vec4(uBlendMux2[3]),vec4(0,1,2,3))));	\n"
			"    lowp vec4 blend2 = muxpm0 * muxa + muxpm2 * muxb;	\n"
			"    clampedColor.rgb = clamp(blend2.rgb, 0.0, 1.0);	\n"
			"  }								\n"
			;
	}
};
```

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
@@ -464,148 +464,50 @@ public:
 class ShaderBlender1 : public ShaderPart
 {
 public:
-	ShaderBlender1(const opengl::GLInfo & _glinfo)
+	ShaderBlender1()
 	{
-		if (_glinfo.isGLES2) {
-			m_part =
-				"  if (uForceBlendCycle1 != 0) {	\n"
-				"    muxPM[0] = clampedColor;		\n"
-			;
-			if (g_cycleType == G_CYC_2CYCLE) {
-				m_part +=
-					"    muxPM[1] = clampedColor;	\n"
-					;
-			}
+		m_part =
+			"  if (uForceBlendCycle1 != 0) {	\n"
+			"    muxPM[0] = clampedColor;		\n"
+		;
+		if (g_cycleType == G_CYC_2CYCLE) {
 			m_part +=
-				"    muxA[0] = clampedColor.a;		\n"
-				"    lowp float muxa;				\n"
-				"    if (uBlendMux1[1] == 0)		\n"
-				"      muxa = muxA[0];				\n"
-				"    else if (uBlendMux1[1] == 1)	\n"
-				"      muxa = muxA[1];				\n"
-				"    else if (uBlendMux1[1] == 2)	\n"
-				"      muxa = muxA[2];				\n"
-				"    else if (uBlendMux1[1] == 3)	\n"
-				"      muxa = muxA[3];				\n"
-				"    muxB[0] = 1.0 - muxa;			\n"
-				"    lowp vec4 muxpm0;				\n"
-				"    if (uBlendMux1[0] == 0)		\n"
-				"      muxpm0 = muxPM[0];			\n"
-				"    else if (uBlendMux1[0] == 1)	\n"
-				"      muxpm0 = muxPM[1];			\n"
-				"    else if (uBlendMux1[0] == 2)	\n"
-				"      muxpm0 = muxPM[2];			\n"
-				"    else if (uBlendMux1[0] == 3)	\n"
-				"      muxpm0 = muxPM[3];			\n"
-				"    lowp vec4 muxpm2;				\n"
-				"    if (uBlendMux1[2] == 0)		\n"
-				"      muxpm2 = muxPM[0];			\n"
-				"    else if (uBlendMux1[2] == 1)	\n"
-				"      muxpm2 = muxPM[1];			\n"
-				"    else if (uBlendMux1[2] == 2)	\n"
-				"      muxpm2 = muxPM[2];			\n"
-				"    else if (uBlendMux1[2] == 3)	\n"
-				"      muxpm2 = muxPM[3];			\n"
-				"    lowp float muxb;				\n"
-				"    if (uBlendMux1[3] == 0)		\n"
-				"      muxb = muxB[0];				\n"
-				"    else if (uBlendMux1[3] == 1)	\n"
-				"      muxb = muxB[1];				\n"
-				"    else if (uBlendMux1[3] == 2)	\n"
-				"      muxb = muxB[2];				\n"
-				"    else if (uBlendMux1[3] == 3)	\n"
-				"      muxb = muxB[3];				\n"
-				"    lowp vec4 blend1 = (muxpm0 * muxa) + (muxpm2 * muxb);	\n"
-				"    clampedColor.rgb = clamp(blend1.rgb, 0.0, 1.0);		\n"
-				"  }								\n"
+				"    muxPM[1] = clampedColor;	\n"
 				;
-		} else {
-			m_part =
-				"  if (uForceBlendCycle1 != 0) {						\n"
-				"    muxPM[0] = clampedColor;							\n"
-			;
-			if (g_cycleType == G_CYC_2CYCLE) {
-				m_part +=
-					"    muxPM[1] = clampedColor;						\n"
-				;
-			}
-			m_part +=
-				"    muxA[0] = clampedColor.a;							\n"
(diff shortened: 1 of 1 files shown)


loganmc10:
This is a little faster than the current GLES2 code (maybe 3-5 VI/s on my laptop in Smash Bros). But it's much cleaner than all the if statements and probably easier for the various GPU's to compile efficiently

loganmc10:
If anyone wants to try this out on Windows, here is a 64-bit mupen64plus build with this change.

[mupen64plus-video-GLideN64.zip](https://github.com/gonetz/GLideN64/files/1319176/mupen64plus-video-GLideN64.zip)

You'll need to get the 64-bit build of mupen64plus from here: https://m64p.github.io/. Just replace the mupen64plus-video-GLideN64.dll file with the one in the zip I provided here.

@AmbientMalice perhaps you can give it a try at higher resolutions? I know you've run into this issue in the past (https://github.com/gonetz/GLideN64/issues/1534)

theboy181:
Not compatible with PJ64?

loganmc10:
The changes are compatible with it, I just don't have the setup to build a PJ64 build for testing.

theboy181:
No worries. if you ever need me to test for PJ, let me know..

theboy181:
Is there a good article explaining the need for the zilmar spec change Mupen needed? Somthing to do with linux?

loganmc10:
> Is there a good article explaining the need for the zilmar spec change Mupen needed? Somthing to do with linux?

I'm sure there is something out there explaining it, but basically the reason (as I understand it) was cross-platform compatibility. It was to remove the need for plugins to worry about what platform it is running on. For instance, the zilmar-spec build of GLideN64 is only compatible with Windows, it won't work on Linux if PJ64 ever gets ported there, and it won't work on Android. Same goes for Azimers audio plugin and every other zilmar spec plugin.

In contrast, the mupen64plus build of GLideN64 works on Windows, Android, and Linux, all using the same code. mupen64plus's cross compatibility is the reason that it had an Android version years before PJ64, and the reason that GLideN64 is available on Android with mupen64plus, but not PJ64 (even though PJ64 has an Android version now)

dankcushions:
builds and runs fine on RPI3. not sure of a good game to really thrash it, but at first glance seemed to be the same performance with legacyblending true or false.

loganmc10:
> builds and runs fine on RPI3. not sure of a good game to really thrash it, but at first glance seemed to be the same performance with legacyblending true or false.

Yeah I'm not really sure either, I just tested a few common games. I'm hoping that perhaps we can just get rid of legacy blending with this change, the performance difference is probably pretty negligible now (it's like 5-7 VI/s difference on my laptop when running at 10x Native res)

AmbientMalice:
Performance seems massively improved. However, enabling MSAA with high resolutions such as x10 native cuases this.

perfect_dark-012

gonetz:
Smart, very smart. @loganmc10 and  @standard-two-simplex - **thanks**!!!
Definitely much better than clumsy 'if else' variant.

Again, this commit conflicts with my current dev branch. I'll add these changes manually. It will be available to test with next WIP build. I hope that users feedback will allow us to decide, is legacy blending option still necessary.

loganmc10:
By the way, I tried:
```
equal(ivec4(uBlendMux1[0]), ivec4(0,1,2,3))
```
Instead of
```
equal(vec4(uBlendMux1[0]), vec4(0,1,2,3))
```

Since all the values are integers, I assumed that using ivec4 might be faster, but it was actually a few VI/s slower at 10x native res. I'm not really sure why, but that why I've left it as all vec4's

gonetz:
Strange, yes. It sounds logical to compare ints as ints, not as floats.
May it is because GPU processors optimized to work with floats, not integers.

AmbientMalice:
I've done further testing. MSAA destroys performance (1fps, basically) at higher resolutions, but Perfect Dark's textures start to corrupt above 8x native (while in High Resolution mode) even without MSAA enabled.

loganmc10:
But is that new with this change? Or does it also happen in the current master?

AmbientMalice:
>But is that new with this change? Or does it also happen in the current master?

Same behavior as current master. I'm just noting it because basically this change isn't fixing performance falling apart at higher resolutions combined with the graphics turning into a flickering mess. It's improving the performance of shader blending sure, but people definitely shouldn't run at high resolutions or use MSAA until this side issue is fixed.

standard-two-simplex:
@loganmc10 What do you use to test performance on mupen64plus? I remember being unable to print the fps to console or screen in the past.

loganmc10:
GLideN64 can print the FPS to the screen, it has a "ShowFPS" and "ShowVIS" option. You can see it in the bottom left corner of AmbientMalice's screenshot

standard-two-simplex:
Ok thanks, I guess I'll have to update the version (xD I haven't downloaded a new one in a long time).

AmbientMalice:
The downside is I can't seem to get m64p's speed limiter to disable, which makes getting the "full" framerate impossible. I'm probably missing something obvious.

loganmc10:
@AmbientMalice do you just click the "toggle speed limiter" option in the GUI? It works on my PC, but you need to disable vsync in the Nvidia control panel if it's enabled

gonetz:
231463e368160

loganmc10:
> Increasing cache size helps somewhat. Fixes texture corruption, at least, without MSAA enabled. I think cache should probably be automatically set based on available VRAM, and you should probably cap the resolution multiplier at something like 12x.

I don't think the resolution multiplier should be capped. I just tested one of the machines at my work and it can run Perfect Dark at 30x Native Res at 160 VI/s (I had to increase the CacheSize to 15000 for the graphics to stay uncorrupted). So the hardware does exist to perform that task, it's just not easily available to most people.

The CacheSize does need to be increased when the resolution gets really high, but I don't think OpenGL provides a way to query the available GPU memory, so I don't know if there is a real good way to automatically set it, but I could be wrong.

loganmc10:
The only real solution would be to get rid of the CacheSize parameter, let it use as much memory as it can, and gracefully deal with ```GL_OUT_OF_MEMORY``` errors if/when they happen.

PR #1600, opened 2017-09-26
Optimize 3point shader

loganmc10:
We recently discovered that variable indexing can be extremely resource intensive. This #ifdef was added for the same reason. Given what we've discovered, I believe it's probably best to use the if/else version

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
@@ -1478,15 +1478,11 @@ public:
 					"#define TEX_OFFSET(off) texture2D(tex, texCoord - (off)/texSize)			\n"
 					"lowp vec4 filter3point(in sampler2D tex, in mediump vec2 texCoord)			\n"
 					"{																			\n"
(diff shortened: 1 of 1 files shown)


loganmc10:
This one is slightly different because it's not indexing based on a uniform value, it's indexing based on a "shader computed variable".  It's only used in GLES2 anyway, so the if/else version is probably best

mudlord:
Why arent you using ternery operators?

loganmc10:
Do you really think it would make a difference?

mudlord:
The code looks ugly as sin otherwise. But no, not really, doubt it will make any performance difference.

loganmc10:
This code is only for GLES2 devices, the Raspberry Pi doesn't allow variable array indexing. I'm not sure what other restrictions it has or how the different variations of code might affect performance, I don't have one to test anymore. I know this code works and performs well.

mudlord:
Oh, in that case, carry on then :slightly_smiling_face:

gonetz:
I don't think that version with ' variable ifs' is any faster than variable indexing. GLSL 'if' works fast with uniforms, not with local variables. However, since both versions are equivalent,  it has sense to keep only one, which works everywhere.

Issue #936, opened 2016-03-24
SDEX games incorrect upscaling in HLE


olivieryuyu:
![sdf](https://cloud.githubusercontent.com/assets/8987490/14011597/1621564e-f1a0-11e5-9c82-51ef95e54d93.jpg)

on top of being slow, there is gfx bug not related to scaling or filtering apparently

AmbientMalice:
Fixed in LLE, kinda. (Game develops different visual errors.)

olivieryuyu:
SD Hiryuu no Ken Densetsu: the same

![sdfsdf](https://cloud.githubusercontent.com/assets/8987490/14013963/c5375e26-f1af-11e5-887b-daa1e7cd23c5.jpg)

olivieryuyu:
Susume! Taisen Puzzle Dama Toukon! Maruma

![qsds](https://cloud.githubusercontent.com/assets/8987490/14014235/a7389546-f1b1-11e5-84a4-dfee571b7e77.jpg)

olivieryuyu:
Pro Mahjong Tsuwamono 64 same

![sdfsdf](https://cloud.githubusercontent.com/assets/8987490/14015437/3aa71594-f1b9-11e5-9721-fef54c2b9dbb.png)

standard-two-simplex:
I corrected sprite sizes and applied the new filtering method. Most of the issues are gone.
44a60e14f81628b8cc080806201b2df084c3d452

olivieryuyu:
interesting! 

Any possibility that you fancy investigating https://github.com/gonetz/GLideN64/issues/63 or https://github.com/gonetz/GLideN64/issues/114?

No sure that there is a link with texrec though :)

olivieryuyu:
@gonetz 

Could it be the usage of the effect of gSPObjRenderMode, G_OBJRM_WIDEN?

I tried to figure out this is working in your code, no idea :(

gonetz:
I don ' t know about that rendermode

olivieryuyu:
http://n64devkit.square7.ch/n64man/gsp/gSPObjRenderMode.htm

G_OBJRM_WIDEN
Enlarges the drawn image by 3/8 pixel in the positive S and T directions. This mode is used to prevent blank space from opening up at the seams when a large rotating object is displayed from texture memory (TMEM) by combining sprites. This mode is effective when combining and drawing opaque sprites. It is unnecessary for translucent sprites because the seams are filled by processing by the blender (BL).

[...]

PR #1604, opened 2017-09-27
Fix sprite sizes

standard-two-simplex:
HLE sprite sizes were 1pixel smaller than LLE sprite sizes when not in copy mode.
Fixes https://github.com/gonetz/GLideN64/issues/936

Edit: I made a mistake while testing yesterday. A second commit was necessary.

### DIFF
--- a/src/gSP.cpp
+++ b/src/gSP.cpp
@@ -2275,12 +2275,8 @@ struct ObjCoordinates
 
 		ulx = frameX;
 		uly = frameY;
-		lrx = frameX + min(imageW/scaleW, frameW) - 1.0f;
-		lry = frameY + min(imageH/scaleH, frameH) - 1.0f;
-		if (gDP.otherMode.cycleType == G_CYC_COPY) {
-			lrx += 1.0f;
-			lry += 1.0f;;
-		}
+		lrx = frameX + min(imageW/scaleW, frameW);
+		lry = frameY + min(imageH/scaleH, frameH);
 
 		uls = imageX;
 		ult = imageY;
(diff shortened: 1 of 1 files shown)


olivieryuyu:
The g[s]SPObjRectangle GBI draws texture for the rectangle area defined by the upper left hand corner screen coordinate (objX, objY), and lower right hand corner screen coordinate (objX+imageW/scaleW-1, objY+imageH/scaleH-1). The drawn texture region will be defined by upper left hand corner (0,0) and lower right hand corner (imageW-1, imageH-1). If scaleW and scaleH are 1 << 10, texture will be drawn with equal proportions, without scaling. 

the **texture** for the rectangle is -1.

standard-two-simplex:
@olivieryuyu Yes, however OpenGL takes the coordinate on the right side of the pixel which is ``(imageW, imageH)``, which is handled in ``gSPDrawObjRect()``. Shifting rectangle position by approximately 0.5 almost gives correct results with ``(imageW, imageH)`` coordinates, but there are precision errors that make it impossible with current code.

gonetz:
I put this fix to 'fix_sprite_size' branch. It needs testing. I foresee regressions in 2D games. Good if I'm wrong.

gonetz:
Debug build with this branch:
https://drive.google.com/file/d/0B0YqMPjGo3B2Z2l3LTBvd3BMZHc/view?usp=sharing

AmbientMalice:
Appears to fix Magical Tetris Challenge:

Before:
gliden64_magical_tetris_001

After:
gliden64_magical_tetris_002

LLE is broken in a different way:
gliden64_magical_tetris_003

olivieryuyu:
done some tests and if regresssion there is, it is not obvious.

gonetz:
I'll include it to next WIP build. May be somebody will find something. Otherwise it will be put to master.

gonetz:
Merged: 89406223f5e271

PR #1605, opened 2017-09-29
ShaderBlender: cast to ivec4 where needed. Fixes #1601

### LINKED ISSUE #1601: Super Mario 64 - green face on Intel Iris Pro GPU (regression)
macOS 10.13
Intel Iris Pro 1536 MB

screen shot 2017-09-26 at 10 31 13 pm

This first appeared in commit https://github.com/gonetz/GLideN64/commit/81cc0854dce4e3914dbc71b0067cf9d7549c936e

Incidentally, that commit fixed **other** Intel issues that first appeared in commit https://github.com/gonetz/GLideN64/commit/231463e3681606615efe56a2a52a3bab0458e593 cc: @loganmc10 so these are somehow related, however Mario's face was not affected:

clobber:
This works around a bug with Intel GPU's on macOS where colors can appear wrong.


loganmc10:
The only potential problem I see with this: based on my reading, ivec4 may not be supported on GLES2. Perhaps someone on a Raspberry Pi could try it out @dankcushions ? Make sure that Legacy Blending is disabled

clobber:
A quick search shows that seems to be the case. So then we have to decide how we want to `#ifdef` around that.

loganmc10:
I'd wait for a test, this metions ivec: https://www.khronos.org/opengles/sdk/docs/reference_cards/OpenGL-ES-2_0-Reference-card.pdf

So it's unclear to me whether it supports it or not. It supports equal() which returns a bvec, so I don't see how it wouldn't support ivec. I just see some Stack Overflow posts with people saying it's not supported

gonetz:
uBlendMux1 is already declared as ivec4:
`uniform lowp ivec4 uBlendMux1`
It is common code for all GL versions. This code works now. It would be reported long ago if GLES2 had problems with it.

gonetz:
Merged: 3c8689d848

PR #1611, opened 2017-10-01
Simplified and complete workaround for #1601

clobber:
This properly worksaround the bug with Intel drivers on the Mac, which initially broke in parts of commits d4dbe7c094509df0f843b94d898f69a57f13d6d7 and 231463e3681606615efe56a2a52a3bab0458e593

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
@@ -463,10 +463,9 @@ public:
 class ShaderBlender1 : public ShaderPart
 {
 public:
-	ShaderBlender1(const opengl::GLInfo & _glinfo)
+	ShaderBlender1()
 	{
 #if 1
-		if (_glinfo.renderer != opengl::Renderer::Intel) {
 			m_part =
 				"  muxPM[0] = clampedColor;													\n"
 				"  lowp vec4 vprobe = vec4(0.0, 1.0, 2.0, 3.0);								\n"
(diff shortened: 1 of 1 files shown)


clobber:
Okay, after pulling my hair from playing whack-a-mole with regressions introduced in the earlier fix https://github.com/gonetz/GLideN64/commit/3c8689d84831dca7d768892826a5de2495ee9633, I went back and looked at @loganmc10's original optimizations https://github.com/gonetz/GLideN64/pull/1581#issue-259288911 and found it strange that it was actually working for me without issues but noticed it looked different from the code that was finally merged in https://github.com/gonetz/GLideN64/commit/231463e3681606615efe56a2a52a3bab0458e593

It was only until I added in the F-Zero X borders fix is when it all went to hell. So what I did was add a `#ifdef` to revert back to a 1-liner that still keeps the F-Zero X fix without breaking anything else.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -480,30 +479,15 @@ class ShaderBlender1 : public ShaderPart
 				"    lowp vec4 blend1 = (muxpm0 * muxa) + (muxpm2 * muxb);					\n"
 				"    clampedColor.rgb = clamp(blend1.rgb, 0.0, 1.0);						\n"
 				"  } else {																	\n"
+// Workaround for Intel drivers for Mac, issue #1601
+#if defined(OS_MAC_OS_X)
+				"      clampedColor.rgb = muxPM[uBlendMux1[0]].rgb;							\n"
+#else
 				"      lowp vec4 muxpm0 = muxPM * vec4(equal(vec4(uBlendMux1[0]), vprobe));	\n"
 				"      clampedColor.rgb = muxpm0.rgb;										\n"
+#endif
```
gonetz: I don't understand.
Does Intrel drivers work ok with 
	"    lowp vec4 muxpm0 = muxPM*vec4(equal(vec4(uBlendMux2[0]), vprobe));		\n"
in "  if (uForceBlendCycle2 != 0) " part of shader,
but fails on the same string in "else" part ?!
It is crazy. @clobber  Are you sure that your Mac is hapy with this version of blender shaders and all green color issues gone with it?

clobber: Yeah I don't understand it either...some really odd bug. I'm sure that all the incorrect texture color issues are gone. I checked 20 games in multiple areas and the issue is gone. It seems @loganmc10's original code in his PR was fine until the F-Zero X fix was added in. Now everything is good though.

clobber: However, if you have an idea of certain games I should try, let me know. Otherwise, I think it all looks good.

gonetz: Ok

loganmc10:
Just a minor style thing, everything should have been shifted left with the removal of the if statement

gonetz:
Did not notice that, thanks

PR #1618, opened 2017-10-02
Use rand_s for MINGW

loganmc10:
I recently discovered that rand() doesn't work in MinGW (it's not really random, it gives the same sequence of numbers each time, regardless of the seed). rand_s works however.

### DIFF
--- a/src/NoiseTexture.cpp
+++ b/src/NoiseTexture.cpp
@@ -1,3 +1,7 @@
+#ifdef MINGW
+#define _CRT_RAND_S
+#endif
+
 #include <thread>
 #include <array>
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


src/NoiseTexture.cpp
```diff
@@ -35,8 +39,15 @@ void FillTextureData(u32 _seed, NoiseTexturesData * _pData, u32 _start, u32 _sto
 	for (u32 i = _start; i < _stop; ++i) {
 		auto & vec = _pData->at(i);
 		const size_t sz = vec.size();
-		for (size_t t = 0; t < sz; ++t)
-			vec[t] = rand() & 0xFF;
+		unsigned int rand_value;
```
gonetz: move this code to 
static u32 Rand(u32 rand_value)
{
#ifdef MINGW
			rand_s(&rand_value);
#else
+			rand_value = rand();
+#endif
  return rand_value;
}

Use:
...
		u32 rand_value;
		for (size_t t = 0; t < sz; ++t) {
			rand_value = Rand(rand_value );
			vec[t] = rand_value & 0xFF;
		}

Thus we will localize #ifdef MINGW in one place.

loganmc10: Done

gonetz:
Thanks!
Merged: 54a1c4a65cff1

Issue #1616, opened 2017-10-02
VidExt_ResizeWindow is called if fullscreen is enabled


gizmo98:
**System:**
- Intel Core i3 with integrated GPU
- Ubuntu 16.04 with hardware enablement stack HWE
- Monitor resolution is 1024x768
- SDL 2.0.4
- fresh compiled mupen64plus and plugins

**Problem:**
If mupen64plus is started with `--fullscreen` display flickers and there is no picture. According to logfile VidExt_ResizeWindow() is called in fullscreen mode multiple times. The requested (fullscreen) resolution of 1024x796 is not supported. 
[runcommand.txt](https://github.com/gonetz/GLideN64/files/1349293/runcommand.txt)

```
Core Error: VidExt_ResizeWindow() called in fullscreen mode.
(EE) Error setting videomode 1024x796
(II) Setting video mode 1024x768...
Core: Setting video mode: 1024x768
```

I don't think VidExt_ResizeWindow() should be called if fullscreen is enabled.

loganmc10:
Do any of the video plugins actually work with --fullscreen?

This works with mupen64plus-gui (https://github.com/m64p/mupen64plus-gui)

gizmo98:
mupen64plus-video-glide64mk2 works. I have not tested any other plugin.

loganmc10:
Does it change your monitor's resolution to match the resolution in the VideoGeneral config section? Or does it upscale the image to match the current monitor resolution?

loganmc10:
But either way, stopping ResizeWindow from being called in fullscreen won't break mupen64plus-gui.

https://github.com/gonetz/GLideN64/blob/master/src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp

m_bFullscreen needs a bit of work, it's not properly tracking the fullscreen state (it doesn't get toggled when _changeWindow is called)

Then inside ```DisplayWindowMupen64plus::_resizeWindow()``` you can just wrap ```CoreVideo_ResizeWindow()``` instead a check for !m_bFullscreen

gizmo98:
This ubuntu SDL2 version tries to change monitor resolution. But resolution change faily if requested resolution is higher than max Monitor resolution.

gizmo98:
@loganmc10 
I don't think mupen64plus-gui is broken. I just see flickering if pc tries to change modes all the time.

[...]

PR #1621, opened 2017-10-03
Refine fullscreen toggle logic

gizmo98:
-Toggle m_bFullscreen if _changeWindow is executed.
-Don’t allow _resizeWindow if m_bFullscreen is true.
-_resizeWindow returns false if fullscreen is enabled.

Changes were tested with Ubuntu 16.10 and Intel Atom CPU. Fullscreen toggling is working. There is no flickering anymore.

https://github.com/gonetz/GLideN64/issues/1616

### DIFF
--- a/src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
+++ b/src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
@@ -131,12 +131,21 @@ bool DisplayWindowMupen64plus::_resizeWindow()
 	m_bFullscreen = false;
 	m_width = m_screenWidth = m_resizeWidth;
 	m_height = m_screenHeight = m_resizeHeight;
-	if (CoreVideo_ResizeWindow(m_screenWidth, m_screenHeight) != M64ERR_SUCCESS) {
-		printf("(EE) Error setting videomode %dx%d\n", m_screenWidth, m_screenHeight);
-		m_width = m_screenWidth = config.video.windowedWidth;
-		m_height = m_screenHeight = config.video.windowedHeight;
-		CoreVideo_Quit();
-		return false;
+	switch (CoreVideo_ResizeWindow(m_screenWidth, m_screenHeight)) 
+	{
+		case M64ERR_INVALID_STATE: 
+			printf("(EE) Error setting videomode %dx%d in fullscreen mode\n", m_screenWidth, m_screenHeight);
+			m_width = m_screenWidth = config.video.windowedWidth;
+			m_height = m_screenHeight = config.video.windowedHeight;
+			break;
+		case M64ERR_SUCCESS:
+			break;
+		default:
+			printf("(EE) Error setting videomode %dx%d\n", m_screenWidth, m_screenHeight);
+			m_width = m_screenWidth = config.video.windowedWidth;
+			m_height = m_screenHeight = config.video.windowedHeight;
+			CoreVideo_Quit();
+			return false;
 	}
 	_setBufferSize();
 	opengl::Utils::isGLError(); // reset GL error.


src/DisplayWindow.cpp
```diff
@@ -75,8 +75,9 @@ bool DisplayWindow::resizeWindow()
 	if (!m_bResizeWindow)
 		return false;
 	m_drawer._destroyData();
-	if (!_resizeWindow())
-		_start();
+	if (!m_bFullscreen)
+		if (!_resizeWindow())
```
loganmc10: Won't this mean that if ```m_bFullscreen == true```, _resizeWindow() will return false, and thus _start() will be called, shouldn't it just do nothing?

gizmo98:
No. It means if m_bFullscreen != true run _resizeWindow. So _resizeWindow will not be called. If it was the other way resizing in window mode wouldn’t work. I have not seen the log message anymore.

loganmc10:
Oh right for some reason I really focused in on thst function, I didn't realize it won't get called in that scenario. I'll test this out in with mupen64plus-gui in a couple minutes

loganmc10:
I still get:

```
Core Error: VidExt_ResizeWindow() called in fullscreen mode.
```
if I try and toggle fullscreen using Alt+Enter in mupen64plus-ui-console. If I start it with --fullscreen, it doesn't properly change my monitor resolution, but that is probably just SDL failing at that. The current master doesn't give me any errors either with --fullscreen, but also doesn't change my monitor resolution

gizmo98:
This should not happen if m_bFullscreen is tracked correctly. Does it happen once? If _resize() fails _start() should read ini settings of m_bFullscreen and run a fresh CoreVideo_SetVideoMode(). m_bFullscreen should be set right then.

loganmc10:
It can only happen once, it segfaults after that message. I'm not sure, maybe because the fullscreen toggle failed? It's never really worked right on my laptop using ui-console.

gizmo98:
I have added a check which toggles m_bFullscreen only if VidExt_ToggleFullScreen() returns success.

loganmc10:
I re-wrote that check, it's not currently an if statement, but I still get the same message/crash. I'll try and spend some time tomorrow figuring out why it's acting this way, I'll try it on Windows too

gizmo98:
To early in the morning. I will rewrite or remove this commit.

loganmc10:
I tested this in Windows and toggling works with ui-console and mupen64plus-gui. It doesn't work in Linux, but it didn't work before, SDL can't change the resolution properly so it gets messed up, but there is nothing wrong with this commit in that respect.

The only thing this breaks is starting a game in fullscreen using mupen64plus-gui. The problem is that CoreVideo_SetMode tries to change the monitor resolution, but mupen64plus-gui doesn't support that. It basically just resizes the window to a fullscreen window, and that triggers a resizeWindow call, but that gets ignored because m_bFullscreen is true.

Can your problem be fixed by simply protecting the CoreVideo_ResizeWindow call?

```
diff --git a/src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp b/src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
index 6272c99..2d9ebbe 100644
--- a/src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
+++ b/src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
@@ -128,15 +128,16 @@ bool DisplayWindowMupen64plus::_resizeWindow()
 {
        _setAttributes();

-       m_bFullscreen = false;
        m_width = m_screenWidth = m_resizeWidth;
        m_height = m_screenHeight = m_resizeHeight;
-       if (CoreVideo_ResizeWindow(m_screenWidth, m_screenHeight) != M64ERR_SUCCESS) {
-               printf("(EE) Error setting videomode %dx%d\n", m_screenWidth, m_screenHeight);
-               m_width = m_screenWidth = config.video.windowedWidth;
-               m_height = m_screenHeight = config.video.windowedHeight;
-               CoreVideo_Quit();
-               return false;
+       if (m_bFullscreen == false) {
+               if (CoreVideo_ResizeWindow(m_screenWidth, m_screenHeight) != M64ERR_SUCCESS) {
+                       printf("(EE) Error setting videomode %dx%d\n", m_screenWidth, m_screenHeight);
+                       m_width = m_screenWidth = config.video.windowedWidth;
+                       m_height = m_screenHeight = config.video.windowedHeight;
+                       CoreVideo_Quit();
+                       return false;
+               }
        }
        _setBufferSize();
        opengl::Utils::isGLError(); // reset GL error.
@@ -145,7 +146,8 @@ bool DisplayWindowMupen64plus::_resizeWindow()

 void DisplayWindowMupen64plus::_changeWindow()
 {
-       CoreVideo_ToggleFullScreen();
+       if (CoreVideo_ToggleFullScreen() == M64ERR_SUCCESS)
+               m_bFullscreen != m_bFullscreen;
 }

 void DisplayWindowMupen64plus::_getDisplaySize()
```

I tested that in Windows and it works in both frontends

gizmo98:
@loganmc10 
If i remove these lines m_bFullscreen is always true if mupen64plus was started with --fullscreen. I modified my code to check the return value of CoreVideo_ResizeWindow. If return value is M64ERR_INVALID_STATE (fullscreen indicator) keep windowWidth and windowHeight.

loganmc10:
Ok, the way it is now looks good to me, that won't change anything with how mupen64plus-gui works

gizmo98:
@loganmc10 
Do you think it is ready to get merged?

loganmc10:
Yes i think it can be merged

gonetz:
Merged: db18139ca55

PR #1623, opened 2017-10-05
Add support for SP_STATUS_REG and RDRAM_SIZE in mupen64plus API

loganmc10:
The mupen64plus API recently gained support for passing SP_STATUS_REG and the RDRAM's size to the GFX plugin.

This (SP_STATUS) should hopefully be enough to properly implement the ucodes for WDC, Stunt Racer, and Gauntlet Legends.

Right now, the RSP plugins set the ```HALT | BROKE | TASKDONE``` bits of SP_STATUS after ProcessDList returns. I am going to modify the RSP plugin to set ```HALT | BROKE | TASKDONE``` before ProcessDList is called, and then the GFX plugin can unset those bits if it needs.

@Gillou68310 @olivieryuyu

EDIT: As should be obvious, GLideN64 will need to make sure SP_STATUS isn't ```nullptr``` before trying to use it when those ucodes are implemented

### DIFF
--- a/src/MupenPlusPluginAPI.cpp
+++ b/src/MupenPlusPluginAPI.cpp
@@ -1,10 +1,15 @@
 #include "PluginAPI.h"
 #include "Types.h"
+#include "mupenplus/GLideN64_mupenplus.h"
+#include "N64.h"
 
 extern "C" {
 
 EXPORT int CALL RomOpen(void)
 {
+	if (rdram_size != nullptr)
+		RDRAMSize = *rdram_size - 1;
+
 	api().RomOpen();
 	return 1;
 }
@@ -43,7 +48,7 @@ EXPORT void CALL SetRenderingCallback(void (*callback)(int))
 {
 	api().SetRenderingCallback(callback);
 }
-	
+
 EXPORT void CALL ResizeVideoOutput(int width, int height)
 {
 	api().ResizeVideoOutput(width, height);
--- a/src/N64.cpp
+++ b/src/N64.cpp
@@ -6,7 +6,7 @@ u8 *IMEM;
 u64 TMEM[512];
 u8 *RDRAM;
 
-u32 RDRAMSize;
+u32 RDRAMSize = 0;
 
 N64Regs REG;
 
--- a/src/N64.h
+++ b/src/N64.h
@@ -32,6 +32,8 @@ struct N64Regs
 	u32 *VI_V_BURST;
 	u32 *VI_X_SCALE;
 	u32 *VI_Y_SCALE;
+
+	u32 *SP_STATUS;
 };
 
 extern N64Regs REG;
--- a/src/RSP.cpp
+++ b/src/RSP.cpp
@@ -219,24 +219,26 @@ void setDepthClearColor()
 
 void RSP_Init()
 {
+	if (RDRAMSize == 0) {
 #ifdef OS_WINDOWS
-	// Calculate RDRAM size by intentionally causing an access violation
-	u32 test;
-	try
-	{
-		test = RDRAM[0x007FFFFF] + 1;
-	}
-	catch (...)
-	{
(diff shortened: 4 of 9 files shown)


loganmc10:
As an example, here is the work that @LegendOfDragoon did to make Gauntlet Legends boot, integrated into this PR (tested and working in HLE with mupen64plus):

https://github.com/loganmc10/GLideN64/commit/e388f3b10570ad7e36ddbebda73e1e974c5a5315

I think @olivieryuyu mentioned that he found a way to get Gauntlet Legends to work without this? So this might not be the correct approach, but perhaps it is

gonetz:
Merged: daa567ec715a72

Thanks for the link on LegendOfDragoon commit. I did not look at Gauntlet Legends issues yet.

olivieryuyu:
This commit should avoid to have a hack for the yielding.

It may not solved the missing graphics and related related potential flickering. Not 100% sure though :)

gonetz:
@loganmc10 could you give me updated build of mupen64plus? I mean your build with gui.

loganmc10:
@gonetz here you go:
[mupen64plus-GLideN64.zip](https://github.com/gonetz/GLideN64/files/1363000/mupen64plus-GLideN64.zip)

That is a 32-bit build, if you try out Gauntlet Legends, the audio is really terrible unless you set CountPerOp=1 (Settings->Core and Plugin Settings->Core, near the bottom). The audio is still pretty crackly with CountPerOp=1, I don't know if it will be any better in PJ64.

@olivieryuyu This does fix the flickering, I can't remember what else was wrong with the image ,but this is what it looks like:

gauntlet_legends-000

gonetz:
> This does fix the flickering,

Could you explain how it works? I missed all that discussions about yielding.

loganmc10:
Usually an RSP task must return with the SP_STATUS_REG bits ```TASKDONE | HALT | BROKE``` set, indicating that the RSP task is complete. There are some tasks that seem to hang, waiting for the CPU to do something before continuing. Since the emulators are single-threaded, the CPU never executes any cycles while the RSP task is executing, so it just hangs forever.

LLE RSP plugins like PJ64's or HatCat's will detect if they are stuck in a loop, and if they are, they will return from the RSP task with the ```TASKDONE | HALT | BROKE``` bits **unset**, indicating that the RSP is still unfinished, and that the CPU should run for a while, then return to the same RSP task.

So this is basically that same system, if it detects it's stuck in a loop, it unsets those SP_STATUS bits, and returns. The CPU will do its thing for a while, then when ProcessDList is called the next time, things are in order for it to continue without getting stuck

Frank-74:
Flickering started when you entered the first level, when power or magic hasn't filled up.  The flickering  stops when the bar is full.  Shoot, the bar will go down and flickering starts again until the bar fills up again.
The bar is above the heart, bottom left.  It's broken in the image above.

This is what happened before on Project64.

gonetz:
@loganmc10 Thanks!

olivieryuyu:
I investigated again the matter and my theory is most likely incorrect :(

So yielding is the primary issue. Remains the issue with the powers indeed.

BigheadSMZ:
So I've been recently getting back into N64 emulation again after years. I'm amazed at how far this plugin has come, many of the problematic games from the past now work great. I've been using the build of mupen64plus and GlideN64 found here: https://m64p.github.io/

It was mentioned here that the flickering for Gauntlet Legends was fixed. Is there some option I need to enable or some specific setup required? I can't seem to figure out how to get the flickering to stop short of using angrylion's plugin. When I enter a stage, it stops there, but the hub world flickers as it always did.

loganmc10:
Thanks for reporting @BigheadSMZ. I found the problem commit and opened a new issue: https://github.com/gonetz/GLideN64/issues/1705

Issue #1626, opened 2017-10-08
Valgrind memory analysis report


fzurita:
I did a valgrind run using mupen64plus in a linux environment. This was with this revision for reference:

e4f29a1eb8d93b548cf39178a71a5d97667e39cd

Here is the output:

[valgrind.txt](https://github.com/gonetz/GLideN64/files/1366629/valgrind.txt)

Things that are note-worthy are the we are using many uninitialized variables at startup. This could be why some games have some issues at startup, like Pokemon Stadium 2 and the background when Depth buffer to RDRAM is enabled, yet it works fine when a save state is loaded.

Edit: I did a cursory check and the issues seem to be real.

fzurita:
This seems to fix https://github.com/gonetz/GLideN64/issues/1582 without having to rely on an entry for GLideN64.custom.ini to set ```DepthBufferToRDRAM=0```.

Issue #1582, opened 2017-09-21
 pokemon stadium 2 gliden64 start screen bug


squalorx:
Question about project 64 with gliden64 (both of them on latest version), when i play pokemon stadium 2 the screen that ask me to press start and the next screen (where you can choose battle, event and white city) are broken, the game works fine, but that 2 screen are broken. The funny thing is, i already played that game before on the same emulator and with the same plugin without this bug,but it was a long time ago(probably at the end of 2016). If i go in and out of full screen the image gets back to normal. This bug doesn't make the game unplayable but it bothers me everytime i start the game. 
i'm using the Ntsc-U version, the same happens if i use the japanese version. If i use the European version this bug won't happen, but sometime the emulator will go black screen close after the intro if i use the europen version ( something like 60% of the times that i try to open it).

PS: sorry for any english mistake, english is not my first language.

screenshots of the problem
https://i.imgur.com/fAC4lmr.png
https://i.imgur.com/g1oXThz.png
https://imgur.com/4x4cjbt

spec: gtx 630 1gb
fx 6300
8 gb memory
 i'm using windows 10 with the creators update

loganmc10:
You need GLideN64.custom.ini

https://github.com/gonetz/GLideN64/blob/master/ini/GLideN64.custom.ini

I think for Project64 you put that in the same folder as the DLL file

squalorx:
thanks for the info, i will try it when i get home

squalorx:
thanks mate, the custom ini file fixed it.

PR #1631, opened 2017-10-10
Fix valgrind issues

fzurita:
Fixes https://github.com/gonetz/GLideN64/issues/1626

Valgrind comes out clean after these changes.

One thing I'm not quite sure about is the default ```size``` attribute for frame buffer textures. ```glsl::UFrameBufferInfo::update``` depends on the size attribute to be set to something, but at startup we are calling this before the size attribute is ever being set. I defaulted it to 2 for now.


fzurita:
This seems to fix https://github.com/gonetz/GLideN64/issues/1582 without having to rely on an entry for GLideN64.custom.ini to set ```copyDepthToRDRAM=0```.

gonetz:
Thanks!

fzurita:
@gonetz Thanks for closing this so quickly, but did you see my question about the initial size parameter value?

gonetz:
> did you see my question about the initial size parameter value?

I did not notice it, sorry.

> I guess it doesn't matter, as long as it prevents the monochrome modes or fixed alpha from being updated on textures that don't need it.

You should be right.

PR #1640, opened 2017-10-25
Fix flickering when rdram to color buffer is enabled

fzurita:
I think clearing RDRAM only makes sense if we are not copying color buffer to RDRAM ourselves.

Without this, if copy color to RDRAM happens at 60 VI/s and copy color buffer to RDRAM happens at a lower rate, we get flickering in the screen. For example Mario Artist - Paint Studio.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ Roms/
 *.user
 projects/msvc15/*db
 projects/msvc15/GeneratedFiles
+projects/msvc15/libGLideNHQ.dir
 projects/msvc12/*db
 projects/msvc12/GeneratedFiles
 src/Revision.h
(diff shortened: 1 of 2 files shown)


loganmc10:
This mostly fixes the flickering in Pokemon Snap when EnableCopyColorFromRDRAM is enabled (so the red dot works), however the borders of the image still flicker when you "target" with Z. Without this change the whole image is constantly flickering if EnableCopyColorFromRDRAM is on

fzurita:
Ok, I can take a look at that tomorrow.

AmbientMalice:
I believe the Wii VC version of this game uses a hack where only the buffer area inside the borders is copied. Regardless, this is a great find and hopefully it helps in other areas, too.

fzurita:
I think I see why the flickering at the borders of the screen when taking a picture is happening. I"ll have to think of a way to fix it.

gonetz:
Merged: 77a777162cff5
Thanks!

PR #1648, opened 2017-11-06
change msvc15 to msvc14

rez-:
Renamed msvc15 directory to msvc14 because Visual Studio 2015 has actually Visual C++ version 14.

Anyhow great work gonetz and all other contributors!


gonetz:
merged: 764275887

fzurita:
This seems like a strange pull request. Most other projects I've seen go by visual studio version, not Visual C++ version.

gonetz:
Visual studio 13 project already in msvc12 and I will not change it. So, msvc14 following the same logic.

gonetz:
I agree, go by visual studio version is less confusing.

gonetz:
I reverted this commit and renamed msvc15 to msvc2015. This is QT style. For example: 
win32-msvc2015
# Written for Microsoft Visual C++ 2015
msvc2015 is less confusing than msvc14.

rez-:
Then shouldn't we update the msvc12 directory to msvc2013 as well?

gonetz:
We should, but we will not. I use MSVS2013 to find regressions. It will be very inconvenient for me to check older commits if project folder will be renamed.

PR #1657, opened 2017-11-11
Copy 4k block in FBInfo Chunk copies

loganmc10:
The spec says:
```
DLL should copy 4KB block content back to RDRAM frame buffer.
Emulator should not call this function again if other memory is read within the same 4KB range
```

Unfortunately the wording is a bit ambiguous. However, I read "4k block" as a 4k aligned block of memory, not ```address + 0x1000```

This is how mupen64 works, and mupen64plus as well. They are the only emulators that actually implement this 4k block checking (1964 never implemented it, it calls FBRead for each read). I think we should stick to this standard, since checking 4k blocks is much easier (can use bit shifting), rather than checking "address + 0x1000", which can't be checked using bit shifting.

I haven't looked too closely in the _copy() function, but it should be making sure that startAddress and endAddress are actually inside the FB, and trim the copy if they aren't. (with this change, startAddress might be outside the FB now)

### DIFF
--- a/src/BufferCopy/ColorBufferToRDRAM.cpp
+++ b/src/BufferCopy/ColorBufferToRDRAM.cpp
@@ -310,7 +310,8 @@ void ColorBufferToRDRAM::copyChunkToRDRAM(u32 _address)
 {
 	if (!_prepareCopy(_address))
 		return;
-	_copy(_address, _address + 0x1000, true);
+	const u32 addr = _address & ~0xfff;
+	_copy(addr, addr + 0x1000, true);
 }
 
 
--- a/src/BufferCopy/DepthBufferToRDRAM.cpp
+++ b/src/BufferCopy/DepthBufferToRDRAM.cpp
@@ -299,6 +299,6 @@ bool DepthBufferToRDRAM::copyChunkToRDRAM(u32 _address)
 	if (!_prepareCopy(_address, true))
 		return false;
 
-	const u32 endAddress = _address + 0x1000;
-	return _copy(_address, endAddress);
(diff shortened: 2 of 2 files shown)


loganmc10:
@gonetz I've been looking into this. Your check (don't FBRead if FBWrite happened on same address) is a good check, but it's not complete.

I am testing Super Mario. Even 1964 misses some of the reads/writes, but if I track every read/write, I can see what is happening.

You only check the first write, then ignore any reads related to that address. You need to check all the writes, and ignore any reads related to those writes. For instance, this happens:

write addr 1
write addr 2
read addr 1
read addr 2

Right now, you only ignore "read addr 1", you need to be ignoring both reads, did I make sense?

loganmc10:
Actually @gonetz the more I think about it, that should be a check performed by the emulator. The emulator shouldn't be sending the GFX plugin bogus FBRead calls. I'm going to implement that check in mupen64plus (don't call FBRead if FBWrite was called for that address). You're free to implement that check in GLideN64 as well of course, but 1964 doesn't track all the reads/writes, so I don't think it will solve the problem in 1964

gonetz:
> You only check the first write, then ignore any reads related to that address. You need to check all the writes, and ignore any reads related to those writes.

Not exactly. Plugin defines frame buffer by FBWrite address. This buffer becomes protected. All FBReads with addresses from this buffer will be ignored. If FBWrite will access another buffer, that buffer will not be protected, and subsequent FBReads to that buffer will break buffer's data. I'll do changes to protect all buffers.

loganmc10:
Sorry I'm using the terms address and buffer a little loosely, being lazy with my words

PR #1658, opened 2017-11-12
Fix Pokemon Snap border flicker

fzurita:
This flicker happens while zooming in to take  a picture while RDRAM to color buffer and color buffer to RDRAM are both enabled.

Also, for some reason I had to switch ```std::min``` to ```min``` after I included FrameBufferInfo.h


gonetz:
> Also, for some reason I had to switch std::min to min after I included FrameBufferInfo.h

The code does not compile on Linux after that.

gonetz:
Corrected and merged: 538674da516

PR #1661, opened 2017-11-14
Fix chuck copy start address

loganmc10:
This is a PR into the fbinfo_fixes branch.

This fixes the freezing in Prof. Oak's check. However, the images still look wrong in the thumbnail (when you look at an individual image they look correct, but the thumbnails are wrong, however the check still works).

The requested start address might be before the actual start of the framebuffer, so this accounts for that possibility.


gonetz:
I see what your fix does, but I don't like that it duplicates code from _prepareCopy, so I refactored it a bit: 4adecb2c4bd7

Regarding wrong images with chunk copy enabled: Mupen64 sometimes does not reports about some buffer reads, thus some parts of buffer are not copied. Good example is Bomberman64 intro: 
gliden64_bomberman64u_000

PR #1667, opened 2017-11-21
Don't unset DP_STATUS_FREEZE

loganmc10:
See the discussion here: https://github.com/ata4/angrylion-rdp-plus/pull/41

The emulator should be handling setting/unsetting this bit, not the GFX plugin

### DIFF
--- a/src/RDP.cpp
+++ b/src/RDP.cpp
@@ -554,7 +554,7 @@ inline u32 READ_RDP_DATA(u32 address)
 void RDP_ProcessRDPList()
 {
 	if (ConfigOpen || dwnd().isResizeWindow()) {
-		dp_status &= ~0x0002;
+		//dp_status &= ~0x0002;
 		dp_start = dp_current = dp_end;
 		gDPFullSync();
 		return;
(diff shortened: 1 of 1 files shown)


src/RDP.cpp
```diff
@@ -554,15 +554,15 @@ inline u32 READ_RDP_DATA(u32 address)
 void RDP_ProcessRDPList()
 {
 	if (ConfigOpen || dwnd().isResizeWindow()) {
-		dp_status &= ~0x0002;
+		//dp_status &= ~0x0002;
```
gonetz: Why not just remove this line?

src/RDP.cpp
```diff
@@ -554,15 +554,15 @@ inline u32 READ_RDP_DATA(u32 address)
...
 		dp_start = dp_current = dp_end;
 		gDPFullSync();
 		return;
 	}
 
 	const u32 length = dp_end - dp_current;
 
-	dp_status &= ~0x0002;
+	//dp_status &= ~0x0002;
```
gonetz: Why not just remove this line?

loganmc10:
I thought it would be useful for "historical" purposes since it's been there for so long, but it could be removed as well.

gonetz:
Git will keep the history for us. I'll remove these lines.

gonetz:
Merged: a2793ace2324

PR #1670, opened 2017-11-23
UI: Gamma correction moved to emulation tab

Aquatakat:
A tab with one option on it feels off to me, so I put it near the other visual stuff on the emulation tab.


gonetz:
Thanks! Merged: d3fbff82f2ed2c1

PR #1672, opened 2017-11-26
Fix rare crash in opengl_BufferedDrawer.cpp

fzurita:
I got a crash in the memcpy in line 84. Should this check be this instead?


gonetz:
Merged. Thanks!

PR #1681, opened 2017-12-09
Further reduce shader logic

fzurita:
This is an ongoing attempt to continue to reduce shader logic as an experiment to improve performance on low end devices. I have removed only a few uniforms per commit to make it easy to find any regressions.

Although, I'm currently stuck, in the last commit where I removed ```uFbMonochrome``` and ```uFbFixedAlpha```, I'm getting this error in CBFD:

conker_bfd-002

Conker has a black spot in his shirt.

All I can say is that ```ShaderFragmentReadTex0```, ```_key.getFbMonochrome0()``` is coming out to 1 instead of zero when it shouldn't. @gonetz Any ideas?

### DIFF
--- a/src/Combiner.cpp
+++ b/src/Combiner.cpp
@@ -325,6 +325,12 @@ void CombinerInfo::setSecondaryParams(CombinerKey& _key)
 
 	//Render target
 	_key.setRenderTarget(gDP.colorImage.address == gDP.depthImageAddress ? gDP.otherMode.depthCompare + 1 : 0);
+
+	//Texture filter mode
+	unsigned int textureFilter = gDP.otherMode.textureFilter;
+	if ((gSP.objRendermode&G_OBJRM_BILERP) != 0)
+		textureFilter |= 2;
+	_key.setTextureFilter(textureFilter);
 }
 
 void CombinerInfo::setCombine(u64 _mux )
--- a/src/Combiner.cpp
+++ b/src/Combiner.cpp
@@ -310,7 +310,18 @@ void CombinerInfo::setSecondaryParams(CombinerKey& _key)
 	}
 	_key.setCvgXAlpha(gDP.otherMode.cvgXAlpha);
 
+	//Alpha compare
 	_key.setAlphaCompareMode(gDP.otherMode.cycleType < G_CYC_COPY ? gDP.otherMode.alphaCompare : 0);
+
+	//Dither modes
+	if (gDP.otherMode.cycleType < G_CYC_COPY) {
+		_key.setAlphaDither(gDP.otherMode.alphaDither);
+		_key.setColorDither(gDP.otherMode.colorDither);
+	}
+	else {
+		_key.setAlphaDither(0);
+		_key.setColorDither(0);
+	}
 }
 
 void CombinerInfo::setCombine(u64 _mux )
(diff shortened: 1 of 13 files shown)


fzurita:
@psyke83 @dankcushions can you test performance with this?

fzurita:
I have tested a few devices so far, old and new, and I'm seeing zero performance difference.

psyke83:
Unfortunately, there's severe graphical distortion and missing textures on Pi 3. I'll try to isolate the specific faulty commit(s) when I have free time.

fzurita:
Can you check for shader complain errors in the GLideN64 log?

fzurita:
I tested a couple more devices.

* A really old device that I have with a very weak CPU and a Mali GPU with GLES 2.0 gave me about a 15% performance drop. I suspect it's because it has to search through more shader keys now.
* A nexus player with GLES 3.1 gave me about a 10% performance improvement.
* Adreno devices with GLES 2.0 or 3.0 or 3.1 show zero performance change.

steeeb:
Tested a few games on a Raspberry Pi with this PR and it works.

psyke83:
Apologies for the delay - I've re-tested again.

Super Mario 64 seems OK with the latest state of this PR, but both Zelda games hang on a black screen (not even showing the N64 logo) due to the latest commit added: "Remove uniforms uFbMonochrome and uFbFixedAlpha". 

Removing that commit allows Majora'a Mask to boot again, but there is still a lot of graphical corruption and slowdown which starts during the part of the intro when an NPC carries a wooden log. The corruption is especially visible in the background. The slowdown/corruption appears to be resolved when removing the "Remove uniform uTextureFilterMode" commit.

fzurita:
@psyke83 Are you able to view the GLideN64 log file? Are you getting any errors?

psyke83:
There's nothing new in the log file when testing this PR:
```
Async color buffer copies are not supported on GLES2
LOD emulation not possible on this device
```

fzurita:
@psyke83 Can you run that last commit with one of the zelda games that give you the black screen and post the GLiden64.log file?

psyke83:
Here you go, tested with Majora's Mask:

https://gist.github.com/psyke83/5998e5452a9aec3b3997e2e489f9dcfc

gonetz:
< Here you go, tested with Majora's Mask:

Hmm...
* "Error in fragment shader ..."
First message must be what the error is. 
Utils::checkShaderCompileStatus or Utils::checkProgramLinkStatus must print the error first, but I don't see it in the log.

* I see body of both  YUV_Convert and TextureFilter functions, while the shader uses only TextureFilter , unconditionally.

fzurita:
@gonetz I made the last commit always print out the shader even though there was no error. Just in case there was an error, but the driver wasn't reporting it for some reason.

 I'm not sure why this code has so much problems in the raspberry pi 3, from the inspection of the log, I saw nothing that jumped out at me as odd.

Either way, the only device where I saw improvements in performance with this code was on PowerVR devices, and the performance increase was minimal, so I think this experiment proved that it didn't make much of a difference to reduce shader logic.

gonetz:
In that case I suggest to close that PR.
We moved texture conversion logic out of shader, lets stop on it.

PR #1690, opened 2017-12-19
Fix texture filter when using GLSL ES.

fzurita:
It looks like we just need to maintain high precision for vTexCoord0 and vTexCoord1.

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_ShaderStorage.h
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_ShaderStorage.h
@@ -20,7 +20,7 @@ namespace glsl {
 		bool _saveCombinerKeys(const graphics::Combiners & _combiners) const;
 		bool _loadFromCombinerKeys(graphics::Combiners & _combiners);
 
-		const u32 m_formatVersion = 0x19U;
+		const u32 m_formatVersion = 0x1AU;
 		const u32 m_keysFormatVersion = 0x04;
 		const opengl::GLInfo & m_glinfo;
 		opengl::CachedUseProgram * m_useProgram;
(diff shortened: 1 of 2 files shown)


src/Graphics/OpenGLContext/GLSL/glsl_ShaderStorage.h
```diff
@@ -20,7 +20,7 @@ namespace glsl {
 		bool _saveCombinerKeys(const graphics::Combiners & _combiners) const;
 		bool _loadFromCombinerKeys(graphics::Combiners & _combiners);
 
-		const u32 m_formatVersion = 0x19U;
+		const u32 m_formatVersion = 0x20U;
```
gonetz: 0x19 + 1 = 0x1A

fzurita: Whoops, I wasn't thinking hex. I'll fix.

fzurita:
Ok, I just tested a GLES 2.0 device and it doesn't fix the problem there. Textures are still pixelated. Did it fix your device? Also, I'm not seeing any performance difference in the GLES 2.0 device.

fzurita:
It seems that not all GL ES 2.0 devices support highp. For those devices, there is no way to fix. Have you tried your old GLES 2.0 device yet?

gonetz:
I have no GLES2 device with Android 4.4 to test.
It seems that standard hardware bilinear filtering is the only option for GLES2 gpu. I need to think, how to return it with minimal changes in code.

gonetz:
@fzurita  I just tested this fix on my Note3. Texture filtering works again. No slowdown noticed. Good work!

fzurita:
Great!

PR #1704, opened 2018-01-08
Adjust font defaults

remyabel:
On Linux systems (Fedora, Ubuntu, OpenSUSE, etc.) it's far more likely that they're going to have `LiberationMono` instead of `FreeSans` installed by default (which comes with a GNU package). Also, I adjusted the font paths because `/usr/share/fonts` is universal, but not all fonts will be installed to the same subdirs.


gonetz:
Merged: 30f56c98bd62

PR #1717, opened 2018-02-12
Implement BOSS ZSort ucode (WDC, Stunt Racer)

Gillou68310:


gonetz:
Thanks!

loganmc10:
A couple things:

The Android Makefile still needs to be updated

I would change the LOG_WARNING to LOG_VERBOSE. That warning isn't useful to an end user, it'll only serve to potentially cause performance issues by writing to the log file mid-game

PR #1720, opened 2018-02-15
Add European version of World Driver Championship.

bsmiles32:


Gillou68310:
@gonetz this is a perfect example for https://github.com/Gillou68310/GLideN64/commit/7f20b53972ed0e64b00e71723b484e59a9577596
Using uc_size will produce the same CRC for both EU and US versions, since both ucodes are identical.

bsmiles32:
Note that ucode size field may not always contain the true ucode size.
Indeed, ucode_boot ignore ucode size field and just copy 0xF7F bytes to imem.

In rsp-hle we checksum  min(uc_size, 0xf80)/2 bytes to identify ucodes. I believe the /2 divider is there to reduce computation time and avoid trailling garbage data polluting the checksum.
Given the current number of ucodes that the rsp-hle has to handle, this approach is sufficient.

gonetz:
> this is a perfect example for Gillou68310/GLideN64@7f20b53

It is a good fix, but I don't have time to update specialMicrocodes list.

Gillou68310:
Yeah don't worry I will update it, I was just waiting for your approval ;-)

gonetz:
There are two ucodes in the list without comments which games they belong. It can be a problem.

PR #1726, opened 2018-02-21
Allow triple buffering in async copy mode

loganmc10:
Some devices are still too slow to handle the current async copies (mobile devices).

This allows using 3 buffers in the ring instead of 2. The config will be backward compatible, since "2" currently means async, but GLideNUI will still need to be updated.

I tested Mario Kart and the scoreboard still looks good, although it will be a frame further behind when using triple buffering. For some reason I can't think of other games that use this.... but I'm sure there are plenty.

Of course some games won't be able to make good use of this, like the puzzle in Banjo-Kazooie, but for other games this will allow better performance while still providing the copies.


loganmc10:
@gonetz 

```
config.frameBufferEmulation.copyToRDRAM != Config::ctSync
```
Is not correct, since that will be true if it is ctDisable. It should be ">" to check for double or triple buffering

gonetz:
Yep, sorry!

PR #1727, opened 2018-02-22
Keep track of FB attachments

loganmc10:
This is a minor optimisation to reduce the number of times glFramebufferTexture2D and glBindFramebuffer is called.

It just keeps track of the attachments and ignores redundant calls. glFramebufferTexture2D itself probably isn't an expensive all, but glBindFramebuffer can be expensive on mobile chipsets and should be avoided when possible

### DIFF
--- a/src/Graphics/Context.cpp
+++ b/src/Graphics/Context.cpp
@@ -90,9 +90,9 @@ ObjectHandle Context::createTexture(Parameter _target)
 	return m_impl->createTexture(_target);
 }
 
-void Context::deleteTexture(ObjectHandle _name)
+void Context::deleteTexture(ObjectHandle _name, bool _isFBTexture)
 {
-	m_impl->deleteTexture(_name);
+	m_impl->deleteTexture(_name, _isFBTexture);
 }
 
 void Context::init2DTexture(const InitTextureParams & _params)
--- a/src/Graphics/Context.h
+++ b/src/Graphics/Context.h
@@ -65,7 +65,7 @@ namespace graphics {
 
 		ObjectHandle createTexture(Parameter _target);
 
-		void deleteTexture(ObjectHandle _name);
+		void deleteTexture(ObjectHandle _name, bool _isFBTexture);
 
 		struct InitTextureParams {
 			ObjectHandle handle;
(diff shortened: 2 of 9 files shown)


LegendOfDragoon:
>It just keeps track of the attachments and ignores redundant calls.

Good work. I like to see these kind of improvements :smile: . Have you noticed any difference in performance?

loganmc10:
> Have you noticed any difference in performance?

I haven't really tested that, I assume the only difference would be on mobile devices, and even then it would really depend on how good the GPU driver is at dealing with these inefficiencies.

I just noticed that it was doing this a lot, for instance, here is the beginning of a frame from Mario Kart 64 (in a race):
old

And here is the same part after this change:
new

So it was calling glFramebufferTexture2D needlessly 4 times every single frame (along with the associated glBindFramebuffer calls).

It all depends on the GPU driver, the driver might have a system to detect needless calls just like I implemented, but it might not, there is no way of knowing.

gonetz:
This optimization causes heavy glitches in Paper Mario. Load this save:
https://drive.google.com/file/d/18PxLxRJkOvUEiH8WZ6RzEbqR5BNp0si6/view?usp=sharing
move right to next room, then back.

Optimization like this is not a good idea in general.
For buffers, there are several color buffers and one depth buffer.
FBO attaches color and depth buffer. Color buffer is always the same for given FBO, depth buffer can change. If program will decide that the old depth buffer is not valid anymore, it will remove it and create a new one. There is high probability that new buffer will have the same name. Name in OpenGL is 32bit unsigned number. Your optimization caches depth buffer attachment. Thus, new depth buffer with the same name will not be attached.

loganmc10:
> Optimization like this is not a good idea in general.

I agree when it's not done properly or has bugs, but if it's done bug free it has the same benefit as all the other caching that is done.

I could reproduce that Paper Mario issue, I had accounted for when the framebuffer is deleted, but not when the texture is deleted, I fixed that here: https://github.com/gonetz/GLideN64/pull/1727/commits/77d5f7de42f8c1951b8623241398421ec752e3a9 that fixes the Paper Mario issue

PR #1728, opened 2018-02-22
Only do depth hack for GLES devices

loganmc10:
This was actually a hack to fix some depth issue on certain Android devices (I want to say it was only PowerVR devices, but I can't remember exactly).

According to at least one GPU vendor (Adreno): https://developer.qualcomm.com/qfile/28557/80-nu141-1_b_adreno_opengl_es_developer_guide.pdf

> The Adreno driver handles glClear() calls by drawing viewport-covering quads into the frame
buffer, using clear color values specified earlier by the application. Clear calls operate at the
normal pixel rate.
Avoid redundant clears as they are costly.

So ideally this would actually just apply to whatever GPU driver had the issue, but at least it should apply to GLES devices only (not OpenGL/Nvidia devices that have good drivers)

### DIFF
--- a/src/Graphics/OpenGLContext/opengl_ContextImpl.cpp
+++ b/src/Graphics/OpenGLContext/opengl_ContextImpl.cpp
@@ -143,8 +143,10 @@ void ContextImpl::clearDepthBuffer()
 	enableScissor->enable(false);
 
 #ifdef OS_ANDROID
-	depthMask->setDepthMask(false);
-	glClear(GL_DEPTH_BUFFER_BIT);
+	if (m_glInfo.isGLESX) {
+		depthMask->setDepthMask(false);
+		glClear(GL_DEPTH_BUFFER_BIT);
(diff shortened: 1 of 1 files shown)


loganmc10:
@fzurita

fzurita:
This fix was only needed for PowerVR devices. It's a good idea to remove the compile time check and limit it only to GLES. Although, I think we should limit it even further to only PowerVR devices.

If you want I can make my own pull request that checks only PowerVR since I have a device handy that exhibits the issue.

fzurita:
Also, if you notice, we set the depth mask to false, so in reality, the GPU driver should be doing nothing after the glClear call.

loganmc10:
> If you want I can make my own pull request that checks only PowerVR since I have a device handy that exhibits the issue.

Yeah that would be best, I'm not sure what the GL vendor string looks like exactly to do that.

PR #1729, opened 2018-02-22
Cache glTexParameter calls

loganmc10:
This PR depends on https://github.com/gonetz/GLideN64/pull/1727. Please only focus on commit https://github.com/gonetz/GLideN64/commit/83b8ba35bf8b87f7f67d115e1610639afb0188b9 for this PR, the other commit is from the other PR.

This furthers the work of optimizing the GL calls.

I profiled the application on my laptop, here are the top 10 GL calls from 60 seconds of GoldenEye (sorted by the total time spent servicing each call):

image

Most of that is pretty normal, except you can see that glTexParameteri is called about 1.5 million times, and it's number 4 on the list.

Here is the summary of my CPU/GPU usage (related to OpenGL):

image

After this PR, glTexParameteri is only called about 10,000 times over the same time period, and here is my CPU/GPU usage:

image

As you can see, the difference isn't massive, but it's measurable, and the benefit may be more pronounced on slower devices.

### DIFF
--- a/src/Graphics/OpenGLContext/opengl_BufferManipulationObjectFactory.cpp
+++ b/src/Graphics/OpenGLContext/opengl_BufferManipulationObjectFactory.cpp
@@ -77,10 +77,16 @@ private:
 class AddFramebufferTexture2D : public AddFramebufferRenderTarget
 {
 public:
-	AddFramebufferTexture2D(CachedBindFramebuffer * _bind) : m_bind(_bind) {}
+	AddFramebufferTexture2D(CachedBindFramebuffer * _bind, FramebufferAttachments * _fbattachments) : m_bind(_bind), m_fbattachments(_fbattachments) {}
 
 	void addFrameBufferRenderTarget(const graphics::Context::FrameBufferRenderTarget & _params) override
 	{
+		FramebufferAttachments::const_iterator iter = m_fbattachments->find(u32(_params.bufferHandle));
+		if (iter != m_fbattachments->end() && iter->second == u32(_params.textureHandle))
+			return;
+
+		(*m_fbattachments)[u32(_params.bufferHandle)] = u32(_params.textureHandle);
+
 		m_bind->bind(_params.bufferTarget, _params.bufferHandle);
 		if (_params.textureTarget == graphics::textureTarget::RENDERBUFFER) {
 			glFramebufferRenderbuffer(GLenum(_params.bufferTarget),
@@ -98,6 +104,7 @@ public:
 
 private:
 	CachedBindFramebuffer * m_bind;
+	FramebufferAttachments * m_fbattachments;
 };
 
 class AddNamedFramebufferTexture : public AddFramebufferRenderTarget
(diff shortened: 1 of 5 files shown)


loganmc10:
By the way, this is even more pronounced in 2D games, for instance here is Mischief Makers, current master:

image

And with this change:

image

In the current master, glTexParameteri is called about 2 millions times in 30 seconds. After this change it's  about 345,000

gonetz:
I tried to make similar cache for texture parameters applied to particular texture. It worked bad. It may work for OpenGL 4.5, but not for older versions of OpenGL. OpenGL prior 4.5 set texture parameters per TMU, not per texture. Thus, caching texture parameters per texture will not work correct.

loganmc10:
I'm a little confused by that, what do you mean per TMU? I'm pretty sure glTexParameter applies to the bound texture, and I delete the cache for a texture if that texture is deleted

gonetz:
TMU - texture map unit, what you set with glActiveTexture
Suppose, we have Texture0 and Texture1, which will be loaded to the same TMU.
Texture0 uses WRAP, Texture1 uses CLAMP.
Scenario:
glActiveTexture(GL_TEXTURE0 + TMU0)
glBindTexture(TEXTURE_2D, Texture0)
SetWrap()
glBindTexture(TEXTURE_2D, Texture1)
SetClamp()
glBindTexture(TEXTURE_2D, Texture0)

Now the question: should we set WRAP for Texture0 again?
Your optimization says: no, it is already set to wrap.
Right answer: yes, because TMU0 currently set to CLAMP.

loganmc10:
I don't think you're right about that, from the spec:

> glTexParameter specifies the texture parameters for the active texture unit, specified by calling glActiveTexture.

So when you call glTexParameter it applies to the active texture, but the parameter sticks to the texture object. The only difference in 4.5 with the Texture* functions is that you can specify the texture object without needing to bind it to the active texture first, but the end result is the same.

gonetz:
From my testing it does not stick to the texture object. I had all kind of texture issues when I tried to cache texture parameters. I did not test your implementation, so I can be wrong.

fzurita:
That would suggest a 14-21% improvement in CPU utilization. For CPU limited devices, that would be a huge deal. I'll check how much difference it makes.

fzurita:
I tried a fast device and a slow device with a very weak GPU. I'm seeing zero change in performance. If it really is that much of an improvement, I would expect to see something. Maybe Android device drivers are more optimized than Intel.

loganmc10:
That test was done on an Nvidia GPU. The difference would theoretically be more noticeable on 2D games, but like you say there is a potential some drivers already deal with this. It also could be that the device has a bottleneck somewhere else, and so this change makes no difference in performance since that bottleneck is capping the FPS at a certain level already

fzurita:
I was using Yoshi's island to test.

loganmc10:
I would maybe test Mischief Makers or Worms Armageddon to see if there is a difference

fzurita:
I'm not seeing much of a difference in mischief makers either. Maybe it will make a difference in the Nvidia shield TV since it also has a Nvidia GPU. I'll try that when I get a chance.

fzurita:
I see a small difference with the NVidia shield TV. Mischief makers idles at around 120fps without this optimization, with the optimization it idles at around 135fps.

That's around a 12% improvement. Too bad it doesn't seem to help low end devices.

loganmc10:
Yeah weird that is seems to help Nvidia devices, usually they are the driver superstars. Although maybe it's like I said, maybe those other devices have some other bottleneck so this optimization makes no difference because they are stalling while doing some other task.

gonetz:
@loganmc10: I rebased your commits to current master. I resolved several conflicts, please check that it is done correctly. The commits are in branch loganmc10-opts. They need to be thoroughly tested, since the optimizations are potentially dangerous.

gonetz:
Build to test:
https://drive.google.com/file/d/1U7vQ6QTamQ-oFm-dmEx5syXi9v-p4Pvm/view?usp=sharing

loganmc10:
The commit itself looks good. My testing hasn't revealed any issues, but hopefully some other people can chime in

PR #1732, opened 2018-02-23
iOS Support + Conditional FreeType Compilation

braindx:
With this PR, GLideN64 can compile and run under iOS using both GLES 2 and GLES 3 contexts. We're currently using GLideN64 in [Provenance](https://github.com/jasarien/Provenance) as part of the Mupen core and I'd love to have iOS support included upstream.

I don't plan on linking FreeType into Provenance at this time, so I've also added a way to compile GLideN64 without this dependency via a preprocessor define. I totally understand if you don't want to merge that bit, but I included it on this PR anyway as I think it makes it easier to get GLideN64 up and running on new platforms. The FreeType dependency can be a bit of a headache.

In case you're wondering what the deal is with that framebuffer caching, the issue is that iOS is always rendering to a framebuffer even when rendering to the "screen" because it all eventually gets composited by the OS. To workaround this, the framebuffer iOS uses is cached off the first time GLideN64 attempts to bind the framebuffer to something else and then that iOS-held framebuffer is bound anytime the plugin attempts to bind the framebuffer to 0.


loganmc10:
> In case you're wondering what the deal is with that framebuffer caching, the issue is that iOS is always rendering to a framebuffer even when rendering to the "screen" because it all eventually gets composited by the OS. To workaround this, the framebuffer iOS uses is cached off the first time GLideN64 attempts to bind the framebuffer to something else and then that iOS-held framebuffer is bound anytime the plugin attempts to bind the framebuffer to 0.

I ran into this issue in the past with RetroArch, which has a similar problem. There is a better way to solve this. You can define ```ObjectHandle::null``` to equal something other than 0. I believe @gonetz looked into this a while ago, he would know better how to accomplish this.

By the way, doing ```glGetIntegerv(GL_FRAMEBUFFER_BINDING, &defaultFramebuffer);``` every time glBindFramebuffer is called has to be bad for performance. I'd be surprised if it wasn't, most the glGetInteger calls aren't meant to be called frequency in the middle of rendering. You should find a way to determine that at the beginning, or only update it when needed, I'm not sure when iOS would change the default framebuffer but I doubt it would happen in the middle of a game.

Besides that, what is with changed all the ""'s into <>'s? That isn't correct

braindx:
Happy to rework it by redefining `ObjectHandle::null` if that makes more sense, but since it's used for more objects than just framebuffers it seems problematic. I'll let @gonetz weigh in on that.

It's definitely not doing `glGetIntegerv(GL_FRAMEBUFFER_BINDING, &defaultFramebuffer);` each frame, only the first time to cache off that value. There's no performance issue there.

The C standard is a bit unhelpful when defining how paths are searched when using angle brackets vs quoted strings. All compilers that I've used seem to handle quoted string includes by first searching for the file as a relative path to the file including it. When that fails, compilers' implementations seem to differ. They are permitted to search in any number of other implementation-defined locations and, only when that search fails, is it required by the standard to fall back to the locations it searches when using angle brackets. In this case, some of the files referenced using quoted strings are not relative pathnames but instead are expected to be found using search paths that include the src or src/include directories. In Xcode, for example, those quoted-string includes ended up including files with the same name from elsewhere in the project rather than from the intended location. My hope was that the angle bracket includes would continue to search these locations for all existing platforms and that this would resolve the issue when compiling via Xcode. If that's not the case, we can solve it other ways such as specifying the file as a relative path.

loganmc10:
> It's definitely not doing glGetIntegerv(GL_FRAMEBUFFER_BINDING, &defaultFramebuffer); each frame, only the first time to cache off that value. There's no performance issue there.

Oh right I misread the if statement, it only happens the first time. Yeah I would work with gonetz to instead redefine ObjectHandle::null for framebuffers, or perhaps create some kind of graphics::defaultFramebuffer variable that defaults to 0 and can be redefined.

The quoted include should search the relative path, if there is something wrong then we should fix the relative path instead of using <>, when you use <> you risk pulling in something from a system path instead of your local include (for example if the user had a /usr/include/N64.h file for some reason, it would be included instead of our N64.h file).

braindx:
I considered the ObjectHandle approach and defaultFramebuffer approach, but ultimately decided against them because:
1. They were more invasive to existing code. Would require fixing up all usage instances.
2. It wouldn't stop people from writing bind(0) even though there's a defined constant they should be using instead, having it work on the device they tested, and then having that break iOS.
Of course, you guys are the ones actively maintaining and working on this repo, so I'll defer to your judgement and do whatever you prefer.

As far as the includes, I would have thought that it would search the project-defined header directories before searching system directories. Surprised to hear that isn't the case. Similarly, if you would prefer I change these all to be relative paths instead and use quotes, I'm happy to do that.

gonetz:
* Conflicts when rebased to current master.

* Whitespace errors in opengl_CachedFunctions.h, Log.cpp, osal_files_ios.mm, 

* #include "N64.h" => #include <N64.h>
Why? The project compiled with gcc and clang, on Mac too.

* Texdrawer.cpp I don't like when source code perforated by OS-specific #ifndef's
Please write TexdrawerStub.cpp for IOS.

JoeMatt:
Hi, I'm Joe co-dev on Provenance where we're pulling in from Gliden. We hope to switch to a submodule system like OpenEMU, which we're loosely based on. I'd like to see the iOS changes merged into mainline so we don't have to maintain a separate fork, especially consider the velocity of updates to Gliden. We're currently the fastest non-jailbreak based N64 emulator on iOS since we don't rely on dynamic recompilation, despite being a FOSS app, we wish to maintain non-jailbreak compatibility. @braindx has lead the way on getting N64 working beyond expectations. 

The issues with header changes is related to some bugs in our Xcode project where it was mis-matching headers from other sub-projects in our repo. I'm going to fix those issues soon and then we can hopefully modify this PR to a state where it can be officially included. We'd love to be able to easily maintain release status with GlideN64 since N64 is one of the most requested systems we've had for some time and this project is a key to performance for us.

I understand most mobile focus has been on Android up until now since there hasn't been a vibrant community around iOS open source N64 emulators, we hope to change that with Provenance. I look forward to working together. I'm committed to resolving any semantic issues while @braindx is the mastermind behind the technical components. Please be patient on this PR but I hope to be able to fix the concerns you have soon.

gonetz:
Hi Joe!

I'm looking forward for fixed PR. Expansion to iOS will increase GLideN64 world domination :)

JoeMatt:
I've updated our fork to the latest Gliden, including the new ARM intrinsics (awesome to have), and made the  Texdrawer.cpp stub class to remove the Macro for free type.  White space errors also fixed.

The header thing is weird, I'll have to look at how your Xcode project handles paths. Instead of using < > brackets, I need to use '../' or Xcode wouldn't find the headers, no matter how I setup the search paths. This is is also a problem with mupen core as they use double quote headers for files in sub-dirs. 

Not sure what the conclusion was on the framebuffer caching calls, I'm not a great C++ or GL dev so I'll leave it up to someone else to figure out what to do there.

I'll make a new PR within the week or so. Just wanted to update that this is being worked on still.

gonetz:
I guess this PR can be closed because of #1776, can not it?

Issue #1561, opened 2017-09-01
Super lag in HLE mode, at least with r600/radeonsi Mesa drivers


Jj0YzL5nvJ:
From the moment of implementation a625225323c902b614ed9601143df3bc51550fc4, this generates some kind of delay in radeon Mesa driver (r600g).

Xubuntu 16.04.3 LTS
glxinfo | grep OpenGL
```
OpenGL vendor string: X.Org
OpenGL renderer string: AMD JUNIPER (DRM 2.43.0 / 4.4.0-93-generic, LLVM 6.0.0)
OpenGL core profile version string: 3.3 (Core Profile) Mesa 17.3.0-devel - padoka PPA
OpenGL core profile shading language version string: 3.30
OpenGL core profile context flags: (none)
OpenGL core profile profile mask: core profile
OpenGL core profile extensions:
OpenGL version string: 3.0 Mesa 17.3.0-devel - padoka PPA
OpenGL shading language version string: 1.30
OpenGL context flags: (none)
OpenGL extensions:
OpenGL ES profile version string: OpenGL ES 3.0 Mesa 17.3.0-devel - padoka PPA
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 3.00
OpenGL ES profile extensions:
```
The CPU consumption is not higher when compared to previous versions. And compiling with `-DCRC_OPT=On` changes almost nothing when it comes to FPS. Examples:

Running SM64 until the moment the white star appears:

dddb3ae1f71afa85ef782d5f5edb9661a1b1b5bd
cmake -DMUPENPLUSAPI=On ../../src/
LIBGL_SHOW_FPS=1 MESA_GL_VERSION_OVERRIDE=3.3COMPAT MESA_GLSL_VERSION_OVERRIDE=420
```
libGL: FPS = 0.5
libGL: FPS = 40.0
libGL: FPS = 40.0
libGL: FPS = 51.0
libGL: FPS = 60.0
libGL: FPS = 4.6
libGL: FPS = 54.0
libGL: FPS = 59.0
libGL: FPS = 59.0
libGL: FPS = 60.0
libGL: FPS = 60.0
libGL: FPS = 60.0
libGL: FPS = 60.0
```

a625225323c902b614ed9601143df3bc51550fc4
cmake -DMUPENPLUSAPI=On ../../src/
[...]

PR #1735, opened 2018-02-26
Fix performance issues on AMD+Mesa GPUs

loganmc10:
@Jj0YzL5nvJ can you please test this and let me know if it resolves all your issues? I'm hoping this also resolves ```EnableCopyColorToRDRAM=2``` for you, so please try that out as well.

Should fix https://github.com/gonetz/GLideN64/issues/1561

### DIFF
--- a/src/Graphics/OpenGLContext/opengl_BufferedDrawer.cpp
+++ b/src/Graphics/OpenGLContext/opengl_BufferedDrawer.cpp
@@ -186,7 +186,7 @@ void BufferedDrawer::_updateTrianglesBuffers(const graphics::Context::DrawTriang
 	if (_params.elements == nullptr)
 		return;
 
-	const GLsizeiptr eboDataSize = sizeof(GLubyte) * _params.elementsCount;
+	const GLsizeiptr eboDataSize = sizeof(GLushort) * _params.elementsCount;
 	Buffer & eboBuffer = m_trisBuffers.ebo;
 	_updateBuffer(eboBuffer, _params.elementsCount, eboDataSize, _params.elements);
 }
@@ -204,8 +204,8 @@ void BufferedDrawer::drawTriangles(const graphics::Context::DrawTriangleParamete
 	}
 
 	if (config.frameBufferEmulation.N64DepthCompare == 0) {
-		glDrawElementsBaseVertex(GLenum(_params.mode), _params.elementsCount, GL_UNSIGNED_BYTE,
-			(char*)nullptr + m_trisBuffers.ebo.pos - _params.elementsCount, m_trisBuffers.vbo.pos - _params.verticesCount);
+		glDrawElementsBaseVertex(GLenum(_params.mode), _params.elementsCount, GL_UNSIGNED_SHORT,
+			(u16*)nullptr + m_trisBuffers.ebo.pos - _params.elementsCount, m_trisBuffers.vbo.pos - _params.verticesCount);
 		return;
 	}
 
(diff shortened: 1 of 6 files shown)


loganmc10:
Here is the comment on the Mesa bug tracker that helped me resolve this issue: https://bugs.freedesktop.org/show_bug.cgi?id=105256#c1

loganmc10:
By the way, I didn't make the GL_UNSIGNED_SHORT thing GPU specific because the person on the bug tracker indicated (https://bugs.freedesktop.org/show_bug.cgi?id=105256#c4) that this doesn't just apply to AMD GPU's but there are a few older Nvidia GPU's as well.

Without a perfect list of all the GPU's that this might affect, I think the larger element size is an OK price to pay to fix compatibility on those cards.

fzurita:
I tested this on a low end Adreno device and saw no difference in performance.

Jj0YzL5nvJ:
In my case this change is a partial fix https://github.com/gonetz/GLideN64/issues/1561#issuecomment-368838242

loganmc10:
@Jj0YzL5nvJ Is it full speed when you have ```EnableCopyColorToRDRAM=0```?

loganmc10:
I added a commit to using client side data for the elements. This is how Dolphin does it and I assume they've done more testing than us.

@Jj0YzL5nvJ I just need to know if change fixes the speed with ```EnableCopyColorToRDRAM=0```, then we know that you just have an issue left with EnableCopyColorToRDRAM or not

gonetz:
The first commit is ok, if it really fixes performance issues on AMD+Mesa GPUs, does it?

But the second one is questionable. It is not the reason to switch to client side data for the elements just because Dolphin does that. It can be caused by specific of Dolphin data flow, and thus has no benefits for us. My point: lets not touch it if it does not bring any performance improvements.

loganmc10:
I'm just waiting to hear back from @Jj0YzL5nvJ , another ATI user reported that it fixed the issue, but @Jj0YzL5nvJ also has issues with EnableCopyColorToRDRAM, so I am just trying to seperate the issues.

As far as the element buffer, it was actually changed during the graphics refactoring, before that it was client side data. @fzurita reported that the switch to VBO/EBO was actually a slight performance hit on Android (https://github.com/gonetz/GLideN64/issues/1561#issuecomment-368550340). I believe buffer objects are only really a benefit if you can reuse the data, since you can just reference the data in the buffer without uploading. But when the data changes every time, I don't think there is any benefit to buffer objects.

The unbuffered drawer is probably the same or faster than buffered drawer, but we have to use VBOs because it's mandated by OpenGL core profile. It isn't mandated for element arrays

BPaden:
@gonetz 

The effects of the ushort patch on early GCN cards and below are pretty dramatic. From ~10-15 FPS reported by Mesa on my machine to ~60.

> But the second one is questionable. It is not the reason to switch to client side data for the elements just because Dolphin does that. It can be caused by specific of Dolphin data flow, and thus has no benefits for us. My point: lets not touch it if it does not bring any performance improvements.

The buffer patch alone does improve the FPS reported by Mesa by about ~5-10.

Jj0YzL5nvJ:
> @Jj0YzL5nvJ Is it full speed when you have EnableCopyColorToRDRAM=0?

loganmc10/GLideN64@9bcfa67
loganmc10/GLideN64@6527537
loganmc10/GLideN64@fa0ab9d

All three give the same results with `EnableCopyColorToRDRAM=0`, because of that I omitted their results.
The best results be achieved with loganmc10/GLideN64@29dc6a1

Even adapt the code on loganmc10/GLideN64@9bcfa67 and loganmc10/GLideN64@6527537 in two new variants of loganmc10/GLideN64@29dc6a1. But both were inferior to loganmc10/GLideN64@29dc6a1 alone.

loganmc10:
I guess I'm getting a little confused, if you set ```EnableCopyColorToRDRAM=0``` and use the commits in this PR, do you get food performance? Is it better than the current master? We can sort about EnableCopyColorToRDRAM later

Jj0YzL5nvJ:
I don't want to offend, but any of your hotfixes is better that 'master'...
I leave you my last benchmark here https://github.com/gonetz/GLideN64/issues/1561#issuecomment-369155714

gonetz:
> The unbuffered drawer is probably the same or faster than buffered drawer, but we have to use VBOs because it's mandated by OpenGL core profile. It isn't mandated for element arrays

Ok, you convinced me.

> The buffer patch alone does improve the FPS reported by Mesa by about ~5-10.

Hard to believe, but I have no Mesa powered devices to test.

loganmc10:
Just the buffer thing caused the issue right? Changing the elements to a short is ok? That was what actually fixed the performance so that's ok

gonetz:
Yes, changing the elements to a short is ok

gonetz:
I put ebo commit to remove_ebo branch, in case somebody want to sort out, what is going on there.

PR #1737, opened 2018-02-27
Reduce some debug GL calls in production

loganmc10:
We are currently calling glGetError() (and glCheckFramebufferStatus()) in the production builds. This could have a negative impact on performance, and we already have a GL_DEBUG definition in other areas of the code.

I also moved the check for depth texture support into GLInfo::init(). Checking for extension support only needs to be done once, and could also have a performance impact.

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_Utils.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_Utils.cpp
@@ -29,6 +29,7 @@ static const GLsizei nShaderLogSize = 1024;
 
 bool Utils::checkShaderCompileStatus(GLuint obj)
 {
+#ifdef GL_DEBUG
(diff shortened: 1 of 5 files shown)


loganmc10:
The reason this is so bad for performance:

If you call glCompileShader, the GPU driver puts that command in its pipeline and executes it at the most optimal time, it probably won't happen right away.

But if you call ```glGetShaderiv(obj, GL_COMPILE_STATUS, &status);``` right after glCompileShader, the only way it can return that call is to compile the shader right then. So you've basically called a "glFinish" and force the driver to synchronise. Some of the glGetError calls won't be as bad, since it can probably quickly verify your inputs, but in general the glGet* calls can be very dangerous for performance.

gonetz:
Thanks!

PR #1738, opened 2018-02-28
Only use ColorBufferReaderWithBufferStorage for GLES

loganmc10:
I have tested sync and async performance on Mesa and Windows+NVidia and saw no performance difference using ColorBufferReaderWithBufferStorage.

Given that we've had performance issues with Intel (https://github.com/gonetz/GLideN64/issues/1312), and now AMD+Mesa (https://github.com/gonetz/GLideN64/issues/1561), I think it would be best to just enable it for GLES

@Jj0YzL5nvJ can you please test this PR with the latest master + ```EnableCopyColorToRDRAM = 2``` and let me know if it resolves your issue and runs at full speed?

### DIFF
--- a/src/Graphics/OpenGLContext/opengl_BufferedDrawer.cpp
+++ b/src/Graphics/OpenGLContext/opengl_BufferedDrawer.cpp
@@ -37,7 +37,6 @@ BufferedDrawer::BufferedDrawer(const GLInfo & _glinfo, CachedVertexAttribArray *
 	glGenVertexArrays(1, &m_trisBuffers.vao);
 	glBindVertexArray(m_trisBuffers.vao);
 	_initBuffer(m_trisBuffers.vbo, m_bufMaxSize);
-	_initBuffer(m_trisBuffers.ebo, m_bufMaxSize);
 	m_cachedAttribArray->enableVertexAttribArray(triangleAttrib::position, true);
 	m_cachedAttribArray->enableVertexAttribArray(triangleAttrib::color, true);
 	m_cachedAttribArray->enableVertexAttribArray(triangleAttrib::texcoord, true);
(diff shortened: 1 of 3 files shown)


Jj0YzL5nvJ:
Sorry for the delay, but I had a "technical mishap" ...my video card is DEAD, stone-dead.
I am responsible for my unwise BIOS configuration (`GART error reporting = Enabled` + `PCIE Spread Spectrum = Auto`) and "pushing stress" on GPU to render at non-standard resolutions (monitor 1080p to 120Hz (max 144Hz), mupen64plus windowed to 1400x1050) using a buggy extension like some kind of "steeplechase generator" for benchmarking. Well, I reaped what I sowed.

Anyway, I had test this pull in `Radeon HD 6350` (in reality is just a recycled `HD 5450` by marketing) and I saw a little improvement from 17 FPS to 20 FPS in [this trial area](https://i.imgur.com/hCEKtz2.png), like the screenshot shows, an average of 26 FPS in such area on the old card. Doing a few maths gives us a hypothetical improvement of 30 FPS on the deceased card...

For me, most wise solution for r600 and probably for radeonsi too, is disable `GL_ARB_buffer_storage` completely (by themselves), the bug there is nasty and dangerous.

loganmc10:
This PR disables the usage of GL_ARB_buffer_storage by the plugin, so if this doesn't help then I'm not sure what would

Lithium64:
@loganmc10 GL_EXT_buffer_storage does not have the performance issue on AMD/Mesa, Is it possible to use this extension instead of the GL_ARB_buffer_storage?

When I disable this extension with MESA_EXTENSION_OVERRIDE=-GL_ARB_buffer_storage, the performance issue is fixed.

loganmc10:
GL_EXT_buffer_storage is a GLES (mobile) extension. GL_ARB_buffer_storage is the desktop variant. Typically GL_EXT_buffer_storage wouldn't be available unless you are running in GLES mode, which nobody really does on a PC, although I think Mesa allows it

Jj0YzL5nvJ:
@Lithium64, how did you tested GL_EXT_buffer_storage?

 I tried GLES in the past with glide64mk2 and rice, in 64-bit they are broken. But on 32-bit, both works even better compared to GLX, even on [crappy hardware](https://github.com/mupen64plus/mupen64plus-core/issues/349).
 I think it would be nice to have the option to use GLES on desktop PC.

PR #1744, opened 2018-03-08
Use glDrawRangeElementsBaseVertex

loganmc10:
The Adreno OpenGL ES developer guide says:

> use indexed ranged draw calls: glDrawRangeElements() tells ES what the min/max values for the index data set of the draw call are. This lets OpenGL ES implementation precache the vertex data for the specified index range, which can improve the rendering performance.

I don't know how much of a difference this will make, but it's easy to implement. I use the range as ```0 -> _params.verticesCount - 1``` because I see here:

https://github.com/gonetz/GLideN64/blob/master/src/GraphicsDrawer.cpp#L739

That triParams.verticesCount is set to triangles.maxElement + 1. @gonetz hopefully you can verify this for me, that the values of the data inside triangles.elements.data() should only range from 0 -> triangles.maxElement.

I didn't implement this for the UnbufferedDrawer because GLES2 doesn't support glDrawRangeElements


tony971:
If you're touching triangles, this might be useful to you. Warning: There is a high probability that I'm talking out of my ass and this is unrelated.

https://forums.cen64.com/viewtopic.php?f=14&p=2735&sid=e21c1d4b5297512c3443b007513e7e34#p2734

loganmc10:
Definitely completely unrelated :) but I'm sure it will be interesting to someone. I don't really know anything about N64 GFX, I'm just trying to optimize the OpenGL backend

Issue #1211, opened 2016-11-05
Excessive clipping in LoZ:OOT in GLES


fzurita:
There is currently excessive clipping when weapons are being held in first person view work with GLES modes work LoZ: OOT. Here is an example:

![the_legend_of_zelda-000](https://cloud.githubusercontent.com/assets/11581687/20030052/4560e358-a333-11e6-8c0b-eb4db64f47ef.png)

Same with Glide64:
![the_legend_of_zelda-001](https://cloud.githubusercontent.com/assets/11581687/20030135/58f773bc-a335-11e6-8f89-576aedd07706.png)

fzurita:
Seems related to this:

https://github.com/gonetz/GLideN64/issues/588

gonetz:
Most likely it is. Desktop build works correct there. Glide64 uses software clipping, so it works correct on mobiles as well.

fzurita:
There is some discussion about how to simulate this fictionally using the vertex shader here:

http://stackoverflow.com/questions/7408855/clipping-planes-in-opengl-es-2-0

It seems to me that using the vertex shader to adjust everything to be within the clipping planes would work.

Is the minimum vertex z position -32768 and max 32767 from the N64?

gonetz:
No. Screen Z is in range [0, 32767]. When it is out of that range {when z/w is out of [-1, 1]), polygon is clipped.

fzurita:
So, this could be easily fixed if these two extensions are supported:
```EXT_clip_cull_distance```
```GL_ARB_depth_clamp```

Unfortunately, most devices don't seem to support GL_ARB_depth_clamp is only supported by full OpenGL and not GLES. Although, most OpenGL ES 3.0+ devices have support for ```EXT_clip_cull_distance``` which gives access to ```GL_CLIP_DISTANCEX``` and ```gl_ClipDistance``` and ```gl_CullDistance``` in shader code. I don't think EXT_clip_cull_distance is enough to fix the problem.

fzurita:
Hmmm.... Changing  ```glDepthRangef``` may be equivalent to enabling/disabling DEPTH_CLAMP in OpenGL ES. I'm going to investigate this.

rcaridade145:
https://github.com/dolphin-emu/dolphin/search?utf8=%E2%9C%93&q=bSupportsDepthClamp Don't know if this can be of any help

fzurita:
[...]

Issue #588, opened 2015-06-22
Missing heart in zelda (GLES)


Gillou68310:
Here's a hack to fix the missing heart in both zelda games:
https://github.com/Gillou68310/GLideN64/commit/1ddecf779935c8d7d9a3200074b283feccdf409f

Weird thing is this hack is only necessary for GLES versions (GLES2 and GLES3, I couldn't check GLES3.1) the desktop version doesn't seems to need it. It might be interesting to understand why this is working fine on desktop.

purplemarshmallow:
the problem is that clipping does not work properly for GLES, the heart and the map arrows are clipped away. Adding a hack is a bad idea, it needs to be fixed properly because other games are affected as well. The characters and vehicles in Star Wars Ep1 racer are cut off. The desktop version does not have these problems.

Gillou68310:
>  Adding a hack is a bad idea, it needs to be fixed properly because other games are affected as well.

I totally agree, I just provided the hack as a possible temporary solution ;-)

gonetz:
@purplemarshmallow - could you explain why do you think that it is clipping and why these particular polygons are clipped?

purplemarshmallow:
I think it's the same problem as in rice video
Rice Video has this comment:

// Hack needed to show flashing last heart and map arrows in Zelda OoT & MM  
// It renders at Z cordinate = 0.0f that gets clipped away  
// So we translate them a bit along Z to make them stick

Gillou68310:
Hehe I made this comment ;-)

gonetz:
I remember similar problem on desktop too. Polygon may not use depth compare at all, as this heart rectangle, but if it has negative z it will be clipped. I tried to force z to zero, but it causes problems with fog. So, I just disabled z-clipping for ucodes, which do not use near-far clipping. Desktop GL allows to do that. GL ES does not. If it clip polygons with Z==0 it's bad. Glide64 uses software clipping and culling, so this problem resolved on plugin's side. It was done because Glide3x API does not provide polygons clipping. I don't want to make custom clipping when I have built in GL one. From the other side, the only alternative for GL ES is a hack. @Gillou68310 - as I understand, your patch does the same trick - translates vertices a bit along Z, right?

Gillou68310:
Exact.

Gillou68310:
Here's the value to fix duke nukem zero hour:
````gSP.matrix.projection[3][2] += 1.f;```

[...]

PR #1747, opened 2018-03-11
Disable clipping using Vertex shader

loganmc10:
This is an attempt to fix https://github.com/gonetz/GLideN64/issues/1211 and https://github.com/gonetz/GLideN64/issues/588

Before:

the_legend_of_zelda-003

After:

the_legend_of_zelda-001

Before:

star_wars_ep1_racer-003

After:

star_wars_ep1_racer-001

As you can see from the Ep I: Racer screenshot, it's not perfect yet. The pilot is still partially missing. I'm kind of stumped on this one, are there other Vertex shaders that need this applied? I see that there are a few "special shaders" do they apply here?

How this works (from the description of glDepthRangef):

> After clipping and division by w, depth coordinates range from -1 to 1, corresponding to the near and far clipping planes

So after clipping, z/w will be between -1 and 1. In the Vertex shader, I check if z/w would be less than -1 (for near plane) or over 1 (for far plane), if it is, I modify z so that z/w will equal -1 or 1, basically I am doing "manual clamping".

We could still use GL_DEPTH_CLAMP for desktops, but I figured the code would be simpler if we just got this working in the vertex shader, then it would work from GLES2 to desktop.

I'm not sure what the issue is with Ep I Racer, I'm hoping somebody will be able to spot the error, I'm also not sure what other games would be a good test for this, so suggestions are welcome. Don't expect this to fix any current clipping issues, this is just to bring the GLES version up to the same level of support as the desktop version.

### DIFF
--- a/src/Graphics/Context.h
+++ b/src/Graphics/Context.h
@@ -17,13 +17,18 @@ namespace graphics {
 
 	enum class SpecialFeatures {
 		Multisampling,
-		NearPlaneClipping,
 		FragmentDepthWrite,
 		BlitFramebuffer,
 		WeakBlitFramebuffer,
 		DepthFramebufferTextures,
 		ShaderProgramBinary,
 		ImageTextures
+	};
+
+	enum class ClampMode {
+		ClippingEnabled,
+		NoNearPlaneClipping,
+		NoClipping
 	};
 
 	class ContextImpl;
@@ -39,7 +44,13 @@ namespace graphics {
 
 		void destroy();
 
+		void setClampMode(ClampMode _mode);
+
+		ClampMode getClampMode();
+
 		void enable(EnableParam _parameter, bool _enable);
+
+		u32 isEnabled(EnableParam _parameter);
 
 		void cullFace(CullModeParam _mode);
 
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
@@ -229,6 +229,7 @@ public:
 			std::stringstream ss;
 			ss << "#version " << Utils::to_string(_glinfo.majorVersion) << Utils::to_string(_glinfo.minorVersion) << "0 es " << std::endl;
 			ss << "# define IN in" << std::endl << "# define OUT out" << std::endl;
+			ss << "OUT highp float vZCoord;" << std::endl << "uniform lowp int uClampMode;" << std::endl;
 			m_part = ss.str();
 		}
 		else {
@@ -316,14 +317,6 @@ public:
 			"      vShadeColor.rgb = vec3(fp);								\n"
 			"  }															\n"
 			;
-		if (!_glinfo.isGLESX) {
-			m_part +=
-				"  gl_ClipDistance[0] = gl_Position.w - gl_Position.z;			\n"
-				;
-		}
-		m_part +=
-			"} \n"
-		;
 	}
 };
 
@@ -374,14 +367,6 @@ public:
 			"      vShadeColor.rgb = vec3(fp);								\n"
 			"  }															\n"
 			;
-		if (!_glinfo.isGLESX) {
-			m_part +=
-				"  gl_ClipDistance[0] = gl_Position.w - gl_Position.z;			\n"
-				;
-		}
-		m_part +=
-			"} \n"
-			;
 	}
 };
 
@@ -405,14 +390,6 @@ public:
 			"  vShadeColor = uRectColor;						\n"
 			"  vTexCoord0 = aTexCoord0;							\n"
 			"  vTexCoord1 = aTexCoord1;							\n"
-		;
-		if (!_glinfo.isGLESX) {
-			m_part +=
-				"  gl_ClipDistance[0] = gl_Position.w - gl_Position.z;			\n"
-				;
-		}
-		m_part +=
-			"} \n"
 			;
 	}
 };
@@ -432,10 +409,24 @@ public:
 			"  gl_Position = aRectPosition;						\n"
 			"  vShadeColor = uRectColor;						\n"
 			;
+	}
+};
+
+class VertexShaderEnd : public ShaderPart
+{
+public:
+	VertexShaderEnd(const opengl::GLInfo & _glinfo)
+	{
 		if (!_glinfo.isGLESX) {
-			m_part +=
-				"  gl_ClipDistance[0] = gl_Position.w - gl_Position.z;			\n"
+			m_part =
+				"  gl_ClipDistance[0] = gl_Position.w - gl_Position.z;	\n"
 				;
+		} else if (config.generalEmulation.enableFragmentDepthWrite != 0 && config.frameBufferEmulation.N64DepthCompare == 0) {
+				m_part =
+					"  vZCoord = gl_Position.z;	\n"
+					"  if (uClampMode > 0)	\n"
+					"    gl_Position.z = 0.0;	\n"
+					;
 		}
 		m_part +=
 			"} \n"
@@ -872,6 +863,14 @@ public:
 			m_part =
 				"void writeDepth();\n";
 			;
+			if (_glinfo.isGLESX) {
+				m_part =
+					"IN highp float vZCoord;	\n"
+					"uniform lowp float uPolygonOffset;	\n"
+					"uniform lowp int uClampMode;	\n"
+					+ m_part
+				;
+			}
 		}
 	}
 };
@@ -1447,7 +1446,19 @@ public:
 						"layout(binding = 0, r32ui) highp uniform readonly uimage2D uZlutImage;\n"
 						"void writeDepth()						        													\n"
 						"{																									\n"
-						"  gl_FragDepth = clamp((gl_FragCoord.z * 2.0 - 1.0) * uDepthScale.s + uDepthScale.t, 0.0, 1.0);	\n"
+						;
+					if (_glinfo.isGLESX && config.frameBufferEmulation.N64DepthCompare == 0) {
+						m_part +=
+							"  highp float z_value = vZCoord * gl_FragCoord.w;	\n"
+							"  if (uClampMode == 1 && (z_value > 1.0)) discard;	\n"
+							"  gl_FragDepth = clamp((z_value - uPolygonOffset) * uDepthScale.s + uDepthScale.t, 0.0, 1.0);	\n"
+							;
+					} else {
+						m_part +=
+							"  gl_FragDepth = clamp((gl_FragCoord.z * 2.0 - 1.0) * uDepthScale.s + uDepthScale.t, 0.0, 1.0);	\n"
+						;
+					}
+					m_part +=
 						"  highp int iZ = gl_FragDepth > 0.999 ? 262143 : int(floor(gl_FragDepth * 262143.0));				\n"
 						"  mediump int y0 = clamp(iZ/512, 0, 511);															\n"
 						"  mediump int x0 = iZ - 512*y0;																	\n"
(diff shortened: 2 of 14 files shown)


gonetz:
I suppose that this code is generally wrong:
```
 if (uClampMode > 0) {
   if (gl_Position.z / gl_Position.w < -1.0) gl_Position.z = -gl_Position.w;
   if (uClampMode == 2) {
     if (gl_Position.z / gl_Position.w > 1.0) gl_Position.z = gl_Position.w;
   }
 }
```

You breaks geometry by changing 3D coordinate of vertex.
Suppose, that our hardware can't clip Y coordinate.
Imagine consequences of this code:
```
   if (gl_Position.y / gl_Position.w < -1.0) gl_Position.y = -gl_Position.w;
   if (gl_Position.y / gl_Position.w > 1.0) gl_Position.y = gl_Position.w;

```

I think that the only valid solution of the problem is software clipping. Glide64 does software clipping and thus has no clipping issues on GL ES devices. I don't want to mess with software clipping in GLideN64, but if it will be hidden in backend, it can be added.

loganmc10:
I'm not sure what you mean, this is exactly what GL_DEPTH_CLAMP does as far as I understand:, from ARB_depth_clamp

> The one major issue is that fragments of a  primitive may extend
    beyond the conventional window space depth range for depth values
    (typically the range [0,1]).  Rather than discarding fragments that
    defy the window space depth range (effectively what near and far
    plane clipping accomplish), the depth values can be clamped to the
    current depth range.

gonetz:
I need a drawing to show what I mean. I'm not good with drawings. I made some ugly sketch. It shows the situation if look on viewing frustum from above, that is  XZ plane.
Green triangle is the original one. One vertex is behind near clipping plane, that is Z/W < -1.
Yellow area must be clipped. Red triangle is what your fix does, as I understand it. X and Y coordinates of vertex remains the same, but Z coordinate changed and this changes the geometry of the triangle.
drawing

loganmc10:
Yes I think you are totally right, but I believe this is what GL_DEPTH_CLAMP does as far as I understand from reading ARB_depth_clamp. I am just trying to mimick the behaviour of GL_DEPTH_CLAMP.

The issue on EP I Racer exists with this method because it uses fragment depth writing. Depth clamping happens right before the depth test, and when you write to gl_fragDepth, the depth test is delayed until after the fragment shader is processed, normally it happens before. So with GL_DEPTH_CLAMP, it clamps the value after the fragment shader runs, but with this method I wrote, it clamps the value before the fragment shader, and thus gl_fragCoord.z has the wrong value when it goes it write gl_fragDepth

gonetz:
I doubt that GL_DEPTH_CLAMP works as you described. "Rather than discarding **fragments** that
defy the window space depth range (effectively what near and far plane clipping accomplish), the depth **values** can be clamped to the current depth range." From my understanding, it clamps depth values of fragments, not vertex. That is, depth value for fragments in yellow area can be clamped instead of whole area clipping. Technique, which alters polygon's geometry sounds weird. I can be wrong though.

loganmc10:
Ah I understand, thanks for clearing that up. I think it would still be possible via shaders to do it the way you describe though:

In the vertex shader, you "copy" the z coordinate, like:

```
OUT float z;

//convert gl_Position.z to a window coordinate between 0 and 1
z = gl_Position.z / gl_Position.w;
z = (z + 1) * 0.5;

gl_Position.z = max(gl_Position.z, -gl_Position.w); //clamp to near plane to disable near plane clipping
```

Then in the fragment shader (writeDepth()), instead of using gl_FragCoord.z, you use "IN float z", like:

```
IN float z
highp float depth = uDepthSource == 0 ? (z * 2.0 - 1.0) : uPrimDepth;
```

This way you've dodged the clipping stage. This seems to work for disabling near plane clipping (you can see the pod and the driver), but something seems broken still about far plane clipping:

star_wars_ep1_racer-001

As you can probably tell, I'm in a little bit over my head here, but it seems like it should be possible to do with shaders.

loganmc10:
I figured out the proper way to do it in the fragment shader, so the geometry isn't messed up (I think).

Thanks for the help @gonetz , I'll update this PR in a bit, and I'll make the solution GLES only just to make sure there are no regressions on the desktop side

gonetz:
@loganmc10 please explain, how you fixed the problem. My point is that any change of gl_Position.z in vertex shader alters polygon geometry. If it can't be avoided, it must be GLES only solution indeed.
May be the problem with depth value can be resolved in geometry shader, which runs before Primitive Clipping in rendering pipeline:
https://www.khronos.org/opengl/wiki/Rendering_Pipeline_Overview

loganmc10:
Sure I'll try to explain my solution, I'll upload the code tomorrow to this PR:

To disable near plane clipping:

In the vertex shader, you check if it will be clipped (gl_Position.z / gl_Position.w < -1.0). If it is going to be clipped, you change gl_Position.z to -gl_Position.w, so the value it right at the near plane. This is the only way to avoid clipping. You also store the unclipped value in a varying float (you store the original value of gl_Position.z / gl_Position.w, converted to window coordinates, it will be less than 0).

In the fragment shader, you check if gl_FragCoord.z == 0 (the fragments at the near plane). If it is, you don't use gl_FragCoord.z when you write to gl_FragDepth, you use the varying value you passed from the vertex shader. This "restores" the un-clipped depth to the fragment. I believe this gives each fragment the correct depth value. Star Wars Ep I Racer renders properly and looks the same to me as the desktop version (but I don't have a great eye for detail).

loganmc10:
I realize now that I read that back that it probably still modifies the geometry of the triangles that should be clipped, however there may not be a better way (for GLES), it's probably not going to be noticeable to most eyes, and it'll be better than the current master (can't see anything)

gonetz:
Ok, lets it be GLES only fix.

"In the fragment shader, you check if gl_FragCoord.z == 0 (the fragments at the near plane). If it is, you don't use gl_FragCoord.z when you write to gl_FragDepth, you use the varying value you passed from the vertex shader. This "restores" the un-clipped depth to the fragment. I believe this gives each fragment the correct depth value. "

Is it really necessary? Which benefit you will get if fragment will have depth < 0 instead of 0? May be you  can omit that step and make the solution GL ES 2 compatible?

loganmc10:
It was necessary to make the driver show up in EP I Racer. Without doing that it is the same as my first solution

loganmc10:
Having the driver show up is some kind of effect related to fragment depth writing, if you disable fragment depth writing on PC, the driver will disappear

gonetz:
I understand: if polygon was fully behind near clipping plane, your fix will project it on XY plane. In that case gl_FragCoord.z correction is really necessary, otherwise depth compare will not work for other polygons, which are wholly or partially behind the near clipping plane.

loganmc10:
I tested that solution on my Nvidia machine and it didn't work, all kinds of flickering.

I updated the PR to use the initial fix I described (where it morphs the geometry to fit inside the clipping plane). It will work for GLES2/3, but it doesn't work for times when depth < 0, like displaying the driver in Ep I Racer.

@fzurita you can give this a try, it should fix the slingshot in Zelda and you should be able to see the pod in Ep I Racer, it should hopefully fix most other (but not all) clipping issues.

fzurita:
Ok, I will test.

loganmc10:
I pushed a commit to fix the driver in Ep I Racer (requires fragment depth writing). Everything looks really  good now, except the floor in OoT flickers (so do the shadows), same with the shadows in SM64. I believe I read about this in SM64 a while ago, and the glPolygonOffset stuff fixed it, I'm not exactly sure what is causing it in this case

loganmc10:
Well I fixed the polygon offset issue and tested this on my phone, it all seems to work properly.

I'll still do some more testing and compare performance, but the problem seems to be fixed

fzurita:
I can confirm that it works and it looks much better on the affected games. Also, episode one racer plays great with a CountPerOp value of 1. It runs at 60fps the whole time.

loganmc10:
I pushed one more commit that fixed the last remaining bug I saw, it was like this:

star_wars_ep1_racer-012

And now it looks proper, like this:

star_wars_ep1_racer-019


To summarize this change (everything is GLES only):

We define 3 "clamp modes", 0=enable clipping, 1=disable near clipping, 2=disable all clipping.

If fragment depth writing is not enabled/supported, we squeeze the Z value into the clipping plane, this will morph the geometry, though isn't really visible to the user. Doesn't work in all cases.

If fragment depth writing is enabled: If clamp mode > 0 (some clipping is disabled), we set gl_Position.z = 0 in vertex shader, and we pass it's original value to the fragment shader. Basically we disable all clipping/z processing in the vertex shader.

Then, in the fragment shader, we handle clipping by discarding clipped fragments ```if (uClampMode == 1 && (z_value > 1.0)) discard```. I had to also pass a ```uPolygonOffset`` uniform into the fragment shader, since bypassing the Z value in the vertex shader disables glPolygonOffset's effect. It actually seems to work better than glPolygonOffset based on my testing.

Hopefully this makes sense, I believe that the version using fragment depth writing should produce the exact same result as GL_DEPTH_CLAMP, without any morphing.

loganmc10:
I have finished bug testing this and I believe it is ready to be reviewed/merged.

src/Graphics/Context.h
```diff
@@ -39,6 +38,10 @@ namespace graphics {
 
 		void destroy();
 
+		void setClampMode(u32 _mode);
```
gonetz: Why not use enum for _mode?

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -229,6 +229,7 @@ class VertexShaderHeader : public ShaderPart
 			std::stringstream ss;
 			ss << "#version " << Utils::to_string(_glinfo.majorVersion) << Utils::to_string(_glinfo.minorVersion) << "0 es " << std::endl;
 			ss << "# define IN in" << std::endl << "# define OUT out" << std::endl;
+			ss << "OUT highp float z;" << std::endl << "uniform lowp int uClampMode;" << std::endl;
```
gonetz: z is a bad name for global variable.
Please choose something else, OriginalZ for example.

gonetz:
Very good work, thanks!
I have just two remarks. I have also one request: please test your fix with games, which rely on primitive depth: Duke Nukem 3D, Silicon Valley, Earth Worm Jim.

loganmc10:
Ok I added a commit to convert ClampMode to an enum, and the "z" variable was actually renamed to "vZCoord" before your review.

I tested those 3 games, I didn't notice anything off, everything looked completely normal to me.

gonetz:
The code looks good. I need to check it on my PC to be sure that nothing is broken.
BTW, do you have Android apk with that fix to test?

loganmc10:
Yes here is an apk with all these commits:

https://s3.amazonaws.com/m64p/m64p/android/mupen64plus-ae.apk

Just remember that for some of the effects to work the device will need to support GL_NV_shader_noperspective_interpolation, which is relatively rare (Adreno devices with Android 8/Oreo). Without that extension this change will just modify the Z value to fit inside the clipping plane in the Vertex shader. With GL_NV_shader_noperspective_interpolation it will correct the Z value in the fragment shader

gonetz:
@loganmc10 Thanks for the apk.
Quick test: Profile: Medium. Zelda OOT. Kokiri forest: problem with ground polygons. I don't know how to explain it: it seems that when you run, some polygons at the bottom are missing for a frame or two.  That is ground polygons blink. No problem, when you stay. I can't make a screen to show the problem. 
I tested the same place with current mupen64plus-fz : it works ok.
So, it is either problem with your build or with your fix.
Could you give me test apk from current master to be sure?

Tested on Samsung Galaxy Note 3, Android 5.0

loganmc10:
Yes you are right, I neglected to test that situation. It's an issue with the polygon offsetting.

I've updated the PR, the "cheating method" of squashing everything into the clipping plane doesn't seem to work, that is what was causing the issue you saw in OoT.

I've updated it so that it only works when using fragment depth writing, that seems to be the only way to properly handle this situation. I also got rid of the need for GL_NV_shader_noperspective_interpolation, so GLES devices with fragment depth writing enabled should look the same as the desktop mode. GLES devices without fragment depth writing enabled should look the same as they currently do.

https://s3.amazonaws.com/m64p/m64p/android/mupen64plus-ae-test1.apk don't forget to delete any binary shaders that might exist.

fzurita:
Does that mean that it will crash on Adreno devices when N64 depth compare is enabled?

loganmc10:
Yes it does. I figured that is the best alternative, since N64 depth compare is broken on all GLES devices currently, and this allows clipping to work properly on all GLES3 devices.

The crash is most definitely a GPU driver bug, do you want me to add a check for "Renderer::Adreno && N64DepthCompare == 1" before enabling this fix?

fzurita:
Yeah, I think that would be a good idea. Just to limit the number of crashes being reported.

gonetz:
My device does not support fragment depth writing: I have black screen with that option.
So, the fix has no impact on my device now.
@fzurita please confirm that the fix actually works.

fzurita:
With the latest as of right now, it seems to work in my device, at least with the slingshot. The ground in Kokiri looks normal to me.

fzurita:
Star Wars Episode I racer looks good as well.

gonetz:
@loganmc10 May be it is better squash all commits to one? Changes in the first commit partially reverted and rewritten in later commits. Do we need that history?

loganmc10:
Ok I squashed it all down to 1 commit, I also disabled this with when N64 Depth Compare is enabled, since it triggers a GPU driver bug in Adreno GPUs that causes a crash. N64 Depth Compare doesn't work in GLES anyway right now, if it ever gets working we can look at enabling this for N64 Depth Compare as well.

gonetz:
I tested it on my PC a bit, no new problems found. Thanks!

PR #1759, opened 2018-03-19
Use included font.ttf for mupen64plus

loganmc10:
mupen64plus is typically distributed with a "font.ttf" file for the built-in OSD. Since it seems a little tricky to always get a valid path from the OS, this will try to use that "font.ttf" file first, if it can't find it, it will still fall back to looking for the TTF file in the system dirs.

### DIFF
--- a/src/TextDrawer.cpp
+++ b/src/TextDrawer.cpp
@@ -25,6 +25,11 @@
 
 #include "TextDrawer.h"
 
+#ifdef MUPENPLUSAPI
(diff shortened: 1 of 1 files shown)


gonetz:
I don't like it.
This way user will not be able to set preferred system font - mupen64plus font will be set if exists.
Please correct the fix to support user-defined font.

loganmc10:
Good point, I corrected it to only use the mupen64plus font.ttf file if it couldn't find the user-defined font

gonetz:
Thanks!

Issue #1754, opened 2018-03-15
Legend of Zelda OoT- vanishing road texture


Kidboy17:
Using Project 64 2.3.2.3 with latest GLideN64 WIP. As soon as you leave Kokiri Forest and start on the dirt path of Hyrule field, the dirt path starts to disappear behind you. I have tried with default settings and with non-default and still happens. Here are some screen shots.

![gliden64_the_legend_of_zelda_002](https://user-images.githubusercontent.com/37382324/37448742-dbd6d392-27e4-11e8-8e71-2cd60053533f.png)

![gliden64_the_legend_of_zelda_001](https://user-images.githubusercontent.com/37382324/37448759-e6067dea-27e4-11e8-9674-d3afb8594d56.png)

![gliden64_the_legend_of_zelda_003](https://user-images.githubusercontent.com/37382324/37448764-ed5c3382-27e4-11e8-9b11-67f993c2be6c.png)

AmbientMalice:
Isn't this how a real N64 behaves?

OldGnashburg:
I've recently played OoT on my Nintendo 64 on a CRT, it does exist but it isn't this pronounced and you literally have to f*ck around with the camera to see it behind you.

loganmc10:
Duplicate of https://github.com/gonetz/GLideN64/issues/541

theboy181:
I have seen this on the real hw. I am not sure if this is the same exact issue but it can happen. It’s more noticeable in HLE I would imagine.

loganmc10:
It can happen on real HW, but it is much more severe in GLideN64, especially at higher resolutions. In fact, I can't get this to happen at 640x480, but if I switch to full screen, it happens constantly.

@gonetz I believe there is a better way to fix this than glPolygonOffset. See this commit: https://github.com/gonetz/GLideN64/pull/1747/commits/d13d2153dbb5cd750d2efe72b48be9e4a64bcfe9

Instead of relying on glPolygonOffset, which seems to have different results at different resolutions, and on different hardware, do this:

When you write to gl_FragDepth, just offset the depth there. Of course this only works with fragment depth writing enabled, so you'd need glPolygonOffset as a fallback, in my testing, this fixes this issue with the roads in Zelda, no matter the resolution.

OldGnashburg:
[...]

Issue #541, opened 2015-05-19
Ocarina of Time: Hyrule Field trail texture disappears behind link. 


WizLizard:
The trail texture in Hyrule Field starts to disappear behind link as he moves. This happens with default settings. Worse in full screen mode. Similar in LLE mode but the game has even more problems. Here's a video demonstrating the bug.

purplemarshmallow:
also happens with angrylion's plugin
![1](https://cloud.githubusercontent.com/assets/7283660/7729167/906557dc-ff13-11e4-8173-c4a36a51fabd.png)

standard-two-simplex:
Strange, I have never seen it on TV (real N64), or at least I can't recall. Anyway, I tried to reproduce and I eventually manged it, but the missing part was considerably smaller.

WizLizard:
@purplemarshmallow I will test this on my wii virtual console copy tonight to see if anything similar happens.

WizLizard:
Has anyone tested with glide64? I'm surprised it happens with angrylion's plugin.

WizLizard:
Glide64 is fine.

purplemarshmallow:
I tested with real hardware to make sure
same as with angrylion's plugin
So GLideN64 is correct
Glide64 uses a hack for the roads

WizLizard:
@purplemarshmallow It seems more severe with the plugin in fullscreen mode than actual hardware. Real hardware you're not able to turn around and see a huge chunk of road gone like in the video.

purplemarshmallow:
That's interesting
the missing part is bigger on higher resolutions
but must be a bug in game code

SwooshyCueb:
Yeah I'm fairly certain this is a bug in the game. I remember taking note of this the first time I played the game on real hardware all those years ago, though it does seem to be worse in the video provided here than I remember.

Still, it would be nice to have a fix, either with a gameshark code (if that's even doable) or a game-specific hack in the gfx plugin, or something

WizLizard:
@gonetz i would really like to hear your input on this when you have the time. Thanks.

gonetz:
[...]

PR #1763, opened 2018-03-23
Offset polygons in fragment shader

loganmc10:
This is a potential fix for https://github.com/gonetz/GLideN64/issues/1754 https://github.com/gonetz/GLideN64/issues/541

If fragment depth writing is enabled, this no longer uses glPolygonOffset to offset the polygons, but simply offsets the fragments in the fragment shader.

If fragment depth writing is not used, the glPolygonOffset is still used.

The number I used for the uniform (0.003f) was just based on testing OoT and Mario 64, it could be adjusted if needed.

### DIFF
--- a/src/Graphics/OpenGLContext/opengl_ContextImpl.cpp
+++ b/src/Graphics/OpenGLContext/opengl_ContextImpl.cpp
@@ -29,6 +29,7 @@ ContextImpl::~ContextImpl()
 void ContextImpl::init()
 {
 	m_clampMode = graphics::ClampMode::ClippingEnabled;
+	m_polygonOffset = false;
 	m_glInfo.init();
 
 	if (m_glInfo.isGLES2) {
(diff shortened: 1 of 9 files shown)


loganmc10:
This should accomplish the same thing as glPolygonOffset, from the spec:

> When GL_POLYGON_OFFSET_FILL, GL_POLYGON_OFFSET_LINE, or GL_POLYGON_OFFSET_POINT is enabled, each fragment's depth value will be offset after it is interpolated from the depth values of the appropriate vertices.

So we are still just offsetting each fragment's depth value after it is interpolated. But glPolygonOffset seems to vary depending on the platform, whereas offsetting the value in the fragment shader should give us consistent results.

src/Graphics/OpenGLContext/opengl_ContextImpl.cpp
```diff
@@ -116,9 +117,17 @@ void ContextImpl::enable(graphics::EnableParam _parameter, bool _enable)
 	m_cachedFunctions->getCachedEnable(_parameter)->enable(_enable);
 }
 
-u32 ContextImpl::isEnabled(graphics::EnableParam _parameter)
+void ContextImpl::polygonOffsetEnable(bool _enable)
 {
-	return m_cachedFunctions->getCachedEnable(_parameter)->get();
+	if (config.generalEmulation.enableFragmentDepthWrite == 0 && config.frameBufferEmulation.N64DepthCompare == 0)
```
gonetz: This condition looks redundant. 
if config.frameBufferEmulation.N64DepthCompare != 0, _enable is always false, see GraphicsDrawer.cpp line 1665
Thus, if (config.generalEmulation.enableFragmentDepthWrite == 0) should be enough.

gonetz:
I noticed a regression with that fix:
fix

Master:
master

Conclusion: fixed polygon offset is not right solution. Right solution would be well working N64 depth compare:
gliden64_super_mario_64_008

loganmc10:
I didn't notice that, but it's not necessarily anything wrong with the fix. uPolygonOffset is set to 0.003f when polygon offsetting is enabled. Changing that value changes how much shadow appears on the trunk of the tree. It's just a matter of finding the right value. I'm not sure if there would be a proper way to determine the value, or if it would just be trial and error.

gonetz:
I don't think that constant offset will work properly for every situation. As you see, amount of shadow on the trunk of the tree depends on distance to the tree when constant offset is used. The larger the distance the greater the shadow. This amount is constant with GL_POLYGON_OFFSET. Most likely polygon offset calculated dynamically in that case, per polygon. N64 depth compare calculates the offset per fragment.

loganmc10:
Yes I see that you are probably right, I don't think it will work as well using this method

PR #1769, opened 2018-03-27
Only write to gl_FragDepth once

loganmc10:
I was trying to fix N64 Depth Compare for GLES using this method. It didn't work (N64 Depth Compare is still broken for GLES).

However, I still think this would be a good change to incorporate. Given that there are drivers that seem very fragile around gl_FragDepth, I think it would be a good idea to write to it only once, and never read from it. It could possibly avoid driver bugs. You never know what kind of optimizations the driver might be trying to perform.

That's what this does. Also, for N64 Depth Compare, gl_FragDepth is never written to now, whereas before it would have been.

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
@@ -861,7 +861,7 @@ public:
 	{
 		if (!_glinfo.isGLES2) {
 			m_part =
-				"void writeDepth();\n";
+				"highp float writeDepth();\n";
 			;
 			if (_glinfo.isGLESX) {
 				m_part =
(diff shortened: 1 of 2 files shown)


loganmc10:
Maybe I misunderstand, but if the fragment is discarded, isn't the write to gl_FragDepth useless, because the fragment is never written?

Also, I only put it in ShaderFragmentRenderTarget because that was at the end of the shader, it will write gl_FragDepth anytime fragment depth writing is enabled and N64 depth compare is disabled

gonetz:
I read in OpenGL docs that if fragment shader writes to gl_FragDepth, then it is responsible for always writing it. May be I misunderstood it and discarded fragments can omit write to gl_FragDepth. In that case write to gl_FragDepth from ShaderFragmentRenderTarget only will be enough. TWINE can be used for testing that. The game will have depth compare issues right at the beginning without gl_FragDepth writes. It need to be tested with and without N64 depth compare, at least on NVidia and AMD.

loganmc10:
I will do more testing, but based on my reading, it says that the depth of the fragment is undefined if gl_FragDepth isn't written, and the depth of the fragment doesn't matter if it's discarded, it's not an error to omit writing it.

I think the issue would be if you had a branch/if statement, and you only wrote gl_FragDepth in one of the branches (and the fragment wasn't discarded)

gonetz:
You must be right - I see no new issues with TWINE.

PR #1768, opened 2018-03-26
Use graphics backend in GLideNHQ

loganmc10:
This is a fix for https://github.com/gonetz/GLideN64/issues/1765

GLideNHQ still references OpenGL directly. This is just a refactor to have it go through the graphics backend.

I had to change the ```switch``` statements to if/else because the compiler would not accept ```case u32(graphics::internalcolorFormat::RGBA8):``` type statements.

PR #1771, opened 2018-03-28
Fix android build after GLideNHQ refactor

fzurita:
https://github.com/gonetz/GLideN64/pull/1768 created a circular reference between the GLideNHQ and GLideN64 libraries. This is a problem in Android because all shared object libraries must not have undefined references at build time. By changing GLideNHQ into a static library, this problem goes away.


gonetz:
Merged:
ed5ebe3047856c

PR #1772, opened 2018-03-28
rpi: limit m_maxCacheSize to fix texture corruption

gizmo98:
Limit m_maxCachSize to fix texture corruption in beetle adventure
racing. I have tested values between 3500 and 8000. Everything >= 4000
produces texture corruption or stuttering after some time.

### DIFF
--- a/src/Textures.h
+++ b/src/Textures.h
@@ -100,7 +100,11 @@ private:
 	u32 m_hits, m_misses;
 	s32 m_curUnpackAlignment;
 	bool m_toggleDumpTex;
+#ifdef VC
+	const size_t m_maxCacheSize = 3500;
+#else
 	const size_t m_maxCacheSize = 8000;
+#endif
 };
 
 void getTextureShiftScale(u32 tile, const TextureCache & cache, f32 & shiftScaleS, f32 & shiftScaleT);


fzurita:
Why not do it for everything? It seems like the number of non-generated textures are usually small, relatively speaking. If the number of cached textures starts getting large, to me it seems like they are generated on the fly textures, so they shouldn't be cached.

loganmc10:
I wouldn't go much lower than 4000, probably some more testing would be needed, but we looked into this a while ago, see:

https://github.com/gonetz/GLideN64/issues/1443#issuecomment-290963608

And other comments in that thread. My conclusions seemed to be that around 4000 was the right number, I'm not sure how we settled on 8000.

The TextureCache class has a ```m_misses``` counter that counts the texture cache misses. Whatever change we make, we should make sure that we aren't increasing the ```m_misses``` count for whatever game we test, if that number is increasing, then we are reducing the effectiveness of the cache.

fzurita:
This is your analysis:

Of those 7396 unique CRC's:

5171 of them were used only once.
1354 of them were used twice
224 of them were used 3 times
282 of them were used 4 times
102 of them were used 5 times
30 of them were used 6 times
20 of them were used 7 times
10 of them were used 8 times
The remaining 200 or so were used quite frequently, with the highest number being textures that are re-used around 5000 times.

For textures that were used less than 10 times, I don't think it would had been worth keeping because they increase the cache size and thus the amount of time searching the cache. It may be faster to just upload the texture again rather than search such a large cache.

So may point is... maybe even lower the number to 500 textures.

loganmc10:
I think you underestimate how slow texture uploads (glTexSubImage2D) can be. Most actual OpenGL applications would never upload textures during gameplay. They would do it all during some kind of loading screen, or they would do it asynchronously (load a texture in a separate thread that you know OpenGL won't need to access for many frames).

Unfortunately for emulators, they have to do it on-the-fly, but it should still be avoided at all costs.

The issue here is just that the Raspberry Pi is running out of VRAM. It has an unusually low amount of VRAM in comparison to most modern hardware. According to their documents it defaults to 64MB (https://www.raspberrypi.org/documentation/configuration/config-txt/memory.md) but that can be increased by the user. I assume setting gpu_mem=512 in /boot/config.txt would fix this problem as well.

fzurita:
I'm curious about how slow texture upload are. I'll do some quick testing with a small cache to compare performance.

fzurita:
Oh, while looking at that old issue again, I was reminded about doing a neon implementation of CRC32. I actually did some experimenting with this. PPSSPP already has a neon implementation of CRC32. It was very easy to bring to GLideN64 since the signature matches the xxHash library exactly. Unfortunately, I didn't notice any performance difference.

fzurita:
Some quick testing, in CBFD at the starting at the waterfall, after jumping around in the water to increase generated textures:

Cache size of 10 textures = ~51fps
Cache size of 50 textures = ~60fps
Cache size of 200 textures = ~65fps
Cache size of 500 textures = ~65fps
Cache size of 8000 textures = ~64fps

After 200 textures in my testing, there wasn't much of an improvement. In my device it didn't hurt performance either. I bet though that in slower devices, the performance penalty would kick in much earlier.

loganmc10:
Right, the problem is there are ~300 games that could have different behaviour. We know (like your test shows), that the 8000 size performs roughly the same as 200, but with 8000 you have a greater chance of reducing texture uploads.

This is why there used to be a MB limit to the texture cache, but then there was the opposite problem, at high resolutions there was corruption because the texture cache was too small to hold everything.

Nvidia and ATI actually have extensions to query the available VRAM, but I don't think any GLES drivers have that

loganmc10:
I think the ifdef for VC is acceptable, there are other more gimmicky ways to try and determine the VRAM size like keep uploading/deleting a larger and larger texture until you get an out of memory error, that kind of thing. But we know that the Raspberry Pi generally has a very low amount of VRAM

Issue #1765, opened 2018-03-26
Advertise the GLES2 macro


hissingshark:
Hi,

I've been building on the Vero4k (Amlogic s905x box running OSMC/debian).

My platform is GLESv2.  I could only build by using `-DEGL=On` and an unadvertised macro `-DGLES2=On`.  The `readme` and `CMakeLists` make no mention of this.  I just found it when going through the source to debug.  Might be worth a mention in those places to aid the unwary in future.

I hope that's helpful to you devs.
Many thanks.

loganmc10:
What happens if you build without -DGLES2=On? You shouldn't need that to get it working with GLES2, the Android version supports GLES2 without doing that.

hissingshark:
`echo $CFLAGS`

> -I/opt/vero3/include -L/opt/vero3/lib -O3 -march=armv8-a+crc -mtune=cortex-a53 -mfloat-abi=hard -mfpu=neon-fp-armv8 -funsafe-math-optimizations

`cmake -DEGL=On -DMUPENPLUSAPI=On -DVEC4_OPT=On -DNEON_OPT=On -DCRC_ARMV8=On -DUSE_SYSTEM_LIBS=On ../../src/`

> -- The C compiler identification is GNU 6.3.0
> -- The CXX compiler identification is GNU 6.3.0
> -- Check for working C compiler: /usr/bin/cc
> -- Check for working C compiler: /usr/bin/cc -- works
> -- Detecting C compiler ABI info
> -- Detecting C compiler ABI info - done
> -- Detecting C compile features
> -- Detecting C compile features - done
> -- Check for working CXX compiler: /usr/bin/c++
> -- Check for working CXX compiler: /usr/bin/c++ -- works
> -- Detecting CXX compiler ABI info
> -- Detecting CXX compiler ABI info - done
> -- Detecting CXX compile features
> -- Detecting CXX compile features - done
> -- Found ZLIB: /usr/lib/arm-linux-gnueabihf/libz.so (found version "1.2.8") 
> -- Found PNG: /usr/lib/arm-linux-gnueabihf/libpng.so (found version "1.6.28") 
> -- Found Freetype: /usr/lib/arm-linux-gnueabihf/libfreetype.so (found version "2.6.3") 
> -- Configuring done
> -- Generating done
> -- Build files have been written to: /home/osmc/GLideN64/projects/cmake

`make`

> Scanning dependencies of target osal
[...]

PR #1774, opened 2018-03-29
Fix GLES2 shader storage support

loganmc10:
This is to fix a crash reported in https://github.com/gonetz/GLideN64/issues/1765 (glProgramParameteri does not exist in GLES2)

I also changed GLES2 to use glProgramBinaryOES and glGetProgramBinaryOES.

Finally, we used to just check for GL_OES_get_program_binary or GL_ARB_get_program_binary, but I added a check for GLES 3.0+ or GL4.1+, since program binaries are not extensions in those versions.


gonetz:
Looks good, thanks!

PR #1775, opened 2018-03-29
Added VERO4K flag - triggering the custom path to libGLESv2.so

hissingshark:
As per #1765 - this helps to fix a segmentation fault at run-time due to the non-standard path to `libGLESv2.so` on the Vero4k.

Also needs #1774 to be merged.
Until then requires `EnableShadersStorage = False` to be set in `mupen64plus.cfg`.

Even then, the shader cache at `/home/.cache/mupen64plus/shaders` must be deleted prior to each run.

### DIFF
--- a/projects/cmake/Readme.txt
+++ b/projects/cmake/Readme.txt
@@ -11,3 +11,5 @@ cmake [-DCMAKE_BUILD_TYPE=Debug] [-DVEC4_OPT=On] [-DCRC_OPT=On] [-DX86_OPT=On] [
 -DNOHQ=On - optional parameter. set to build without realtime texture enhancer library (GLideNHQ).
 -DUSE_SYSTEM_LIBS=On - optional parameter. set to use system provided libraries for libpng and zlib.
 -DMUPENPLUSAPI=On - required parameter. currently cmake build works only for mupen64plus version of the plugin.
+-DODROID=On - set if you need to build on an Odroid board.
+-DVERO4K=On - set if you need to build on the OSMC Vero4k.
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -9,6 +9,7 @@ option(PANDORA "Set to ON if targeting an OpenPandora" ${PANDORA})
 option(ODROID "Set to ON if targeting an Odroid" ${ODROID})
 option(MUPENPLUSAPI "Set to ON for Mupen64Plus plugin" ${MUPENPLUSAPI})
 option(MESA "Set to ON to disable Raspberry Pi autodetection" ${MESA})
+option(VERO4K "Set to ON if targeting a Vero4k" ${VERO4K})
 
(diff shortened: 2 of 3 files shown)


gonetz:
Please add a notice about new cmake option to projects/cmake/Readme.txt.
Please add that change to this commit .

hissingshark:
Apologies.  The irony isn't lost on me that I came here in the first place to helpfully report an undocumented build flag!
I've also added the Odroid flag as it too wasn't in the readme.txt, but does much the same thing for that platform.

projects/cmake/Readme.txt
```diff
@@ -4,10 +4,13 @@ cmake [-DCMAKE_BUILD_TYPE=Debug] [-DVEC4_OPT=On] [-DCRC_OPT=On] [-DX86_OPT=On] [
...
--DNOHQ=On - optional parameter. set to build without realtime texture enhancer library (GLideNHQ).
--DUSE_SYSTEM_LIBS=On - optional parameter. set to use system provided libraries for libpng and zlib.
--DMUPENPLUSAPI=On - required parameter. currently cmake build works only for mupen64plus version of the plugin.
+-DX86_OPT=On - optional parameter. set it if you want to enable additional X86 ASM optimization (can cause additional bugs).
+-DNOHQ=On - build without realtime texture enhancer library (GLideNHQ).
+-DMUPENPLUSAPI=On - currently cmake build works only for mupen64plus version of the plugin.
+-DUSE_SYSTEM_LIBS=On - set to use system provided libraries for libpng and zlib.
+-DODROID=On - set if you need to build on an Odroid board.
+-DVERO4K=On - set if you need to build on the OSMC Vero4k.
+
```
gonetz: Why so many changes when you just need to add two options and thus two lines of text?

hissingshark: Sorry, I've not added to a commit before.
I thought I did just that, committed to master with the 2 lines at the bottom.  Simple enough as you say.
But then Git declared there was a merge conflict, and showed the lines all shuffled for some reason.  So having corrected the order again this is where we are.  I'm not proud of it, but there it is.
How would you like to proceed?

gonetz: Please take your original commit, which is without corrections in Readme.txt, and rebase it to current master. Then add these two lines to Readme.txt, update the commit and PR it there.

loganmc10:
The file change list looks good now, all you need to do is squash it all into one commit, lookup "git squash" if you've never done it before

hissingshark:
Will do.  Learning all the way.  
I see what happened earlier now.  Since putting in a PR for my first commit the very file I wanted to append had been cleaned up, hence the shuffling of the lines...  Needed to have re-synced with upstream in the meantime.
Thank you all for your patience.

gonetz:
Looks good, thanks!

PR #1776, opened 2018-03-30
iOS Support Fixed

JoeMatt:
Recreation of @braindx 's PR #1732 with the requested fixes.

~~Some header paths still had to be changed for when we build with our Xcode project.
I combined those changes into a single commit though at least so it can be reverted out if need be.~~

An updated make file would need to be created for iOS users looking to build for other front ends.

I've had success with the following C flags
```
-mfpu=neon -pthread -fPIC -D__unix__ -D__NEON_OPT -D__VEC4_OPT -DMUPENPLUSAPI -DTXFILTER_LIB -DOS_IOS -DGLESX
```

### DIFF
--- a/src/TextDrawerStub.cpp
+++ b/src/TextDrawerStub.cpp
@@ -1,0 +1,30 @@
+/* Stub for TextDrawer.h
+ * Use to replace remove freetype library requirement.
+ */
+
+#include "TextDrawer.h"
+
+TextDrawer g_textDrawer;
+
+struct Atlas {
+};
+
+void TextDrawer::init()
+{
+}
+
+void TextDrawer::destroy()
+{
+}
+
+void TextDrawer::drawText(const char *_pText, float _x, float _y) const
+{
+}
+
+void TextDrawer::getTextSize(const char *_pText, float & _w, float & _h) const
+{
+}
+
+void TextDrawer::setTextColor(float * _color)
+{
+}
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -40,7 +40,7 @@ void Config::resetToDefaults()
 	generalEmulation.enableNativeResTexrects = 0;
 	generalEmulation.enableLegacyBlending = 0;
 	generalEmulation.hacks = 0;
-#ifdef OS_ANDROID
+#if defined(OS_ANDROID) || defined(OS_IOS)
 	generalEmulation.enableFragmentDepthWrite = 0;
 	generalEmulation.enableBlitScreenWorkaround = 0;
 	generalEmulation.forcePolygonOffset = 0;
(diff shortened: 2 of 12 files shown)


src/TextDrawerStub.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include "Graphics/Context.h"
+#include "Graphics/Parameters.h"
+
+#include "TextDrawer.h"
+
+#ifdef MUPENPLUSAPI
+#include "mupenplus/GLideN64_mupenplus.h"
+#include <osal_files.h>
+#endif
+
```
gonetz: I guess that most of includes are useless there. Please clean up.

src/TextDrawerStub.cpp
```diff
@@ -0,0 +1,64 @@
+/* Draw text on screen.
+ * Requires freetype library.
+ * Code is taken from "OpenGL source examples from the OpenGL Programming wikibook:
+ * http://en.wikibooks.org/wiki/OpenGL_Programming"
+ */
+
```
gonetz: This comment does not correspond to the code below. Better remove it at all.

src/Log.cpp
```diff
@@ -16,6 +16,9 @@ void LOG(u16 type, const char * format, ...) {
 
 #ifdef OS_WINDOWS
 	FILE *dumpFile = _wfopen(logPath, wst("a+"));
+#elif defined(OS_IOS)
+    FILE *dumpFile = stdout;
+    fputs("Gliden: ", dumpFile);
```
gonetz: GLideN64:

src/Neon/gSPNeon.cpp
```diff
@@ -2,7 +2,7 @@
 #include <math.h>
 #include <algorithm>
 #include <assert.h>
-#include "N64.h"
+#include "../N64.h"
```
gonetz: Have  you tried to force Xcode to look at source folder for headers?
May be CMAKE_CURRENT_SOURCE_DIR can help?
For example, replace
include_directories( . inc )
by
include_directories( ${CMAKE_CURRENT_SOURCE_DIR} inc )

It looks strange that it can't find headers located in folders contained in includes path.

JoeMatt: It was finding headers in other folders like Rice and causing conflicts. I had to use 'Always search user paths' combined with 'User header maps' to 'off' which is a setting I had to look up but it makes Xcode not go around looking all over the project dir for headers of the same name. I'll revert these changes in a short bit.

gonetz:
The code looks good now, thanks! However, when I loaded PR to my local repo and opened files you modified, I noticed that you use spaces for indentations. The project uses "Tabs only" policy for indentations, thus your changes look out of style. Please correct it. Also, please rebuild your chain of commits: squash commits, which modify the same files to one commit; remove "Use relative paths for some headers that XCode was not finding" and its revert from the chain.

JoeMatt:
I had to do a hard reset on my remote to get the squash to work on this PR, which auto-closed it, but it's reopened with my clean git history

gonetz:
Looks clean now, thanks!

PR #1779, opened 2018-04-01
Support stereoscopic rendering

justinmichaud:
See comment below

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -116,8 +116,7 @@ void Config::resetToDefaults()
 
 	debug.dumpMode = 0;
 	stereo.enabled = false;
-	stereo.separation = 0;
-	stereo.convergence = 0;
+	stereo.eyedistance = 0;
 }
 
 bool isHWLightingAllowed()
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -115,6 +115,7 @@ void Config::resetToDefaults()
 	onScreenDisplay.pos = posBottomLeft;
 
 	debug.dumpMode = 0;
+	stereo.enabled = 0;
 }
 
 bool isHWLightingAllowed()
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -116,7 +116,8 @@ void Config::resetToDefaults()
 
 	debug.dumpMode = 0;
 	stereo.enabled = false;
-	stereo.eyedistance = 0;
+	stereo.separation = 0;
+	stereo.convergence = 0;
 }
 
 bool isHWLightingAllowed()
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -115,7 +115,8 @@ void Config::resetToDefaults()
 	onScreenDisplay.pos = posBottomLeft;
 
 	debug.dumpMode = 0;
-	stereo.enabled = 0;
+	stereo.enabled = false;
+	stereo.eyedistance = 0;
 }
 
 bool isHWLightingAllowed()
--- a/src/Config.h
+++ b/src/Config.h
@@ -175,7 +175,7 @@ struct Config
 
 	struct {
 		bool enabled;
-		f32 convergence, separation;
+		f32 eyedistance;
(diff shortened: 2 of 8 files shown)


fzurita:
Cool, yes, we really need to fix NEON before it can be merged. I can see you fixed the black sky issue, that's good. Look at the option of "Render 2d elements in N64 resolution". That may help you not render 2d elements twice.

justinmichaud:
@fzurita Unfortunately, the black sky issue is still present, but since the viewport is never rotated it is not common. It seems that super mario 64 uses a billboard of some kind to draw the sky, and so the game would need to be patched to fix that.

I will take a look at the option you mentioned. What should it help with? Right now, the issue isn't that I have to draw the elements twice as much as that they no longer line up correctly with 3d game objects. Might this option potentially help with that?

fzurita:
That option may or may not help with that.  I'm just saying that you should try toggling it to see what it does. It's good for games like Mario Kart 64. It may help in your case, but I don't really know, and it's worth a try.

justinmichaud:
@fzurita I am having a bit of a strange problem with both Mario Kart 64 and Zelda. It seems that for these games, the model view matrix transforms points into world space, and then the projection matrix handles the camera rotation, scaling and projection. As a result, the eye translation happens before the points are transformed into camera space.

Do you have any ideas for how I could get just the perspective matrix alone (or rather, its inverse)?

fzurita:
You got me, you are venturing into an area I'm not as familiar with. @gonetz can probably help though.

justinmichaud:
It looks like http://developer.download.nvidia.com/whitepapers/2010/NVIDIA%203D%20Vision%20Automatic.pdf might be a better solution. I am going to look into it.

This is what dolphin uses: https://en.dolphin-emu.org/blog/2015/05/13/a-second-perspective/

gonetz:
> @gonetz can probably help though.

Hardly. RSP uses two kind of matrices: model and projection. There is stack of model matrices and one projection matrix. Vertices are transformed to eye space using combined matrix, which is result of multiplication of model and projection matrix. RSP also can use ready combined matrix, which is loaded from RDRAM with ForceMatrix command.
Most of 2D elements has only screen coordinates, with no translation of vertices from world space to screen space.

justinmichaud:
screenshot_20180405-223342
Separation: 8, convergence: -10000

This demonstrates the current problem, which is that each camera is tilted in rather than pointing straight ahead. I am going to do more research to see how to fix that.

theboy181:
Fingers crossed.  You join discord yet?

justinmichaud:
@theboy181 Would it be possible for me to join the discord?

theboy181:
https://discord.gg/sxf5kc

gonetz:
@justinmichaud I'm busy atm with other tasks. I'll check your PR as soon I as complete with urgent stuff.

fzurita:
I'm seeing strange depth issues in Super Mario 64 in Android with a Adreno 540:

super_mario_64-001

Also, performance really suffers, I'm only getting around 35fps max in Super Mario 64. This game can usually hit around 160fps or higher.

fzurita:
Another thing, "widescreen" should be forced to enabled I think when stereoscopic rendering is enabled.

justinmichaud:
@fzurita That image is a little unsettling :P

I think it probably a precision issue that has something to do with the projection matrix that I use to re-project the vertices. I will take a look to see if I can make it better,

The performance drop is a little worrying. I would have expected each frame to take twice as long, not 5 times. Do you see any way to make this perform better, or an easy way to move these computations to the gpu?

fzurita:
I can't think of an easy way to move it to the GPU. Maybe a vertex shader program could do it though.

I can think of a very simple fragment shader that would simply duplicate the screen twice, side by side. I'm not sure how I would do it with a vertex shader though.

BiatuAutMiahn:
You might be able to test with Reshade/SuperDepth3D Shader

Issue #1777, opened 2018-03-31
Beetle Adventure Racing broken menu background


fzurita:
This commit broke the Beetle Adventure Racing background in menus:

https://github.com/gonetz/GLideN64/commit/7aeb1018ea19c29389fcc2839033d136453893ec

It looks like this after that commit:

![beetle_adventure_rac-001](https://user-images.githubusercontent.com/11581687/38159703-213fb2e8-347c-11e8-9bd7-12e124c885fd.png)

fzurita:
@gonetz This is never succeeding because ```clamps``` is always ```1```

https://github.com/gonetz/GLideN64/blob/master/src/FrameBuffer.cpp#L439

Why do we check for ```clamps``` or ```clampt``` to equal to zero before getting the "sub texture"?

gonetz:
@fzurita I just tested BAR on my PC - it works with no problem. Do you have this issue on mobile device?

Regarding check for clamps or clampt. Usually plugin can re-map texture coordinates to read data from original buffer texture. However, if S or T clamping is not set, texture wrapping is possible. The wrapping will not work correctly with whole buffer, thus necessary part of it is copied to a new sub-texture. Thus, checks for clamps and clampt save us from redundant texture copy.

gizmo98:
@fzurita
Please test without NEON_OPT. I have seen equal problems if CopyMatrix() was not implemented correctly.

fzurita:
@gonetz Yes, this is on a mobile device. @gizmo98 I'll check without neon.

fzurita:
@gizmo98 Same issue with NEON and VEC4 optimizations disabled.

This issue was specifically introduced by https://github.com/gonetz/GLideN64/commit/7aeb1018ea19c29389fcc2839033d136453893ec

loganmc10:
@fzurita does this happen on the Shield? I'm curious if it's GLES related or ARM/Android related

fzurita:
It doesn't happen with the Shield TV. So it's not an ARM issue, it's a GLES issue.

gonetz:
@fzurita what about frame buffer effects in other games?
For example, subscreen in Zelda OOT uses the same code.
Also, please check BAR menu with fb emulation disabled.

fzurita:
BAR with FBE disabled:
![beetle_adventure_rac-002](https://user-images.githubusercontent.com/11581687/38170338-4795b94e-3551-11e8-9e93-aaba5510d366.png)

[...]

PR #1784, opened 2018-04-05
Remove Image Texture requirement for some shaders

loganmc10:
This removes the Image Texture requirement for the fog effect in Beetle Adventure Racing and the depth in Resident Evil 2.

Tested on Windows + Intel. I haven't tested this on GLES.

@fzurita This should make RE2 work in GLES I would assume. It should also make the BAR fog work on all devices. Hopefully it might fix the bug in https://github.com/gonetz/GLideN64/issues/1777

### DIFF
--- a/src/Graphics/Context.cpp
+++ b/src/Graphics/Context.cpp
@@ -5,9 +5,6 @@ using namespace graphics;
 
 Context gfxContext;
 
-bool Context::imageTextures = false;
-bool Context::multisampling = false;
-
 Context::Context() {}
 
 Context::~Context() {
@@ -20,8 +17,6 @@ void Context::init()
 	m_impl.reset(new opengl::ContextImpl);
 	m_impl->init();
 	m_fbTexFormats.reset(m_impl->getFramebufferTextureFormats());
-	imageTextures = isSupported(SpecialFeatures::ImageTextures);
-	multisampling = isSupported(SpecialFeatures::Multisampling);
 }
 
 void Context::destroy()
--- a/src/Graphics/Context.cpp
+++ b/src/Graphics/Context.cpp
@@ -5,6 +5,15 @@ using namespace graphics;
 
 Context gfxContext;
 
+bool Context::Multisampling = false;
+bool Context::BlitFramebuffer = false;
+bool Context::WeakBlitFramebuffer = false;
+bool Context::DepthFramebufferTextures = false;
+bool Context::ShaderProgramBinary = false;
+bool Context::ImageTextures = false;
+bool Context::IntegerTextures = false;
+bool Context::ClipControl = false;
+
 Context::Context() {}
 
 Context::~Context() {
@@ -17,6 +26,14 @@ void Context::init()
 	m_impl.reset(new opengl::ContextImpl);
 	m_impl->init();
 	m_fbTexFormats.reset(m_impl->getFramebufferTextureFormats());
+	Multisampling = m_impl->isSupported(SpecialFeatures::Multisampling);
+	BlitFramebuffer = m_impl->isSupported(SpecialFeatures::BlitFramebuffer);
+	WeakBlitFramebuffer = m_impl->isSupported(SpecialFeatures::WeakBlitFramebuffer);
+	DepthFramebufferTextures = m_impl->isSupported(SpecialFeatures::DepthFramebufferTextures);
+	ShaderProgramBinary = m_impl->isSupported(SpecialFeatures::ShaderProgramBinary);
+	ImageTextures = m_impl->isSupported(SpecialFeatures::ImageTextures);
+	IntegerTextures = m_impl->isSupported(SpecialFeatures::IntegerTextures);
+	ClipControl = m_impl->isSupported(SpecialFeatures::ClipControl);
 }
 
 void Context::destroy()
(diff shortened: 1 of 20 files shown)


loganmc10:
Ok I disabled it for GLES2, GLES2 doesn't support integer textures either, so the shader would need to be modified as well as a few other things.

fzurita:
I just tried it. RE2 still does work, at least with Adreno. It doesn't fix BAR menus with Adreno either.

loganmc10:
Did you have fragment depth writing enabled for RE2? That is very strange.

The BAR thing is really bizarre. Either way I think this is a good change since it removes the Image Texture requirement, it's too bad that bug still exists though.

fzurita:
Yes, I had fragment depth write enabled.

loganmc10:
Well the good news is now that it doesn't depend on Image Textures, I can test this GLES code on my laptop (supports GLES3.0). I pushed a change that may fix the BAR menus, they were broken on my laptop in GLES mode (the usampler2D needed a precision set).

I notice that RE2 has clipping disabled, and it is also broken on my laptop in GLES mode. I assume it might need the recent depth clamp emulation to work (I'll have to test on my Adreno device that supports that soon)

fzurita:
The menu still doesn't work with Adreno, I think that's just a driver bug.

fzurita:
Why did you switch the precision from highp to lowp for zlut and tlut images?

loganmc10:
lowp is the default precision qualifier for a sampler2D. usampler2D has no default, so I assumed lowp would also be a good default. My testing showed no difference. I'm not actually sure what the difference between lowp and highp is when you're dealing with integers. But everything I read says lowp is better for performance.

fzurita:
My assumption would be something like 8/16/32 bit. Higher bits allow for bigger numbers. But you are right, normally the scale here is 0 to 255, so lowp would be fine.

loganmc10:
Well that uniform is actually just holding the name of the texture (like the ID number), it's not storing the texture data, so its just the GL identifier for the texture. I don't think high precision would do anything differently.

fzurita:
Yeah, another good point.

loganmc10:
Anyway we now know that the BAR menu works with Mesa GLES, Mali, Nvidia, and other desktop drivers. That combined with the weird glBlit behaviour makes it pretty conclusive that there is some really weird Adreno bug with the BAR menu.

fzurita:
Yep, I completely agree. I have an idea of a workaround that is efficient as well, I just need to find the time to implement it.

loganmc10:
I pushed one unrelated commit since it will fix a regression caused by https://github.com/gonetz/GLideN64/commit/af6460ff4f40f0d0145afcd80ae8707d9dd33b6f

Since that commit only fixes clipping on GLES when GL_NV_shader_noperspective_interpolation is supported, the clipping hack in GraphicsDrawer needs to be enabled always for GLESX.

I also changed the check for if FragmentDepthWrite is supported to if it is enabled. We shouldn't really care if it is supported, it can be enabled or disabled, and we force it to be disabled for GLES2 anyway

src/Graphics/Context.h
```diff
@@ -17,12 +17,13 @@ namespace graphics {
...
 	enum class SpecialFeatures {
 		Multisampling,
-		FragmentDepthWrite,
 		BlitFramebuffer,
 		WeakBlitFramebuffer,
 		DepthFramebufferTextures,
 		ShaderProgramBinary,
-		ImageTextures
+		ImageTextures,
+		LUTTextures,
```
gonetz: LUTTextures is a bad name. Can you pick more general name, which reflect essence of that special feature?

src/Graphics/Context.cpp
```diff
@@ -20,8 +17,6 @@ void Context::init()
 	m_impl.reset(new opengl::ContextImpl);
 	m_impl->init();
 	m_fbTexFormats.reset(m_impl->getFramebufferTextureFormats());
-	imageTextures = isSupported(SpecialFeatures::ImageTextures);
-	multisampling = isSupported(SpecialFeatures::Multisampling);
```
gonetz: Why? These variables cached most often requested special features.

gonetz: You may add variable for LUTTexture as well.

src/Graphics/OpenGLContext/opengl_TextureManipulationObjectFactory.cpp
```diff
@@ -191,7 +191,6 @@ namespace opengl {
 				}
 
 				if (_params.ImageUnit.isValid()) {
-					assert(IS_GL_FUNCTION_VALID(glBindImageTexture));
```
gonetz: Why removed?

src/Graphics/OpenGLContext/opengl_TextureManipulationObjectFactory.cpp
```diff
@@ -273,7 +272,6 @@ namespace opengl {
 				_params.data);
 
 			if (_params.ImageUnit.isValid() && _params.internalFormat.isValid()) {
-				assert(IS_GL_FUNCTION_VALID(glBindImageTexture));
```
gonetz: Why removed?

gonetz:
> Ok I changed all the Special Features checks to use Context:: style checks, this should be faster and keep everything consistent. Is that alright?

Yes, looks good. Thanks!

PR #1785, opened 2018-04-06
Only call glTexSubImage2D is there is data

loganmc10:
Sometimes TextDrawer calls Update2DTexture will a nullptr for the data.

Since we don't use PBO's for texture uploads anymore, a nullptr is never valid for glTexSubImage2D. The Intel Windows driver was flagging this as "undefined behavior"


gonetz:
> Sometimes TextDrawer calls Update2DTexture will a nullptr for the data.

May be fix TextDrawer then? When and how it happens?

loganmc10:
The very first update call has a nullptr, so I assume g->bitmap.buffer is null for some reason. I don't know enough about how the text drawing works to know why that is

gonetz:
Thanks, I'll check it.

gonetz:
I found it. Space character, ASCII 0x20 , has no bitmap.

gonetz:
c57eaf52850ac

Issue #1702, opened 2018-01-04
Majora's Mask - flashback cutscene glitched


Nerrel:
This problem has occurred in every release of GlideN64 I've used, even predating the 2.0 release, but I did not see an open issue for it.  

There is a flashback near the beginning of the game where Zelda gives Link the ocarina of time.  It's in black and white and seems to have the framebuffer blur effect in heavy use along with a grain filter of sorts.  In past versions, the scene would not render at all and there would only be a black image, but text and 2D elements would still display.

In the current release, things are improved and the 3D objects are visible, but heavily glitched.  Textures and portions of models flicker in and out of visibility, and the black widescreen bars don't show up.  

https://c1.staticflickr.com/5/4593/38797285084_10ca0732ac_c.jpg
https://c1.staticflickr.com/5/4643/24637828057_60a1c9b8c1_c.jpg

It does not seem to matter what settings I use... aside from setting the aspect ratio to "try to fit" and using a higher resolution, the settings are at default.  If it helps, I can upload a savestate from just before the scene takes place later when I get time.

gonetz:
It is a problem on your side. This cut scene worked with 2.0 release and it works with the current one:
![gliden64_zelda_majora s_mask_000](https://user-images.githubusercontent.com/8983092/34597438-37e921dc-f219-11e7-9f28-bd995505624a.png)

No special settings needed. Note: the game may not work correctly if you use hacked rom. Plugin uses some hacks to emulate this game, and these hacks enabled if internal name of the rom is ZELDA_MAJORA'S_MASK. Also, be sure that you run the game in HLE mode. Frame buffer emulation work incorrect in LLE for this game.

fzurita:
@Nerrel can you post a save state?

gonetz:
@fzurita PJ64 save:
https://drive.google.com/file/d/1sBMU7VJV9J3C5U-5PRCz0dlV7lDWYKuA/view?usp=sharing

fzurita:
In Android, this scene is very glitched. Much worse than what @Nerrel is showing. I can take a look.

Nerrel:
I'm unable to get the scene to work... the rom (USA version) isn't hacked, and I've always used HLE.  I also tried setting everything to default, like the aspect ratio and resolution, just in case.  Anything else I can try?

fzurita:
Is this scene using the monochrome effect hack?

loganmc10:
@fzurita can you provide a mupen64plus save state?

[...]

PR #1786, opened 2018-04-06
Workaround for Adreno issues with buggy FBOs

fzurita:
If we are updating texture parameters on a frame buffer texture, and the previous frame buffer texture for which we are updating parameters has not been blit at any point, then we will perform a zero size blit to a fbo that was allocated just for this workaround.

This is because with Adreno, the color buffer for an FBO is getting thrown away if it's not used at least once, so when we go and use the FBO's texture in a shader, there is nothing there.

### DIFF
--- a/src/BufferCopy/ColorBufferToRDRAM.cpp
+++ b/src/BufferCopy/ColorBufferToRDRAM.cpp
@@ -101,11 +101,14 @@ void ColorBufferToRDRAM::_initFBTexture(void)
 	}
 	{
 		Context::FrameBufferRenderTarget bufTarget;
-		bufTarget.bufferHandle = ObjectHandle(m_FBO);
+		bufTarget.bufferHandle = m_FBO;
 		bufTarget.bufferTarget = bufferTarget::DRAW_FRAMEBUFFER;
 		bufTarget.attachment = bufferAttachment::COLOR_ATTACHMENT0;
 		bufTarget.textureTarget = textureTarget::TEXTURE_2D;
 		bufTarget.textureHandle = m_pTexture->name;
+		m_pTexture->fbo = m_FBO;
+		m_pTexture->isFboValid = true;
+		m_pTexture->isDepthAttachment = false;
 		gfxContext.addFrameBufferRenderTarget(bufTarget);
 	}
 
--- a/src/BufferCopy/DepthBufferToRDRAM.cpp
+++ b/src/BufferCopy/DepthBufferToRDRAM.cpp
@@ -113,10 +113,16 @@ void DepthBufferToRDRAM::init()
 	bufTarget.attachment = bufferAttachment::COLOR_ATTACHMENT0;
 	bufTarget.textureTarget = textureTarget::TEXTURE_2D;
 	bufTarget.textureHandle = m_pColorTexture->name;
+	m_pColorTexture->fbo = m_FBO;
+	m_pColorTexture->isFboValid = true;
+	m_pColorTexture->isDepthAttachment = false;
 	gfxContext.addFrameBufferRenderTarget(bufTarget);
 
 	bufTarget.attachment = bufferAttachment::DEPTH_ATTACHMENT;
 	bufTarget.textureHandle = m_pDepthTexture->name;
+	m_pDepthTexture->fbo = m_FBO;
+	m_pDepthTexture->isFboValid = true;
+	m_pDepthTexture->isDepthAttachment = true;
 	gfxContext.addFrameBufferRenderTarget(bufTarget);
 
 	// check if everything is OK
--- a/src/DepthBuffer.cpp
+++ b/src/DepthBuffer.cpp
@@ -198,6 +198,10 @@ void DepthBuffer::setDepthAttachment(ObjectHandle _fbo, BufferTargetParam _targe
 		params.textureHandle = m_depthRenderbuffer;
 		params.textureTarget = textureTarget::RENDERBUFFER;
 	}
+
+	m_pDepthBufferTexture->fbo = _fbo;
+	m_pDepthBufferTexture->isFboValid = true;
+	m_pDepthBufferTexture->isDepthAttachment = true;
 	gfxContext.addFrameBufferRenderTarget(params);
 
 	m_copied = false;
@@ -235,6 +239,9 @@ CachedTexture * DepthBuffer::resolveDepthBufferTexture(FrameBuffer * _pBuffer)
 	targetParams.bufferTarget = bufferTarget::DRAW_FRAMEBUFFER;
 	targetParams.textureHandle = m_pResolveDepthBufferTexture->name;
 	targetParams.textureTarget = textureTarget::TEXTURE_2D;
+	m_pResolveDepthBufferTexture->fbo = _pBuffer->m_resolveFBO;
+	m_pResolveDepthBufferTexture->isFboValid = true;
+	m_pResolveDepthBufferTexture->isDepthAttachment = true;
 	gfxContext.addFrameBufferRenderTarget(targetParams);
 
 	Context::BlitFramebuffersParams blitParams;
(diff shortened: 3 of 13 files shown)


loganmc10:
I tested this (with Majora's Mask) in VMware (by enabling buggyFBO) since you mentioned this fixed https://github.com/gonetz/GLideN64/issues/1702 for Adreno. The issue is still there in VMware.

This seems like a lot of code to implement a really bizarre fix for a device specific problem in a couple games. Let's at least sit on this for a couple days, give me some time to see if I can use this information to try and maybe figure out a solution that would fix the issue on other affected GPUs as well.

fzurita:
I'm doing Adreno specific checks. You would have to remove those for VMware.

loganmc10:
Yeah, I have an apitrace file from the Majora's Mask scene, it plays back normally on my Windows PC, and it plays back garbled in VMware. The answer to that bug (and it seems like the BAR bug might be related), must be in that apitrace file. I've been trying to get someone at Mesa to take a look at it, but haven't had any luck yet.

loganmc10:
Hey @fzurita can you try putting a ```glFlush()``` at the beginning of ```setTextureParameters()```? That fixes it for the VMware driver, it's possible that the 0 size glBlitFramebuffer caused a similar flush that fixed the issue.

loganmc10:
I believe at it's core, this is a synchronization issue. Something like this: https://www.opengl.org/discussion_boards/showthread.php/174844-Texture-output-input-sync-problem

Basically we are writing to the framebuffer texture, and then trying to read from it too soon. The empty glBlitFramebuffer call tells GL to "finish up" with whatever rending it needs to do on that framebuffer, because it's going to be read (even though nothing is actually read).

There are functions to deal with this like glTextureBarrier:

> The values of rendered fragments are undefined when a shader stage fetches texels and the same texels are written via fragment shader outputs, even if the reads and writes are not in the same drawing command. To safely read the result of a written texel via a texel fetch in a subsequent drawing command, call glTextureBarrier between the two drawing commands to guarantee that writes have completed and caches have been invalidated before subsequent drawing commands are executed.

But it's not available on GLES or even many desktop GPUs. But it does seem like some kind of overlapping texel read/write situation.

fzurita:
You are right, glFlush does fix it, but obviously, performance drops to a crawl.

loganmc10:
Yes, I think using glFenceSync would fix this issue. Like right after you finishing writing to the texture, do you glFenceSync, then right before you want to read from that same texture, you do glWaitSync (I don't think you need glClientWaitSync). That will make sure that you aren't overlapping. I'm not really familiar enough with the rendering pipeline to know where these commands should go though.

loganmc10:
The other solution would probably be: instead of reading and writing to the same FBO, you copy it using glBlitFramebuffer, then read from that. That copy operation should ensure that you aren't overlapping anything

fzurita:
Well, glFenceSync and glWaitSync do work... Unfortunately, they also kill performance.

loganmc10:
I don't think that is the right spot, that is called so many times. We need to just call it in this particular circumstance, I'm hoping @gonetz will have some insight on when it might be needed

loganmc10:
Also I think it should be glWaitSync, not ClientWaitSync

fzurita:
Yes, see my edit.

fzurita:
Also, I'm only calling it when we are setting parameters on framebuffer textures.

loganmc10:
SwapBuffers causes a glFlush, so the situation we must be running into is:

Rendering into an FBO texture and then using a function like texelFetch to read from that same texture, all in the same frame. This is my assumption. 

We need to put some kind of barrier between the FBO render and the texture read. The other option would be to render the FBO, then blit to a temp FBO, and do the texelFetch from that FBO, it shouldn't have anything to do with setting the texture parameters.

loganmc10:
I don't think this is something you would solve in the graphics backend, you would solve it in the actual emulation code for this particular situation, either by adding an extra blit or introducing some gfxContext.synx type functions

fzurita:
Yeah, I'm aware it has nothing to do with setting texture parameters. That was just the easiest GL call I could find that happens when we "activate" a texture.

fzurita:
> Rendering into an FBO texture and then using a function like texelFetch to read from that same texture, all in the same frame.

This is making the assumption that the rendering to the FBO and texelFetch happen in parallel rather than serially. Is this really the case? GL commands are supposed to happen serially.

loganmc10:
The API commands happen serially, but in order to make use of the large parallelism of the GPU, it'll batch together a bunch of draw calls and execute them all at once. So you might issue 3 draw calls one after another, but it could execute the shaders for those draw calls all at the same time.

Obviously it will make decisions based on what can be batched together, but like the thing I quoted earlier says:

> The values of rendered fragments are undefined when a shader stage fetches texels and the same texels are written via fragment shader outputs, even if the reads and writes are not in the same drawing command

loganmc10:
@fzurita can you try something for me?

Right at the beginning of this function:

https://github.com/gonetz/GLideN64/blob/master/src/Graphics/OpenGLContext/GLSL/glsl_SpecialShadersFactory.cpp#L436

Can you add:
```
glMemoryBarrier(GL_TEXTURE_FETCH_BARRIER_BIT);
```

That won't fix Zelda, but I'm hoping it would fix the BAR menu on your phone. If that works, it's probably the most performance effective way to do it, and it could be implemented for the Zelda monochrome shader as well.

fzurita:
I just tried it, it doesn't fix the BAR menu.

loganmc10:
Sorry can you try the same thing with GL_FRAMEBUFFER_BARRIER_BIT? It should be in the very beginning of the activate() function. I'm surprised that doesn't work

fzurita:
Still no luck.

fzurita:
This is what it's looking like, just to make sure:

```
		void activate() override {
			glMemoryBarrier(GL_FRAMEBUFFER_BARRIER_BIT);
			ShadowMapShaderBase::activate();
			glUniform4fv(m_loc, 1, &gDP.fogColor.r);
			glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
			g_paletteTexture.update();
		}
```

loganmc10:
Yeah that looks right, it probably needs to be called between draw calls I guess (like is done for N64 Depth Compare) and not just when the shader is first activated

gonetz:
> Basically we are writing to the framebuffer texture, and then trying to read from it too soon.

Correction: BAR menu does not actually read from framebuffer texture. I commented content of FrameBuffer_ActivateBufferTexture and menu still works without issues. I explained how it works:
https://github.com/gonetz/GLideN64/issues/1777#issuecomment-377957160

gonetz:
Only depth texture is actually used. I have an idea. @fzurita , could you also comment body of FrameBuffer_ActivateBufferTexture(u32 t, u32 _frameBufferAddress)  and check BAR menu  again?

fzurita:
@gonetz commenting it the body of ```FrameBuffer_ActivateBufferTexture(u32 t, u32 _frameBufferAddress)``` doesn't fix the issue.

gonetz:
It seems that Logan is right and this is synchronization issue.

loganmc10:
@gonetz is there a way in the ShadowMapFragmentShader to use gl_FragCoord.z instead of texelFetching the data from the depth texture?

I notice in the RE2 shader, it uses: 
```
highp float FragDepth = clamp((gl_FragCoord.z * 2.0 - 1.0) * uDepthScale.s + uDepthScale.t, 0.0, 1.0);	
highp int iZ = FragDepth > 0.999 ? 262143 : int(floor(FragDepth * 262143.0));
```

And the BAR shader does:
```
highp float bufZ = texelFetch(uDepthImage,coord, 0).r;	\n"
highp int iZ = bufZ > 0.999 ? 262143 : int(floor(bufZ * 262143.0));
```

I tried to use gl_FragCoord.z in the BAR shader, even using the uDepthScale stuff as well, but it didn't work, if we could use gl_FragCoord.z instead of texelFetch it would probably solve the problem

loganmc10:
I think I see the difference, the texelFetch in the ShadowMap shader and Zelda Monochrome shader are meant to grab the depth of the previous fragment, not current, for blurring I guess.

In GLES, extensions exist to support this, ```GL_ARM_shader_framebuffer_fetch_depth_stencil```, this provides a ```gl_LastFragDepthARM``` variable that is guaranteed to not be corrupt, there is a similar extension for color data. We should probably use these on GLES (even though its an ARM extension, it's supported on Adreno 5XX devices).

There are still a few desktop drivers that would be affected though, there is a discussion here on some possible alternatives on the desktop https://www.opengl.org/discussion_boards/showthread.php/183629-Equivalent-for-gl_LastFragData-0-on-OpenGL

loganmc10:
I think my hunch was correct, @fzurita when you get a chance can you test: https://github.com/loganmc10/GLideN64/commit/da61d0c320c36f22b47e43bc0ca290845570d854

That fixes the BAR menu and fog on my phone. If that works a similar same thing can be done to fix Majora's Mask

fzurita:
Yep that worked.

fzurita:
Closing due to this https://github.com/gonetz/GLideN64/pull/1787

gonetz:
> I think I see the difference, the texelFetch in the ShadowMap shader and Zelda Monochrome shader are meant to grab the depth/color of the previous fragment, not current, for blurring I guess.

What do you mean by previous fragment?
How depth fog works on N64:
* Render whole scene. Fog blender mode is not used - no fogged objects yet.
* Take current depth buffer and render it as indexed IA texture over our scene using texrect commands. Upper byte of depth value is used as palette index. Since depth buffer rendered with the current color buffer, there is 1-1 correspondence between depth texture texels and output pixels. That is, for example, fog for pixel [10, 20] calculated with texel [10, 20] of depth image.

Plugin emulates it exactly this way. Whole scene is rendered. Then, when depth fog started, texrect code detects depth fog rendering and substitutes current color shader and color texture with ShadowMapShader and depth buffer texture. The shader ignores texture coordinates of fragments and uses fragment screen coordinates to fetch from depth texture.

Monochrome shader works similarly. Its task is to convert color pixel to monochrome one. It also ignores texture coordinates and uses fragment screen coordinates to fetch from input texture.

What is "previous fragment" in these cases?

loganmc10:
Sorry when I say previous fragment I mean the current depth or color buffer

PR #1787, opened 2018-04-08
[WIP] use gl_LastFragData/DepthARM when possible

loganmc10:
This is a fix for the issues with the BAR fog and Majora's Mask monochrome scene in GLES.

I haven't tested the Majora's Mask scene yet, so this shouldn't be merged yet, just opening for review.

If/when this is tested and merged, a similar fix should be applied for desktop GPUs. We haven't seen anyone on the desktop have issues with the BAR fog, but we have had a report of issues with Majora's Mask scene.

On the desktop, this should be possible to fix via ARB_texture_barrier/NV_texture_barrier, but I'll probably deal with that in a separate pull request.

These extensions are available on all Mali GPUs, and Adreno 5XX. I don't know of any way to fix this on pre 5XX Adreno devices. We would probably have to resort to ugly hacks.

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
@@ -455,6 +455,16 @@ public:
 		} else if (_glinfo.isGLESX) {
 			std::stringstream ss;
 			ss << "#version " << Utils::to_string(_glinfo.majorVersion) << Utils::to_string(_glinfo.minorVersion) << "0 es " << std::endl;
+
+			if (_glinfo.ext_fetch) {
+				ss << "#extension GL_EXT_shader_framebuffer_fetch : enable" << std::endl;
+			} else if (_glinfo.arm_fetch) {
+				ss << "#extension GL_ARM_shader_framebuffer_fetch : enable" << std::endl;
+			}
+			if (_glinfo.fetch_depth) {
+				ss << "#extension GL_ARM_shader_framebuffer_fetch_depth_stencil : enable" << std::endl;
+			}
+
 			ss << "# define IN in" << std::endl
 				<< "# define OUT out" << std::endl
 				<< "# define texture2D texture" << std::endl;
@@ -462,6 +472,9 @@ public:
 		} else {
 			std::stringstream ss;
 			ss << "#version " << Utils::to_string(_glinfo.majorVersion) << Utils::to_string(_glinfo.minorVersion) << "0 core " << std::endl;
+			if (_glinfo.ext_fetch) {
+				ss << "#extension GL_EXT_shader_framebuffer_fetch : enable" << std::endl;
+			}
 			if (_glinfo.imageTextures && _glinfo.majorVersion * 10 + _glinfo.minorVersion < 42) {
 				ss << "#extension GL_ARB_shader_image_load_store : enable" << std::endl
 					<< "#extension GL_ARB_shading_language_420pack : enable" << std::endl;
@@ -771,8 +784,11 @@ public:
 			"IN highp vec2 vTexCoord1;\n"
 			"IN mediump vec2 vLodTexCoord;\n"
 			"IN lowp float vNumLights;	\n"
-			"OUT lowp vec4 fragColor;	\n"
-		;
+			;
+		if (_glinfo.ext_fetch && !_glinfo.isGLES2)
+			m_part += "inout lowp vec4 fragColor;	\n";
+		else
+			m_part += "OUT lowp vec4 fragColor;	\n";
 	}
 };
 
@@ -836,7 +852,11 @@ public:
 		m_part +=
 			"IN lowp vec4 vShadeColor;	\n"
 			"IN lowp float vNumLights;	\n"
-			"OUT lowp vec4 fragColor;	\n";
+			;
+		if (_glinfo.ext_fetch && !_glinfo.isGLES2)
+			m_part += "inout lowp vec4 fragColor;	\n";
+		else
+			m_part += "OUT lowp vec4 fragColor;	\n";
 	}
 };
 
@@ -1006,8 +1026,22 @@ public:
 					"#define READ_TEX(name, tex, texCoord, fbMonochrome, fbFixedAlpha)	\\\n"
 					"  {																\\\n"
 					"  if (fbMonochrome == 3) {											\\\n"
-					"    mediump ivec2 coord = ivec2(gl_FragCoord.xy);					\\\n"
-					"    name = texelFetch(tex, coord, 0);								\\\n"
+					;
+				if (m_glinfo.ext_fetch) {
+					shaderPart +=
+						"    name = fragColor;	\\\n"
+						;
+				} else if (m_glinfo.arm_fetch) {
+					shaderPart +=
+						"    name = gl_LastFragColorARM;	\\\n"
+						;
+				} else {
+					shaderPart +=
+						"    mediump ivec2 coord = ivec2(gl_FragCoord.xy);	\\\n"
+						"    name = texelFetch(tex, coord, 0);								\\\n"
+						;
+				}
+				shaderPart +=
 					"  } else {															\\\n"
 					"    if (uTextureFilterMode == 0) name = texture(tex, texCoord);	\\\n"
 					"    else TEX_FILTER(name, tex, texCoord);			 				\\\n"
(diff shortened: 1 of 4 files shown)


fzurita:
Hete is the core dump:
```
********** Crash dump: **********
Build fingerprint: 'razer/cheryl/cheryl:7.1.1/NMF26X-RZR-180118/3005:user/release-keys'
pid: 19322, tid: 19431, name: ServiceStartArg  >>> paulscode.android.mupen64plusae.GameActivity <<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x1c
Stack frame #00 pc 000bc05a  /vendor/lib/egl/libGLESv2_adreno.so (_ZN10EsxContext19DrawArraysInstancedE11EsxPrimTypeijj+37)
Stack frame #01 pc 000eb7d7  /vendor/lib/egl/libGLESv2_adreno.so (_ZN21EsxGlApiParamValidate12GlDrawArraysEP11EsxDispatchjii+102)
Stack frame #02 pc 0001799c  /data/app/org.mupen64plusae.v3.alpha-2/lib/arm/libmupen64plus-video-gliden64.so: Routine GraphicsDrawer::drawTexturedRect(GraphicsDrawer::TexturedRectParams const&) at /home/fzurita/git_projects/mupen64plus/mupen64plus-ae/mupen64plus-video-gliden64/./upstream/src/GraphicsDrawer.cpp:1351
Stack frame #03 pc 0003923c  /data/app/org.mupen64plusae.v3.alpha-2/lib/arm/libmupen64plus-video-gliden64.so: Routine RDRAMtoColorBuffer::copyFromRDRAM(unsigned int, bool) at /home/fzurita/git_projects/mupen64plus/mupen64plus-ae/mupen64plus-video-gliden64/./upstream/src/BufferCopy/RDRAMtoColorBuffer.cpp:302
Stack frame #04 pc 00031ba4  /data/app/org.mupen64plusae.v3.alpha-2/lib/arm/libmupen64plus-video-gliden64.so: Routine VI_UpdateScreen() at /home/fzurita/git_projects/mupen64plus/mupen64plus-ae/mupen64plus-video-gliden64/./upstream/src/VI.cpp:176
Stack frame #05 pc 000b5e4c  /data/app/org.mupen64plusae.v3.alpha-2/lib/arm/libmupen64plus-core.so
Stack frame #06 pc 00041160  /data/app/org.mupen64plusae.v3.alpha-2/lib/arm/libmupen64plus-core.so
Stack frame #07 pc 000b72e8  /data/app/org.mupen64plusae.v3.alpha-2/lib/arm/libmupen64plus-core.so

```

loganmc10:
The shader is failing to compile, it says "gl_LastFragData is an unknown identifier". It's quite strange actually. I think it might be because it's inside the macro. Tomorrow I'll try putting it in its own function and call that function from the macro. Gotta love glsl compilers...

fzurita:
Yep...

gonetz:
Is "gl_LastFrag" a last value written to color/depth buffer at current fragment position?

loganmc10:
Yes correct, it is the current color or depth value of the fragment

gonetz:
Cool! I hope you will resolve that issue with shader compilation.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -1006,8 +1032,21 @@ class ShaderFragmentHeaderReadTex : public ShaderPart
...
+					shaderPart +=
+						"    name = gl_LastFragColorARM;	\\\n"
+						;
+				} else {
+					shaderPart +=
+						"    mediump ivec2 coord = ivec2(gl_FragCoord.xy);	\\\n"
+						"    name = texelFetch(tex, coord, 0);								\\\n"
+						;
+				}
 					"  } else {															\\\n"
```
fzurita: You are missing ```shaderPart +=``` here.

fzurita:
So according to this, the same extension can be supported by OpenGL ES and full OpenGL.

https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_shader_framebuffer_fetch.txt

loganmc10:
Yes it is supported by Intel on Windows (on Linux it's supported by Intel but only in GLES mode), I don't think Nvidia or AMD support it. I'm not sure about the Shield in GLES mode, I didn't see it on the list though.

fzurita:
From the spec:
```
    "By default, gl_LastFragData is declared with the mediump precision
    qualifier. This can be changed by redeclaring the corresponding variables
    with the desired precision qualifier.
    
    Redeclarations are done as follows
    
    // Redeclaration that changes nothing is allowed
    mediump vec4 gl_LastFragData[gl_MaxDrawBuffers]; 
```

So maybe just adding ```    mediump vec4 gl_LastFragData[gl_MaxDrawBuffers]; ``` to the header of the shader will fix it.

loganmc10:
Yeah I'll have to try later today, did you try after adding the ```ShaderPart +=``` that I missed? I'm surprised that didn't throw a compiler error

fzurita:
It threw a warning. It didn't fix the issue though with gl_LastFragData bring undeclared.

fzurita:
Lol, but when I try to redefine it, I get this:
ERROR: 0:40: 'gl_LastFragData' : can not redefine gl_LastFragData in es3.0 or above 
    ERROR: 1 compilation errors.  No code generated.

fzurita:
https://developer.apple.com/library/content/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/AdoptingOpenGLES3/AdoptingOpenGLES3.html

> The EXT_shader_framebuffer_fetch extension works differently. GLSL ES 3.0 removes the gl_FragData and gl_FragColor builtin fragment output variables in favor of requiring fragment outputs to be declared in the shader. Correspondingly, the gl_LastFragData builtin variable is not present in GLSL ES 3.0 fragment shaders. Instead, any fragment output variables you declare with the inout qualifier contain previous fragment data when the shader runs. For more details, see Fetch Framebuffer Data for Programmable Blending.

loganmc10:
Oh well look at me blaming Adreno when it was my fault lol, so if I'm reading that correctly we should replace glLastFragData[0] with ```fragColor```

loganmc10:
And fragColor might have to be changed from an OUT to an INOUT

fzurita:
I tried it and it seems to work. Oddly though, there is a weird line in the middle of the monochrome effect scene in Zelda.

loganmc10:
I think the code we are dealing with is !GLES2, and I assume the desktop will work the same as GLES3 since there is no gl_FragColor there either

fzurita:
BAR menu is not working though.

fzurita:
It's because this is missing fromthe ShadowMap shader:
 ``` "#extension GL_ARM_shader_framebuffer_fetch_depth_stencil : enable	\n" ```

fzurita:
One advantage of all this is that BAR performance improves by around 30%.

loganmc10:
Alright I fixed up the shader and tested this on my phone, everything looks normal now! I tested the Majora's Mask scene and BAR. I didn't see any line in the Majora's Mask scene, I wonder if it is the combination of some other setting you've got set that causes it? If so, it probably does that on the PC as well, but we can test.

I think it's ready to be merged assuming it looks good to @fzurita

loganmc10:
It might be worth testing this on a Mali GPU as well, I don't have one to test, and so the ARM_framebuffer_fetch codepath hasn't been tested at all.

fzurita:
I can test both.

fzurita:
Here is a screenshot of the line
zelda_majora s_mask-002

Also, the blur effect seems to be missing.

BAR seems ok though.

fzurita:
The line only shows up at 3x native resolution.

fzurita:
I tested a Galaxy S6 (Mali GPU) and it worked exactly the same as Adreno.

loganmc10:
So I'm not too worried about the line, that is probably unrelated (I'm assuming). But yes the blur effect does seem to be missing.

Do you have a device where the Majora's Mask cutscene worked before this change? Like a Mali device or something? I'm wondering if the blur is broken because of this, or its just always been broken on GLES because of the "WeakBlitFramebuffer" or something.

fzurita:
The blur effect works correctly in this branch. Also, no line in this branch either:

https://github.com/gonetz/GLideN64/pull/1786

loganmc10:
Huh ok, I'll need to look into it more closely tomorrow. I'm not sure what is going on

loganmc10:
Closed in favor of https://github.com/gonetz/GLideN64/pull/1790

PR #1790, opened 2018-04-09
Fix BAR fog

loganmc10:
After playing around with all sorts of exotic extensions, I realized there is a very simple fix for this problem, and the function to fix it already exists.

Basically, you should never ever ever sample from a texture that is attached to a FBO. You need to copy it first. This is explained in the page for glFramebufferTexture2D:

> Special precautions need to be taken to avoid attaching a texture image to the currently bound framebuffer while the texture object is currently bound and potentially sampled by the current vertex or fragment shader. Doing so could lead to the creation of a "feedback loop" between the writing of pixels by rendering operations and the simultaneous reading of those same pixels when used as texels in the currently bound texture. In this scenario, the framebuffer will be considered framebuffer complete, but the values of fragments rendered while in this state will be undefined. The values of texture samples may be undefined as well.

So what this PR does, is copy the depth texture and then attaches that copy instead.

A similar fix is needed for FrameBuffer_ActivateBufferTexture/FrameBuffer_ActivateBufferTextureBG. That will fix the Majora's Mask cutscene. I will submit a separate PR for that.

### DIFF
--- a/src/DepthBuffer.cpp
+++ b/src/DepthBuffer.cpp
@@ -313,7 +313,7 @@ CachedTexture * DepthBuffer::copyDepthBufferTexture(FrameBuffer * _pBuffer)
 
 void DepthBuffer::activateDepthBufferTexture(FrameBuffer * _pBuffer)
 {
-	textureCache().activateTexture(0, resolveDepthBufferTexture(_pBuffer));
+	textureCache().activateTexture(0, copyDepthBufferTexture(_pBuffer));
 }
 
 void DepthBuffer::bindDepthImageTexture()


fzurita:
I liked your older solution that worked just fine with the BAR fog. It was a 50% performance improvement in BAR.

loganmc10:
Yes that worked well, what I'm mostly concerned about is being complaint with the OpenGL spec, what we are doing currently is undefined behavior (sampling from an FBO texture).

We can add a SpecialFeature::DepthFetch for devices that support GL_ARM_shader_framebuffer_fetch_depth_stencil, in that case they can skip activateDepthBufferTexture altogether, but this PR is still how it should work for other devices.

gonetz:
If it is not safe to fetch from texture attached to FBO, may be it is simpler to detach the texture from FBO before attaching it to shader?

loganmc10:
It's possible, but nothing that I read indicates that detaching a framebuffer attachment forces it to flush any pending writes. I don't see anything in the spec saying that it must work that way

loganmc10:
@gonetz I believe you ran into this same issue when you were working on N64 Depth Compare (http://gliden64.blogspot.com/2014/07/depth-buffer-emulation.html) to quote you:

> Quick dig in docs revealed the sad truth - I lost my time for a wrong idea. It is prohibited to Sampling and Rendering to the Same Texture. Documentation says that result is undefined in that case. AMD and NVidia gave me different results, but both results were bad. So, I had to find another way.

We are running into the same issue here. Your only real options are to call glTextureBarrier() (not available on GLES) between when you render to the texture and when you sample from it, or copy the texture. I'm not really aware of any other solutions.

gonetz:
"Sampling and Rendering to the Same Texture" is definitely not related to BAR fog.
For shader samples from depth texture but renders to color buffer.
AFAIR, #1702 also not related to that issue: shader reads from aux color buffer and writes to the main one. Both effects works well on desktop GPUs. N64 depth compare did not work with OpenGL until I switched to ImageTexture with glMemoryBarrier.

I suggest to try "detach textures from FBO" approach. Any hackish way, just to see how it works.
I would try it myself, but I don't have device to test.

BTW, you mentioned that the problem reproduced with VMware Mesa.
Do you run Windows or Linux with VMware?

loganmc10:
> For shader samples from depth texture but renders to color buffer

Yes but how did the data get into the depth texture? It was put there by a shader, that is the problem, you can't sample from a texture that was written by a shader.

> shader reads from aux color buffer and writes to the main one.

Same situation here, you can't read from the aux color buffer because it was written by a shader.

I'll try detaching the attachment tomorrow, but I don't think we should rely on that, since behaviour could be different depending on the driver.

I run Linux inside VMware, VMware itself is running on Windows with an Intel GPU

gonetz:
>you can't sample from a texture that was written by a shader.

All frame buffer textures written by shaders. Most of fb effects works.

gonetz:
> But these are the only instances where you sample (texelFetch type calls) from the FB textures.

Do you mean that there are situations, when you can call sample with texture(), but not with texelFetch()? Sounds weird.
The cite in your first post, about "Special precautions", implies that it is not allowed to sample (with any function) from texture, for which there are unfinished writing processes. As soon as texture is detached, writing to it must be impossible, thus reading must be safe.

As I remember, there are situations, where N64 reads and writes to the same buffer. Plugin does the same. However, since read and write pixels are not the same, it works. IIRC, it happens in Mario Tennis.
So, even if texture detach will help to fix BAR fog and #1702, it should not became general solution.

loganmc10:
The actual proper way to fix this on a desktop is to use glTextureBarrier (https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTextureBarrier.xhtml). Even though that only became core in 4.5, the extension is extremely common on most GPUs.

That is probably what we should be doing in OpenGL, the problem is there isn't a similar function for GLES

loganmc10:
https://github.com/gonetz/GLideN64/pull/1794

PR #1791, opened 2018-04-09
Fix Zelda MM cutscene

loganmc10:
This is a companion PR to https://github.com/gonetz/GLideN64/pull/1790

I would say the BAR fog PR if very safe, I know that code is specific only to the BAR fog scene.

This PR could affect games other than Zelda MM, I'm not sure what other games use this feature.

It does the same thing though: when we used to sample from an FBO texture, we now copy that texture using glBlitFramebuffer, and sample from the copied texture. This fixes the monochrome scene for me in VMware and on my phone.

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -57,6 +57,8 @@ FrameBuffer::FrameBuffer()
 	m_loadTileOrigin.uls = m_loadTileOrigin.ult = 0;
 	m_pTexture = textureCache().addFrameBufferTexture(config.video.multisampling != 0);
 	m_FBO = gfxContext.createFramebuffer();
+	m_pFrameBufferCopyTexture = textureCache().addFrameBufferTexture(config.video.multisampling != 0);
+	m_copyFBO = gfxContext.createFramebuffer();
 }
 
 FrameBuffer::~FrameBuffer()
@@ -64,10 +66,12 @@ FrameBuffer::~FrameBuffer()
 	gfxContext.deleteFramebuffer(m_FBO);
 	gfxContext.deleteFramebuffer(m_resolveFBO);
 	gfxContext.deleteFramebuffer(m_SubFBO);
+	gfxContext.deleteFramebuffer(m_copyFBO);
 
 	textureCache().removeFrameBufferTexture(m_pTexture);
 	textureCache().removeFrameBufferTexture(m_pResolveTexture);
 	textureCache().removeFrameBufferTexture(m_pSubTexture);
+	textureCache().removeFrameBufferTexture(m_pFrameBufferCopyTexture);
 }
 
 void FrameBuffer::_initTexture(u16 _width, u16 _height, u16 _format, u16 _size, CachedTexture *_pTexture)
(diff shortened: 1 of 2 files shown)


fzurita:
Well... this ends up looking a lot like my original solution.... My original solution did something similar, so it explains why it fixed the issues as well.

loganmc10:
The blit actually doesn't happen every time, it only happens when FrameBuffer_ActivateBufferTexture or FrameBuffer_ActivateBufferTextureBG is called. So it'll happen every time during these types of scenes, but not generally.

The 0 size blit didn't seem to trick the VMware driver, it still had the issue, so I don't think we can count on that kind of trick. Also, the OpenGL spec says that sampling from an FBO texture is undefined behavior, so I think the application should work like this (copying the whole FBO to a new texture) in order to be compliant with the spec, even though we seem to have gotten lucky on most desktop drivers.

loganmc10:
It is never written to via a shader. Sorry I should have specified. If you write to an FBO via a shader, you are writing to the FBO texture via the shader. There is no way in OpenGL to know when that shader is done executing, or in what order, etc... Basically you can't trust the contents of the texture.

If you do a glBlitFramebuffer and read from that FBO, it forces OpenGL to finish up with whatever shader might be writing to it, because it needs that data. I believe this is why the glBlitFramebuffer size 0 worked, since it caused the driver to flush any pending writes to that texture, but the Mesa driver seems to be smart enough to know that it was a dummy blit

fzurita:
There may be an even simpler solution to this:

https://medium.com/@igorpener/overcome-incoherent-memory-access-in-opengl-es-3-0-3b7cc32c4b51

Just calling ```glMemoryBarrier(GL_FRAMEBUFFER_BARRIER_BIT), ``` before activating a different texture... Wait, did we try that already?

loganmc10:
Yes I believe I had you try that. I think the glMemoryBarrier stuff is really only supposed to be used with Image Textures, it's kind of unclear but I don't know if its really "general purpose". Even if it was, it would require Image Textures support to work.

fzurita:
Another thing here... Should we invalidate any FBOs after the copy is complete. Technically, we shouldn't need m_FBO any more, so it should be invalidated.

loganmc10:
Potentially yes, unless it is used for other things (like another blit later on or something). Everything I've read about mobile devices says using glClear() and glInvalidateFramebuffer() will really help performance, but you really have to be certain about when you no longer need the FBO contents. Probably an item for a separate PR, but something to look into.

fzurita:
Did you look at all at ```GL_INTEL_fragment_shader_ordering``` It seems to also attempt to fix the same thing.

loganmc10:
Yes, it's very similar to ```GL_ARB_fragment_shader_interlock```, it probably would fix the problem, but they are still GL4+ extensions, so not available everywhere. The hardest problem to solve is the Majora's Mask cutscene.

I thought ext_framebuffer_fetch would work, but the Majora's Mask scene actually samples from the FBO texture of a different FBO. ext_framebuffer_fetch only works to get the current value of the current FBO, not a different FBO. I'm not really sure of an extension or any other way to deal with that in GLES besides copying the texture or something even worse like using glFlush

loganmc10:
There may be a way to reduce the number of times _copyFrameBufferTexture() needs to be called, for instance in the already existing copyDepthBufferTexture(), there is an m_copied variable, that (I believe) makes sure the copy only happens once per frame. A similar technique could be employed for _copyFrameBufferTexture(), but I would rely on @gonetz to know whether that is the right optimization to make.

Just to be clear: _copyFrameBufferTexture() only happens in very specific circumstances. I tested a good number of games, and the only times I ever saw it executed was during the spinning mask in Zelda MM, during the monochrome cutscene, and when there was fog in BAR. I never saw it executed any other time, although I'm sure there are other games that use this feature, but it's not common.

loganmc10:
https://github.com/gonetz/GLideN64/pull/1794

Issue #692, opened 2015-08-30
Mario Party 3 (U): minigame Curtain Call issues


Mastergatto:
The character you select appears in the cloud, but the other choices are still visible outside of it, in line. 

![marioparty3u](https://cloud.githubusercontent.com/assets/1028191/9567671/40e8fca6-4f34-11e5-8da9-5c99ed060614.png)

Since I can't use the software rendering on Linux, so here is this preview of the minigame:
![curtaincall](https://cloud.githubusercontent.com/assets/1028191/9567683/870e8bce-4f34-11e5-9062-c5d8336227d5.png)

Here is the save state:
https://mega.nz/#!u4xFEDzR!7Z9xCdev1YGukH9t-3hsQpEM9KRpm5jwDhp3h7eexDA

purplemarshmallow:
correct with angrylion's plugin
No other plugin seems to be correct

olivieryuyu:
It works with N64 depth compares option

Fixed :)

purplemarshmallow:
Yes it really works with N64 depth compare that's good
>Fixed

But can it be enabled in the ini? This option is very demanding. Performance is also important. Does the rest of the game work correctly with N64 depth compare?

olivieryuyu:
it can be implemented in ini. However it will slow down for sure. The implementation of this depth compare is working per triangle as far as I understood. May be there is way to speed up the matter but Sergey got not luck here yet.

purplemarshmallow:
LLE is also slow (compared with z64gl) because triangles are not optimized. If one optimization could solve both that would make GLideN64 clearly the best plugin available superior to everything else.

purplemarshmallow:
I noticed in HLE N64 depth compare completely kills performance but in LLE it makes hardly any difference. I think with the right optimizations both LLE and N64 depth compare will be fast

gonetz:
> If one optimization could solve both
It can't. LLE optimization is to draw as many triangles with one GL command as possible. That kills N64 depth compare shader.

olivieryuyu:
please close as it is fixed

purplemarshmallow:
it's not fixed with default settings
And I don't recommed setting it in the ini. I can't run it fullspeed with N64 depth compare enabled.

purplemarshmallow:
>It can't. LLE optimization is to draw as many triangles with one GL command as possible. That kills N64 depth compare shader.

[...]

PR #1792, opened 2018-04-10
Rewrite N64 Depth Compare for better performance

loganmc10:
OpenGL extensions exist to ensure fragment shader read/writes don't overlap with each other.

See the performance difference below (tested on Intel+Windows):

Before:

After:

Right now this requires one of 3 extensions: GL_ARB_fragment_shader_interlock , GL_NV_fragment_shader_interlock , GL_INTEL_fragment_shader_ordering. Therefore it leaves out GLES devices, and Mesa, probably some older Nvidia and AMD cards as well. I'm still researching other ways to do this, the most promising is an "atomic spinlock" (https://stackoverflow.com/questions/21538555/broken-glsl-spinlock-glsl-locks-compendium), but I haven't gotten this working yet. Therefore this probably isn't ready to be merged, but I thought I'd leave this here for review.


fzurita:
Pretty impressive, the performance penalty of n64 depth compare with this is zero from testing so far.

AmbientMalice:
It would be ideal if N64 depth compare could be sped up to the point that enabling it on all necessary games by default would be non-controversial.

fzurita:
One strange issue though, it locks up when multisampling is disabled with the NVidia Shield TV.

loganmc10:
> It would be ideal if N64 depth compare could be sped up to the point that enabling it on all necessary games by default would be non-controversial.

I'm hoping that with improved performance, more people will be inclined to use it and test it, but I don't think it's ready to be the default yet. While I was testing the PR I noticed at least 2 issues that exist in the current master: the background in FIFA 99 is completely missing, and most the Bomb-omb's in Mario Tennis (they stand behind the court in a match) are missing. So N64 Depth Compare still needs some more work in terms of accuracy.

gonetz:
Impressive speed boost!

"N64 Depth Compare still needs some more work in terms of accuracy."

Most likely these are not accuracy issues. There are tricks with frame buffer, which do not take Image depth texture into account.

Also, this mode is not compatible with MSAA. Which lead as to necessity to support post-processing AA. Which leads to general support of post-processing shaders. Which is our bright future, for which I currently has no resources.

mudlord:
@loganmc10

a little nugget of advice, I doubt you will even fucking take it:

https://pastebin.com/tbu5ygym

You are welcome. >_>

oddMLan:
I wish Reshade could get fixed to play nicely with GLideN64

loganmc10:
I added a commit that adds another rendering method, using GL_EXT_shader_framebuffer_fetch. Using this extension, Image Textures aren't used at all, instead the depth data is read/written to framebuffer color attachments.

This extension is supported by Intel on Windows and Linux (starting in Mesa 18.1). It is also supported by PowerVR and Adreno on newer chipsets. I've been testing this with @fzurita , it works properly on Intel and PowerVR, but it doesn't work properly on Adreno...

loganmc10:
So I'm pretty happy with how this PR looks now. This would actually reduce the number of people that could use N64 Depth Compare, since the GPU would need to support Image Textures + one of these extensions, instead of just Image Textures. I think this is the right thing to do though, here is why:

I think it's important that N64 Depth Compare always performs well, right now, we have N64DepthCompare=1 in the INI file for TWINE, just to fix the shadows. So if we allow a fallback to the current (slow) method, a lot of people can't run the game at full speed. It's better to have incorrect shadows and a playable game.

So this PR will try to provide a fast implementation of N64 Depth Compare, otherwise N64 Depth Compare will be disabled. This will allow us to set N64DepthCompare=1 in the INI for games that need it, while still allowing people to play games at full speed if their card doesn't support these extensions.

AmbientMalice:
Would it be worth implementing the z mode compare less hack Glide64 used as a fallback for GPUs that cant handle N64 depth compare? That way games like Toy Story 2 would have correct HUD rendering.

loganmc10:
@fzurita can you try using "NV_pixel_interlock_ordered" instead of "pixel_interlock_ordered". I just realized the NV version exists

fzurita:
Yeah, sure, once I get home tonight.

fzurita:
I get this: 
```
E/GLideN64: shader_compile error: 0(5) : error C3008: unknown layout specifier 'nv_pixel_interlock_ordered'
```

And before you ask, I'm spelling it as ```NV_pixel_interlock_ordered```

loganmc10:
Huh ok weird, I was just reading https://www.khronos.org/registry/OpenGL/extensions/NV/NV_fragment_shader_interlock.txt and I noticed they mentioned both versions in that document, but that's obviously not correct

fzurita:
So apparently the shield TV doesn't support the fragment_interlockNV extension. It's reporting false in "Utils::isExtensionSupported```

loganmc10:
So how is N64 depth compare even working? It should disable it if none of those extensions are enabled

fzurita:
```GL_ARB_fragment_shader_interlock``` does report as being supported.

**Comment by fzurita** [comment]
I did a quick experiment. Commenting out
```m_part += "  if (should_discard) discard;	\n";```
avoids the lockup. Obviously, it creates bad graphic bugs.

**Comment by loganmc10** [comment]
To give some background:GL_ARB_fragment_shader_interlock did not allow me to put the discard statement inside the "block" which is why I created the "should_discard" variable and I do the discard outside the block, perhaps there is a different method of doing it to avoid the crash.

**Comment by fzurita** [comment]
It definitely has to do with ```beginInvocationInterlockARB``` and ```endInvocationInterlockARB```. Commenting those out prevents the lockup.

**Comment by loganmc10** [comment]
Ok yeah so they're both supported, they are essentially the same thing, it's just that only one should be enabled at a time, which is why I added that check

**Comment by fzurita** [comment]
Forcing ```fragment_interlockNV``` to true and ```fragment_interlock``` didn't fix the lockup.

**Comment by olivieryuyu** [comment]
interesting fast depth compare would be great :)

**Comment by gonetz** [comment]
I see no problems with code. I see no regressions - the feature works properly. Performance is really good. Great work!!!

However, locks up issue on NVidia Shield TV must be fixed somehow.
@fzurita "it locks up when multisampling is disabled with the NVidia Shield TV"
N64 depth compare must not be activated at all when MSAA enabled.

**Comment by fzurita** [comment]
Yep, that is true, but that is the only way I could get it to not lock up.

**Comment by fzurita** [comment]
I could force the Android version to always use GLES. There is really no advantage any more to the full OpenGL version with this change. I think this would work with gles because it uses ext_shader_framebuffer_fetch.

**Comment by loganmc10** [comment]
I don't think Nvidia supports ext_shader_framebuffer_fetch, and GL_NV_fragment_shader_interlock should be supported in GLES as well on the Shield. We might just have to exclude the X1 from supporting N64 Depth Compare

**Comment by fzurita** [comment]
That's a pain since it used to work before, at least with the slow method. Maybe we should keep the slow method around for devices that support image textures.

**Comment by loganmc10** [comment]
The problem with keeping the slow method around is you could never have N64 Depth Compare as a default option. It would ruin the performance on some devices. If you only support the fast method then you could turn it on by default, and devices that can't run it fast can just run without it, and still get a pretty good experience.

Is the current/slow method usable on the X1? Do games run at full speed?

**Comment by fzurita** [comment]
It depends on the game, but generally yes. It is around a 50% loss in performance.

**Comment by oddMLan** [comment]
What's the performance difference between no depth compare and the optimized N64 depth compare method?

**Comment by fzurita** [comment]
Depends on the device. In the Nexus player, it was about 10% performance drop. On the Shield tv, the was no performance difference. I'm assuming that on desktop devices, there is no performance penalty.

**Comment by oddMLan** [comment]
Holy moly... that's an astonishing difference! It would make N64 depth compare viable as the default and fix tons of problematic games and little innacuracies with the shadows in many others. However, no MSAA. I guess one could get around it by using a very high internal resolution, at least for now.

**Comment by fzurita** [comment]
The lockup may be fixed by some of the stuff I did in this branch:

https://github.com/fzurita/GLideN64/commits/further_reduce_shader_logic

File: src/Graphics/OpenGLContext/opengl_GLInfo.cpp [hunk 1]
```diff
@@ -79,12 +82,23 @@ void GLInfo::init() {
 		}
 	}
 
+	bool ext_draw_buffers_indexed = isGLESX && (Utils::isExtensionSupported(*this, "GL_EXT_draw_buffers_indexed") || numericVersion >= 32);
 #ifdef EGL
 	if (isGLESX && bufferStorage)
 		g_glBufferStorage = (PFNGLBUFFERSTORAGEPROC) eglGetProcAddress("glBufferStorageEXT");
+	if (isGLESX && numericVerison < 32) {
```
fzurita: There is a typo: "Verison"

loganmc10: Whoops, fixed

fzurita:
Including my one commit from the other branch didn't help anything. There is one more extension that can be used that NVidia may support better on the shield TV. It's this: https://www.khronos.org/registry/OpenGL/extensions/KHR/KHR_blend_equation_advanced.txt

loganmc10:
That extension seems widely supported but I'm not sure how it would work for us. It seems to guarantee the coherency of blending operations, but as far as I understand, blending happens after the fragment shader, so it doesn't necessarily guarantee the coherency of fragment shader operations. It's a little unclear to me how you enable coherent behaviour using that extension, it seems to say that it will be enabled by default if it is supported

fzurita:
Yeah, I agree, I realized that after I read more closely into it.

fzurita:
I think I give up at this point I think this can be merged, I'll make a new pull request to remove support from Tegra X1 since it has driver bugs.

fzurita:
Hmmm... I wonder if this can be closed with this: https://github.com/gonetz/GLideN64/issues/692

loganmc10:
Yes I think that would be an example of a game where we could finally just set N64DepthCompare=1 in the INI file for that game and not have to worry about ruining the performance for anyone

gonetz:
>> I think I give up at this point I think this can be merged, I'll make a new pull request to remove support from Tegra X1 since it has driver bugs.

Ok.

tony971:
I'd like to see it on by default for all games unless someone selects MSAA. With no performance impact and a wide array of graphical fixes, it makes sense to just leave it on.

loganmc10:
If anyone wants to try out this new N64 Depth Compare stuff, https://m64p.github.io/ has been updated

To enable N64 Depth Compare you go Settings->Core and Plugin Settings->Video Plugin and look for "EnableN64DepthCompare"

oddMLan:
> I think I give up at this point I think this can be merged, I'll make a new pull request to remove support from Tegra X1 since it has driver bugs.

so no N64 depth compare option for the Nintendo Switch? 😛

fzurita:
Lol, or shield tv, or pixel c

loganmc10:
My phone just got Android 9 and N64 Depth Compare seems to work properly now (Adreno 540)

@fzurita if you have an Android 9 Adreno device you may want to test, I seem to recall this was all messed up on Adreno with Android 8.1

fzurita:
Yep, my phone hasn't been updated. So it really sounds like a driver bug if it's working with Android 9. It is cool too know that once Android 9 comes out for my phone I will have working N64 depth compare though.

PR #1794, opened 2018-04-13
use gl_LastFragDepthARM and glTextureBarrier

loganmc10:
Sorry for opening yet another PR for this problem. This is to fix the BAR fog and Majora's Mask cutscene.

For GLES, the BAR fog can be properly emulated using gl_LastFragDepthARM, so that is the method that is used.

From the description of glTextureBarrier:

> The values of rendered fragments are undefined when a shader stage fetches texels and the same texels are written via fragment shader outputs, even if the reads and writes are not in the same drawing command. To safely read the result of a written texel via a texel fetch in a subsequent drawing command, call glTextureBarrier between the two drawing commands to guarantee that writes have completed and caches have been invalidated before subsequent drawing commands are executed.

So, if we are going to attach a FB texture, we must call glTextureBarrier to ensure that all drawing commands that might write to that texture have completed.

glTextureBarrier is only available on desktop, therefore I still don't have a good solution for the Majora's Mask cutscene on GLES, where it is taking the FB texture from one FBO, and trying to sample from that texture and render into another FBO.


gonetz:
The fix implemented very accurate, thanks!
@fzurita please confirm that BAR fog works on your GLES devices and does not cause performance issues. I'll test it on my desktop this evening.

@loganmc10 Regarding Majora's Mask cutscene: if detaching fb texture from FBO can solve this problem on GLES devices, we can do it. For example, we can add special code to UFrameBufferInfo::update , where monochrome mode is set.

fzurita:
I can confirm it works with my GLES device. Also, the game runs 50% faster, at least in the menus. Very nice performance increase.

fzurita:
Also, the shader version should be incremented.

loganmc10:
It's not just the Majora's Mask cutscene, anytime FrameBuffer_ActivateBufferTexture/FrameBuffer_ActivateBufferTextureBG is called we need to detach the framebuffer attachment (assuming that works), I just don't know where/when to re-attach it

loganmc10:
@fzurita you said the Majora's scene was messed up on the Shield right? This PR should fix that

gonetz:
Cool, thanks!

gonetz:
> FrameBuffer_ActivateBufferTexture/FrameBuffer_ActivateBufferTextureBG is called we need to detach the framebuffer attachment (assuming that works), I just don't know where/when to re-attach it

It is good question, which needs careful cogitation. First, I need to know that texture detaching really helps to fix the problem on GLES GPUs.

fzurita:
@loganmc10 sure, I'll retest that scene.

fzurita:
Yep, it's working correctly in the shield tv now

gonetz:
Works correctly on my Radeon.

PR #1798, opened 2018-04-16
Copy FBO texture if glTextureBarrier is unsupported

loganmc10:
This fixes the Majora's Mask scene for GLES where glTextureBarrier is unsupported. I added an "m_copied" variable, similar to what was done for depth texture copies, to improve performance. The "m_copied" greatly reduces the number of copies needed in the Majora's Mask scene and it still looks correct.

Tested on an Adreno 540.

To summarize this fix: If we are trying to read from the currently attached FBO texture, we execute glTextureBarrier if we can, if not, we copy the texture and read from that copy.

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -53,6 +53,9 @@ FrameBuffer::FrameBuffer()
 	, m_pResolveTexture(nullptr)
 	, m_resolved(false)
 	, m_pSubTexture(nullptr)
+	, m_copied(false)
+	, m_pFrameBufferCopyTexture(nullptr)
+	, m_copyFBO(ObjectHandle::null)
 {
 	m_loadTileOrigin.uls = m_loadTileOrigin.ult = 0;
 	m_pTexture = textureCache().addFrameBufferTexture(config.video.multisampling != 0);
@@ -64,10 +67,12 @@ FrameBuffer::~FrameBuffer()
 	gfxContext.deleteFramebuffer(m_FBO);
 	gfxContext.deleteFramebuffer(m_resolveFBO);
 	gfxContext.deleteFramebuffer(m_SubFBO);
+	gfxContext.deleteFramebuffer(m_copyFBO);
 
 	textureCache().removeFrameBufferTexture(m_pTexture);
 	textureCache().removeFrameBufferTexture(m_pResolveTexture);
 	textureCache().removeFrameBufferTexture(m_pSubTexture);
+	textureCache().removeFrameBufferTexture(m_pFrameBufferCopyTexture);
 }
 
 void FrameBuffer::_initTexture(u16 _width, u16 _height, u16 _format, u16 _size, CachedTexture *_pTexture)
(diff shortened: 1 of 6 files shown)


fzurita:
This doesn't fix Majora's mask monochrome scene with GLES devices. I tried getting rid of the optimization that only copies of it hasn't been cooped already and it didn't fix it either. 

What else is different with this fix compared to your last one?

fzurita:
So apparently, texture barrier is being reported as being supported by Adreno 530.

loganmc10:
Weird, so if you force ```texture_barrier``` to false does this PR work? ARB_texture_barrier is supposed to be an OpenGL only extension, I wonder if it's reporting support in error

fzurita:
Maybe the TextureBarrier function returned by EGL is not valid, I can't check now though.

loganmc10:
I think I'll just add a ```& !isGLESX``` to ```texture_barrier```, it should only be supported by full GL devices. texture_barrierNV says it can work on GLES though, like the Shield devices in GLES mode

loganmc10:
OK I made that change

fzurita:
It's working ok now.

src/FrameBuffer.cpp
```diff
@@ -176,20 +181,26 @@ void FrameBuffer::init(u32 _address, u16 _format, u16 _size, u16 _width, bool _c
 
 	const u16 maxHeight = VI_GetMaxBufferHeight(_width);
 	_initTexture(_width, maxHeight, _format, _size, m_pTexture);
+	_initTexture(_width, maxHeight, _format, _size, m_pFrameBufferCopyTexture);
```
gonetz: I don't like that copy texture created for every buffer, even if it will never use it.
Lets do lazy initialization of copy texture and copy FBO.  For example, FrameBuffer::_copyFrameBufferTexture() can check for copy texture and init if necessary.

loganmc10:
Ok I made that change, that is how the depth texture copy worked, so now they both look very similar.

src/FrameBuffer.cpp
```diff
@@ -418,6 +423,42 @@ CachedTexture * FrameBuffer::_getSubTexture(u32 _t)
 	return m_pSubTexture;
 }
 
+CachedTexture * FrameBuffer::_copyFrameBufferTexture()
+{
+	if (m_copied)
+		return m_pFrameBufferCopyTexture;
+
+	if (m_pFrameBufferCopyTexture == nullptr) {
```
gonetz: Texture initialization better move to separate method for readability.
_initCopyTexture().

src/FrameBuffer.cpp
```diff
@@ -53,21 +53,26 @@ FrameBuffer::FrameBuffer()
 	, m_pResolveTexture(nullptr)
 	, m_resolved(false)
 	, m_pSubTexture(nullptr)
+	, m_copied(false)
+	, m_pFrameBufferCopyTexture(nullptr)
 {
 	m_loadTileOrigin.uls = m_loadTileOrigin.ult = 0;
 	m_pTexture = textureCache().addFrameBufferTexture(config.video.multisampling != 0);
 	m_FBO = gfxContext.createFramebuffer();
+	m_copyFBO = gfxContext.createFramebuffer();
```
gonetz: Move m_copyFBO creation to _initCopyTexture(), see below.

src/FrameBuffer.cpp
```diff
@@ -418,6 +423,42 @@ CachedTexture * FrameBuffer::_getSubTexture(u32 _t)
...
+	blitParams.srcX1 = m_pTexture->realWidth;
+	blitParams.srcY1 = m_pTexture->realHeight;
+	blitParams.dstX0 = 0;
+	blitParams.dstY0 = 0;
+	blitParams.dstX1 = m_pTexture->realWidth;
+	blitParams.dstY1 = m_pTexture->realHeight;
+	blitParams.mask = blitMask::COLOR_BUFFER;
+	blitParams.filter = textureParameters::FILTER_NEAREST;
+
+	gfxContext.blitFramebuffers(blitParams);
```
gonetz: What if GPU does not support blitFramebuffers? GL ES 2.0 for example.
May be use GraphicsDrawer::blitOrCopyTexturedRect ?

loganmc10: I don't think copyTexturedRect will work to solve this problem. In order to use that, you still need to sample from m_pTexture to do the copy, which is the whole problem this is trying to solve. glBlitFramebuffer is the only way to do it. I should disable this solution if glBlitFramebuffer isn't supported though

src/FrameBuffer.cpp
```diff
@@ -461,15 +514,24 @@ CachedTexture * FrameBuffer::getTexture(u32 _t)
...
 CachedTexture * FrameBuffer::getTextureBG(u32 _t)
 {
-	m_pTexture->scaleS = m_scale / (float)m_pTexture->realWidth;
-	m_pTexture->scaleT = m_scale / (float)m_pTexture->realHeight;
+	CachedTexture *pTexture = m_pTexture;
+
+	if (this == frameBufferList().getCurrent()) {
+		if (Context::TextureBarrier)
+			gfxContext.textureBarrier();
+		else if (Context::BlitFramebuffer)
```
loganmc10: I added a check for ```Context::BlitFramebuffer``` here, so this won't work with GLES2 (GLES2 devices may still render these scenes properly, they may not, it'll just be like it is now)

gonetz:
Looks good for me, thanks!

I need to test it on my PC, hopefully this evening.

PR #1819, opened 2018-05-22
Allow non-0 default framebuffer

loganmc10:
@gonetz this should fix the issue compiling on 64-bit mupen64plus


loganmc10:
Just a reminder that there are a few cases in the overscan branch of using ObjectHandle::null which should be changed to ObjectHandle::defaultFramebuffer now

PR #1821, opened 2018-05-23
UI: Crop removed, overscan controls now in hideable tab widget

Aquatakat:
image


oddMLan:
I think "reduce overscan" is not the adequate vocabulary here (because they are contradictory terms; an oxymoron). Just "Overscan" would suffice. That's how most TV's name that option. It should have a tooltip that says something like "When enabled, it crops part of the image on screen. Useful to get rid of annoying black borders in some games".

gonetz:
@Aquatakat Looks and works great, thanks!
Merged: 2b9aebbfd0b

PR #1828, opened 2018-05-31
Minor improvements

gizmo98:
-GBI.h: Add better float reciproc values
-use _FIXED2FLOAT() and _FIXED2FLOATCOLOR() if possible
-use FIXED2FLOATRECIP16 if possible
-add FIXED2FLOATRECIPCOLOR5
-gSP.cpp, gSPInsertMatrix(): add comments. Remove int cast to preserve fraction value. Use copysignf() to transfer sign from fraction to newvalue.


src/uCodes/ZSortBOSS.cpp
```diff
@@ -519,9 +519,9 @@ void ZSortBOSS_Lighting( u32 _w0, u32 _w1 )
 	for(u32 i = 0; i < num; i++) {
 		SPVertex & vtx = pVtx[i];
 
-		vtx.nx = ((s8*)DMEM)[(nsrs++)^3] * 0.00390625f;
```
gonetz: @Gillou68310 I just noticed that. Normal is signed char. Should not it be divided by 127 instead of 256?

src/uCodes/ZSortBOSS.cpp
```diff
@@ -605,17 +605,17 @@ void ZSortBOSS_TransformLights( u32 _w0, u32 _w1 )
 		gSP.lights.rgb[i][G] = (f32)(((u8*)DMEM)[(addr+8+1)^3]) * 0.0039215689f;
 		gSP.lights.rgb[i][B] = (f32)(((u8*)DMEM)[(addr+8+2)^3]) * 0.0039215689f;
 
-		gSP.lights.xyz[i][X] = (f32)(((s8*)DMEM)[(addr+16+0)^3]) * 0.00390625f;
```
gonetz: @Gillou68310 There is the same: light vector is signed char. Why you divide it by 256 when convert to float?

src/gSP.cpp
```diff
@@ -524,7 +524,7 @@ void gSPLightVertexCBFD(u32 v, SPVertex * spVtx)
 			const f32 vy = (vtx.y + gSP.vertexCoordMod[9])*gSP.vertexCoordMod[13] - gSP.lights.pos_xyzw[l][Y];
 			const f32 vz = (vtx.z + gSP.vertexCoordMod[10])*gSP.vertexCoordMod[14] - gSP.lights.pos_xyzw[l][Z];
 			const f32 vw = (vtx.w + gSP.vertexCoordMod[11])*gSP.vertexCoordMod[15] - gSP.lights.pos_xyzw[l][W];
-			const f32 len = (vx*vx + vy*vy + vz*vz + vw*vw) / 65536.0f;
+			const f32 len = _FIXED2FLOAT((vx*vx + vy*vy + vz*vz + vw*vw),16);
```
gonetz: Purpose of _FIXED2FLOAT macro is to convert integer value to float. Here you use it to divide float value by 65536.
I know that result will be correct, but this use of that macro is confusing for code reader. 
Better use the same approach as in your previous commit:
f32 recip = FIXED2FLOATRECIP16;

gonetz:
Looks good for me now, thanks! Merged.
I still need comments from @Gillou68310 regarding conversion of signed char to float.

Gillou68310:
Looks weird indeed, I honestly don't remember why I'm doing this :-(

gonetz:
Could you check, how the microcode processes that data?

PR #1835, opened 2018-06-09
Minor improvements 2

gizmo98:
-use a u32 s32 union to load, store and insert matrix values
-ZSortBOSS_Audio2(): use absolut value of fraction

### DIFF
--- a/src/GBI.h
+++ b/src/GBI.h
@@ -69,6 +69,11 @@
 #define _SHIFTR( v, s, w )	\
 	(((u32)v >> s) & ((0x01 << w) - 1))
 
+union s32u32 {
+	s32 s;
+	u32 u;
+};
+
 // BG flags
 #define	G_BGLT_LOADBLOCK	0x0033
 #define	G_BGLT_LOADTILE		0xfff4
--- a/src/RSP_LoadMatrix.cpp
+++ b/src/RSP_LoadMatrix.cpp
@@ -3,6 +3,8 @@
 
 void RSP_LoadMatrix( f32 mtx[4][4], u32 address )
 {
+    s32u32 value;
+
     struct _N64Matrix
     {
         s16 integer[4][4];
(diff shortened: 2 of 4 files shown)


theboy181:
What improvements can be witnessed from this commit?

theboy181:
Thank! improvements are always rewarding. I'm sure there are benefits. :)

gonetz:
@gizmo98 The idea is good, thanks!
I moved matrix element conversion to separate functions.
The code is in https://github.com/gonetz/GLideN64/tree/gizmo98-minor-improvements-2 branch.
Please take a look. Do you know, which games use gSPInsertMatrix?
It would be good to test that function.

gizmo98:
@gonetz 
Thanks! Your modifications look much more polished. Nice job. I‘m on vacation atm. I will test your branch in two weeks.

According to this old thread Super Smash Bros uses gSPInsertMatrix. http://www.emutalk.net/threads/12288-glNintendo64()-v0-3/page8

Just another question. We load a s32 4x4 matrix from RDRAM and convert it to 4x4 float matrix. Float has a 23Bit mantissa. Float cannot store all possible s32 values. Wouldn‘t it be better to use a 4x4 double matrix?

weinerschnitzel:
IIRC, it is Kirby's Dreamland stage that uses insert matrix with the trees.

gonetz:
I'll test SSB and Kirby, thanks!

Regarding double for matrix: I'm sure that float precision is enough for 3D graphics, at least of N64 level. We do not need to keep any possible s32 value in float. Values of matrix elements are quite limited.

gonetz:
Tested. Works correct. Merged.

theboy181:
What does this fix? Any visual improvements?” To SSB like the Kirby stage?

gonetz:
@theboy181 You already asked this question. This commit is about code improvements.

Issue #133, opened 2014-11-13
Parlor! Pro 64: texture issue


olivieryuyu:
![screen68](https://cloud.githubusercontent.com/assets/8987490/5035823/a49c2f96-6b7a-11e4-9439-61100a27ac02.jpg)

LLE Software

![fs](https://cloud.githubusercontent.com/assets/8987490/5035825/a7257d94-6b7a-11e4-9e87-24e1f6cdab52.jpg)

Rice seems the only one to avoid this issue, may be with a hack, cannot say

olivieryuyu:
is it fixable? :(

gonetz:
Have not check it yet.

olivieryuyu:
4bit CI texture i guess :)

gonetz:
Do you know something special about 4bit CI textures?

olivieryuyu:
well yes and no. What i can see that 4bit texture are treated differently in the sprite microcode.

I am not sure however how i should interpret them. The code is a bit messy, even with so many comments.

olivieryuyu:
the command used to draw the 2D are just horrendous. 800 lines of asm just for one commmand... it is a big as the ground generation command in SWRS!

@gizmo98 : as you are interested in S2DEX, any idea where the issue could be? it should be in command 0xC2 and/or 0x05.

gizmo98:
@olivieryuyu i can take a look if i have fixed my current PR problem.

olivieryuyu:
thanks

i think i have a lead actually, hopefully it is the source of the issue :)

gonetz:
The problem rectangle is rendered by Sprite command 0xC2
https://level42.ca/projects/ultra64/Documentation/man/pro-man/pro25/index25.5.html

It uses the following formulas for texture coordinates:
```
x' = A * x + B * y + X
y' = C * y + D * y + Y
```
where A, B, C, D, X, Y are values from the 2D matrix.

2D matrix for this sprite is
```
A	0.0
B	-44.0
C	1.0
D	0.0
X	257.0
Y	150.0
```

(objX, objY) = (0, 0)
ImageW = 8, ImageH = 1
Upper left vertex has coordinates (objX, objY) = (0, 0)
Lower right vertex has coordinates (objX + ImageW/scaleW, objY + ImageH/scaleH) = (8, 1)
If you substitute the values to the formulas above, you will coordinates
(213, 150) - (257, 158)
[...]

PR #1847, opened 2018-06-26
Fix Worms Armageddon one pixel garbage line under names and menu items

gizmo98:
There is a one pixel garbage line under every menu item and under every worm name . If gSP.bgImage.height is imageH - 1 there is no garbage. I have tested this fix with multiple games and have not seen any regression.

I also replaced imageW modulo operation with & 0xFFFFFFFE.

### DIFF
--- a/src/gSP.cpp
+++ b/src/gSP.cpp
@@ -2288,8 +2288,8 @@ void _loadBGImage(const uObjScaleBg * _bgInfo, bool _loadScale)
 		gSP.bgImage.width = *REG.VI_WIDTH;
 		gSP.bgImage.height = (imageH * imageW) / gSP.bgImage.width;
 	} else {
-		gSP.bgImage.width = imageW - imageW%2;
-		gSP.bgImage.height = imageH - imageH%2;
+		gSP.bgImage.width = imageW & 0xFFFFFFFE;
+		gSP.bgImage.height = imageH - 1;
 	}
 	gSP.bgImage.format = _bgInfo->imageFmt;
 	gSP.bgImage.size = _bgInfo->imageSiz;


src/gSP.cpp
```diff
@@ -2288,8 +2288,8 @@ void _loadBGImage(const uObjScaleBg * _bgInfo, bool _loadScale)
 		gSP.bgImage.width = *REG.VI_WIDTH;
 		gSP.bgImage.height = (imageH * imageW) / gSP.bgImage.width;
 	} else {
-		gSP.bgImage.width = imageW - imageW%2;
-		gSP.bgImage.height = imageH - imageH%2;
+		gSP.bgImage.width = imageW & 0xFFFFFFFE;
+		gSP.bgImage.height = imageH - 1;
```
gonetz: I suppose that imageH - imageH%2 was done for purpose.
`imageH & 0xFFFFFFFE` would cause no questions.
I'm afraid that `imageH - 1` will cause regressions. I need to test it first.

olivieryuyu:
Does it work ok in lle?

gizmo98:
Comparison:
At the moment
org1
org2
org3

With fix
fix1
fix2
fix3

src/gSP.cpp
gizmo98: @gonetz 
There can be regressions but i have seen none. I have tested beetle adventure racing, sm64, mario tennis, smash bros, paper mario, yoshis story, mischief makers, bomberman and many other games, re2, doom, mario kart and goldeneye.

olivieryuyu:
this is a HLE issue as in LLE not such an issue.

I am working on the ucode (slowly but surely, i will take my time on this) but from there a proper fix should be found for all games with S2DEX errors.

theboy181:
I wonder if this is related to the issue I’m seeing in F-zero with the mph numbers showing up at the bottom of the texture.

theboy181:
Is there a binary available so I can test the changes?

DonelBueno:
Mario Tennis pause menu also has a line of garbage. I'm sure I've seen this in other games too, but Mario Tennis is the only one that comes to my mind.

tony971:
Mario Kart 64 has a garbage line to the right of the split screen race finish screen (on master. Dont know if this PR affects it.)

gizmo98:
@theboy181 
I have only a armv7 mp64 binary.

@olivieryuyu @gonetz 
According to SDEX.h imageW is 8byte aligned (https://github.com/gonetz/GLideN64/blob/master/src/uCodes/S2DEX.h#L22). imageH is not https://github.com/gonetz/GLideN64/blob/master/src/uCodes/S2DEX.h#L29. So if imageSiz is G_IM_SIZ_32b and imageW is uneven we need one more texel to get a even number of texels. The modulo operation takes care we have a even number of texels. (Do we need to clamp the extra texel?)

> ALL textures must Align to 64bit boundries in each row. 
> eg. a 32bit texture MUST use multiples of 2 texels, while a 4bit texture MUST use multiples of 16 texels. 
http://fgfc.ddns.net/PerfectGold/Textures.htm

Quite interesting. Do we ned more than one fill up texel if imageSiz  is G_IM_SIZ_16b, G_IM_SIZ_8b or G_IM_SIZ_4b?

I can write my PR a little bit different to get the same result:
`gSP.bgImage.width = (imageW - 1 ) + (imageW - 1 )%2;`
`gSP.bgImage.height = imageH - 1;
 `

olivieryuyu:
Interesting!

_Do we need to clamp the extra texel?_

See we need gSPObjRenderMode in manual

I invite also  you have a look at some other games with S2DEX issues: https://github.com/gonetz/GLideN64/issues/936 and https://github.com/gonetz/GLideN64/issues/133

gizmo98:
I have found more information:
https://level42.ca/projects/ultra64/Documentation/man/pro-man/pro25/25-05.html
> BG image's width, imageW must be aligned to 8 bytes. Since the actual values used for imageW and imageH are in (u10.2) format, the values to be assigned must be multiplied by 4. The following chart shows the imageW's value constraints, taking (u10.2) format into consideration and multiplying by 4. **There is no need to align imageH values.**

- When G_IM_SIZ_4b: | imageW is a multiple of 64
- When G_IM_SIZ_8b: | imageW is a multiple of 32
- When G_IM_SIZ_16b: | imageW is a multiple of 16
- When G_IM_SIZ_32b: | W is a multiple of 8

So if modulo operation is there to calculate the right imageW it is not necessary to correct imageH.

olivieryuyu:
this issue did not exit in Sep 2017. It is a new one. Seems we are turning around an issue...

gizmo98:
@olivieryuyu if it is a regression i could try to bisect the commit which changes behaviour.

olivieryuyu:
it was not here in September 2017. I guess something happened after.

gizmo98:
Reverting https://github.com/gonetz/GLideN64/commit/feb0e53034d54095464687dbae17854c9dc3caa6 but keeping https://github.com/gonetz/GLideN64/commit/89406223f5e271caab3687e550a48c487dbca9fd fixes it.

gizmo98:
I will close this for now. I get equal or better results if i revert above mentioned commit and modify ObjCoordinates creation.

Issue #1863, opened 2018-07-16
Super Smash Brothers Regression


theboy181:
with latest build SSB64 has a regression with the 2D stage elements. I think this was found in some of the older plugins too.

![image](https://user-images.githubusercontent.com/9299743/42766536-9b244b3a-88cf-11e8-8325-f386ddc3e652.png)

weinerschnitzel:
Looks like insert matrix issue that Rice Video used to have. Does #1835 affect this?

theboy181:
I tired to ask what this would do, and no one took the time to explain.. If this is caused by this commit, then I fear for GLideN64's future. There have been many regressions lately, and I think that the project is just accepting things without properly testing first.

I do hope there is some traction on the many regressions that have been reported lately.

weinerschnitzel:
It's a regression prone type of project, but there are some competent people fixing things and keen eyed people reporting issues properly. And consistently I'm confident that we are ahead of the regression game. No worries here.

gizmo98:
@theboy181 
Savestate please! I will try to bisect the wrong commit.

theboy181:
All you need to do is go to vs mode and pick dreamland stage

gizmo98:
@theboy181
Thank you! I will take a look.

gizmo98:
@theboy181 
I have found a small difference between my PR and gonetz refactored code. Results are different if fractional element is to be changed and the current matrix value is negativ. I need to check this modification later:
https://github.com/gonetz/GLideN64/commit/6993042c4ce6143a16de088eba3ee9fc2f7b3213#diff-69d8715ec7f9fd627ec4f5516edd003d

theboy181:
Your code doesn’t have this issue, the refactoring does?

gizmo98:
Yeah. Gonetz ask me to test the refactoring but i had no time because i was on vacation. So it's more or less my problem...

I checked the above posted commit. It fixes the problem.

PR #1864, opened 2018-07-17
gSPInsertMatrix: fix insert fractional part if old value is negativ

gizmo98:
**Example:** Matrix value is -2.5f and new fraction is 0x44 

**Current implementation of change fractional part:**
s32 integer = static_cast<s32>(-2.5f) = -2
GetFloatMatrixElement(static_cast s16 (integer), 0x44) = -1,999

**Fix:**
s32 integer = (-2.5f * 65535.0f) >> 16 = (163837) >> 16 = GetIntMatrixElement(-2.5f).first =-3
GetFloatMatrixElement(static_cast s16 (integer), 0x44) = -2,999

Fixes https://github.com/gonetz/GLideN64/issues/1863.

### DIFF
--- a/src/gSP.cpp
+++ b/src/gSP.cpp
@@ -1593,13 +1593,13 @@ void gSPInsertMatrix( u32 where, u32 num )
 		if (where < 0x20) {
 			// integer elements of the matrix to be changed
 			const s16 integer = static_cast<s16>(pData[i ^ 1]);
-			const s32 fract = static_cast<s32>(gSP.matrix.combined[0][index + i] * 65536.0f);
-			gSP.matrix.combined[0][index + i] = GetFloatMatrixElement(integer, static_cast<u16>(fract&0xFFFF));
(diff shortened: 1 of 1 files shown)


gizmo98:
N64 uses a s16.u16 representation for matrix values. GlideN64 uses float values.  https://github.com/gonetz/GLideN64/commit/3ff1478103c1ccf6d375252730e29c456eeb06e8 introduced `float GetFloatMatrixElement(s16 _int, u16 _fract)` and `std::pair<s16, u16> GetIntMatrixElement(f32 _elem)` for save conversion between both representations. During review it was overseen that at one location `static_cast_s16(f32 _elem)` was used instead of `GetIntMatrixElement(f32 _elem)` to get the integer value of float.  `static_cast_s16(f32 _elem)` produces wrong integer values if float value is negativ.

gonetz:
Thanks!

PR #1881, opened 2018-08-12
ObjCoordinates: disable coord correction if hack_ZeldaMM is active

gizmo98:
Fix Zelda MM motion blur effect for now. 

Motion blur effect could be broken due special Zelda MM hacks in gSPBgRect1Cyc(). Disable G_OBJRM_BILERP coordinate correction if `hack_ZeldaMM` is set.


gizmo98:
I don't think this is the right solution. Just close it for now.

PR #1910, opened 2018-09-27
tile functions: Replace magic numbers with constants

gizmo98:
Just refactoring. Replace magic numbers with constants. No functional
changes.


gonetz:
Thanks!

PR #1960, opened 2018-12-10
Fix performance with radeonsi.

orbea:
This takes `Donkey Kong 64` from 12 fps to full speed with framebuffer emulation enabled.

I recently upgraded my Nvidia GTX 780 Ti card to an AMD RX Vega 56 and while everything was much faster, GLideN64 became far slower with most games. After a lot of help from #radeon @ freenode I found the bottleneck with perf.
```
    67.80%  libc-2.28.so                   [.] __memcpy_ssse3
     5.33%  [kernel]                       [k] deactivate_slab.isra.69
     3.75%  [processor]                    [k] init_module
     3.25%  [amdgpu]                       [k] amdgpu_mm_rreg
     0.65%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f6e
     0.53%  [kernel]                       [k] apic_timer_interrupt
     0.46%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f61
     0.38%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f6b
     0.37%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f76
     0.32%  [kernel]                       [k] native_load_gs_index
     0.29%  [kernel]                       [k] __switch_to
     0.27%  mupen64plus-video-GLideN64.so  [.] ColorBufferToRDRAM::_RGBAtoRGBA16
     0.27%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f5d
     0.23%  [kernel]                       [k] _raw_spin_lock_irqsave
     0.22%  [kernel]                       [k] _raw_spin_lock
     0.22%  mupen64plus-video-GLideN64.so  [.] ColorBufferToRDRAM::_copy
     0.20%  [kernel]                       [k] __x86_indirect_thunk_rax
     0.20%  [kernel]                       [k] check_preemption_disabled
     0.20%  [kernel]                       [k] __sched_text_start
     0.18%  [kernel]                       [k] read_tsc
     0.17%  mupen64plus-video-GLideN64.so  [.] FrameBuffer::isValid
     0.16%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f54
     0.16%  [kernel]                       [k] entry_SYSCALL_64
     0.14%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f64
     0.14%  [kernel]                       [k] ___slab_alloc.constprop.82
     0.13%  [kernel]                       [k] native_sched_clock
     0.13%  [kernel]                       [k] menu_select
     0.13%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f72
     0.13%  [kernel]                       [k] preempt_count_add
     0.13%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f45
     0.12%  mupen64plus-video-GLideN64.so  [.] CRC_Calculate
     0.12%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f2b
```
More specifically its this single `memcpy`.

https://github.com/gonetz/GLideN64/blob/f7a4cf0c7532c27f73106ce3a9f33841ae0af68f/src/Graphics/ColorBufferReader.cpp#L55

The committed change was suggested which takes GLideN64 back to full speed on my system.

My very basic understanding is that without this its reading that memcpy with VRAM which is very slow and that there is no reason to not add this.

### DIFF
--- a/src/Graphics/OpenGLContext/opengl_ColorBufferReaderWithBufferStorage.cpp
+++ b/src/Graphics/OpenGLContext/opengl_ColorBufferReaderWithBufferStorage.cpp
@@ -31,7 +31,7 @@ void ColorBufferReaderWithBufferStorage::_initBuffers()
 	for (u32 index = 0; index < m_numPBO; ++index) {
 		m_bindBuffer->bind(Parameter(GL_PIXEL_PACK_BUFFER), ObjectHandle(m_PBO[index]));
 		m_fence[index] = 0;
-		glBufferStorage(GL_PIXEL_PACK_BUFFER, m_pTexture->textureBytes, nullptr, GL_MAP_READ_BIT | GL_MAP_PERSISTENT_BIT | GL_MAP_COHERENT_BIT);
+		glBufferStorage(GL_PIXEL_PACK_BUFFER, m_pTexture->textureBytes, nullptr, GL_MAP_READ_BIT | GL_MAP_PERSISTENT_BIT | GL_MAP_COHERENT_BIT | GL_CLIENT_STORAGE_BIT);
 		m_PBOData[index] = glMapBufferRange(GL_PIXEL_PACK_BUFFER, 0, m_pTexture->textureBytes, GL_MAP_READ_BIT | GL_MAP_PERSISTENT_BIT | GL_MAP_COHERENT_BIT);
 	}
 


Jj0YzL5nvJ:
[This is missing background...](https://bugs.freedesktop.org/show_bug.cgi?id=102204#c7)

` Matias N. Goldberg 2018-11-03 16:51:03 UTC`
```
Disclaimer: I'm not a Mesa dev.

I saw this ticket by accident and since I'm a heavy user of GL_ARB_buffer_storage on AMD+Mesa, I took a look in case there's something that could affect me.

After glancing through the problems, it appears that the problem is "user error" and the real bug here are the lack of performance warnings to report via KHR_debug.

GL_ARB_buffer_storage can backfire if you do not use it well.
It would appear that you're performing several operations that are not supported in HW (such as the use of GL_UNSIGNED_BYTE which requires converting all indices to GL_UNSIGNED_SHORT in SW) and must be processed to work correctly.

Or do stuff that is a minefield, like issuing a loop of glMemoryBarrier+glDrawElementsBaseVertex PER TRIANGLE while reading from a persistent mapped buffer. That is just not gonna work fast.

The reason GL_CLIENT_STORAGE_BIT improves things is because the buffer is stored on a CPU buffer (i.e. a good ol' malloc), and when it's time to render, Mesa does the SW conversions for you, then uploads the data to GPU.
When that flag is not present, Mesa must be constantly downloading & uploading data from GPU back & forth to accommodate whatever you're doing that is not natively supported.

The rule of thumb is that persistent mapped memory should be treated as write only, read once memory.
If I'm writing view matrices and then render that once, I write directly to the the persistent mapped buffer and use it directly on the shader.
If I'm writing material/pass data that is reused multiple times on multiple draws, I write it once to the persistent mapped buffer and then perform a glCopyBufferSubData to another buffer that is not CPU visible at all.

Btw avoid using GL_MAP_COHERENT_BIT. Your calls to glReadPixels to directly store into the PBO backed by coherent memory could be hurting you a lot. glReadPixels must do deswizzling; it's not a raw memcpy on the GPU side. Keeping both the CPU & GPU caches in sync is foolishness here.
If you insist on using GL_MAP_COHERENT_BIT for your glReadPixels; perform the glReadPixels onto a PBO buffer that is not CPU visible (to perform deswizzling), then perform a glCopyBufferSubData from that PBO into another PBO (that is now CPU visible).
```

` Matias N. Goldberg 2018-11-03 18:25:08 UTC`
```
OK this left me wondering.

I looked further into what happens when you call glBufferStorage, and it seems that buffer_usage in st_cb_bufferobjects.c https://github.com/anholt/mesa/blob/master/src/mesa/state_tracker/st_cb_bufferobjects.c#L226 stores all buffers in VRAM unless GL_CLIENT_STORAGE_BIT is used because it always returns PIPE_USAGE_DEFAULT.

The chosen pipe_resource_usage will end up in si_init_resource_fields (https://github.com/anholt/mesa/blob/master/src/gallium/drivers/radeonsi/si_buffer.c#L103) which ends up putting the buffers all buffers in VRAM with Write Combining (unless GL_CLIENT_STORAGE_BIT is set)

That is... an odd choice. This is specially bad for buffers requested with GL_MAP_READ_BIT flags, which clearly should not be stored in VRAM with WC bits set.

This indeed looks like a bug to me. Unfortunately, these hint flags don't map really well to HW.
```

mudlord:
**....Certain people should have profiled this first**. I'll leave it at that.

gonetz:
Ok, thanks for the info and fix. Lets add it.

loganmc10:
> **....Certain people should have profiled this first**. I'll leave it at that.

Lol most developers are going to have access to 1 type of GPU and 1 OS. This only affected Linux + AMD GPUs. An open source project like this is going to depend on people with access to different kinds of hardware contributing (notice how you don't have the "contributer" badge)

gonetz:
@mudlord it will be reverted or made Linux specifix if it breaks performance for someone else.

orbea:
> This only affected Linux + AMD GPUs.

Its kind of strange that nouveau is fast here, llvmpipe is fast because it has no VRAM and nvidia is likely choosing the fast path regardless of what the app wants... I'm not sure about intel.

I suspect this might offer perf improvements in other cases, but AMD + mesa really hit it hard.

loganmc10:
GL_CLIENT_STORAGE_BIT is just a hint, saying "please don't store this in VRAM". A smart GPU driver can probably detect when it is best to store something in VRAM or in RAM. I don't know a lot about Intel, but I was under the impression that there is a pretty blurry line between RAM and VRAM with an integrated GPU

On the flip side, there could be drivers that ignore GL_CLIENT_STORAGE_BIT and still store the buffer in VRAM.

orbea:
I understand that intel and android are taking a different code path here, I was also able to avoid the performance issue by forcing the `ColorBufferReaderWithPixelBuffer` path in `src/Graphics/OpenGLContext/opengl_ContextImpl.cpp`.

https://github.com/gonetz/GLideN64/blob/67e68ed1f63adf1d0ab53736c8db23ef0ddfbacf/src/Graphics/OpenGLContext/opengl_ContextImpl.cpp#L346

orbea:
Also note that there are some other smaller potential bottlenecks left.
```
    20.10%  [processor]                    [k] init_module
     3.21%  mupen64plus-video-GLideN64.so  [.] CRC_Calculate
     2.15%  mupen64plus-video-GLideN64.so  [.] ColorBufferToRDRAM::_RGBA
     1.88%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f6e
     1.86%  libc-2.28.so                   [.] __memcpy_ssse3
     1.57%  mupen64plus-video-GLideN64.so  [.] Rasterize
     1.52%  mupen64plus-video-GLideN64.so  [.] ColorBufferToRDRAM::_copy
     1.41%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f61
     1.27%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f76
     1.24%  [kernel]                       [k] deactivate_slab.isra.69
     1.24%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f6b
     1.04%  mupen64plus-video-GLideN64.so  [.] UnswapCopyWrap
     1.01%  mupen64plus-video-GLideN64.so  [.] FrameBuffer::isValid
     0.85%  mupen64plus-video-GLideN64.so  [.] renderTriangles
     0.78%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f5d
     0.61%  libmupen64plus.so.2.0.0        [.] 0x0000000000046055
     0.48%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f64
     0.43%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f54
     0.41%  mupen64plus-video-GLideN64.so  [.] FrameBufferList::fillRDRAM
     0.40%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f5a
     0.39%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f3b
     0.38%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f72
     0.37%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f38
     0.37%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f27
     0.36%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f68
     0.36%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f2b
     0.36%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f45
     0.36%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f51
     0.35%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f49
     0.34%  [kernel]                       [k] _raw_spin_lock_irqsave
     0.33%  libc-2.28.so                   [.] __memset_sse2_unaligned_erms
     0.30%  [kernel]                       [k] update_blocked_averages
     0.30%  mupen64plus-video-GLideN64.so  [.] ColorBufferToRDRAM::_RGBAtoRGBA16
     0.29%  libpthread-2.28.so             [.] __pthread_mutex_lock
     0.26%  [kernel]                       [k] _raw_spin_lock
     0.25%  libasound.so.2.0.0             [.] 0x0000000000086963
     0.24%  [kernel]                       [k] __sched_text_start
     0.23%  [kernel]                       [k] check_preemption_disabled
     0.21%  libasound.so.2.0.0             [.] 0x0000000000086950
     0.21%  libspeexdsp.so.1.5.0           [.] 0x0000000000009f7c
     0.20%  [amdgpu]                       [k] amdgpu_mm_rreg
     0.19%  [kernel]                       [k] preempt_count_add
     0.19%  [kernel]                       [k] __x86_indirect_thunk_rax
     0.19%  [kernel]                       [k] menu_select
     0.18%  libmupen64plus.so.2.0.0        [.] 0x0000000000046058
     0.18%  mupen64plus-video-GLideN64.so  [.] gDPLoadBlock
     0.18%  [kernel]                       [k] __switch_to
     0.16%  [kernel]                       [k] native_load_gs_index
     0.16%  [kernel]                       [k] __update_load_avg_cfs_rq
     0.15%  [kernel]                       [k] find_busiest_group
     0.15%  libmupen64plus.so.2.0.0        [.] 0x0000000000024e5f
```

gonetz:
@orbea thanks! Rasterize can be moved to separate thread since it s result is not needed until the end of display list processing. It would save some CPU time and mobile devices could get performance benefits. I just don't have time to implement it.

PR #1983, opened 2019-01-14
Android makefile updates for 64 bit

fzurita:
Android makefile updates for 64 bit


gizmo98:
Could be android uses default -march=armv8-a instead of -march=armv8-a+crc. Intrinsics can only be used if a cpu with crc support is selected or if generic -armv8-a+crc is set.

Issue #1933, opened 2018-10-18
Fix bilinear filtering


theboy181:
![image](https://user-images.githubusercontent.com/9299743/47132633-d3ac3200-d259-11e8-92f4-a2c9058062f8.png)
OK looks a lot better. Are the sprites also a bug? or something different?

theboy181:
![image](https://user-images.githubusercontent.com/9299743/47132776-91cfbb80-d25a-11e8-90ce-6353ed0c3404.png)
Something else is wrong. The output of the pixels here are smaller than the others. and you can see that its happening throughout all games.

theboy181:
![image](https://user-images.githubusercontent.com/9299743/47133118-1cfd8100-d25c-11e8-98b8-4829836a506d.png)

gonetz:
> Shouldn't all the pixels be square?

In native res and without VI post-processing - yes, they should, but there is no such mode in GLideN64.

theboy181:
Can this issue be fixed for X2+ resolutions, and VI post-processing? How are other emulators able to double the resolution without issue?

gizmo98:
Just a note about 2D elements which consist of more then one tile and the use of higher resolutions. I'm not an expert but i think N64 splits big textures in tiles because texture memory is limited. Most of the time tile borders are visible (those ugly lines in textures) because texture filter cannot sample from the surrounding tiles. A red texel at coord 1.0 from tile0 and a white texel at coord 0.0 from tile1 will not be interpolated to rose. If texture clamp is enabled You will see a red or white pixel at the tile border.

If you know all tiles you could reconstruct the original texture and sample from this one. But this is no easy task.

theboy181:
Probably big part of the issue and even know native res was mentioned as not having this issue, is simply does.

gizmo98:
I have found a thread about tiles and seams:
https://gamedev.stackexchange.com/questions/74420/seamless-tilemap-rendering-borderless-adjacent-images

theboy181:
I wondered if it wasn’t locked to intergerscaling. This thought to impliment

weinerschnitzel:
I think it would be useful for comparisons and also nice to have if we could disable VI

gonetz:
Fix with use of premultiplied alpha color values works great where it applicable.
However, the idea to smooth alpha for filtered textures is questionable. Stefan's comment: "improves kirby64 and zelda mm text". Lets look at Zelda MM text:
[...]

PR #1996, opened 2019-02-03
Add bilinear filtering modes which remove halos around transparent te…

gizmo98:
…xtures.

0=N64 3point, 1=standard, 2=N64 3point with color bleeding, 3=standard with color bleeding and premultiplied alpha.

New options fix halos around transparent textures in pokemon puzzle league, mario kart, body harvest, puyo pyuo etc.

New bilerp options are not implemented for gles2 devices.
https://github.com/gonetz/GLideN64/issues/1933


tony971:
Since the optimal setting is game dependent, is this something that could possibly make its way to the gliden64.custom.ini?

gonetz:
Merged, thanks!

gonetz:
> Since the optimal setting is game dependent, is this something that could possibly make its way to the gliden64.custom.ini?

Yes. It is usual config setting. Just add
`texture\bilinearMode=2`
or
`texture\bilinearMode=3`
to custom section of your game.

tony971:
Sorry, I wasn't very specific. What I meant was having an "auto" option and the best setting for each game loaded into the default custom.ini

gonetz:
"best setting" is individual imo.
Some people prefer PC-style filtering, others like N64-style more.
I think, it will be better to add an option "remove halos", which will enable fixed bilinear filtering for user-selected filtering style. In that case it will be possible to enable that option in custom.ini for all games, which suffer from "halos" issue.

PR #1997, opened 2019-02-04
Add support for Android EGL Image Public API

fzurita:
This adds support for the EGL image public API introduced in Android API level 26. This makes copy color from to RDRAM very fast in Async mode.

Also I did some refactoring of how glEGLImageTargetTexture2DOES and added a new macro definition to let the plugin know if the API is available.

Eventually, I should do run time loading of the API to avoid having to use the macro definition.


fzurita:
That wasn't too bad actually. I switched it to use run time loading of the public API using chromium as an example.

src/Graphics/OpenGLContext/GraphicBufferPublicApi/android_hardware_buffer_compat.h
```diff
@@ -0,0 +1,75 @@
...
+	void** out_virtual_address);
+	int RecvHandleFromUnixSocket(int socketFd, AHardwareBuffer** outBuffer);
+	void Release(AHardwareBuffer* buffer);
+	int SendHandleToUnixSocket(const AHardwareBuffer* buffer, int socketFd);
+	int Unlock(AHardwareBuffer* buffer, int32_t* fence);
+
+private:
+
+	AndroidHardwareBufferCompat();
+    ~AndroidHardwareBufferCompat() = default;
```
fzurita: Should be tab

src/Graphics/OpenGLContext/opengl_ColorBufferReaderWithEGLImagePrivate.h
```diff
@@ -0,0 +1,33 @@
...
+	~ColorBufferReaderWithEGLImagePrivate() override = default;
+
+	const u8 * _readPixels(const ReadColorBufferParams& _params, u32& _heightOffset, u32& _stride) override;
+
+	void cleanUp() override;
+
+private:
+	void _initBuffers();
+
+    CachedBindTexture * m_bindTexture;
```
fzurita: Should be tabs

src/mupen64plus-video-gliden64.mk
```diff
@@ -5,7 +5,9 @@ include $(CLEAR_VARS)
 LOCAL_PATH := $(JNI_LOCAL_PATH)
 SRCDIR := ./$(BASE_DIR)/src
 
+
```
fzurita: Get rid of extra spaces

fzurita:
@gonetz This should be ready to merge, unless you have any comments

fzurita:
I have some ideas to clean this up even more... I can create a wrapper class that will automatically use the public or private API depending on API level.

gonetz:
These changes are mostly out of my competence and localized in files which I almost never touch.
I hope that it will really help to Android users, so, merged.  Francisco, thanks!

BTW, which Android version/ GLES version is required to get that 'fast color buffer copy'?

fzurita:
Android Oreo (8.0) or higher is required. Or a devices that has marshmallow (6.0) or less and doesn't support GLES 3.0.

fzurita:
Also, for some reason it doesn't work with the shield devices, but I suspect it's because they are using full GL.

gonetz:
Ok, thanks. I don't have a device, which supports that feature.

Issue #2005, opened 2019-02-13
Building fails on RPI due to f209304d


joolswills:
```
-- The C compiler identification is GNU 6.3.0
-- The CXX compiler identification is GNU 6.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
bcm_host.h found
-- Found ZLIB: /usr/lib/arm-linux-gnueabihf/libz.so (found version "1.2.8") 
-- Found PNG: /usr/lib/arm-linux-gnueabihf/libpng.so (found version "1.6.28") 
-- Found Freetype: /usr/lib/arm-linux-gnueabihf/libfreetype.so (found version "2.6.3") 
-- Configuring done
-- Generating done
-- Build files have been written to: /home/pi/RetroPie-Setup/tmp/build/mupen64plus/GLideN64/projects/cmake
Scanning dependencies of target osal
[  1%] Building C object osal/CMakeFiles/osal.dir/osal_files_unix.c.o
[  2%] Linking C static library libosal.a
[  2%] Built target osal
Scanning dependencies of target GLideNHQ
[  3%] Building CXX object GLideNHQ/CMakeFiles/GLideNHQ.dir/TextureFilters_2xsai.cpp.o
[  3%] Building CXX object GLideNHQ/CMakeFiles/GLideNHQ.dir/TextureFilters.cpp.o
[  4%] Building CXX object GLideNHQ/CMakeFiles/GLideNHQ.dir/TextureFilters_hq2x.cpp.o
[  5%] Building CXX object GLideNHQ/CMakeFiles/GLideNHQ.dir/TextureFilters_hq4x.cpp.o
[  6%] Building CXX object GLideNHQ/CMakeFiles/GLideNHQ.dir/TextureFilters_xbrz.cpp.o
[  7%] Building CXX object GLideNHQ/CMakeFiles/GLideNHQ.dir/TxCache.cpp.o
[...]

PR #2006, opened 2019-02-14
Fix build errors when using CMake with RPi

fzurita:
Should fix https://github.com/gonetz/GLideN64/issues/2005 or at least attempt to.

@joolswills can you please test.


gonetz:
I'll wait for @joolswills confirmation.

joolswills:
Sorry for delay. I'm probably not going to have time to test till tomorrow. Thanks.

joolswills:
Apologies again for the delay - I forgot :-)

I can confirm this is building now - thanks for the fix.

PR #2011, opened 2019-02-22
Fix shader compilation in GLSL

fzurita:
The latest shaders updates don't seem to compile with OpenGL ES devices, this seems to work at the cost of some code duplication.


gonetz:
My fix does not look as nice as before, but if it works with OpenGL ES devices now its ok.

PR #2012, opened 2019-02-23
Fix shader storage with Android 4.4

fzurita:
wchars don't seem to work very well on anything Android 4.4 and under. This fixes that.

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_ShaderStorage.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_ShaderStorage.cpp
@@ -21,31 +21,44 @@
 
 using namespace glsl;
 
-#define SHADER_STORAGE_FOLDER_NAME L"shaders"
+#define SHADER_STORAGE_FOLDER_NAME "shaders"
 
 static
-void getStorageFileName(const opengl::GLInfo & _glinfo, wchar_t * _shadersFileName, const wchar_t * _fileExtension)
+std::string getStorageFileName(const opengl::GLInfo & _glinfo, const char * _fileExtension)
 {
 	wchar_t strCacheFolderPath[PLUGIN_PATH_SIZE];
 	api().GetUserCachePath(strCacheFolderPath);
+
+	// Convert wchar string to char string
+	char strCacheFolderPathChar[PLUGIN_PATH_SIZE];
+	std::wcstombs(strCacheFolderPathChar, strCacheFolderPath, PLUGIN_PATH_SIZE);
+
+	std::stringstream path;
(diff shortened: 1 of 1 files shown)


fzurita:
Fixed some left over logging and formatting issues.

gonetz:
@fzurita Just tested. The fix works incorrect. For example when I run current master build for mupen64plus, I have the following shader cache files :
```
%USER%\AppData\Roaming\Mupen64Plus\shaders\GLideN64.91e9293e.OpenGL.keys 
%USER%\AppData\Roaming\Mupen64Plus\shaders\GLideN64.91e9293e.OpenGL.shaders
```
With your fix I get
%USER%\AppData\Roaming\Mupen64Plus\GLideN64.2447976766.OpenGL.keys 
%USER%\AppData\Roaming\Mupen64Plus\GLideN64.2447976766.OpenGL.shaders

That is file names are wrong and they saved in a wrong place.

Also, I have doubts that your fix will work correct if path to shader storage folder will contain non-ASCII symbols. Please test it too.

fzurita:
Ok, thanks, it looks like I forgot to change decimal to hex on the file name. Not sure what happened with the folder.

Also, you are right, this won't work if the user config has non ASCII character. That won't be a problem in Android 4.4 since it doesn't support non ASCII characters. I will have to think how to handle the case where supporting is available.

fzurita:
@gonetz The existing implementation already won't work with non ASCII characters. You are converting wchar_t to chat right before loading it saving the files. For example:

https://github.com/gonetz/GLideN64/blob/164b44e8e56422cb71cbf13d5cccacbea1aaad10/src/Graphics/OpenGLContext/GLSL/glsl_ShaderStorage.cpp#L67-L68

fzurita:
Ok, it doesn't seem like Linux supports supports wchars in filenames from the reading I have done so far. Filenames in Linux seem to be treated as null terminated byte arrays. So if a wchar contains a \0, then that will be treated as the end of the file name.

fzurita:
@gonetz do you have a Linux PC? If so, does making a file with Russian characters work?

gonetz:
Yes, now it works correct, thanks!

> The existing implementation already won't work with non ASCII characters.

Yes. Locale need to be set from default "C" to the locale settings selected in the environment. I'll do it.

> You are converting wchar_t to chat right before loading or saving the files. For example:

wcstombs converts wchar_t to multibyte char. It is the way Nix systems works with unicode characters.
Thus, it is wrong idea to set size of  output buffer the same as input buffer size. wchar_t character can be translated to sequence of several chars.

PR #1452, opened 2017-04-02
Threaded OpenGL API calls

fzurita:
This is the ongoing work to get a threaded implementation of the OpenGL API going.

This is not threaded yet, but it's the refactoring I have had to do so far that has been needed to support that.

Currently this does seem to have a performance regression, I have to figure out where it's coming from.

PR #2014, opened 2019-02-23
Threaded OpenGL API calls

fzurita:
This is a redo of the implementation of https://github.com/gonetz/GLideN64/pull/1452. This version has almost no impact on the main code and it just replaces the gl calls with the threaded version since the API is now the same.

This was specifically tailored to work with GLideN64, but it could theoretically work with other projects with minor modification.

Downsides are that not all OpenGL calls are implemented, only the ones the GLideN64 uses. If new OpenGL calls are needed, they will have to be added, but that is no different than before. Adding calls does become a little more difficult though. Also, if threading is disabled, it will behave identically as before.

Most everything is in the src/Graphics/OpenGLContext/ThreadedOpenGl folder, so it's kept out of the way.

And one last thing, I only checked compilation in Windows. I haven't tested this version in Windows, but it should work the same as the older pull request.


src/Graphics/OpenGLContext/opengl_ColorBufferReaderWithBufferStorage.cpp
```diff
@@ -30,7 +30,6 @@ void ColorBufferReaderWithBufferStorage::_initBuffers()
 	// Initialize Pixel Buffer Objects
 	for (u32 index = 0; index < m_numPBO; ++index) {
 		m_bindBuffer->bind(Parameter(GL_PIXEL_PACK_BUFFER), ObjectHandle(m_PBO[index]));
-		m_fence[index] = 0;
```
fzurita: No longer using fences since they can hurt performance and this is not really trying to be that accurate.

src/Graphics/OpenGLContext/opengl_CachedFunctions.cpp
```diff
@@ -152,9 +173,7 @@ void CachedTextureUnpackAlignment::setTextureUnpackAlignment(s32 _param)
 /*---------------CachedFunctions-------------*/
 
 CachedFunctions::CachedFunctions(const GLInfo & _glinfo)
-: m_bindFramebuffer(GET_GL_FUNCTION(glBindFramebuffer))
```
fzurita: We don't want to get any function pointers directly.

fzurita:
Performance improvements vary by settings, but with my phone and synced color buffer copies to RDRAM enabled, performance is 30% to 40% faster in the base case scenario. With async, I'm seeing about a 10-15% performance improvement.

fzurita:
Performance could also be further improved. I'm still newing things up a lot. I also found a thread safe memory pool implementation which helped significantly, but if I used it too much, it actually slowed things down due to synchronization.

I think eventually I will replace that thread safe memory pool with a lock free ring buffer which should further improve performance since allocating in the heap is so expensive.

gonetz:
> I haven't tested this version in Windows

I just tested it. 32bit mupen64plus build crashes on start of any rom.
Compilation of 64bit version gives me tons of warnings like:
`src\graphics\openglcontext\threadedopengl\opengl_WrappedFunctions.h(991): warning C4244: 'initializing': conversion from '__int64' to 'unsigned long', possible loss of data (compiling source file ..\..\src\Graphics\OpenGLContext\ThreadedOpenGl\opengl_WrappedFunctions.cpp)`
The build crashes too.

fzurita:
Ok, thanks, I'll investigate. Android works correctly though. I have the belong that is probably something wrong with the initialization of GL function pointers.

fzurita:
I started debugging this in Windows. It doesn't crash for me, but it does hang at startup. I'll try to figure out what that's about.

fzurita:
Ok, it now correctly runs in Windows and there are no warnings in the 64 bit build.

I am encountering a strange issue that I didn't see in the previous version of this. Disabling the speed limiter actually makes it run slower.... Trying to debug.

fzurita:
Ok, it was running slow because I was doing a debug build instead of release build. In Windows though, threaded mode seems to run slower then not threaded. I have the feeling it's the synchronized memory pool I'm using. I'm going to have to to move a lock free ring buffer memory pool.

fzurita:
Even after introducing a lock free ring buffer pool, performance still seems to be about 10-20% worse in Windows when threaded mode is enabled.... I don't quite understand why. It seems to be the opposite in Android.

gonetz:
Sure, I'll check it.

gonetz:
@fzurita How do you test performance?

fzurita:
Turn threaded mode on and off. It can be done using the GUI or the
GLideN64.ini.

gonetz:
> I usually just compare FPS.
It is not reliable. FPS counter constantly jumps.

fzurita:
I guess a more reliable way would be total frames rendered over a longer period of time, like frames per minute. There is no easy way to get that though.

Also, I know the source of the slow down on windows though. The opengl command queue consumer is slow to respond to the producer. It's fast in Linux. It seems like mutexes and condition variables are not as efficient in Windows (in my use case). I'm going see if I can figure out a workaround.

fzurita:
Also, for N64 depth compare, which seem really slow according to you in Banjo Tooie, you should be able to perceive a difference if the change in performance is large enough. I think it could help this scenario since it's a GPU limited case.

fzurita:
Performance seems to be ok in threaded mode in Windows now, only around 5% performance loss by just looking at the FPS counter with no limiter.

It should help though in cases where GPU drivers have to wait, which doesn't seem to happen in Windows at all it seems.

loganmc10:
I believe some OpenGL drivers are already threaded, so this kind of thing would probably be mostly duplicated in the driver, which is why in some instances you probably won't see much difference or even slightly worse results

fzurita:
Yeah, I'm aware of that. Which is probably why I don't see much of a performance improvement in Windows. I was hoping it would help in cases where the GPU is stalling, like N64 depth compare.

fzurita:
Ok, it would help if I had frame buffer emulation enabled.

I'm seeing a significant performance improvement in windows with sync color buffer to RDRAM enabled. About a 25% improvement.

fzurita:
With frame buffer emulation enabled, enabling threading is definitely faster in all cases I have seen so far in Windows.

fzurita:
After finding the bottleneck in Windows and fixing it there. It has vastly improved android performance. It went from a 10-15% performance improvement to about a 40% performance improvement in Android.

gonetz:
@fzurita I tried to test the latest version of Threaded GLideN64. With my imprecise FPS counter I can't say for sure that I noticed any good or bad impact. Btw, I can't reproduce slowdown with N64 depth compare in BT anymore, with master or with your version.

Of course, 40% performance improvement in Android as a strong argument, and the way it is made now it does not make things much more complicated. The only problem I see now is that I can't compile your code with VS 2013 because of:
`	static constexpr GLbitfield WRITE_ACCESS_ASYNC = GL_MAP_WRITE_BIT | GL_MAP_UNSYNCHRONIZED_BIT;
`
VS 2013 does not support constexpr. I'm planning to drop VS 2013 support after new public release, but not now. Can you rewrite that line?

fzurita:
Ok, done

gonetz:
Thanks!
Unfortunately, now it just crashes again when threaded option enabled.
I fixed some compilation errors, tabulation and made other minor correction before test.
I don't want to have it lost, so I pushed your commit with my corrections into `fzurita-threaded_GLideN64_v2` branch. Please fix the crash using that branch.

Note: I built it with Visual Studio 2013. May be the crash is related to this version of compiler. I'll test with other compilers later.

fzurita:
You are right, something I did recently is causing windows to crash. I'll see what's up.

fzurita:
Ok, figured it out. Fix coming in a sec.

fzurita:
Sorry about that, fixed.

gonetz:
@fzurita Just built from rev. 85cd640 with VS 2015. mupen64plus build still crashes :(
Besides: "I don't want to have that work lost, so I pushed your commit with my corrections into [fzurita-threaded_GLideN64_v2 branch](https://github.com/gonetz/GLideN64/tree/fzurita-threaded_GLideN64_v2). **Please fix the crash using that branch**." 
Please do it. I don't want to spend another hour fixing these issues.

fzurita:
Ok, I will take a look at your branch. I was able to run without issue this morning in windows.

gonetz:
Project64 works just fine.
I tested it with debug build as well to be sure that threaded_wrapper is used.

fzurita:
I'm confused, so it is working now?

gonetz:
mupen64plus build crashes with threaded enabled.

fzurita:
In your branch change this 1024*1024*1 to 1024*1024*5 here:
https://github.com/gonetz/GLideN64/commit/5f223abf8de005e4f1882f45dd1fcbd36f3591d0#diff-02d09eae392f4fb086f8e216121f88a4R9

Maybe that's the issue. I have only tried mupen64plus in Android, not in Windows. I'll have to check it out.

gonetz:
> In your branch change this 1024_1024_1 to 1024_1024_5 here:

Done: 3e394849dc

fzurita:
Any improvement?

gonetz:
You have that fix in your 85cd64030 commit and it crashes. Of course it still crashes after I added it to my local branch.

fzurita:
Ok, I understand. I will debug mupen64plus for Windows tonight.

fzurita:
@gonetz Are you using m64p? I just tried it and I can reproduce the crash there. It's happening because m64p is doing this:

https://github.com/m64p/mupen64plus-gui/blob/master/vidext.cpp#L42

It seems to be using the QT rendering thread to create the OpenGL context. @loganmc10 is that correct?

I'm not sure how I would work around that QT issue since it's trying to own the OpenGL context.

fzurita:
I found someone who is trying to do what I'm trying to do here: https://stackoverflow.com/questions/24894635/qt-managing-opengl-context-in-a-separate-thread. @loganmc10 any way you could help?

loganmc10:
Have you tried it with vanilla mupen64plus (mupen64plus-ui-console)?

In m64p, the game (including the OpenGL context) are created in their own thread, seperate from the main Qt thread. This allows th GUI to remain responsive while the game is running.

I haven't looked too carefully through the code in this pull request, but typically you can not run OpenGL in multiple threads. You need to call *makeCurrent (depends on the platform) in the thread that you want to run OpenGL calls, before you do any OpenGL work. And *makeCurrent was called in the thread that started the game, not the extra thread that was created for this work.

I'm kind of surprised it works in Windows and Android without any kind of magic with the makeCurrent calls. I'd be surprised as well if it worked in mupen64plus-ui-console and not m64p

gonetz:
> @gonetz Are you using m64p?

Yes.

> @gonetz If it's not easy to make work with m64p, can we still merge with the caveat that it's not compatible with that way of creating the OpenGL context?

Well, threaded version works ok with mupen64plus when threaded disabled. Of course, it would be better to have some compatibility check. Is it possible to test that we can use threaded API calls before enabling that mode?

fzurita:
@loganmc10 makeCurrent is being called for Android here:

https://github.com/mupen64plus-ae/mupen64plus-ae/blob/master/mupen64plus-ui-console/ae-bridge/ae_vidext.cpp#L111

This call is happening in the same thread as the rendering thread. In m64p, it's the same, the makeCurrent call is happening in the same thread as the rendering thread, but you are using QT api calls to do it. I think it must be doing something strange.

The previous version of this has worked fine with just ui-console as tested by someone who has a raspberry pi, the only difference is that this version is not efficient and has less impact.

@gonetz I have to think of a way to determine if the call was successful and potentially fallback to non-threaded mode.

fzurita:
@loganmc10 can you check why it's crashing here https://github.com/m64p/mupen64plus-gui/blob/master/vidext.cpp#L42 When reading is enabled?

I don't have an easy way to compile and debug m64p.

loganmc10:
Yes it seems like Qt handles threads a bit differently, see:

https://doc.qt.io/qt-5/qopenglcontext.html#thread-affinity

> QOpenGLContext can be moved to a different thread with moveToThread(). Do not call makeCurrent() from a different thread than the one to which the QOpenGLContext object belongs

loganmc10:
It's almost certainly crashing because SetVideoMode is called from inside the new thread, and that includes a call to MakeCurrent before MoveToThread was called specifying the new thread

fzurita:
Ok, that sounds like the exact problem. Any way you can easily fix on the interest of getting this to work with m64p?

loganmc10:
I'll make a note to test. Maybe there is a way inside SetVideoMode to figure out what thread it's currently 'in" and call MoveToThread before calling MakeCurrent

fzurita:
Potentially something like ``` MoveToThread(QThread::currentThread()) ``` ?

gonetz:
@fzurita your code does not compile on Linux: 
https://travis-ci.org/gonetz/GLideN64/jobs/507925683

fzurita:
@gonetz please try again.

loganmc10:
I tried to compile this with MinGW and get these errors during linking:

```
msys/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: CMakeFiles/mupen64plus-video-GLideN64.dir/objects.a(opengl_Command.cpp.obj):opengl_Command.cpp:(.text.startup+0x11): undefined reference to `opengl::RingBufferPool::RingBufferPool(unsigned long long)'
msys/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: CMakeFiles/mupen64plus-video-GLideN64.dir/objects.a(opengl_Wrapper.cpp.obj):opengl_Wrapper.cpp:(.text+0x3ec6): undefined reference to `opengl::RingBufferPool::createPoolBuffer(char const*, unsigned long long)'
msys/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: CMakeFiles/mupen64plus-video-GLideN64.dir/objects.a(opengl_Wrapper.cpp.obj):opengl_Wrapper.cpp:(.text+0x3f5f): undefined reference to `opengl::PoolBufferPointer::operator=(opengl::PoolBufferPointer const&)'
```
there are many more of these

gonetz:
> please try again.

https://travis-ci.org/gonetz/GLideN64/jobs/508335098

fzurita:
Yeah, I can see it's still failing. I'll have to check this compilation errors later.

fzurita:
@loganmc10 It seems like I need to add a few cpp files to the CMake build script.

fzurita:
@loganmc10 Please try again with minigw. I had a missing cpp file in the CMakeLists

fzurita:
@gonetz I reset my branch to your branch. Also, linux build seems fixed now, untested though.

fzurita:
Ok, looks like it's passing all the build tests now.

fzurita:
@dankcushions You tested the previous implementation of this. Any way you could test this in the RPI to make sure it still works correctly after all the changes?

loganmc10:
@fzurita it compiles in MinGW now, but I ran into another problem with MoveToThread:

> This function is not thread-safe; the current thread must be same as the current thread affinity. In other words, this function can only "push" an object from the current thread to another thread, it cannot "pull" an object from any arbitrary thread to the current thread.

I can't call MoveToThread from SetVideoMode, since that is running inside your thread, it has to be "pushed" from the main thread into your thread. I can't think of a way off the top of my head to do this besides maybe extending the VidExt API that mupen64plus has with a new function like ```VidExt_setThread(thread_handle)``` where you can pass the handle of the rendering thread (and that VidExt_setThread() function would need to be run from the main thread)

fzurita:
I can potentially call CoreVideo_Init within the main thread instead of the rendering thread. I was only calling it within the rendering thread to be consistent with other calls, but there are no opengl calls being made there, so it could be called within the main thread.

You could call MoveToThread there. But you will still need a thread handle, which I have no idea how to pass to vidext.

loganmc10:
I'll try some things tonight, I think I may have a way to make it work. When CoreVideo_Init is called, I can grab the handle of the current thread, and then use that thread when creating the OpenGL context, I think that will work.

fzurita:
As currently coded, CoreVideo_Init is happening in the rendering thread. What you say will only work if I move that call to the main thread instead. Is that something you can do just to try out, or do you want me to update this pull request to do that?

loganmc10:
I think I got it working, I've attached an updated mupen64plus-gui.exe (64-bit windows) if you want to try it out, seems to work with threaded video.

When CoreVideo_Init is called (inside your rendering thread), I grab the handle of the thread. Then when SetVideoMode is called, the main OpenGL window is created in the main Qt thread (that's how Qt works), when it is created I call MoveToThread and point it to your thread.

[mupen64plus-gui.zip](https://github.com/gonetz/GLideN64/files/2985361/mupen64plus-gui.zip)

fzurita:
Ok, cool, interesting. I'll try it out. Do you see any performance impact? I suspect it will depend on what drivers you have.

loganmc10:
I'll do some testing, here is the commit that made it work if you're interested:

https://github.com/m64p/mupen64plus-gui/commit/c9ff50e36767f40f6be18de0050aefa782163319

fzurita:
Ok, I see what you did.

loganmc10:
Testing a few games on my Windows laptop (intel card), it's hard to say if it's any faster. It looks like it may be a little faster using threaded mode, but I am testing by unlocking the frame limiter and watching the VI/S. The number jumps around so much it's hard to be sure. It certainly doesn't seem to hurt performance, but it would probably be a modest increase if there is one.

Like you mentioned, counting total number of frames processed within a given time period would be much more accurate way of testing.

fzurita:
Try using operation that we know will cause the GPU to slow down, like async or sync color copies.

loganmc10:
Still not noticing much of a difference (tried increasing rendering resolution, toggling sync copies and N64 depth compare), but like you say, it may depend on the drivers.

fzurita:
Yeah, it may not be much of a difference in Windows since newer drivers are generally well optimized. In Android, where drivers are terrible, the difference is fairly substantial, so at least it helps that. It may help low power Linux devices as well where drivers are not as good as desktop.

fzurita:
@loganmc10 I tried your build, but I'm not having much luck. It's taking about 20 seconds to load a game, and it just crashes after it's done loading the shader cache. There are no issues if I disable threaded mode.

loganmc10:
I got it to work in Windows, it may be that I can't just share the mupen64plus-gui.exe, the assosicated Qt dll files may have been changed as well, I'll post a complete build in a bit.

I tested in Windows earlier, and now I just tested on Linux. I can say on Linux that the "Mario Head" scene at the beginning of Super Mario 64 is consistently 30 VI/S faster using the threaded video mode (260 vs 230). This was again using a laptop with an Intel card. So it's probably either that the threading optimizations don't provide the same performance benefits on Windows, or the GL driver behaves differently.

fzurita:
Ok, on my phone, looking at the fps counter

Mario head:
No threading: ~120fps (around 240VI/s)
Threading: ~150fps (around 300VI/s)

Inside peaches Castle, looking at the stairs at the entrance:
No threading: ~125fps
Threading: ~180fps

fzurita:
It's odd though, on my PC, m64p seems to run shower then on my phone. Why is that? Are you using interpreter or CXD4 by default?

loganmc10:
No it uses the x64 new dynarec by default, and only includes the HLE RSP plugin. I'm not sure why it would be slower. I would compare it to PJ64 on PC. GLideN64 probably has more intensive defaults on PC than you use for Android (buffer swap every VI for example)

One other issue I noticed in m64p using your branch (using threaded or non-threaded video): If I try to open a 2nd game (stop 1st game and start a new one), the emulator crashes. Doesn't happen with the current master. I'll try to get a backtrace if I can.

gonetz:
> @gonetz I reset my branch to your branch. 

I force pushed your changes as one commit to  fzurita-threaded_GLideN64_v2 branch.

fzurita:
@loganmc10 Sounds like something is not being cleaned up correctly.

fzurita:
@loganmc10 Any luck with the backtrace? Also, I'm surprised it affect non-threaded mode, that should behave identically to master.

fzurita:
One additional change to make the RingBufferPool more thread safe in the case where it runs out of memory.

fzurita:
@loganmc10 I was thinking about the crash in m64p in Linux when opening a second game.

Each time the plugin is initialized, a new thread is created, is it possible that you are not updating the rendering thread id the second time the plugin is initialized?

loganmc10:
The rendering thread is set during the VidExt_Init() call, which should happen everytime ROMOpen is called. Plus it also crashed when using non-threaded mode. In hoping to get some time to look at this this weekend.

Usually this kind of thing happens when a global/static variable isn't properly reset each time a new game starts. Even though a static variable might get initialized to 0 the first time, the "life of the plugin" can span multiple game/ROM sessions.

In hoping it isn't too hard to track down, sometimes these issues disappear in debug mode because of the optimization differences

fzurita:
So I do have a lot of static data not reset when the plugin starts. I will go ahead and reset all of it on initialization of the plugin.

loganmc10:
My apologies, it seems like the crashing is related to the new dynarec, not your modifications, I opened an issue with mupen64plus-core (https://github.com/mupen64plus/mupen64plus-core/issues/640). It was very inconsistent to reproduce so I was looking into the wrong thing

fzurita:
Ok. I have experienced similar things in Android. That's the reason I automatically kill the app when a game is exited.

fzurita:
Rebasing with master

fzurita:
I did some testing in the Nexus Player. I'm seeing a 78% performance increase with threaded mode.

dankcushions:
I haven't forgotten this :) I'll try and test it on the rpi3 B+ tonight. I still don't have a great way of benchmarking differences in performance (I know many games with slowdowns on the pi3, but it's not always clear what bottleneck is at play)

fzurita:
@dankcushions you can disable the frame limiter and enable the fps display in GLideN64.

fzurita:
Some additional fixes, added GL_LUMINANCE as a valid color format, aligned the Ring Buffer to 4 bytes, it turns out that some devices don't like OpenGL commands with data that is not 4 byte aligned.

fzurita:
Added fix for https://github.com/gonetz/GLideN64/issues/2029.

The fix is to increase the buffered drawer buffer size from 4MB to 8MB.

fzurita:
There seems to be some kind of issue with ```GlDrawElementsUnbufferedCommand``` that seems to manifest only sometimes. I'm trying to debug.

fzurita:
Figured out the problem... In cases where the OpenGL command queue gets rather big due to the GPU driver falling behind, the device is literally running out of memory.  ```GlDrawElementsUnbufferedCommand``` just happens to be the command that is called the most. I'm going to have to figure out a way to limit it.

fzurita:
Ok, all known issues should be fixed.

src/Graphics/OpenGLContext/ThreadedOpenGl/atomicops.h
```diff
@@ -0,0 +1,675 @@
+﻿// ©2013-2016 Cameron Desrochers.
+// Distributed under the simplified BSD license (see the license file that
+// should have come with this header).
```
gonetz: Should we have a copy of BSD license file in our repo?

fzurita:
Ok, I can pull it in.

fzurita:
The license file has been added for that library.

gonetz:
Thanks! I think it is ready for merge now.
Great work, Francisco!

fzurita:
Great, thanks! :-)

dankcushions:
I'm trying to get some real benchmarking data on the raspberry pi 3 for this one so I can justify turning it on by default for everyone in RetroPie, however i can't seem to get --nospeedlimit working via mupen64plus. we don't use a GUI so it's sent as a command line argument. i am using the dummy audio plugin, VerticalSync is False in my mupen64plus.cfg, but still the games will never exceed 60VI/sec, even in title screens (eg Mario Kart 64), or staring at a wall in goldeneye.

If I switch to gles2n64 video plugin, the speed of MK64 is around double.

Weirdly, if I try the latest beta (ie, on their respective github releases pages) versions of mupen64plus and gliden64 for windows, it runs without a speed limit.

Either this is broken in the latest master, or there's something strange going on with RPI ?

fzurita:
It could be vertical sync being enabled somehow.

dankcushions:
I built with the verbose flag, hoping that might shed some light:
```
2019/06/28,15:06:39.564,mupen64plus_DisplayWindow.cpp:53,VERBOSE, "[gles2GlideN64]: _setAttributes"
2019/06/28,15:06:39.614,mupen64plus_DisplayWindow.cpp:97,VERBOSE, "[gles2GlideN64]: Create setting videomode 320x240"
2019/06/28,15:06:39.615,opengl_GLInfo.cpp:27,VERBOSE, "OpenGL ES major version: 2"
2019/06/28,15:06:40.510,opengl_GLInfo.cpp:28,VERBOSE, "OpenGL ES minor version: 0"
2019/06/28,15:06:40.510,opengl_GLInfo.cpp:31,VERBOSE, "OpenGL vendor: Broadcom"
2019/06/28,15:06:40.511,opengl_GLInfo.cpp:50,VERBOSE, "OpenGL renderer: VideoCore IV HW"
2019/06/28,15:06:40.512,opengl_GLInfo.cpp:130,WARNING, "Async color buffer copies are not supported on GLES2"
2019/06/28,15:06:40.512,opengl_GLInfo.cpp:136,WARNING, "LOD emulation not possible on this device"
2019/06/28,15:06:40.523,TextDrawer.cpp:191,VERBOSE, "Generated a 747 x 14 (10 kb) texture atlas"
2019/06/28,15:06:47.936,GBI.cpp:372,VERBOSE, "Load microcode type: 0 crc: 0x6932365f romname: SUPER MARIO 64"
```
not very useful, although i can confirm at least that

a) the --nospeedlimit is being respected at the mupen64plus, as all of the 'waiting' messages have dissappeared from the mupen64plus consoleoutput

and b) that on my pi vsync is always on - i switched the display mode to 1080 24Hz, and mariokart runs at 24fps. I think the reason i saw speed ups on gles2n64 plugin, even with vsync locked on, is because it had enough headroom to go from 30fps to 60fps, but gliden64 does not.

I'll get my pi fully updated and see if I can do some sort of standalone gl test to see if vsync is functional.

dankcushions:
so i finally figured out my issue! the problem was that VSync was always on because of this issue: https://github.com/gonetz/GLideN64/pull/2070

with that fix, I'm able to exceed vsync with --nospeed limit, so here's some figures for an rpi3B+:

Threaded True:
mario64 title screen: 95-105 VI/S

False:
mario64 title screen: 72-85 VI/S

fzurita:
Great! Sounds like a good improvement.

Issue #2040, opened 2019-04-09
Crash on savestate load with Threaded Video enabled.


gonetz:
Project 64 2.4
GLideN64 current master, default settings + Threaded Video.
Crash on savestate load. Tested with Zeldas, Wave Race and others.

@fzurita please check.

fzurita:
Ok, I will check. I have not experienced this in Android and while testing on my PC.

Also, keep in mind that the threaded version will abort if you compile with GL_DEBUG and any GL errors happen.

fzurita:
I only see the build in the Debug build, not Release build. 

This is in the GLideN64 log:

```
 OpenGL error: 0x502, on function: Shutdown
```

This led to the crash. It's odd though, Shutdown is not a GL function....

fzurita:
@gonetz After disabling the abort on GL_ERROR only for the shutdown function, it still crashes. But this is not a problem with the threading. Here is a back trace as generated by Visual Studio:
```
 	GLideN64.dll!graphics::Context::deleteFramebuffer(graphics::ObjectHandle _name) Line 193	C++
 	GLideN64.dll!TexrectDrawer::destroy() Line 100	C++
 	GLideN64.dll!GraphicsDrawer::_destroyData() Line 1801	C++
 	GLideN64.dll!DisplayWindow::stop() Line 25	C++
 	GLideN64.dll!PluginAPI::RomClosed() Line 189	C++
>	GLideN64.dll!RomClosed() Line 34	C++
 	[External Code]	
```

fzurita:
The problem seems to be that "RomClosed" is getting called twice in a row.

gonetz:
Thanks for the info. Yes, I met this problem with Debug buld. I did not notice, that your code works different when GL_DEBUG enabled.

> The problem seems to be that "RomClosed" is getting called twice in a row.

As I see, the problem is not because of two RomClosed. I set log level to LOG_APIFUNC:
```
ProcessDList
UpdateScreen
RomClosed
 OpenGL error: 0x502, on function: Shutdown
RomClosed

```
That is, the problem rises in the first RomClosed call. Call stack:
```
GLideN64.dll!opengl::OpenGlCommand::performCommandSingleThreaded() Line 23	C++
GLideN64.dll!opengl::FunctionWrapper::executeCommand(std::shared_ptr<opengl::OpenGlCommand> _command) Line 27	C++
GLideN64.dll!opengl::FunctionWrapper::windowsStop() Line 1401	C++
[...]

PR #2041, opened 2019-04-10
Fix issues with threaded OpenGL when debug is enabled

fzurita:
This should fix https://github.com/gonetz/GLideN64/issues/2040 assuming that this was happening with GL_DEBUG enabled.

Also, protect against RomClosed being called more than once.


gonetz:
The problem is fixed, thanks!

PR #2042, opened 2019-04-10
Improve logging

fzurita:
Improves logging so that it looks like this:
```
2019/04/10,00:27:44.279,opengl_glinfo.cpp:27,VERBOSE, "OpenGL major version: 4"
```
Every time the app is restarted, a new log file will be created in this format:
```
20190410_002743_gliden64.log
```

Also, all the log statements that had ```\n``` were updated to not have it since it's automatically added by the logging function.

### DIFF
--- a/src/DebugDump.cpp
+++ b/src/DebugDump.cpp
@@ -35,7 +35,7 @@ BufferedLog::BufferedLog(u32 _mode) : m_mode(_mode)
 		m_logBuffer.resize(1024*1024);
 		m_log.rdbuf()->pubsetbuf(&m_logBuffer.front(), m_logBuffer.size());
 	} catch(std::bad_alloc&) {
-		LOG(LOG_ERROR, "Failed to alloc memory for log buffer\n");
+		LOG(LOG_ERROR, "Failed to alloc memory for log buffer");
 	}
 
 	wchar_t logPath[PLUGIN_PATH_SIZE + 16];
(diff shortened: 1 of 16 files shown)


fzurita:
Also, what do you think about creating a logging thread instead of logging in the same thread as the logging call was made? Writes to disk can be very slow, which halts the main thread.

The one disadvantage I can think of is that if the application crashes, some logging could be lost. This can be worked around by adding a signal handler, but that's platform specific.

fzurita:
Hmm.... wchar issues again, hopefully what I did is alright in regards to that.

gonetz:
> Also, what do you think about creating a logging thread instead of logging in the same thread as the logging call was made? Writes to disk can be very slow, which halts the main thread.

I don't see such necessity. Plugin writes nothing to log if everything goes right. Few lines of information are not expensive to write. If something goes wrong, logging performance is the least matter which troubles user at this moment.

gonetz:
Conflicts after previous merge.

fzurita:
The conflict has been resolved.

Also, it seems that ```std::put_time``` is not implemented in gcc 4.8.4, but it is in 5.0+. Can we change travis-ci to use gcc 5.0+?

fzurita:
I tried to updated the .travis.yml, but if I make any change, it doesn't seem to run, so I can't verify if it works. I did try to use this example to change the gcc version to a higher one: https://docs.travis-ci.com/user/languages/cpp/

gonetz:
@fzurita I switched to Xenial and that fixed the problem with gcc.

I have one additional request to this fix. New code creates new log file on every rom run.
I can't say that I'm happy with that. I mostly prefer old behavior where log appended to existing file or overwrites existing file. Please return old file creation code under some preprocessor define, and enable that define for Visual Studio projects.

fzurita:
Ok, I will change that behavior.

fzurita:
I couldn't think of an easy way to not make a new file after starting a new game while keeping the date on it, so I just got rid of the date on the file name.

Issue #2037, opened 2019-04-07
WIP Builds 4


gonetz:
Thread for WIP builds links. No feedback in that thread please.

gonetz:
Current master + Threaded mode by fzurita:
https://drive.google.com/file/d/1a9IFzTfsVNPHR6tEx-5-hqvutLg3JVQC/view?usp=sharing

dezraj:
Hello gonetz! Are there any new wip builds?

gonetz:
Not yet. You may find experimental builds here: #1885

gonetz:
Master + file tex storage + refreshed GUI:
https://drive.google.com/file/d/1ZmVmiiEuCrY22-5NkqCtAh7B1FyK-1AZ/view?usp=sharing

DonelBueno:
@gonetz What's the difference between this build and HD fix 7-2, on #1885?

gonetz:
@DonelBueno added file tex storage option; GUI refreshed.

dezraj:
gonetz, the information tips for antialiasing are not showing in this version.

JustNoon:
Can you explain what the difference between "Use file storage instead of memory cache", "Use file storage instead of memory cache", and "Save enhanced texture cache to hard disk" are? 

I thought this was a "disable hires texture preloading" kind of feature but I seem to be mistaken (and possibly bugged. I'll make a full bug report since it's a full WIP Build now.)

gonetz:
@JustNoon I tried to explain it in #2079
In short: "Use file storage instead of memory cache" enables texture file storage, which loads textures from HDD. Memory cache keeps all textures in RAM. This option can be set separately for enhanced textures and for hi-res ones. "Save enhanced texture cache to hard disk" saves memory/file cache to disk. Without it, the cache will be dropped after rom closed.

gonetz:
@DonelBueno I'm close to completion of #1548. I'll make new WIP build when it will be finished.

gonetz:
Master + file tex storage + refreshed GUI + depth copy option:
https://drive.google.com/file/d/15IFiEiLesOb7Ad5F8FxdWkqc0wdGXe6E/view?usp=sharing

Depth copy is for Reshade, see #1548
[...]

PR #2067, opened 2019-07-06
UI refresh

Aquatakat:
I've adjusted the main page a bit:

image

Profiles changed to avoid the ambiguous state between typing a name and clicking the Add button:

image

New texture enhancements page allows copy/pasting a path without opening a new window:

image

And a bunch of other little changes. I hope you like it!

### DIFF
--- a/src/GLideNUI/ConfigDialog.cpp
+++ b/src/GLideNUI/ConfigDialog.cpp
@@ -375,8 +375,8 @@ void ConfigDialog::setRomName(const char * _romName)
 {
 	const bool bRomNameIsEmpty = _romName == nullptr || strlen(_romName) == 0;
 	m_romName = bRomNameIsEmpty ? nullptr : _romName;
-	ui->profilesFrame->setEnabled(bRomNameIsEmpty);
-	ui->customSettingsWarningFrame->setVisible(!bRomNameIsEmpty && config.generalEmulation.enableCustomSettings != 0);
+	ui->customSettingsCheckBox->toggle();
+	ui->customSettingsCheckBox->setChecked(config.generalEmulation.enableCustomSettings != 0);
 }
 
 ConfigDialog::ConfigDialog(QWidget *parent, Qt::WindowFlags f) :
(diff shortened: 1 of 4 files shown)


CrankySupertoon:
Dont Pull Request in Master, other then that i think its good

gonetz:
@Aquatakat 
Your modifications look good, as usual :) Thanks!
I have few notes about AA-related parameters.
* You made AA mode selection via radio buttons.
They are working not quite consistent.
If I have MSAA enabled and switch to No AA, MSAA reset to 'off'
If I have MSAA enabled and switch to FXAA, MSAA does not reset.
Why? FXAA also means that MSAA will be disabled.
* MSAA scale is from 'off' to 'high'. When user switches from No AA to MSAA, MSAA slider set to 'off' and user need to do another click to select necessary AA. I suggest to automatically set MSAA to "high" level when MSAA selected. Most users prefer max MSAA level. What do you think?

Aquatakat:
That's probably a good idea @gonetz, I think the only reason I have those behave the way they do is because that's how the settings ini is set up. Leaving MSAA at max when the no aa/FXAA are checked is probably a good idea, I'll work on that in a few days.

gonetz:
AA works good with latest commit, thanks!
I have two questions:
* You moved "Store compiled shaders" options at the bottom. Why? It looks alone here
* Configuration panel has plenty of blank space at the bottom, at least with my build. Can it be shrunk?
StoreShaders

Aquatakat:
@gonetz, I just wanted to let you know I'm still working on this. I wanted to add a feature to allow a user to use per-game settings but only update a settings profile while a game is running.

image

It works well but some of the behaviour needs a little adjusting.

Also file_tex_storage is supported:

image

gonetz:
@Aquatakat Hi Ryan,

I put your work to https://github.com/gonetz/GLideN64/tree/file_tex_starage+new_gui branch.
Test build available in WIP Builds 4, #2037
Imo, the new GUI looks good. "Texture enhancement" tab is more convenient to use now.
There is one feedback note https://github.com/gonetz/GLideN64/issues/2037#issuecomment-520398657
that information tips for antialiasing options is missing.

gonetz:
@Aquatakat Thanks!
The changes are in master now.

PR #2070, opened 2019-07-12
M64P_GL_SWAP_CONTROL (and others?) requires a context, so needs to be…

dankcushions:
… after CoreVideo_SetVideoMode (which gets one)

Otherwise the plugin effectively ignores the VerticalSync mupen64plus option - it's always on.

I don't know if there's any consequences of setting these other attributes after this point, but it seems to run fine in my testing.

### DIFF
--- a/src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
+++ b/src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
@@ -78,7 +78,6 @@ bool DisplayWindowMupen64plus::_start()
 	FunctionWrapper::setThreadedMode(config.video.threadedVideo);
 
 	FunctionWrapper::CoreVideo_Init();
-	_setAttributes();
 
 	m_bFullscreen = config.video.fullscreen > 0;
 	m_screenWidth = config.video.windowedWidth;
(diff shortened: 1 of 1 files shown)


loganmc10:
Actually this is incorrect, according to the mupen64plus docs (https://github.com/mupen64plus/mupen64plus-core/blob/master/doc/emuwiki-api-doc/Mupen64Plus-v2.0-Core-Video-Extension.mediawiki) for VidExt_SetVideoMode:

> This function creates a rendering window or switches into a fullscreen video mode. Any desired OpenGL attributes should be set before calling this function. 

Also under VidExt_GL_SetAttribute:
> This function is used to set certain OpenGL attributes which must be specified before creating the rendering window with VidExt_SetVideoMode. 

Basically the frontend application should be tracking what you ask for when you call VidExt_GL_SetAttribute, and when VidExt_SetVideoMode is called, it should create a proper context based on the attributes you requested.

dankcushions:
i think the issue might be that the SWAP_CONTROL is not really an attribute in SDL, like these others are. it needs to be set AFTER the window/context, but the rest before: http://forums.libsdl.org/viewtopic.php?p=43308

it seems like it shouldn't be part of VidExt_GL_SetAttribute at all, and instead part of VidExt_SetVideoMode

dankcushions:
anyway, will close as GLideN64 is respecting the API spec, but any input welcome :)

loganmc10:
Yeah it does seem like a bit of a bug/strange implementation in SDL. Qt allows you to set the vsync mode before creating the GL context.

This could definitely be fixed in mupen64plus-core though:
https://github.com/mupen64plus/mupen64plus-core/blob/master/src/api/vidext.c#L429-L434

instead of trying to call SDL_GL_SetSwapInterval during VidExt_GL_SetAttribute, you could just set a variable to "true" and then call SDL_GL_SetSwapInterval in VidExt_SetVideoMode based on whatever the variable was set to

PR #2091, opened 2019-09-28
Fix android logging

fzurita:
It looks like when I updated the logging in Windows, I forgot to back and fix it for Android.


gonetz:
500 pull request!

fzurita:
Great!

Issue #2087, opened 2019-09-15
GlideN64 on gles 3.0 on rpi4


Askmewho:
Hi, its possible to run this plugin on gles 3.0 under x11?  because ive tried with egl on and with SDL2 (with full gl profile and it gives a segfault as expected) but with a non full profile gl SDL2 it claims than it needs glx profile.

Askmewho:
@gonetz @gizmo98 we chat this topic before I believe, Ive compiled every single mupen module + sdl2 to not use opengl, but rather gles, and Ive get a segfault (which I will debug if you need so, but its a bit difficult because I use a python frontend) . At least it doesn't say than it cant create gl context.

m4xw:
@Askmewho Fixed in https://github.com/libretro/mupen64plus-libretro-nx/commit/3dae33e95cb4d860cdc45dcb6227b5c8f5f69197
You can backport it.

Askmewho:
fixed with that! push it!

PR #2092, opened 2019-09-29
Fix gles2

fzurita:
This fixes various graphical issues in GLSL for GLES 2.0. We don't have guaranteed highp precision there for floating point numbers, which means we can't do good math with floating point numbers with significantly different values.

A value of 10.0 for wrap when there is no wrapping seems reasonable to me.

Also, a potential fix for failure to get function pointers on X windows.


fzurita:
@gonetz You think this is a good idea?
```
"highp vec2 clampWrapMirror(in highp vec2 vTexCoord, in highp vec4 vClamp,		\n"
			"	in highp vec2 vWrap, in lowp vec2 vMirror, in highp vec2 vScale)			\n"
			"{																				\n"
			"  highp vec2 texCoord = clamp(vTexCoord, vClamp.xy, vClamp.zw);				\n"
			"  lowp vec2 one = vec2(1.0);													\n"
			"  lowp vec2 clamped = step(vClamp.zw, texCoord);								\n"
			"  lowp vec2 notClamped = one - clamped;										\n"
			"  texCoord = clamped * texCoord + notClamped * mod(texCoord, vWrap);			\n"

		"  lowp vec2 wrapped = step(vWrap , texCoord);								\n"
			"  lowp vec2 notWrapped = one - wrapped;										\n"
			"  texCoord = clamped * texCoord + notClamped *(wrapped*mod(texCoord, vWrap) + notWrapped*texCoord);			\n"

			"  highp vec2 intPart = floor(texCoord);															\n"
			"  highp vec2 fractPart = fract(texCoord);								\n"
			"  lowp vec2 needMirror = step(vec2(0.5), mod(intPart, vec2(2.0))) * vMirror;	\n"
			"  texCoord = clamped * texCoord + notClamped * fractPart;						\n"
			"  texCoord = (one - vMirror) * texCoord + vMirror * fractPart;					\n"
			"  texCoord = (one - texCoord) * needMirror + texCoord * (one - needMirror);	\n"
			"  texCoord *= vScale;															\n"
			"  return texCoord;																\n"
			"}																				\n"
			;
	}
```

I think it may avoid the floating point math issue since the mod operation will hey zeroed out if it's not necessary. Although, I should probably try it.

m4xw:
@fzurita Wanna backport this while you are at it?
https://github.com/libretro/GLideN64/commit/4064af53adb88966a822f6be91a080e2180d9065
Would fix #2087, however you know the eglInitialize placement better for upstreaming it, don't want it to conflict with stuff you do in FZ, potentially.
The queryString + getString sanity checks seem widely used.

fzurita:
Yeah, sure, I can take a look at that.

m4xw:
Also thanks for working on the fixes, very appreciated!

fzurita:
No problem!

src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
```diff
@@ -77,6 +77,22 @@ bool DisplayWindowMupen64plus::_start()
...
+	EGLDisplay display;
+	if ((display = eglGetDisplay(EGL_DEFAULT_DISPLAY)) == EGL_NO_DISPLAY) {
+		LOG(LOG_ERROR, "eglGetDisplay() returned error %d", eglGetError());
+		return false;
+	}
+	if (!eglInitialize(display, nullptr, nullptr)) {
+		LOG(LOG_ERROR, "eglInitialize() returned error %d", eglGetError());
+		return false;
+	}
+	eglInitialize(eglGetDisplay(EGL_DEFAULT_DISPLAY), nullptr, nullptr);
```
m4xw: @fzurita forgot to remove that?

m4xw: Also maybe add [GlideN64] for consistency?

m4xw:
cc @hhromic btw, upstreaming the RPi fix!

gonetz:
@gonetz You think this is a good idea?

Well, texCoord should be in range [0, 1], so having default wrap 10 should be enough.
However, I'd set it to 100 to be sure that it will work for all cases. 
Will mod operation get zeroed with 100 as wrap value?

fzurita:
I did try 100 and it didn't quite fix the issue either, 10 did fx it. The proposed second fix I think may be better since it would avoid the issue completely.

I think if we go with the second approach I proposed, it wouldn't matter what the vWrap value is. We could make it 100 or 100000.

src/Graphics/OpenGLContext/GLFunctions.cpp
```diff
@@ -45,7 +45,7 @@ static void* AppleGLGetProcAddress (const char *name)
 	return (image ? dlsym(image, name) : NULL);
 }
 #define glGetProcAddress AppleGLGetProcAddress
-#define GL_GET_PROC_ADR(proc_type, proc_name) ptr##proc_name = (proc_type) glGetProcAddress(#proc_name)
+#define GL_GET_PROC_ADR(proc_type, proc_name) ptr##proc_name = (proc_type) glGetProcAddress("gl"#proc_name)
```
gonetz: Current code works on MacOS. Why you changed it? Have you tested that change?

fzurita: I have not tested the change, I don't have the right devices to be able to test it. Apparently this fixed Linux though, and the same mistake _appears_ to be in MacOS as well.

I changed it because I see what looks like a mistake. Without the "gl" in front, it looks like it will attempt to get the function name incorrectly. Maybe MacOS is ok with that?

gonetz:
>I think if we go with the second approach I proposed, it wouldn't matter what the vWrap value is. We could make it 100 or 100000.

I did not notice your changes in the shader. The idea looks good. It make the code more reliable indeed. Please update your PR with it.

fzurita:
I'll update it once I test it.

fzurita:
Ok, I'll update the pull request.

src/Graphics/OpenGLContext/GLFunctions.cpp
gonetz: Hmm, yes, it looks wrong without "gl".
However, I have tested my own build of GLideN64 on a Macbook several month ago and it worked.
Ok, lets fix only Linux for now. I'll try to get a Macbook again and test with and without that fix.

fzurita:
@gonetz I attempted the second method for fixing the shader that doesn't rely on floating point precision. It works just as well, and I think it's a better approach. I went ahead and updated the pull request with it instead.

gonetz:
@fzurita Thanks!
I tested the cases with wrapped textures and Nerrel's texture pack. Everything works as before.
I put all but two last commits to master.
Mac and IOS initialization fixes are in fix_mac_init branch.
I want to test fix for Mac on real Mac first.

It seems that your fixes solved many problems on non-desktop devices. Thanks!

fzurita:
No problem!

src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
```diff
@@ -77,6 +77,21 @@ bool DisplayWindowMupen64plus::_start()
...
+	// do this here anyways. Calling eglInitialize has no negative effect according to the spec
+
+	EGLDisplay display;
+	if ((display = eglGetDisplay(EGL_DEFAULT_DISPLAY)) == EGL_NO_DISPLAY) {
+		LOG(LOG_ERROR, "eglGetDisplay() returned error %d", eglGetError());
+		return false;
+	}
+	if (!eglInitialize(display, nullptr, nullptr)) {
+		LOG(LOG_ERROR, "eglInitialize() returned error %d", eglGetError());
+		return false;
```
psyke83: This causes breakage on RPI4 and RPI3 in firmware kms mode, running in a KMS context via SDL2's kmsdrm driver. The plugin segfaults, leaving this message in the gliden64.log:
`2019/10/02,01:28:12.143,mupen64plus_DisplayWindow.cpp:90,ERROR, "eglInitialize() returned error 12289"`

fzurita:
That would tell me that it may not support EGL.

psyke83:
Building without EGL support seemed to work OK on RPI4, but on RPI3, while it does allow the plugin to load, it causes graphical corruption (such as black squares in place of character shadows in Mario 64).

If I build with EGL enabled, but remove the `return false` from your `eglInitialize` check above, it allows the plugin to work as before (without graphical glitches) on RPI3. This may be a quirk that only occurs for KMS configurations. It's late here, so I'll follow this up after I've had more time to test, perhaps in a new PR. Thanks.

fzurita:
Ok, that's interesting. We may need a specific #ifdef just for RPi 3 configurations, or potentially a startup configuration to ignore an invalid eglInitialize return.

m4xw:
@fzurita The fix made was specifically to fix rpi3/4, @hhromic got some insight?

m4xw:
btw @fzurita fwiw I just saw mupen next's fix never checked the return value and we didn't encounter issues
image

fzurita:
Interesting... The error number 12289 means ```EGL_NOT_INITIALIZED```

It sounds like driver quirkiness.

m4xw:
@psyke83 bigbrain strat: just ignore the error and see if it works :P

m4xw:
However I encountered some weird driver issues too with gbm/drm on some other software on Pi4, it will fail to query screens until a reboot sometimes.

fzurita:
I got a Rpi4 coming soon. I should be able to play with it then to see what's wrong.

fzurita:
@psyke83 weekdays installed on your RPi 4?

psyke83:
@fzurita 

Making a RPI3-specific call would not fix the problem as it happens also on RPI4. To re-iterate, the behaviour after this PR was merged:

EGL enabled:
* RPI3 (fkms) & RPI4 both segfault.

EGL disabled:
* RPI3 (fkms) works, but has rendering errors
* RPI4 works with no apparent rendering errors.

Although RPI4 looks to render OK, it's probably not a good idea to build with EGL disabled in this way due to the regression on RPI3. I'm not certain why there's a difference, but keep in mind that the RPI4's VC6 hardware uses a different Mesa driver (V3D vs VC4) and supports GLES3.

The best resolution for both board revisions is to treat the `eglInitialize` call failure as a warning only (which I have already tested and confirmed to resolve the issue, @m4xw). It would probably be a good idea to understand why the call is failing in a KMS context, however.

psyke83:
I suspect that EGL is already initalized by SDL2's internal video driver:
1: https://github.com/spurious/SDL-mirror/blob/77d8ec0dbc894270e35741d65730890394e27a27/src/video/kmsdrm/SDL_kmsdrmopengles.c#L39-L41
2: https://github.com/spurious/SDL-mirror/blob/b40d67ae2117af053688781bfb8fdd68f4767451/src/video/SDL_egl.c#L489-L493

m4xw:
Actually the real question is why the sane failure segfaults and not sanely aborts.

psyke83:
@m4xw,

Yeah, not sure why that's happening either. Here's a backtrace:

```
Starting program: /opt/retropie/emulators/mupen64plus/bin/mupen64plus --noosd --windowed --resolution 1360x768 --rsp mupen64plus-rsp-hle.so --gfx mupen64plus-video-GLideN64.so --audio
 mupen64plus-audio-sdl.so --configdir /opt/retropie/configs/n64 --datadir /opt/retropie/configs/n64 /home/pi/RetroPie/roms/n64/Super\ Mario\ 64\ \(U\)\ \[\!\].z64                     
[Thread debugging using libthread_db enabled]                                                                                                                                          
Using host libthread_db library "/lib/arm-linux-gnueabihf/libthread_db.so.1".                                                                                                          
 __  __                         __   _  _   ____  _                                                                                                                                    
|  \/  |_   _ _ __   ___ _ __  / /_ | || | |  _ \| |_   _ ___                                                                                                                          
| |\/| | | | | '_ \ / _ \ '_ \| '_ \| || |_| |_) | | | | / __|                                                                                                                         
| |  | | |_| | |_) |  __/ | | | (_) |__   _|  __/| | |_| \__ \                                                                                                                         
|_|  |_|\__,_| .__/ \___|_| |_|\___/   |_| |_|   |_|\__,_|___/                                                                                                                         
             |_|         https://mupen64plus.org/                                                                                                                                      
Mupen64Plus Console User-Interface Version 2.5.9                                                                                                                                       
                                                                                                                                                                                       
UI-Console: attached to core library 'Mupen64Plus Core' version 2.5.9                                                                                                                  
UI-Console:             Includes support for Dynamic Recompiler.                                                                                                                       
Core: Using full mem base                                                                                                                                                              
[New Thread 0x8f5db460 (LWP 22694)]                                                                                                                                                    
Core: Goodname: Super Mario 64 (U) [!]                                                                                                                                                 
Core: Name: SUPER MARIO 64                                                                                                                                                             
Core: MD5: 20B854B239203BAF6C961B850A4A51A2                                                                                                                                            
Core: CRC: 635A2BFF 8B022326
Core: Imagetype: .z64 (native)
Core: Rom size: 8388608 bytes (or 8 Mb or 64 Megabits)
Core: Version: 1444
Core: Manufacturer: Nintendo
Core: Country: USA
UI-Console Status: Cheat codes disabled.
UI-Console: using Video plugin: 'GLideN64' v2.0.0
UI-Console: using Audio plugin: 'Mupen64Plus SDL Audio Plugin' v2.5.9
Input: Using auto-config file at: '/opt/retropie/configs/n64/InputAutoCfg.ini'
UI-Console: using Input plugin: 'Mupen64Plus SDL Input Plugin' v2.5.9
UI-Console: using RSP plugin: 'Hacktarux/Azimer High-Level Emulation RSP Plugin' v2.5.9
Core: input plugin did not specify a render callback; there will be no on screen display by the input plugin.
Input: 0 SDL joysticks were found.
Input: N64 Controller #1: Forcing default keyboard configuration
Input: Using auto-config file at: '/opt/retropie/configs/n64/InputAutoCfg.ini'
Input: 1 controller(s) found, 1 plugged in and usable in the emulator
Input Warning: Couldn't open rumble support for joystick #1
Input Warning: Couldn't open rumble support for joystick #2
Input Warning: Couldn't open rumble support for joystick #3
Input Warning: Couldn't open rumble support for joystick #4
Input: Mupen64Plus SDL Input Plugin version 2.5.9 initialized.
RSP: RSP Fallback disabled !
Core: Using video capture backend: dummy
Core: Game controller 0 (Standard controller) has a Memory pak plugged in
Core: Game controller 1 (Standard controller) has a Memory pak plugged in
Core: Game controller 2 (Standard controller) has a Memory pak plugged in
Core: Game controller 3 (Standard controller) has a Memory pak plugged in
Core: Using CIC type X102
[New Thread 0x8d20d460 (LWP 22695)]

Thread 3 "mupen64plus" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x8d20d460 (LWP 22695)]
0x00000000 in ?? ()
(gdb) bt
#0  0x00000000 in ?? ()
#1  0x8e8e1788 in opengl::GlGetStringCommand::commandToExecute() () from /opt/retropie/emulators/mupen64plus/lib/mupen64plus/mupen64plus-video-GLideN64.so                            
#2  0x8e8e6518 in opengl::OpenGlCommand::performCommandSingleThreaded() () from /opt/retropie/emulators/mupen64plus/lib/mupen64plus/mupen64plus-video-GLideN64.so                     
#3  0x8e8d83a0 in opengl::OpenGlCommand::performCommand() () from /opt/retropie/emulators/mupen64plus/lib/mupen64plus/mupen64plus-video-GLideN64.so                                   
#4  0x8e8d8788 in opengl::FunctionWrapper::commandLoop() () from /opt/retropie/emulators/mupen64plus/lib/mupen64plus/mupen64plus-video-GLideN64.so                                    
#5  0xaf7389b0 in ?? () from /usr/lib/arm-linux-gnueabihf/libstdc++.so.6
#6  0xb6e53494 in start_thread (arg=0x8d20d460) at pthread_create.c:486
#7  0xb6dd6578 in ?? () at ../sysdeps/unix/sysv/linux/arm/clone.S:73 from /lib/arm-linux-gnueabihf/libc.so.6                                                                          
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

fzurita:
The specific RPi drivers must not be meeting the EGL spec. According to this:
https://www.google.com/search?q=eglinitialize&rlz=1C1GCEA_enUS842US842&oq=eglinitialize&aqs=chrome..69i57j0l5.4278j1j7&sourceid=chrome&ie=UTF-8

```
Initializing an already initialized EGL display connection has no effect besides returning the version numbers.
```

psyke83:
@fzurita,

No, I'm not convinced it's isolated to RPI's VC4/V3D drivers. As a test, I compiled mupen64plus + GLideN64 for my laptop (Intel Haswell i965) - with EGL enabled via cmake & with a kmsdrm-enabled SDL2 build installed. Dropping to a VT and running mupen64plus so that it runs in a KMS context, it has the same segfault problem and message in the log:
`2019/10/02,17:13:57.592,mupen64plus_DisplayWindow.cpp:90,ERROR, "eglInitialize() returned error 12289"`

The issue is once again resolved by removing the `return false;` call. These errors appear in the log:
`
2019/10/02,17:20:02.328,mupen64plus_DisplayWindow.cpp:90,ERROR, "eglInitialize() returned error 12289"
2019/10/02,17:20:02.428,opengl_Utils.cpp:60,WARNING, "Could not query EGL extensions on this device"
2019/10/02,17:20:02.428,opengl_Utils.cpp:60,WARNING, "Could not query EGL extensions on this device"
`
So I would suggest that this is an odd interaction with SDL2's kmsdrm driver.

fzurita:
It sounds like we should be following this guide when running under KMS:

http://virtuousgeek.org/blog/index.php/jbarnes/2011/10/31/writing_stanalone_programs_with_egl_and_

We are definitely not doing all those things.

psyke83:
But the plugin is not standalone, in the sense that it's using SDL2 to create the EGL context in the first place.

Perhaps we shouldn't be trying to initialize EGL ourselves? As for EGL extension queries, there's a `SDL_EGL_HasExtension` function that might work in in place of querying the driver directly.

fzurita:
I would hate top bring SDL2 into the plug-in. I agree, we shouldn't be initializing it ourselves. That is what the VidExt mupen API is for. The "fix"is more like a hack to help libretro. This is how wet so it in Android where we do implement the API:

https://github.com/mupen64plus-ae/mupen64plus-ae/blob/master/mupen64plus-ui-console/ae-bridge/ae_vidext.cpp

The VidExt needs to be implement for KMS in my opinion. At least I think it wouldn't be hacky.

m4xw:
Yea it doesn't make sense tho from libretro's perspective.
However I know the Issue happens with some standalones too, with gliden as plugin, there without the NULL check it will crash, however the pi's supports a bunch of extensions and eglInitialize fixed it there.
So something would be broken there even with the initalize removed.

@hhromic knows better, lets let explain him.

m4xw:
@fzurita check this pr & comments https://github.com/libretro/mupen64plus-libretro-nx/pull/82#issuecomment-518326937

fzurita:
Yes, I agree, it will have the same issue. But with standalone, we can fix this issue by using the VidExt API fairly easily.

Can we do the same with libretro as well?

psyke83:
BTW, I'm the person that @hhromic was referencing in that message. It was previously necessary to bypass the extensions check, but this PR did resolve that particular crash (once you bypass the segfault caused by the `eglInitialize` fail).

m4xw:
@psyke83 i c.
@fzurita i think we should check why it crashes.
It shouldn't crash in any case if eglInitialize fails with a valid error, just abort/stop.
Right now this fixes it if we just do it without the result check, but I guess the result cant be differentiated from real errors..

m4xw:
(Also theres more reason why I don't use the VidExt stuff and the design choices, but I don't have time right now to explain, you should poke me later on that)

m4xw:
I should note tho, it's no problem for me if you want that in VidExt, I just wanted the fix upstreamed because it happens in standalone too.
The whole core is basically about static integration of the (multi-)plugins while also keeping easy rebase-ability via shallow forks.

fzurita:
Yes, I think we can be more graceful. We should definitely have a graceful shutdown without crashing if initialization fails. We don't want the whole frontend to crash, that's annoying.

We can also fix the hack so that it doesn't fail, but we should still fix the crashing when initialization fails anyways.

In the end, I do think eglInitialize should be moved to the front end using the VidExt API since that's how it was designed to be done by the original mupen64plus authors.

m4xw:
> In the end, I do think eglInitialize should be moved to the front end using the VidExt API since that's how it was designed to be done by the original mupen64plus authors.

@fzurita That's so last year ;)

hhromic:
Hi guys, sorry to be late to the party.
First of all, I want to clarify that I'm by no means an expert on EGL/GBM/KMS/DRM/etc. So take my comments with a grain of salt :)

However I see there is a bit of confusion and some context missing. Everything started when the plugin started to crash because of `eglQueryString()` returning `nullptr`. The reason for this is still unknown. I believe it has to do with `eglGetDisplay(EGL_DEFAULT_DISPLAY)` not always returning the correct display, i.e. in RPI with KMS/DRM drivers, but again I'm no expert on the subject.

Then at some point it was stated that a call to `eglInitialize()` might be required (https://github.com/libretro/mupen64plus-libretro-nx/pull/92#issuecomment-528243306). However after testing it was confirmed that actually `eglInitialize()` doesn't solve the problem either (https://github.com/libretro/mupen64plus-libretro-nx/pull/92#issuecomment-528602449).

However, later @fzurita commented that `eglInitialize()` is a required call (https://github.com/libretro/mupen64plus-libretro-nx/pull/92#issuecomment-529004283), and here is where the wires got crossed imho. For **Android**, they do call `eglInitialize()` but properly in VidExt Init.

Then, after discussing with @m4xw in Discord, we came to the conclusion that the safest approach was to add `eglInitialize()` *somehow* and, in the **libretro** context, the best place seemed to be DisplayWindow (we know is not the best place). Furthermore, I didn't guard the call with an error-check effectively making it "no-op" in case of errors. Also it doesn't segfault in the libretro version.

So, to summarise, @fzurita implemented my PR https://github.com/libretro/mupen64plus-libretro-nx/pull/94 for the libretro version here in standalone (with error-checks) but it happens to segfault in standalone. Having said all that, I'm not sure at all if the call to `eglInitialize()` should be made here in **standalone** (or even in the libretro version). And definitively not in DisplayWindow.

So my non-expert opinion would be to simply not call `eglInitialize()` here, specially considering it didn't seem to be necessary all this time.

m4xw:
@hhromic thanks, I missremembered a bunch, however I told you to use displaywindow because our vidext is just  a stub and I don't even call it there.
It's only in place to not cause compiler errors :P
https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/custom/mupen64plus-core/api/vidext_libretro.c
I'd prefer dropping it in my fork because I don't like it for the static design I use by choice and it's counterproductive for the use-case.
It just really doesn't make sense to me with the integration I do, i would have to maintain a file anyway since I can't use mupen's (way too bloated for libretro purposes, we don't use SDL2 either).

hhromic:
@m4xw that's fine I think.

The important point is that we are not really sure in which contexts the `eglInitialize()` call is really necessary. For example in the libretro case, RetroArch already initialises EGL. For the standalone case, SDL2 already initialises EGL too. Seems that for Android @fzurita already does it in VidExt Init as well.

So given the fact that it wasn't necessary before, imho we should keep it out for now to fix this PR.

psyke83:
To re-iterate what I mentioned in a previous post, `SDL_EGL_LoadLibrary` will call `eglInitialize` [itself](https://github.com/spurious/SDL-mirror/blob/b40d67ae2117af053688781bfb8fdd68f4767451/src/video/SDL_egl.c#L489-L493). This function is called as part of most if not all the major SDL2 internal video drivers that support EGL context creation:
```
$ git grep SDL_EGL_LoadLibrary src/video/*/
src/video/android/SDL_androidgl.c:    return SDL_EGL_LoadLibrary(_this, path, (NativeDisplayType) 0, 0);
src/video/cocoa/SDL_cocoaopengles.m:        return SDL_EGL_LoadLibrary(_this, NULL, EGL_DEFAULT_DISPLAY, 0);
src/video/cocoa/SDL_cocoaopengles.m:        if (SDL_EGL_LoadLibrary(_this, NULL, EGL_DEFAULT_DISPLAY, 0) < 0) {
src/video/kmsdrm/SDL_kmsdrmopengles.c:    return SDL_EGL_LoadLibrary(_this, path, ((SDL_VideoData *)_this->driverdata)->gbm, EGL_PLATFORM_GBM_MESA);
src/video/mali-fbdev/SDL_maliopengles.c:    return SDL_EGL_LoadLibrary(_this, path, EGL_DEFAULT_DISPLAY, 0);
src/video/raspberry/SDL_rpiopengles.c:    return SDL_EGL_LoadLibrary(_this, path, EGL_DEFAULT_DISPLAY, 0);
src/video/vivante/SDL_vivanteopengles.c:    return SDL_EGL_LoadLibrary(_this, path, displaydata->native_display, 0);
src/video/wayland/SDL_waylandopengles.c:    ret = SDL_EGL_LoadLibrary(_this, path, (NativeDisplayType) data->display, 0);
src/video/windows/SDL_windowsopengles.c:        return SDL_EGL_LoadLibrary(_this, NULL, EGL_DEFAULT_DISPLAY, 0);
src/video/windows/SDL_windowsopengles.c:        if (SDL_EGL_LoadLibrary(_this, NULL, EGL_DEFAULT_DISPLAY, 0) < 0) {
src/video/winrt/SDL_winrtopengles.cpp:    if (SDL_EGL_LoadLibrary(_this, path, EGL_DEFAULT_DISPLAY, 0) != 0) {
src/video/winrt/SDL_winrtopengles.cpp:         * platforms, this would probably get done by SDL_EGL_LoadLibrary,
src/video/x11/SDL_x11opengles.c:    return SDL_EGL_LoadLibrary(_this, path, (NativeDisplayType) data->display, 0);
```
I can't speak as to whether other platforms such as Android really need this call within GLideN64, but as-is, this call is superfluous for all SDL2 targets (and in the case of RPI4, actually breaks `kmsdrm`, presumably due to `eglInitialize` not referencing the correct EGLDisplay connection for KMS/GBM context and thus not being a mere no-op as expected).

fzurita:
I created this pull request:

https://github.com/gonetz/GLideN64/pull/2096

PR #2096, opened 2019-10-04
Don't crash on failed initialization

fzurita:
This adds a configurable hack that should allow EGL to work on most platforms. This includes platforms that should have the mupen64plus VidExt API implemented, but don't.

Also, this fixes a crash when DisplayWindowMupen64plus::_start() fails.

### DIFF
--- a/src/BufferCopy/ColorBufferToRDRAM.cpp
+++ b/src/BufferCopy/ColorBufferToRDRAM.cpp
@@ -81,9 +81,12 @@ void ColorBufferToRDRAM::_initFBTexture(void)
 	m_pTexture->height = VI_GetMaxBufferHeight(m_lastBufferWidth);
 	m_pTexture->textureBytes = m_pTexture->width * m_pTexture->height * fbTexFormat.colorFormatBytes;
 
+    TextureTargetParam target = Context::EglImage ? textureTarget::TEXTURE_EXTERNAL : textureTarget::TEXTURE_2D;
+
 	{
 		Context::InitTextureParams params;
 		params.handle = m_pTexture->name;
+		params.target = target;
 		params.width = m_pTexture->width;
 		params.height = m_pTexture->height;
 		params.internalFormat = fbTexFormat.colorInternalFormat;
@@ -94,7 +97,7 @@ void ColorBufferToRDRAM::_initFBTexture(void)
 	{
 		Context::TexParameters params;
 		params.handle = m_pTexture->name;
-		params.target = textureTarget::TEXTURE_2D;
+		params.target = target;
 		params.textureUnitIndex = textureIndices::Tex[0];
 		params.minFilter = textureParameters::FILTER_LINEAR;
 		params.magFilter = textureParameters::FILTER_LINEAR;
@@ -105,7 +108,7 @@ void ColorBufferToRDRAM::_initFBTexture(void)
 		bufTarget.bufferHandle = ObjectHandle(m_FBO);
 		bufTarget.bufferTarget = bufferTarget::DRAW_FRAMEBUFFER;
 		bufTarget.attachment = bufferAttachment::COLOR_ATTACHMENT0;
-		bufTarget.textureTarget = textureTarget::TEXTURE_2D;
+		bufTarget.textureTarget = target;
 		bufTarget.textureHandle = m_pTexture->name;
 		gfxContext.addFrameBufferRenderTarget(bufTarget);
 	}
--- a/src/DisplayWindow.cpp
+++ b/src/DisplayWindow.cpp
@@ -8,15 +8,18 @@
 #include "PluginAPI.h"
 #include "FrameBuffer.h"
 
-void DisplayWindow::start()
+bool DisplayWindow::start()
 {
-	_start(); // TODO: process initialization error
+	if (!_start())
+		return false;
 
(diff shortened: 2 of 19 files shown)


hhromic:
Hi @fzurita, thanks for this PR and trying to solve the current breakage with RPI4.

However, I still strongly think that a hack for EGL Init is not necessary and moreover the need for `eglInitialize()` is unclear at this time. Imho I would remove that entirely as there is not a single concrete known example of where this is actually needed. In other words, is a premature fix.

To be clear, the original problem was `eglQueryString()` returning `nullptr` in some cases and that is already fixed in your last PR (by checking for `nullptr` instead of segfaulting).

> This adds a configurable hack that should allow EGL to work on most platforms. This includes platforms that should have the mupen64plus VidExt API implemented, but don't.

If you are referring to the libretro platform, `eglInitialize()` is **not** required there as far as I know.

TL;DR: I would drop entirely the call and hack for `eglInitialize()` until we know for sure where is really needed and why.

fzurita:
@hhromic This is mostly to address the KMS/DRM issues that we are seeing right now with RPi3 and RPi4 with @psyke83 is reporting. Apparently it affects other KMS/DRM platforms as well.

hhromic:
With all respect, I think you got confused :(

What @psyke83 reported is that *adding and calling* `eglInitialize()` broke the plugin for RPI in KMS/DRM mode. In these scenarios, SDL2 with drm/kms is used and handles all initialisation. In other words `eglInitialize()` should not be used at all in the plugin (as you suggest too).

I agree that a more elegant implementation via VidExt is desirable, however keep in mind that the plugin never had the call to `eglInitialize()` before and worked normally.

fzurita:
I'm ok with being called out on being confused :)

Either way, then what was this fix about? It looks like I'm not getting it.

https://github.com/libretro/mupen64plus-libretro-nx/commit/3dae33e95cb4d860cdc45dcb6227b5c8f5f69197

fzurita:
Also, maybe I'm mis interpreting this:
https://github.com/gonetz/GLideN64/pull/2092#issuecomment-537313625

@psyke83 seems to suggest there that the eglInitialize is needed for RPi3.

hhromic:
> Either way, then what was this fix about? It looks like I'm not getting it.

That `eglInitialize()` call there was added thinking it was necessary for Android. However after re-viewing all the discussion around the subject I realised that this is actually not the case. There is no reported/clear need for `eglInitialize()` for Android, so it turned out to be a red herring.
Please ignore that part of the commit. The important part (real fix) was the check for `nullptr`.
I admit my part of the confusion there, hope it is clarified now.

> psyke83 seems to suggest there that the eglInitialize is needed for RPi3.

He wrote:

> Building without EGL support seemed to work OK on RPI4, but on RPI3, while it does allow the plugin to load, it causes graphical corruption (such as black squares in place of character shadows in Mario 64).
> If I build with EGL enabled, but remove the return false from your eglInitialize check above, it allows the plugin to work as before (without graphical glitches) on RPI3.

What he means with "EGL enabled/disabled" is configuring the build for the compiler to use `-DEGL` or not. Therefore when not passing `EGL`  then the call to `eglInitialize()` is avoided, however that also affects other codepaths in the plugin probably causing the glitches. In other words, `EGL` should  be enabled but `eglInitialize()` should *not* be called.

I hope all makes more sense now.

fzurita:
```
EGL enabled:

RPI3 (fkms) & RPI4 both segfault.
EGL disabled:

RPI3 (fkms) works, but has rendering errors
RPI4 works with no apparent rendering errors.
```

Above seems to suggest that if we don't turn on EGL at all, he was getting rendering errors on RPi 3?

fzurita:
Actually, I think I get it. I'll remove that specific commit.

hhromic:
Oh yes, leave the crash fix. That seems good indeed!

Just make sure to remove any traces around `eglinitialize()`, leaving the plugin like it was before (no calls to `eglInitialize()`.

If we ever find an actual concrete case for `eglInitialize()` to be needed, we can handle that separatedly without confusions and knowing the specific case.

Thanks!

psyke83:
@fzurita,

Apologies for not being clearer re: "EGL enabled". By that I meant the `-DEGL=On` cmake configuration, unrelated to the `eglInitialize` call that was added to the other merged PR. Although the plugin never called `eglInitialize` itself, when the `EGL` cmake configuration is active, it does cause different behaviour that could explain the rendering problems on RPI3; just one [example](https://github.com/gonetz/GLideN64/blob/master/src/Graphics/OpenGLContext/opengl_GLInfo.cpp#L108-L125). My point was just that disabling the `EGL` cmake configuration was not a viable workaround for RPI3.

This PR looks like the right solution. If it's true that Android also uses SDL2, then SDL2's internal `android` driver is also responsible for initializing EGL, so this call in the plugin was probably unnecessary for all platforms. As @hhromic said, the part of your other change that checked for `nullptr` when querying EGL extensions was the part necessary to fix the segfault on KMS targets (which is also reproducible on Intel i965 via KMS, so it's KMS-specific and not a Pi-specific driver issue per se).

fzurita:
Also added some fixes for EGL image for Mali devices.

gonetz:
Impressive, thanks!

Issue #2064, opened 2019-06-26
Ogre Battle 64 segfault


Pickle:
Im getting a seg fault in the CRC_Calculate when Update_Textures is run.
I did try and put a null pointer check in the function so im not sure what the issue is.
Im running with the latest sources for the rest up mupen.

I tried to start with no GlideN64 section in the cfg file.
Other games are working fine.

gonetz:
Yes, OB64 is broken by commit 2d712f2ae0
Thanks for the report.

gonetz:
@fzurita Hi Francisco,

I have very strange issue with Ogre Battle 64.
I use 32bit mupen64plus-gui.
ThreadedVideo option is disabled.
When I run debug build of the current GLideN64 master branch, Ogre Battle 64 freezes right after the N logo. Release build runs fine.

I found that the freeze is result of interrelation of two commits:
* Fix issues with threaded OpenGL when debug is enabled 92bcbe598
* Rewrite textures loading and mapping 2d712f2a

Reverting any of them fixes the freeze.
Then I found that reverting only this part of 92bcbe598 also fixes the freeze:
```
+#ifdef GL_DEBUG
+		m_threaded_wrapper = true;
+		m_shutdown = false;
+#else
 		if (_threaded == 1) {
 			m_threaded_wrapper = true;
 			m_shutdown = false;
 			m_commandExecutionThread = std::thread(&FunctionWrapper::commandLoop);
-		} else {
+
+		}
+		else {
 			m_threaded_wrapper = false;
 			m_shutdown = true;
 		}
+#endif 

```
Could you check it? I have no idea what is going on there and how to debug it.

fzurita:
Sure, I will look at it. I'm on business travel right now, so it may not be for a while.

gonetz:
It is not urgent. Thanks in advance :)

Pickle:
@gonetz did you experience the crash at the function I reported or just
happened to experience something else?

gonetz:
I had no crash, but I tested only first logos.
When the game crashes for you? GLideN64 has no Update_Textures.

Pickle:
```
Thread 1 "mupen64plus" received signal SIGSEGV, Segmentation fault.
0x00007fffce7e81ae in CRC_Calculate(unsigned int, void const*, unsigned int) ()
   from /usr/local/lib/mupen64plus/mupen64plus-video-GLideN64.so
(gdb) where
[...]

PR #2103, opened 2019-10-14
Fix threaded OpenGL lockup when trying to remove invalid buffer

fzurita:
This should fix the lockup you are experiencing here:

https://github.com/gonetz/GLideN64/issues/2064

I don't know about that crash though since I can't recreate it.


gonetz:
Thanks!

PR #2117, opened 2019-11-01
Fix some incorrectly set android makefile options

fzurita:
Some android makefile options are incorrectly added to LOCAL_CFLAGS instead of LOCAL_CPPFLAGS. This fixes that.


fzurita:
Also, added a few GL debug functions to the threaded OpenGL implementation.

PR #2119, opened 2019-11-02
Add logging of OpenGL profiling data

fzurita:
Log data that can be used for profiling OpenGL calls


gonetz:
I have GL_DEBUG defined for debug builds, but I don't want every call to be profiled.
Please hide profiling under GL_PROFILE define.

fzurita:
Ok

fzurita:
This should be all set as well.

PR #2130, opened 2019-11-08
Shaders cleanup

standard-two-simplex:
Cleanup combiner and blender shaders. It uses macros to make code neater, but the operations are almost the same. In theory, there should be no changes to image or performance.
I changed the problematic `equal` function for two `step` functions. If anyone can test if there are no problems on Mac would be great.

Do not merge. Cleanup needed.
I placed the `WRAP` function there because I didn't find an appropriate place. Any suggestions as to where it may go? 
Blender code looks almost as old code left as reference. Worth keeping it?

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
@@ -622,7 +622,7 @@ public:
 	ShaderSignExtendColorC()
 	{
 		m_part =
-			" color1 = WRAP(color1, -1.00, 1.01); \n"
+			" color1 = WRAP(color1, -1.01, 1.01); \n"
 			;
 	}
 };
@@ -633,7 +633,7 @@ public:
 	ShaderSignExtendAlphaC()
 	{
 		m_part =
-			" alpha1 = WRAP(alpha1, -1.00, 1.01); \n"
+			" alpha1 = WRAP(alpha1, -1.01, 1.01); \n"
 			;
 	}
 };
@@ -654,7 +654,7 @@ public:
 	ShaderSignExtendColorABD()
 	{
 		m_part =
-			" color1 = WRAP(color1, -0.50, 1.51); \n"
+			" color1 = WRAP(color1, -0.51, 1.51); \n"
 			;
 	}
 };
@@ -665,7 +665,7 @@ public:
 	ShaderSignExtendAlphaABD()
 	{
 		m_part =
-			"  alpha1 = WRAP(alpha1, -0.50,1.51); \n"
+			"  alpha1 = WRAP(alpha1, -0.51,1.51); \n"
 			;
 	}
 };
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
@@ -1251,22 +1251,22 @@ public:
 		m_part =
 			"void main() \n"
 			"{			 \n"
-		;
+			;
 		if (!_glinfo.isGLES2) {
 			m_part +=
 				"  highp float fragDepth = writeDepth();	\n"
-			;
+				;
 		}
 		m_part +=
 			"  lowp vec4 vec_color;				\n"
 			"  lowp float alpha1;				\n"
 			"  lowp vec3 color1, input_color;	\n"
-		;
+			;
 		m_part += "#define WRAP(x, low, high) mod((x)-(low), (high)-(low)) + (low) \n"; // Return wrapped value of x in interval [low, high)
 		// m_part += "#define WRAP(x, low, high) (x) - ((high)-(low)) * floor(((x)-(low))/((high)-(low)))  \n"; // Perhaps more compatible?
-		// m_part += "#define WRAP(x, low, high) (x) - ((high)-(low)) * (ceil(((x)-(low))/((high)-(low))) - 1.0) \n"; // Return wrapped value of x in interval (low,high]
-
+		// m_part += "#define WRAP(x, low, high) (x) + ((high)-(low)) * (1.0-step(low,x)) - ((high)-(low)) * step(high,x) \n"; // Step based version. Only wraps correctly if input is in the range [low-(high-low), high + (high-low)). Similar to old code.
 	}
+
 };
 
 class ShaderFragmentMain2Cycle : public ShaderPart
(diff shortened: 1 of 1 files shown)


gonetz:
Very good refactor, thanks!
I have no objections for the code.
The PR itself should be cleaned. There should be two commits: one for color WRAP and another one for blender, the final version.

> I placed the WRAP function there because I didn't find an appropriate place. Any suggestions as to where it may go?

FragmentShaderHeader is ok, but ShaderFragmentMain is probably better:
```
m_part =
	"void main() \n"
	"{			 \n"
	"#define WRAP(x, low, high) mod((x)-(low), (high)-(low)) + (low) \n"; // Return wrapped value of x in interval [low, high)
```
> Blender code looks almost as old code left as reference. Worth keeping it?

Yes. With it in  place I see clearly that the new code is the same as the reference one.

standard-two-simplex:
I refactored the code.

> No need to left the old code commented.

Do you want this sort of explanations out too, or do I leave them there?
```
/*
N64 sign-extension for C component of combiner
if (c > 1.0)
return c - 2.0;

return c;
*/
```

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -506,41 +506,19 @@ class ShaderBlender1 : public ShaderPart
...
+				"  #define MUXPM(pos) muxPM*(STVEC(pos))				\n"
+				"  muxPM[0] = clampedColor;								\n"
+				"  if (uForceBlendCycle1 != 0) {						\n"
+				"    muxA[0] = clampedColor.a;							\n"
+				"    muxB[0] = 1.0 - MUXA(uBlendMux1[1]);				\n"
+				"    lowp vec4 blend1 = MUXPM(uBlendMux1[0]) * MUXA(uBlendMux1[1]) + MUXPM(uBlendMux1[2]) * MUXB(uBlendMux1[3]);	\n"
+				"    clampedColor.rgb = clamp(blend1.rgb, 0.0, 1.0);	\n"
+				"  } else clampedColor.rgb = (MUXPM(uBlendMux1[0])).rgb;	\n"
 				;
-#else
```
gonetz: I guess we have misunderstanding here. I asked to keep the old code for reference. With it in place I see that the new code is correct.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -1234,6 +1234,10 @@ class ShaderFragmentMain : public ShaderPart
 			"  lowp float alpha1;				\n"
 			"  lowp vec3 color1, input_color;	\n"
 		;
+		m_part += "#define WRAP(x, low, high) mod((x)-(low), (high)-(low)) + (low) \n"; // Return wrapped value of x in interval [low, high)
+		// m_part += "#define WRAP(x, low, high) (x) - ((high)-(low)) * floor(((x)-(low))/((high)-(low)))  \n"; // Perhaps more compatible?
+		// m_part += "#define WRAP(x, low, high) (x) - ((high)-(low)) * (ceil(((x)-(low))/((high)-(low))) - 1.0) \n"; // Return wrapped value of x in interval (low,high]
+
```
gonetz: May be add the old implementation for reference? It can be useful to quick check that a problem is not caused by the new code.

standard-two-simplex:
> I guess we have misunderstanding here. I asked to keep the old code for reference. With it in place I see that the new code is correct.

Ok. I think I got it right now. Blender code keeps the old code.

> May be also add the old implementation for reference? It can be useful to quick check that a problem is not caused by the new code.

> please just remove old code in the WRAP commit

About wrap code, I'm not sure what you meant. I left the explanations in the wrap code but removed the old code. I hope that's what you meant.

standard-two-simplex:
I put `WRAP` in both `ShaderFragmentMain` and `ShaderFragmentMain2Cycle `, otherwise shaders didn't compile in 2 cycle mode. 
I would appreciate it if it were tested before any merge. I tested it in many games but

gonetz:
> I left the explanations in the wrap code but removed the old code. I hope that's what you meant.

Yes, it is what I meant. I also asked to add another WRAP define, with the old code, that is via step, for testing purposes. It is not necessary, but can be handy to quick compare between old and new variant without reverting new code.

> I would appreciate it if it were tested before any merge.

The PR looks good to me, thanks! You commits uploaded into shaders_cleanup branch:
https://github.com/gonetz/GLideN64/commits/shaders_cleanup
I made WIP build with these changes: #2037
If there will be no new issues, it will be included to the master.

standard-two-simplex:
> I also asked to add another WRAP define

So I added another define using the step function. However, all approaches were not identical in the old code, so it is not completely accurate to old code.
Furthermore, in order to make the wrap define useful for any situation, I wrote `(high-low)` were `2.0` was before. 

In any case, function `WRAP` should be useful to emulate masking integers (i.e., operations like `&0xff`) in floating points and it is not restricted to its use in the combiner.

If you want I can squash the last commit and force push the branch.

GhostlyDark:
zelda_majora's_mask-003
This happens in the new GLideN64 WIP build. I assume it is caused by this commit. The fire arrow effect in Majora's Mask has some issues.

Nerrel:
You can also see the issue right at the title screen (the smoke effect looks like it is shooting particles out) and when charging the sword attack:
image of title

image of sword

gonetz:
"Write wrap function" causes the issues.
"Cleanup blender code." is ok.

gonetz:
@standard-two-simplex  Switch to Step based version of WRAP does not help to fix the issues.

standard-two-simplex:
I suspected it would be wrap.
You may try changing (high -low) by 2.0 in the stepped version but I don't think thats the problem.
I'll investigate further this afternoon. 
One question? Does it happen without HD textures too?

GhostlyDark:
Yes and no. The fire arrow effect looks normal without HD textures, but the issues described by Nerrel do happen with vanilla textures as well.

standard-two-simplex:
@GhostlyDark  Please, test the high res case with this build.
[test_build.zip](https://github.com/gonetz/GLideN64/files/3837574/test_build.zip)

>  Switch to Step based version of WRAP does not help to fix the issues.

Unfortunately, the previous code was not too consistent from one case to another, so it is not possible to write one macro to do every line removed in the exact same way. I tried to write a step function that was as close as possible.

- Some cases were guarded with a slightly greater/smaller value, some other cases weren't. The new implementation doesn't touch the lowest bound, and since `low` is used also as multiplier in `high-low` I tried keeping it as close to the theoretical value as possible. This seemed to be the mistake: an alpha of approximatedly -1 was wrapped to 1 making some textures opaque instead of transparent.
- Cases ABD were wrapped above and below, case C only above. This is not possible with a generic wrap function. In order to avoid writing another different wrap function, now it wraps both above and below. I checked angrylions sources and he uses `SIGNF(x,9)` function, which wraps both above and below, so it shouldn't be a problem (at least in theory).

gonetz:
I tested new version. The issues reported above are fixed.

GhostlyDark:
I can confirm. The issues are gone.

gonetz:
Merged to master. @standard-two-simplex thanks!

PR #2131, opened 2019-11-12
[appveyor] Add Zilmar-spec

tim241:
This is still WIP at the moment

personally I'd prefer to put the whole `build_script` stuff in a batch script and just call that from the `appveyor.yml`, would that be fine with you?

Also are you fine with the packaging changes? This'll package it like the WIP builds you publish

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,5 @@
+image: Visual Studio 2017
+
 version: 1.0.{build}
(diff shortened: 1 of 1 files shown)


tim241:
I'm gonna close this PR due to forgetting to make a new branch(Github doesn't support re-targeting of the branch sadly)

PR #2132, opened 2019-11-12
[appveyor] Add Zilmar-spec

tim241:
This is still WIP at the moment

personally I'd prefer to put the whole `build_script` stuff in a batch script and just call that from the `appveyor.yml`, I'm not sure what you'd prefer however.

Also are you fine with the packaging changes? This'll package it like the WIP builds you publish

also note the `QT_BUILD_BASE_URL` variable, I uploaded the static qt builds (vs2017 x86 & x64) to github releases because downloading files from google drive is kinda hard without an external tool(which also needs to be maintained), so if you'd like full control over the static qt builds I recommend you make a `qt_build` release with the static qt builds and just change the variable

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,5 @@
+image: Visual Studio 2017
+
 version: 1.0.{build}
 pull_requests:
   do_not_increment_build_number: true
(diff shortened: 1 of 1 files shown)


gonetz:
Tim, thanks! The commit looks good to me. I have no preferences regarding build scripts. If it works it is good.

> Also are you fine with the packaging changes? This'll package it like the WIP builds you publish

I've downloaded GLideN64.7z artifact. It contains:
- mupen64plus
- Mupen64Plus_x64
- zilmar-spec
- Zilmar-spec_x64

that is some folder names started from lowercase and others from uppercase. I can't see, why.

Also, could you add another step and remove .lib and .exp files from the result? These files are useless.

Everything else is fine.

>  I recommend you make a qt_build release with the static qt builds and just change the variable

Ok.

Jj0YzL5nvJ:
Very good, thanks to your code I was able to compile my own build from CLI.

Can you implement something like this?
```
for /f "tokens=3,4 delims=_" %a in ('git describe --tags') do set rev1=%a& set rev2=%b
7z a GLideN64_v%rev1%.%rev2%.7z *
```
Similar to https://github.com/PCSX2/pcsx2/blob/master/appveyor.yml

oddMLan:
You can get the current revision from revision.h, so I guess you could leverage that in the packaging step.

gonetz:
I downloaded result of 0a6c0aea. It looks fine except for the lowercase/uppercase issue.
Tim, please add the feature suggested by @Jj0YzL5nvJ . It is useful when users download builds from several commits to compare.

tim241:
I'm not sure how that casing issue can exist, I specify the correct directory path in the variables, maybe it has to do with Windows it's case insensitive file system.

also I've changed the artifact filename to `GLideN64-$(git describe --always).7z` which should be good enough because `git describe --tags` it's output was rather long

gonetz:
Very cool, thanks!

PR #2135, opened 2019-11-13
Update appveyor.yml

oddMLan:
Allow parallel builds for improved building speed. Current build time is 10-12 minutes.
EDIT: With this change, build time is about 9 minutes. (9m21s)


tim241:
oh I totally forgot about parallel building, thank you for fixing it :+1:

PR #2144, opened 2019-11-28
Add glGetTexLevelParameteriv threaded function

fzurita:
This adds the glGetTexLevelParameteriv as a function available for threading.

Also, for some reason, I'm getting a black screen with the latest master, I'll update this pull request once I figure out why.


fzurita:
Also, glGetTexLevelParameteriv is a synced function, which will halt the whole threaded pipeline until OpenGL returns. If this is a function that is being called often, it would be good to find a way to not do it through a GL command.

fzurita:
```
"#version 320 es 
    # define IN in
    # define OUT out
    # define texture2D texture
    #if __VERSION__ < 130                                                              
    uniform ivec2 uTextureSize;                                                        
    #endif                                                                             
                                                                                       
    ivec2 getTextureSize(in sampler2D tex)                                             
    {                                                                                  
    #if __VERSION__ < 130                                                              
        return uTextureSize;                                                           
    #else                                                                              
        return textureSize(tex, 0);                                                    
    #endif                                                                             
    }                                                                                  
    vec2 norm2denorm(in vec2 uv, in ivec2 texture_size)                                
    {                                                                                  
           return uv * texture_size - 0.5f;                                            
    }                                                                                  
    ivec2 denorm2idx(in vec2 d_uv)                                                     
    {                                                                                  
           return ivec2(floor(d_uv));                                                  
    }                                                                                  
    ivec2 norm2idx(in vec2 uv, in ivec2 texture_size)                                  
    {                                                                                  
           return denorm2idx(norm2denorm(uv, texture_size));                           
    }                                                                                  
    vec2 idx2norm(in ivec2 idx, in ivec2 texture_size)                                 
    {                                                                                  
           vec2 denorm_uv = vec2(idx) + 0.5f;                                          
           return denorm_uv / texture_size;                                          
    }                                                                                  
    vec4 texel_fetch(in sampler2D tex, in ivec2 idx, in ivec2 texture_size)            
    {                                                                                  
           vec2 uv = idx2norm(idx, texture_size);                                      
           return texture2D(tex, uv);                                                  
    }                                                                                  
    vec4 hybridFilter(in sampler2D tex, in vec2 uv)                                    
    {                                                                                  
           ivec2 texSize = getTextureSize(tex);                                        
           vec2 denorm_uv = norm2denorm(uv,texSize);                                   
           ivec2 idx_low = denorm2idx(denorm_uv);                                      
           vec2 ratio = denorm_uv - idx_low;                                           
           ivec2 rounded_idx = idx_low + ivec2(step(0.5f, ratio));                     
                                                                                       
           ivec2 idx00 = idx_low;                                                      
           vec4 t00 = texel_fetch(tex, idx00, texSize);                                
                                                                                       
           ivec2 idx01 = idx00 + ivec2(0, 1);                                          
           vec4 t01 = texel_fetch(tex, idx01, texSize);                                
                                                                                       
           ivec2 idx10 = idx00 + ivec2(1, 0);                                          
           vec4 t10 = texel_fetch(tex, idx10, texSize);                                
                                                                                       
           ivec2 idx11 = idx00 + ivec2(1, 1);                                          
           vec4 t11 = texel_fetch(tex, idx11, texSize);                                
                                                                                       
           /*                                                                          
           * radius is the distance from the edge where interpolation happens.         
           * It's calculated based on  how big a fragment is in denormalized           
           * texture coordinates.                                                      
           *                                                                           
           * E.g.: If a texel maps to 5 fragments, then each fragment is               
           * 1/5 texels big. So the smooth transition should be between one and        
           * two fragments big, since there are enough fragments to show the full      
           * color of the texel.                                                       
           *                                                                           
           * If a fragment is larger than one texel, we don't care, we're already      
           * sampling the wrong texels, and should be using mipmaps instead.           
           */                                                                            
                                                                                         
           // Here, fwidth() is used to estimte how much denorm_uv changes per fragment. 
           // But we divide it by 2, since fwidth() is adding abs(dx) + abs(dy).         
           vec2 fragment_size = fwidth(denorm_uv) / 2.0f;                                
                                                                                         
           float is_frag_gt1, radius;                                                    
           // Do nothing if fragment is greater than 1 texel                             
           // Don't make the transition more than one fragment (+/- 0.5 fragment)        
           is_frag_gt1 = step(1.0f, fragment_size.s);                                    
           radius = min(fragment_size.s, 0.5f);                                          
           ratio.s = ratio.s * is_frag_gt1 + smoothstep(0.5f - radius,                   
               0.5f + radius,	ratio.s) * (1.0 - is_frag_gt1);                           
           is_frag_gt1 = step(1.0f, fragment_size.t);                                    
           radius = min(fragment_size.t, 0.5f);                                          
           ratio.t = ratio.t * is_frag_gt1 + smoothstep(0.5f - radius,                   
               0.5f + radius,	ratio.t) * (1.0 - is_frag_gt1);                           
                                                                                         
           // interpolate first on S direction then on T.                                
           vec4 top = mix(t00, t10, ratio.s);                                            
           vec4 bottom = mix(t01, t11, ratio.s);                                         
           return mix(top, bottom, ratio.t);                                             
    }                                                                                    
    IN mediump vec2 vTexCoord0;							
    uniform sampler2D uTex0;								
    OUT lowp vec4 fragColor;								
    														
    void main()											
    {														
    	fragColor = hybridFilter(uTex0, vTexCoord0);		
    } 
```

    2019-11-28 10:53:43.634 30698-30768/paulscode.android.mupen64plusae.GameActivity E/GLideN64: glsl_Utils.cpp:40,ERROR, "shader_compile error: 0:19: S0001: Type mismatch in arithmetic operation between 'vec2' and 'ivec2'"
2019-11-28 10:53:43.634 30698-30768/paulscode.android.mupen64plusae.GameActivity E/GLideN64: glsl_Utils.cpp:65,ERROR, "Error in fragment shader"

fzurita:
Ok, shader compilation has been fixed in GLSL 3.0 +

fzurita:
So as far as I can, this will cause a stall in threaded opengl for GLES 2.0 devices.

fzurita:
I got rid of glGetTexLevelParameteriv usage to determine texture size and instead use passed in parameters. That should get rid of halting the threaded GL pipeline.

fzurita:
One additional note, there is still an issue with GLES 2.0 that I hadn't noticed before
GL_DRAW_FRAMEBUFFER is not a valid target for glFramebufferTexture2D in GLES 2.0
and we are doing that in here: https://github.com/gonetz/GLideN64/blob/2b5e8786d1b98ac83f4422f2ac4860f9ee160023/src/BufferCopy/ColorBufferToRDRAM.cpp#L115

this is producing a GL error on glFramebufferTexture2D usage in that place.

gonetz:
> One additional note, there is still an issue with GLES 2.0 that I hadn't noticed before
GL_DRAW_FRAMEBUFFER is not a valid target for glFramebufferTexture2D in GLES 2.0

Yes, but we know that GLES 2.0 does not support GL_DRAW_FRAMEBUFFER:
```
	if (m_glInfo.isGLES2) {
		// Correct buffer target parameters, since GLES2 knows only GL_FRAMEBUFFER
		graphics::bufferTarget::DRAW_FRAMEBUFFER = graphics::bufferTarget::FRAMEBUFFER;
		graphics::bufferTarget::READ_FRAMEBUFFER = graphics::bufferTarget::FRAMEBUFFER;
	}

```
Thus, it must not be the source of the GL error on glFramebufferTexture2D.

gonetz:
@fzurita I merged your fixes, thanks!
As I understand, the last your commit makes the first one useless, so I did not include it.
Also, I corrected one line:
`_params.combiner->setTextureSize(_params.dstWidth, _params.dstHeight);`
It should be
`_params.combiner->setTextureSize(_params.srcWidth, _params.srcHeight);`

fzurita:
Ok, thanks, it seemed to work correctly with the destination width and height, but maybe they just happen to be the same values.

fzurita:
@gonetz the new shader kills performance with low end GLES 2.0 devices.

The majority of the time is spent on ```glDrawElements``` in UnbufferedDrawer.

gonetz:
What about average GLES 3 devices?

As for GLES2, I can use old simple shader for it. In that case all setTextureSize changes becomes useless. It is probably for good. We planned to drop GLES 2 support, but instead we do massive changes to support new functionality with it.

fzurita:
I don't see a performance difference with GLES 3.0 devices.

I don't see a problem with dropping GLES 2.0 support and simplifying the baseline. I could see Raspberry Pi users complaining about that though since the plugin does seem to get reasonable performance with the Mesa drivers.

fzurita:
Also, I just discovered that Adreno devices have a hard time compiling the new shader... trying to figure out why now.

fzurita:
The performance penalty could be related to using mediump precision. Do you think we could get by with lowp?

PR #2146, opened 2019-11-29
Fix android issues

fzurita:
Fix shader compilation errors with Adreno devices.

Also, remove call being made GL_DEBUG was enabled since it creates too much of a slowdown in Android.


gonetz:
Merged, thanks!

I also removed code for supporting hybrid filer with GLES 2.0.

fzurita:
Ok

PR #2150, opened 2019-11-30
Remove Android makefiles and use CMakeLists for Android instead

fzurita:
As the title says. Android will now use the same CMakeLists file as for unix builds instead of Android specific makefiles.


fzurita:
I have no good way of verifying that I didn't break anything in other platforms that depend on the CMakeLists file.

What happened to the automated builds?

oddMLan:
fzurita force pushed changes to this branch

fzurita:
Yep, I added automatic generation of Revision.h

gonetz:
Mac build failed

/Users/travis/build/gonetz/GLideN64/src/TextDrawer.cpp:13:10: fatal error: 'ft2build.h' file not found
#include <ft2build.h>
         ^~~~~~~~~~~~
1 error generated.

fzurita:
Hmmm, freetype issues, maybe one of my conditionals wasn't right

fzurita:
I found the problem, this should be ```else``` not ```elseif```:
https://github.com/gonetz/GLideN64/commit/d69b794c0a27b3bbc9fce58fe23ae8b44df1c800#diff-95e351a3805a1dafa85bf20b81d086e6R324

gonetz:
Yes, I found it too :)
I force updated the commit.

fzurita:
Ok, great, thanks!

fzurita:
Potentially, you could do even more consolidation:

https://docs.microsoft.com/en-us/cpp/build/cmake-projects-in-visual-studio?view=vs-2019

Visual studio can also use CMake as well it seems.

gonetz:
May be someday. Current VS projects are good.

PR #2151, opened 2019-12-02
Fix CMakeLists error in Android conditional

fzurita:
I think this should fix the error you saw.

@gonetz Any reason why we are populating the parenthesis for ```else``` and ```endif```?


fzurita:
Closing, you already took care of it.

PR #2154, opened 2019-12-14
Use a different library name when building for Android

fzurita:
Android works much more easily if a library starts with "lib".

### DIFF
--- a/src/getRevision.sh
+++ b/src/getRevision.sh
@@ -1,5 +1,12 @@
 SCRIPT_DIRECTORY=`dirname $0`
 rev=\"`git rev-parse --short HEAD`\"
+lastrev=$(head -n 1 $SCRIPT_DIRECTORY/Revision.h | awk -F'PLUGIN_REVISION ' {'print $2'})
+
 echo current revision $rev
-echo "#define PLUGIN_REVISION $rev" > $SCRIPT_DIRECTORY/Revision.h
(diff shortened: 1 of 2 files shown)


src/getRevision.sh
```diff
@@ -1,5 +1,13 @@
 SCRIPT_DIRECTORY=`dirname $0`
 rev=\"`git rev-parse --short HEAD`\"
+lastrev=$(cat $SCRIPT_DIRECTORY/revision.txt)
```
gonetz: It is a clumsy solution.
We already have current revision in Revision.h. It can be parsed as
sed -n '1{p;q}' $SCRIPT_DIRECTORY/Revision.h | awk -F'"' {'print $2'}

gonetz: or 
head -n 1 $SCRIPT_DIRECTORY/Revision.h | awk -F'"' {'print $2'}

fzurita: Yes, you are right, scripting is not my forte all I was just going with something that worked for me. I can update it to the second option.

fzurita: This command produces the revision with double quotes:
```git rev-parse --short HEAD```
so I had to change your command to this and then it works:
```head -n 1 $SCRIPT_DIRECTORY/Revision.h | awk -F'PLUGIN_REVISION ' {'print $2'}```

fzurita:
I updated with suggested changes.

gonetz:
Thanks!

PR #2172, opened 2020-02-08
Make the hybrid texture filter optional through configuration

fzurita:
I'm getting a lot of complaints saying that GLideN64 is running a lot slower now. This is probably because of the new hybrid texture filter. I'm guessing that weak GLES 3 devices are having problems running it with acceptable performance.

I'm thinking the best solution is to make it optional, but enabled by default. Then in my front end, I can disable it by default and only allow it through configuration.

Also, I left it always disabled for GLES 2.0

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -43,6 +43,7 @@ void Config::resetToDefaults()
 	generalEmulation.enableCustomSettings = 1;
 	generalEmulation.enableShadersStorage = 1;
 	generalEmulation.enableLegacyBlending = 0;
+	generalEmulation.enableHybridFilter = 1;
 	generalEmulation.hacks = 0;
 #if defined(OS_ANDROID) || defined(OS_IOS)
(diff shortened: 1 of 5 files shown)


oddMLan:
> Also, I left it always disabled for GLES 2.0

The GLES 2.0 version of the codepath was removed a while back iirc

fzurita:
I think the GLSL code may still be compatible with the older GLSL version, the path is just disabled when running under GLES 2.0.

oddMLan:
Also I'm thinking, all the filter does is applying bilinear interpolation where the texels don't perfectly match with the output resolution, and the rest is nearest. This is the same as oversampling the texture first (2x,3x,4x nearest, or whichever integer is bigger than the output resolution), and then applying bilinear interpolation to the result. Would that be faster for the driver to do, @fzurita?

fzurita:
That's a good question. It may or may not be faster. I'm sure it will depend on the device. I think we would have to try.

PR #2180, opened 2020-02-21
Enabled FBInfo for Rat Attack

loganmc10:
I finally got this game working with mupen64plus. Strangely, unless FBInfo is enabled the game will lock up before trying to go in-game, sometimes complaining about an unknown opcode.

### DIFF
--- a/ini/GLideN64.custom.ini
+++ b/ini/GLideN64.custom.ini
@@ -1,6 +1,6 @@
 ; Custom game settings
 [General]
-version=12
+version=13
 
 [TWINE]
 Good_Name=007 - The World Is Not Enough (E)(U)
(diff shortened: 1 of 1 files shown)


gonetz:
> Strangely, unless FBInfo is enabled the game will lock up before trying to go in-game, sometimes complaining about an unknown opcode.

Yes, it is strange. Have you tried to disable fb emulation? May be the plugin writes something wrong into RDRAM.

loganmc10:
Yea disabling FB emulation also fixes the problem, but the graphics look all wrong. I didn't try too many other options

It's just a bit strange because it actually seems to affect the execution of the game when it gets it wrong

loganmc10:
"Copy auxiliary color buffers to N64 memory" also fixes the problem, instead of FBInfo

gonetz:
> Yea disabling FB emulation also fixes the problem, but the graphics look all wrong. 

Most likely the plugin writes something wrong into RDRAM and the game hangs. It needs to be investigated, but I hardly will find time on it in near future. I'll accept this change, but someday it should be reconsidered.

PR #2182, opened 2020-02-26
Fix CMake warnings presented by Android Studio

fzurita:
The new Android Studio 3.6 is generating some CMake warnings that is preventing compilation. This fixes them.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,9 +1,5 @@
 cmake_minimum_required(VERSION 2.6)
 
-if ((${CMAKE_VERSION} VERSION_EQUAL 3.9) OR (${CMAKE_VERSION} VERSION_GREATER 3.9))
-  cmake_policy(SET CMP0069 NEW)
-endif()
-
 option(EGL "Set to ON if targeting an EGL device" ${EGL})
 option(PANDORA "Set to ON if targeting an OpenPandora" ${PANDORA})
 option(ODROID "Set to ON if targeting an Odroid" ${ODROID})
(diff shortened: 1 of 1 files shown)


loganmc10:
I believe this actually disables IPO altogether, now when I compile, linking is almost instant, instead of taking a while to do because of the IPO optimizations. If you read:

https://cmake.org/cmake/help/v3.9/prop_tgt/INTERPROCEDURAL_OPTIMIZATION.html#prop_tgt:INTERPROCEDURAL_OPTIMIZATION

It says ```This property is initialized by the CMAKE_INTERPROCEDURAL_OPTIMIZATION variable if it is set when a target is created.```

So I believe in order for it to work, that variable would need to be set before the ```add_library``` call, not after it

fzurita:
Ok, without it, it was telling me that it was disabling IPO in a warning message. So it was not working for me before either. So it was not working for me before, and this at least got rid of the warning for me.

loganmc10:
What happens if you change that line back to ```set_property(TARGET ${GLideN64_DLL_NAME} PROPERTY INTERPROCEDURAL_OPTIMIZATION TRUE)```

Does the compilation fail, or it just produces a warning?

fzurita:
I get a warning but the build succeeds without IPO. The only required line for me was this one:
```set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -O3")```

loganmc10:
Ok, can we set it back to ```set_property(TARGET ${GLideN64_DLL_NAME} PROPERTY INTERPROCEDURAL_OPTIMIZATION TRUE)```?

I'm not sure how you would tell on Android, but I'm guessing the reason that the warning went away is because IPO is disabled altogether now

Issue #2173, opened 2020-02-08
Lack of dithering in N64 framebuffer causing color banding


ToruTheRedFox:
Dithering is honestly necessary to produce a correct (and actually somewhat good looking) N64 framebuffer, as it's quantized with no dithering, which deviates from hardware, which does infact have dithering

gonetz:
Do you mean that the plugin should apply color dithering to 16bit color buffer, which is stored in RDRAM?

ToruTheRedFox:
for accuracy purposes, yes

gizmo98:
Which matrix types does original hardware use? Bayer matrix 3x3, 4x4 or 8x8? How does the ordered grid magic square pattern look like?

ToruTheRedFox:
it's a 4x4 matrix for both, and the matrices are 
static const uint8_t bayer_matrix[16] =
{
     0,  4,  1, 5,
     4,  0,  5, 1,
     3,  7,  2, 6,
     7,  3,  6, 2
};


static const uint8_t magic_matrix[16] =
{
     0,  6,  1, 7,
     4,  2,  5, 3,
     3,  5,  2, 4,
     7,  1,  6, 0
};
(taken from Angrylion RDP Plus)

gizmo98:
This looks good. Where is the right location to add dithering?

Add if( uColorDitherMode == 0) and if (uColorDitherMode == 1) here?
https://github.com/gonetz/GLideN64/blob/0692abea4aa4ed33b7cd3e7def1f24560e513bcd/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp#L701

Or here before color quantization?
https://github.com/gonetz/GLideN64/blob/44ce554d5f9d22e49e60a74e39b73fbae991b733/src/BufferCopy/WriteToRDRAM.h

ToruTheRedFox:
I implemented those into a shader, and my method anyways requires it to be done before the quantization.
Example code I made which also does the 21-bit quantization (albeit without dithering this time), which works in Unity just fine:
```HLSL

Pass
        {
            CGPROGRAM
            #pragma vertex vert
            #pragma fragment frag
			#include "UnityCG.cginc"

            struct appdata
            {
                float4 vertex : POSITION;
                float4 uv : TEXCOORD0;
            };

            struct v2f
            {
                float4 uv : TEXCOORD0;
                float4 vertex : SV_POSITION;
				float4 screenPosition : TEXCOORD1;
            };

[...]

PR #2183, opened 2020-02-26
Add RDRAM dithering: Dither image if RGBA16 is written to RDRAM

gizmo98:
ColorbufferToRDRAM RGBA to RGBA16 conversion produces color banding.
Implement dithering according to n64 documentation to hide color
banding.

https://github.com/gonetz/GLideN64/issues/2173

### DIFF
--- a/src/BufferCopy/ColorBufferToRDRAM.cpp
+++ b/src/BufferCopy/ColorBufferToRDRAM.cpp
@@ -258,7 +258,7 @@ u16 ColorBufferToRDRAM::_RGBAtoRGBA16(u32 _c, u32 x, u32 y) {
 	union RGBA c;
 	c.raw = _c;
 	
-	if(gDP.otherMode.colorDither <= 1) {
+	if(gDP.otherMode.colorDither <= 1 && config.frameBufferEmulation.nativeResFactor != 1 && config.generalEmulation.enableDithering == 0) {
 		s32 threshold = 0;
 		
 		switch(gDP.otherMode.colorDither){
--- a/src/BufferCopy/ColorBufferToRDRAM.cpp
+++ b/src/BufferCopy/ColorBufferToRDRAM.cpp
@@ -234,17 +234,51 @@ bool ColorBufferToRDRAM::_prepareCopy(u32& _startAddress)
 	return true;
 }
 
-u8 ColorBufferToRDRAM::_RGBAtoR8(u8 _c) {
+// Precalculated 4x4 bayer matrix values for 5Bit
+const s32 thresholdMapBayer [4][4] = {
+	{-4,  2, -3,  4 },
+	{ 0, -2,  2, -1 },
+	{-3,  3, -4,  3 },
+	{ 1, -1,  1, -2 }
+};
+
+// Precalculated 4x4 magic square matrix values for 5Bit
+const s32 thresholdMapMagicSquare [4][4] = {
+	{-4,  2,  2, -1 },
+	{ 3, -2, -3,  1 },
+	{-3,  0,  4, -2 },
+	{ 3, -1, -4,  1 }
+};
+
+u8 ColorBufferToRDRAM::_RGBAtoR8(u8 _c, u32 x, u32 y) {
 	return _c;
 }
 
-u16 ColorBufferToRDRAM::_RGBAtoRGBA16(u32 _c) {
-	RGBA c;
+u16 ColorBufferToRDRAM::_RGBAtoRGBA16(u32 _c, u32 x, u32 y) {
+	union RGBA c;
 	c.raw = _c;
-	return ((c.r >> 3) << 11) | ((c.g >> 3) << 6) | ((c.b >> 3) << 1) | (c.a == 0 ? 0 : 1);
+	
+	if(gDP.otherMode.colorDither <= 1) {
+		s32 threshold = 0;
+		
+		switch(gDP.otherMode.colorDither){
+			case G_CD_BAYER:
+				threshold = thresholdMapBayer[x & 3][y & 3];
+				break;
+			case G_CD_MAGICSQ:
+				threshold = thresholdMapMagicSquare[x & 3][y & 3];
+				break;
+		}
+		
+		c.r = (u8)std::max(std::min((s32)c.r + threshold, 255), 0);
+		c.g = (u8)std::max(std::min((s32)c.g + threshold, 255), 0);
+		c.b = (u8)std::max(std::min((s32)c.b + threshold, 255), 0);
+	}
+		
+	return ((c.r >> 3) << 11) | ((c.g >> 3) << 6) | ((c.b >> 3) << 1) | (c.a == 0 ? 0 : 1);	
 }
 
-u32 ColorBufferToRDRAM::_RGBAtoRGBA32(u32 _c) {
+u32 ColorBufferToRDRAM::_RGBAtoRGBA32(u32 _c, u32 x, u32 y) {
 	RGBA c;
 	c.raw = _c;
 	return (c.r << 24) | (c.g << 16) | (c.b << 8) | c.a;
--- a/src/BufferCopy/ColorBufferToRDRAM.cpp
+++ b/src/BufferCopy/ColorBufferToRDRAM.cpp
@@ -258,7 +258,9 @@ u16 ColorBufferToRDRAM::_RGBAtoRGBA16(u32 _c, u32 x, u32 y) {
 	union RGBA c;
 	c.raw = _c;
 	
-	if(gDP.otherMode.colorDither <= 1 && config.frameBufferEmulation.nativeResFactor != 1 && config.generalEmulation.enableDithering == 0) {
+	if(gDP.otherMode.colorDither <= 1 && 
+		((config.frameBufferEmulation.nativeResFactor != 1 	&& config.generalEmulation.ditheringMode >= 3) 
+		|| config.generalEmulation.ditheringMode < 3)) {
 		s32 threshold = 0;
 		
 		switch(gDP.otherMode.colorDither){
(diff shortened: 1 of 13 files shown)


gonetz:
I added that commit into https://github.com/gonetz/GLideN64/tree/dithering branch.
Please add your shader-based dithering and enableDithering option.

gizmo98:
I kept the enableNoise option as is and added enableDithering for ordered grid dithering and ditheringMode to enable or disable 16 Bit quantization after dithering for noise, bayer and magic square dithering.

the_legend_of_zelda-063
the_legend_of_zelda-064
beetle_adventure_rac-009
ridge_racer_64-009
conker_bfd-202

src/BufferCopy/ColorBufferToRDRAM.cpp
```diff
@@ -258,7 +258,7 @@ u16 ColorBufferToRDRAM::_RGBAtoRGBA16(u32 _c, u32 x, u32 y) {
 	union RGBA c;
 	c.raw = _c;
 	
-	if(gDP.otherMode.colorDither <= 1) {
+	if(gDP.otherMode.colorDither <= 1 && config.frameBufferEmulation.nativeResFactor != 1 && config.generalEmulation.enableDithering == 0) {
```
gonetz: Never dither for config.frameBufferEmulation.nativeResFactor == 1? Even when config.generalEmulation.enableDithering == 0?

May be:
if (gDP.otherMode.colorDither <= 1 && (config.generalEmulation.enableDithering == 0 || config.frameBufferEmulation.nativeResFactor != 1)

?

gonetz:
> I kept the enableNoise option as is and added enableDithering for ordered grid dithering and ditheringMode to enable or disable 16 Bit quantization after dithering for noise, bayer and magic square dithering

I would prefer only one setting **ditheringMode** for dithering with options:
* disable
* enable
* enable with 16bit quantization

There is a problem with only setting: it is impossible to enable only noise dithering and 16bit quantization. It can be solved. We can make noise dithering be controlled by this setting as well, so its options will be:
* disable
* only noise dithering (default)
* noise dithering with 16bit quantization
* noise and ordered grid dithering
* dithering with 16bit quantization

So, generalEmulation.enableNoise will control only use of Noise color input.

gizmo98:
@gonetz
Is it ok if i enable noise with enableNoise = true in config::validate if ditheringMode >=1? At least i need uNoiseTex for noise dithering.

gonetz:
> Is it ok if i enable noise with enableNoise = true in config::validate if ditheringMode >=1? At least i need uNoiseTex for noise dithering.

If ditheringMode will control noise dithering, no.
enableNoise will control only noise color input, and it would be wrong to make it dependent on noise dithering. I suggest to add a new method:
```
bool Config::isNoiseEnabled
{
  return generalEmulation::enableNoise != 0 || // check for ditheringMode be here
}
```
isNoiseEnabled should be used to check is noise textures init is necessary and in some other places where generalEmulation::enableNoise is currently used to check that Noise is enabled.

gizmo98:
Ok. Dithering can be selected only with ditherMode now and enableNoise is no longer necessary.

src/BufferCopy/ColorBufferToRDRAM.cpp
```diff
@@ -234,17 +234,53 @@ bool ColorBufferToRDRAM::_prepareCopy(u32& _startAddress)
...
-	RGBA c;
+u16 ColorBufferToRDRAM::_RGBAtoRGBA16(u32 _c, u32 x, u32 y) {
+	union RGBA c;
 	c.raw = _c;
-	return ((c.r >> 3) << 11) | ((c.g >> 3) << 6) | ((c.b >> 3) << 1) | (c.a == 0 ? 0 : 1);
+	
+	if(gDP.otherMode.colorDither <= 1 && 
+		((config.frameBufferEmulation.nativeResFactor != 1 	&& config.generalEmulation.ditheringMode >= 3) 
+		|| config.generalEmulation.ditheringMode < 3)) {
+		s32 threshold = 0;
```
gonetz: I don't understand this condition.
gDP.otherMode.colorDither <= 1 is clear, the rest is not.

This condition needed to disable double dithering, that is dithering already dithered image.
Image can be already dithered only when native res is enabled and config.generalEmulation.ditheringMode >= 3
because dithering disappears when hi-res image scaled down. So, the condition is
```
gDP.otherMode.colorDither <= 1 // ordered grid dithering enabled in othermode
&& !(config.frameBufferEmulation.nativeResFactor == 1 && config.generalEmulation.ditheringMode >= 3) // image is NOT already dithered

```
or
```
gDP.otherMode.colorDither <= 1 // ordered grid dithering enabled in othermode
&& (config.frameBufferEmulation.nativeResFactor != 1 || config.generalEmulation.ditheringMode < 3)

```

gonetz:
I've put this PR into "dithering" branch and added some corrections:
https://github.com/gonetz/GLideN64/tree/dithering
I haven't tested that branch yet.

gizmo98:
Thank you!

RDRAM dithering should always happen if there is no shader dithering `(config.generalEmulation.ditheringMode < 3)` or if shader dithering is enabled and nativeRes is != 1 because dithering disappears when hi-res image scaled down `((config.frameBufferEmulation.nativeResFactor != 1 && config.generalEmulation.ditheringMode >= 3)`.

Your second suggestion could do the trick.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -696,11 +696,59 @@ class ShaderCallDither : public ShaderPart
...
+					"    quantizeRGB(clampedColor.rgb);											\n"
+					"  }																		\n"
+					"  if (uAlphaDitherMode < 2) {												\n"
+					"    lowp float thresholdPattern = mix(mSquareThreshold,bayerThreshold,clamp(float(uColorDitherMode),0.0,1.0));\n"
+					"    thresholdPattern = mix(thresholdPattern,mSquareThreshold,clamp(float(uColorDitherMode - 1),0.0,1.0));\n" // uColorDitherMode = 2
+					"    thresholdPattern = mix(thresholdPattern,bayerThreshold,clamp(float(uColorDitherMode - 2),0.0,1.0));\n" // uColorDitherMode = 3
+					"  																			\n"
+					"    lowp float thresholdNotPattern = mix(bayerThreshold,mSquareThreshold,clamp(float(uColorDitherMode),0.0,1.0));\n"
+					"    thresholdNotPattern = mix(thresholdNotPattern,bayerThreshold,clamp(float(uColorDitherMode - 1),0.0,1.0));\n" // uColorDitherMode = 2
+					"    thresholdNotPattern = mix(thresholdNotPattern,mSquareThreshold,clamp(float(uColorDitherMode - 2),0.0,1.0));\n" // uColorDitherMode = 3
```
gonetz: Why some many calculations for alpha dither?
Why alpha dither depends on uColorDitherMode?

gizmo98: I just wanted to replace branches with mix functions. Should not be necessary for newer gpus. 

Alpha dither depends on uColorDitherMode if alpha dither mode is G_AD_PATTERN or G_AD_NOTPATTERN according to this docu:

https://hack64.net/wiki/doku.php?id=rcpstructs

gonetz: > I just wanted to replace branches with mix functions. Should not be necessary for newer gpus.

Since uColorDitherMode is an uniform, we can use it in if..else statement without noticeable performance penalty. We can even use switch(uColorDitherMode) since you disabled shader dithering for GLES2.

 >Alpha dither depends on uColorDitherMode if alpha dither mode is G_AD_PATTERN or G_AD_NOTPATTERN according to this docu:

Nice docu, thanks!

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -696,11 +696,59 @@ class ShaderCallDither : public ShaderPart
...
+					"                            0.5, 0.25, 0.6875, 0.4375,						\n"
+					"                            0.125, 0.875, 0.0625, 0.8125,					\n"
+					"                            0.625, 0.375, 0.5625, 0.3125);					\n"
+					"  lowp mat4 mSquare = mat4( 0.0, 0.6875, 0.75, 0.4375,						\n"
+					"                              0.875, 0.3125, 0.125, 0.5625, 				\n"
+					"                              0.1875, 0.5, 0.9375, 0.25,					\n"
+					"                              0.8125, 0.375, 0.0625, 0.625);				\n"
+					"  																			\n"
+					"  lowp float bayerThreshold = bayer[position.x][position.y];				\n"
+					"  lowp float mSquareThreshold = mSquare[position.x][position.y];			\n"
```
gonetz: We will have performance problems here. It is known that fetch from matrix by dynamically calculated indices is slow. I already had that problem when I implemented shader-based blending. This part need to be rewritten using the same approach, as for the blending.

gizmo98:
Two 4x4 textures and two texelFetch() calls would do the same thing. Or switch statements with 16 cases.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
gizmo98: Two 4x4 textures and two texelFetch() calls would do the same thing. Or switch statements with 16 cases.

gonetz: > Two 4x4 textures and two texelFetch() calls would do the same thing.

Yes, it would be fastest, but I don't want to write special code for special textures.

> Or switch statements with 16 cases.

I suppose it will not be faster if you switch on dynamic value.

I made it as for blending, with matrix-vertex multiplication and dot, commit 2f1cda1654

gonetz:
In the master now. @gizmo98 thanks!

PR #2184, opened 2020-02-28
Fix IPO build on Linux

loganmc10:
@fzurita can you let me know if this works on Android?

The changes to the cmake file disabled IPO for Windows and Linux for me.

This change bumps the required cmake version to 3.9, 3.9 came out in 2017, I think most Linux distros (and Android) use it now.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.6)
+cmake_minimum_required(VERSION 3.9)
 
 option(EGL "Set to ON if targeting an EGL device" ${EGL})
 option(PANDORA "Set to ON if targeting an OpenPandora" ${PANDORA})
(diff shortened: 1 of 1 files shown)


fzurita:
Sure, I will try it when I can, I've been busy the last few days.

fzurita:
Yep, this works fine. And it's taking a lot longer at link time, so IPO is actually working. I would merge this.

gonetz:
I merged this commit, but it causes compilation error on Linux, so I reverted it back. The error:
```
/Users/travis/build/gonetz/GLideN64/src/./GraphicsDrawer.h:102:26: error: expected '(' for function-style cast or type construction

                Textures tex = Textures{ { nullptr, nullptr } };

                               ~~~~~~~~^

```

fzurita:
Yikes... That's a very weird side effect.

fzurita:
Here are a few possible solutions: https://stackoverflow.com/questions/12362502/cmake-not-working-with-c11-on-osx

gonetz:
This commit changes minimal required cmake version and options for Release build.
Since both Debug and Release got broken, it seems that the problem is with the cmake version.
However, as I found, Travis CI uses cmake 3.11.3 for Mac:
https://docs.travis-ci.com/user/reference/osx/
I don't understand, what happens.

>I think we want at least c++14 or even c++17 support
I don't mind to set c++14/17 flags for cmake builds, since it does not affect the code.
But I don't see how it is related to the problem.

gonetz:
I added 
`set(CMAKE_CXX_STANDARD 11)`
and it fixed compilation on Mac.

PR #2193, opened 2020-03-22
WTL UI

project64:
This is not complete, there is still more work to do. 

I have been working on a WTL version of the UI instead of the QT version. This will make compiling quicker and simpler as well as dramatically reduce the file size. 

With the huge amount of changes I am sure things are not going to be perfect and there is a lot more testing i need to do. 

Some of the functionality missing is:
- Profile
- Translations
- Tool Tips
- Screen shots

making the pull request to start the discussion on this work


oddMLan:
Test build, for PJ64
[GLideN64.zip](https://github.com/gonetz/GLideN64/files/4364779/GLideN64.zip)

gonetz:
Hi Zilmar!

It is good that this project finally reached PR status.

Some general thoughts first:
* I personally see no major benefits in use of WTL instead of QT.
** This will make compiling quicker - true, but full rebuild of QT version of GLideNUI takes several seconds and full rebuild of WTL version also takes several seconds, just a few less. Build time of rest of the plugin's code does not change.
** and simpler - QT requires more pre-requisites, but when everything is tuned the build process is  as simple as possible - select the project and invoke Build.
** dramatically reduce the file size - yes, WTL version is almost 5 times smaller, or 9mb less. Does it matter nowadays?
* I don't mind to have WTL-based GUI, especially if it is the way for GLideN64 to be included in Project64 distributive.
* Two years ago I would not mind to remove QT completely and switch to WTL since WTL is better for windows-only GUI. However, @loganmc10 already ported this GUI to his m64p project. m64p is multi-platform. It already brings GLideN64 GUI to Linux users, and  I think that Mac support is also possible.

Thus, your approach to remove QT-based GUI is wrong imo. The right way is to left current GLideNUI as it is and create another project, say GLideNUI-WTL. Since wtl-based GUI uses the same API from GLideNUI.h, the rest of the plugin's code will not change. We will need just to add new solution configurations: Release-qt, Release-wtl, Debug-qt, Debug-wtl. Users, who do not need to build QT versions will not install and setup QT prerequisites, as it is not necessary now for mupen64plus configurations. I think it is not hard to do since you already have working WTL-based GLideNUI, and the only thing to do is to add new project for it.

Now to lesser problems:
* The new GUI looks wrong on Win10:
GUI
No spaces between combo-boxes and other elements, wrong "Save and Close" button.
* You use spaces instead of tabs. When the project will be merged to master, it have to be changed to conform to project code style.

loganmc10:
Yea my project uses the Qt GUI and makes it available for Linux users, which I think is pretty useful and justifies keeping the Qt GUI. I would prefer the WTL GUI be added alongside the existing Qt GUI

oddMLan:
I agree with all your points. Cosmetic problems can be fixed without too much hassle and I will personally make sure that the WTL GUI is 1:1 to Qt GUI. There are a couple other advantages of WTL (at least on Windows). 

- Less issues with Qt interfering with emulation threads. It is my understanding some workarounds had to be added to prevent Qt interfering with the video context with multithreaded, though I might be mistaken. There is an issue with Jet Force Gemini and opening the config dialog iirc, WTL might not have that issue. 
- A long requested "Apply" button might be possible as well with WTL. It worked with Glide64 and the WTL version of it bundled with Project64.
- WTL UI scales better with high-dpi displays. By default they use either 125% or 150% scaling. The Qt version of the About dialog looks wrong with scaling, though that's minor. WTL version looks correct.

Now, the reason options look too close together in current version is because he uses a high resolution display and the options look like that to him with the Qt GUI, so he didn't know there was anything wrong when he attempted to recreate it in WTL. When it's fixed, the scaled WTL GUI will look even more faithful to the original Qt GUI than the actual Qt GUI looks scaled.

Finally, if a second UI option is added, who is going to take the burden of mantaining both UIs? Maybe Gonetz and Ryan Rosser will mantain the Qt GUI and keep adding new features to it while zilmar and myself do our best to keep up with the changes made to Qt UI. Because I don't know if you are familiar with WTL, but it's much more low-level than Qt so adding new things to it if you're not used to working with it might be harder. You can look at the current code in this PR to get an idea. Or if it simply isn't in your plans to touch it, I understand.

This situation is a bit ironic since the Qt UI was made for Project64 only in mind at the time. A GUI for linux and mac was only recently made possible thanks to @loganmc10's efforts to bring a cross-platform UI to Mupen64plus through m64p. To me this shows that Mupen64plus had an unsuccessful approach with letting the emulator handle all the UI components. libretro was able to do the same concept better.

gonetz:
> There is an issue with Jet Force Gemini and opening the config dialog iirc, WTL might not have that issue.

Hardly.

> A long requested "Apply" button might be possible as well with WTL.

That would be great.

> When it's fixed, the scaled WTL GUI will look even more faithful to the original Qt GUI than the actual Qt GUI looks scaled.

Great

> Maybe Gonetz and Ryan Rosser will mantain the Qt GUI and keep adding new features to it while zilmar and myself do our best to keep up with the changes made to Qt UI

Good plan. I hope that GUI changes will be seldom.

> This situation is a bit ironic since the Qt UI was made for Project64 only in mind at the time. 

For zilmar-spec emulators, including Mupen64, which has 32bit Linux port.

project64:
I deleted my branch and created a new one https://github.com/project64/GLideN64/tree/wtl-ui this has both the qt and the wtl version there. Once I get more done then I will create a new pull request. The UI was done on my machine which has a high def screen so things get scaled, The QT version is not always the best so matching that has some negative side effects in layout as you mentioned. I am working with @oddMLan to get it to look right and match the qt version on a non scaled machine.

PR #2195, opened 2020-03-29
Mask RDRAM address in ProcessRDPList

loganmc10:
Avoids a crash in Donkey Kong 64 (U) [f2]

See in Angrylion:
https://github.com/ata4/angrylion-rdp-plus/blob/f7cad30db1fdfce23957fcc3c7dfc11e69aed04a/src/core/n64video/rdp/rdram.c#L7
https://github.com/ata4/angrylion-rdp-plus/blob/f7cad30db1fdfce23957fcc3c7dfc11e69aed04a/src/core/n64video/rdp/rdram.c#L88-L89

The address the game tries to use is 0xA0150000, the 0xA needs to be masked away or it crashes

### DIFF
--- a/src/RDP.cpp
+++ b/src/RDP.cpp
@@ -564,7 +564,7 @@ inline u32 READ_RDP_DATA(u32 address)
 	if (dp_status & 0x1)          // XBUS_DMEM_DMA enabled
 		return rsp_dmem[(address & 0xfff)>>2];
 	else
-		return rdram[address>>2];
+		return rdram[(address & 0xffffff)>>2];
 }
 
 void RDP_ProcessRDPList()


loganmc10:
By the way @gonetz the new_lle branch actually breaks Donkey Kong 64 (U) [f2].

The f2 version of the ROM is a cracked version with the CIC/anti-piracy stuff removed. When the game boots it has this little screen added by the cracker:

donkey_kong_64-000

That screen isn't rendered at all by GLideN64 (it works with Angrylion), and the emulator actually calls ProcessRDPList, even in HLE mode. With the current master, it all works, except you can't see that little crack banner. With the new_lle branch, when you hit "A" to move on to the game, it's all invisible. I assume this is because it assumes the rest of the game should be in LLE?

loganmc10:
Actually sorry, I guess in general I don't get any graphics at all with the new_lle branch, HLE or LLE, I tested Windows and Linux, maybe I'm doing something wrong with that branch?

PR #2196, opened 2020-03-30
Update to XXH3 for CRC_OPT

loganmc10:
XXH3 is an updated version of the xxHash algorithm. It uses 64 bit hashes, so I had to update the CRCs to u64 across the board.

I'm not really sure the NEON version of CRC_OPT needs to exist anymore, XXH3 appears to have NEON optimizations built in, but I didn't remove the NEON version.

Everything with XXH3 is inlined, and it's supposed to work faster with small data, which we see a lot of in this plugin.


src/Textures.cpp
```diff
@@ -1252,7 +1252,7 @@ u32 _calculateCRC(u32 _t, const TextureParams & _params, u32 _bytes)
 		_bytes >>= 1;
 	const u32 tMemMask = (gDP.otherMode.textureLUT == G_TT_NONE && !rgba32) ? 0x1FF : 0xFF;
 	const u64 *src = (u64*)&TMEM[gSP.textureTile[_t]->tmem & tMemMask];
-	u32 crc = 0xFFFFFFFF;
+	u64 crc = 0xFFFFFFFF;
```
gonetz: u64 crc = 0xFFFFFFFFFFFFFFFF;

src/Graphics/OpenGLContext/opengl_BufferedDrawer.cpp
```diff
@@ -109,7 +109,7 @@ void BufferedDrawer::_updateRectBuffer(const graphics::Context::DrawRectParamete
 		return;
 	}
 
-	const u32 crc = CRC_Calculate(0xFFFFFFFF, _params.vertices, dataSize);
+	const u64 crc = CRC_Calculate(0xFFFFFFFF, _params.vertices, dataSize);
```
gonetz: `const u64 crc = CRC_Calculate(0xFFFFFFFFFFFFFFFF, _params.vertices, dataSize);`

or better define some constant for that, like
#define CRC_SEED 0xFFFFFFFFFFFFFFFF

src/Textures.cpp
```diff
@@ -1350,7 +1350,7 @@ void TextureCache::activateMSDummy(u32 _t)
 void TextureCache::_updateBackground()
 {
 	u32 numBytes = gSP.bgImage.width * gSP.bgImage.height << gSP.bgImage.size >> 1;
-	u32 crc;
+	u64 crc;
 
 	crc = CRC_Calculate( 0xFFFFFFFF, &RDRAM[gSP.bgImage.address], numBytes );
```
gonetz: `crc = CRC_Calculate( 0xFFFFFFFFFFFFFFFF, &RDRAM[gSP.bgImage.address], numBytes );`

src/CRC32.cpp
```diff
@@ -32,7 +32,7 @@ void CRC_Init()
 	}
 }
 
-u32 CRC_Calculate( u32 crc, const void * buffer, u32 count )
+u64 CRC_Calculate( u64 crc, const void * buffer, u32 count )
 {
 	u8 *p;
 	u32 orig = crc;
```
gonetz: It looks suspicious now.
I suppose, it would be safer to rewrite it as:
```
        u32 crc32 = static_cast<u32>(crc);
        u32 orig = crc32;
	p = (u8*) buffer;
	while (count--)
		crc32 = (crc32 >> 8) ^ CRCTable[(crc32 & 0xFF) ^ *p++];
	return crc32 ^ orig;
}
```

The same is for CRC_CalculatePalette bellow and for functions from CRC32_ARMV8.cpp

loganmc10:
Ok I've updated the PR with your suggestions.

I used UINT64_MAX for the CRC seed, it is defined in stdint.h:

```
#define UINT64_MAX  0xffffffffffffffff
```

gonetz:
Thanks!

dankcushions:
cool! is there a good game/scene to benchmark this? i could test to see if it's better than NEON on a pi4 (and it not, we can retire the neon stuff)

loganmc10:
From what I recall, 2D games (like Worms Armageddon) are quite affected by the texture cache

dankcushions:
my rpi4 was using CRC_ARMV8 (although that also appears to be a seperate path). i tried forcing it to use CRC_OPT by using that compile flag, and i couldn't see a regression in worms, but i couldn't really tell if I was even hitting that as a bottleneck. i was getting single digit VI/s during static screen fade in/fade out transitions and on the first few frames of a new map.

@gizmo98 did the ARMV8 stuff, so maybe they have a view.

Issue #2155, opened 2019-12-17
Zelda MM Beneath Graveyard Tiling floor textures wrong


IronYoshi64:
In Majora's Mask, Beneath the Graveyard, the floor textures don't seem to be blending/bluring/merging/tiling correctly in latest GLideN64 rev build. I'm using version GLideN64-de754772. I notice on Public Release 4.0, this doesn't seem to be an issue, so I guess this is a regression. Playing on latest PJ64 dev. Here are some screenshots of what I'm talking about.
GLideN64-de754772 incorrect tiling:
![GLideN64_ZELDA_MAJORA'S_MASK_000](https://user-images.githubusercontent.com/57850880/70964593-0e724b80-2052-11ea-8b93-4ac3599bfddc.png)
![GLideN64_ZELDA_MAJORA'S_MASK_002](https://user-images.githubusercontent.com/57850880/70964602-13cf9600-2052-11ea-824a-7c5e2fa392f6.png)
GLideN64 Public Release 4.0 correct tiling:
![GLideN64_ZELDA_MAJORA'S_MASK_001](https://user-images.githubusercontent.com/57850880/70964677-4c6f6f80-2052-11ea-9074-cf3cadfb6469.png)
![GLideN64_ZELDA_MAJORA'S_MASK_003](https://user-images.githubusercontent.com/57850880/70964679-5002f680-2052-11ea-993a-09458e1f4bcb.png)

gonetz:
Please give me a savestate from this place.

IronYoshi64:
Sure! Here is 2 saved states for beneath the graveyard. One for 1st night area, and one for the 2nd night area, so you can see both areas. For Zelda Majora's Mask (U) [!] version.
[zelda mm beneath graveyard night 1.pj.zip](https://github.com/gonetz/GLideN64/files/3973772/zelda.mm.beneath.graveyard.night.1.pj.zip)
[zelda mm beneath graveyard night 2.pj.zip](https://github.com/gonetz/GLideN64/files/3973776/zelda.mm.beneath.graveyard.night.2.pj.zip)

gonetz:
Thanks!
The regression caused by commit 2d712f2ae0
It is bad news since this commit changes a lot.

IronYoshi64:
You are welcome! Also bad news? You mean that fixing this 1 issue will cause so many other things to be broken now, or what do you mean?

gonetz:
[...]

PR #2202, opened 2020-04-07
Texture coordinate handling refactor

standard-two-simplex:
NOT TO BE MERGED YET.
It is missing GLES part. I can write it if you want but I have no way to test it, so I'll make mistakes and someone will have to correct them. 

Quote from the [manual](http://n64devkit.square7.ch/pro-man/pro12/12-04.htm)
> The Texture Engine takes s/w, t/w, 1/w, and LOD values for a pixel and fetches the onboard texture memory (TMEM: special memory of texture built-in RDP) for the four nearest texels to the screen pixel. 

This refactors how texture coordinates are handled (not calculated) and tries to make the code more modular. I implemented the texture engine module trying to adjust to the programming manual and leaving the texture filter only the task of mixing colors. 

It uses non-normalized texture coordinates to keep the values closer to that of a N64. Most of the changes preserve the math, so the picture should be the same as in current master. Only clamp, wrap and mirror part has been changed considerably, in view of https://github.com/gonetz/GLideN64/issues/2155 (fixed).

It needs testing and it can probably use some more cleanup.


standard-two-simplex:
@gizmo98 I took this out because I couldn't understand what the algorithm is doing. Could you explain what it is doing?
```
"    if(p0q0.a > p1q0.a) p1q0.rgb = p0q0.rgb;		\\\n"
"    if(p1q0.a > p0q0.a) p0q0.rgb = p1q0.rgb;		\\\n"
"    if(p0q1.a > p1q1.a) p1q1.rgb = p0q1.rgb;		\\\n"
"    if(p1q1.a > p0q1.a) p0q1.rgb = p1q1.rgb;		\\\n"
"    if(p0q0.a > p0q1.a) p0q1.rgb = p0q0.rgb;		\\\n"
"    if(p0q1.a > p0q0.a) p0q0.rgb = p0q1.rgb;		\\\n"
"    if(p1q0.a > p1q1.a) p1q1.rgb = p1q0.rgb;		\\\n"
"    if(p1q1.a > p1q0.a) p1q0.rgb = p1q1.rgb;		\\\n"
```
It seems to me that either all samples will end up with the same rgb or either two with one rgb and two with another. Also the algorithm changes the result completely if two alpha channels are almost equal, I.e., result is completely different if `p0q0.a=0.71` and `p0q1.a= 0.72` and viceversa which is a little bit supicious.

gonetz:
Very interesting and promising approach!
 #2155 is fixed indeed. However, some mirror-related things are broken.
Here my saves for Zelda MM for Project64, where texture issues are obvious:
https://drive.google.com/file/d/1zgNm66oQ3yHgvklt2dhuNeOQtewBLmcY/view?usp=sharing
Please check them.

standard-two-simplex:
> some mirror-related things are broken

Ok. I will check.

I also notice that the menus in Vigilante 8 are broken. Do you know if it uses YUV? I'm not sure I supported it.

gonetz:
> I also notice that the menus in Vigilante 8 are broken. Do you know if it uses YUV? I'm not sure I supported it.

Those, which now look as a pile of pixels are YUV.

standard-two-simplex:
> some mirror-related things are broken

I'm not sure which effects you mean. Is some setting breaking the effects? Can you post a pic?

gizmo98:
@standard-two-simplex
I let the texture bleed like suggested here:
http://www.adriancourreges.com/blog/2017/05/09/beware-of-transparent-pixels/

Certainly there is room for improvement.

standard-two-simplex:
> I let the texture bleed like suggested here:
http://www.adriancourreges.com/blog/2017/05/09/beware-of-transparent-pixels/

I read the article and using premultiplied alpha seems enough. Supressing rgb of a texel with 0 alpha is not neccessary with that method and it can be problematic if a texture uses partial transparency and different colors in nearby textures.

gonetz:
> I'm not sure which effects you mean. Is some setting breaking the effects? Can you post a pic?

Just load my saves for Zelda MM with master build and with your fix - the difference is obvious.

standard-two-simplex:
> Just load my saves for Zelda MM with master build and with your fix - the difference is obvious.

I think the issue is with HD textures. I also found out that frambuffer textures are giving some trouble with non-native resolutions, probably because non-normalized texture coordinates are not adapted to the new texture size.

Can you confirm this is what we're speaking about?

gonetz:
> Can you confirm this is what we're speaking about?

Yes. I did not notice that HD textures are enabled, sorry.
Your fix works correctly without HD textures.

> probably because non-normalized texture coordinates are not adapted to the new texture size.

Most likely they are not.

gonetz:
> Support YUV textures

Fixed!

standard-two-simplex:
I diagnosed the issue and effectively the problem is that non-normalized coordinates are too small for the new texture. I need the ratio `HD texture width / regular texture width` to fix the problem, but I'm not sure how to compute it.

gonetz:
I think the simplest way is to store the ratio HD texture width / regular texture width in CachedTexture.
You need to add two fields, say 
```
f32 hdRatioS; //HD texture width / regular texture width
f32 hdRatioT; //HD texture hight / regular texture height
```
fill them in _updateCachedTexture
and pass them as uniforms to shaders.

To get current textures:
```
TextureCache & cache = textureCache();
cache.current[0] // Tex0
cache.current[1] // Tex1
```

standard-two-simplex:
Thanks. Got HD textures working, but I need to initialize hdRatio to 1 for regular textures. What is the right place for that?

gonetz:
At the end of TextureCache::update  and TextureCache::_updateBackground(), where pCurrent is initialized.

standard-two-simplex:
Ok. Hires textures are now working.
I have some trouble with framebuffered textures though. I initialized `hdRatio` in `_initFrameBufferTexture()` but there seem to be uninitialized frame buffer textures yet. Is there any other place where they are set up?

gonetz:
_initSubTexture
_initDepthBufferTexture
_initDepthImageTexture

gonetz:
TextureCache::_initDummyTexture

standard-two-simplex:
It turns out the one resisting wasn't frambuffer texture but RDRAMtoColor texture. I think it is working now.

gonetz:
All texrects are broken if hires textures enabled:
GLideN64_ZELDA_MAJORA'S_MASK_028

gonetz:
@standard-two-simplex Fixed now. Super mega cool!
I've tested this branch a bit. It works super well now, very impressive. I found only one new bug, Paper Mario:
PMFlame

It is a frame buffer effect, since aux buffer is used to render the flame.
I suppose that not all buffer scaling issues are solved.
Save state:
https://drive.google.com/file/d/1P4dpFIqM0EmBMUU6pQZOZSBzjq-UXiXP/view?usp=sharing

I guess there should be other frame buffer related issues. Your approach works well in general, so I hope these issues are not so hard to fix.

olivieryuyu:
bug in intro of Conker (1st screen)

GLideN64_CONKER_BFD_000

standard-two-simplex:
>  I found only one new bug, Paper Mario

I haven't guessed what's causing it yet. It looks ok in native so it must be due to bigger than expected buffer size. 

Do you know how it renders that scene? Like does it use one auxiliary framebuffer to render both (red and orange) flames or does it use two?

> bug in intro of Conker (1st screen)

I see Conker intro correctly. It is likely due to old shader cache.

gonetz:
> Do you know how it renders that scene?

I'm not sure. I investigated it long ago, but can't find my article about it.
It seems that it is one texture. It created dynamically in an aux buffer.
Two textures are used to create image of the flame in the 8bit aux buffer.
The resulted texture is 8bit intensity, and color combiner blends it with constant colors. It becomes orange in its center because of color overflow, see N64 color wrap and clamp on floats in src\Graphics\OpenGLContext\GLSL\glsl_CombinerProgramBuilder.cpp
You should remember it, since you helped me to implement this functionality. May be the problem is with these functions?

gonetz:
I also see Conker intro correctly.

gonetz:
> May be the problem is with these functions?

No, it is hardly possible. Color becomes orange because of ShaderSignExtendColorC(), which just wraps the output of color combiner. So, either input texture is wrong or it is fetched somehow incorrect.

standard-two-simplex:
0729cc1 isn't completely correct. It breaks some backgrounds in Zelda.

standard-two-simplex:
> I found only one new bug, Paper Mario:
Fixed in 581ae2c4e58da50c475330b143259de04197178f

The only remaining known issue is Mario Tennis intro.


It seems that the texture has fewer texels than (1<<mask). Therefore, the texture coordinates are out of bounds. In master, `texture()` function is used to fetch texels  so hardware wrapping conceals the issue. Since I changed functions to `texelFetch()`, fetching the texture out of bounds returns a black pixel.

I got this "fixed" in e1f9a57 by setting a maximum wrap width to `pTexture->width`, but it creates issues with ISS64. I need to investigate further, but I see that sometimes you take the power of two of this variable. @gonetz  Is there a chance that `pTexture->width` doesn't hold the actual width?

gonetz:
> Is there a chance that pTexture->width doesn't hold the actual width?

No, pTexture->width should hold the actual texture width. TextureCache::_getTextureDestData loads texture data of 'width x height' size.

> I see that sometimes you take the power of two of this variable. 

It is probably for weird cases when non-pot texture is wrapped.

standard-two-simplex:
It seems that ISS64 sets the mask to 9 and then uses texture coordinates slightly smaller than 1024 = 2^10, even if the texture is much smaller. This way, they are sort of using negative numbers while working with unsigned integers and they get the mirroring effect they wanted. So it seems unwise to do e1f9a57. 

I can hide the issue by falling back to `texture()` call. I see 3 options.
- Use `texture()` everywhere for consistency reasons.
- Use `texelFetch()` everywhere except for bilinear filter, where this issue happens. Open an issue about this and add a comment in the code that says it is a workaround.
- Use `texelFetch()` everywhere. Open an issue about this and deal with the graphical glitch until it is solved.

Personally, I like using `texelFetch()` because it is meant to be a texture lookup, not a filtered texel fetch. The second approach has the benefit of showing a decent image. The third one has the benefit of possibly surfacing other related issues, but comes at the cost of a graphical glitch. It's your call.

gonetz:
We are not in a rush. I prefer to make things right.
Unfortunately, I not always can find proper solution for texturing problems, so I add workarounds.
These workarounds often cause other problems, sometimes well hidden.
So, it will be better to reveal other related issues in hope that a solution will be found.
I suggest to apply a mix of the second and third option:
use texelFetch() everywhere but add a workaround under define, like
```
#if 1
texelFetch()
#else
workaround()
#endif

```

gizmo98:
Dumb question. Has it something to do with 64bit alignment? So texture width is actually a multiple of 2,4,8,16?http://fgfc.ddns.net/PerfectGold/Textures.htm

olivieryuyu:
what i can confirm is that any RDRAM access is done in 64bit, whatever it is for RSP or RDP. Fetching a texture should fall under this feature.

standard-two-simplex:
> use texelFetch() everywhere but add a workaround under define

Done. I think the branch is ready for the testing. I will still add some cleanup though.

> Dumb question. Has it something to do with 64bit alignment? So texture width is actually a multiple of 2,4,8,16?http://fgfc.ddns.net/PerfectGold/Textures.htm

Current implementation (both master and this branch) wraps at power of 2 boundaries indicated by mask, i.e., `2^maskValue`. The problem is maskValue is greater than 4 but the texture has less than 16 texels in each direction. This means that the texture coordinates aren't guaranteed to be inside the texel after wrapping. The question is, why is it wrapping at byte 16, 32 or greater when it has fewer texels?

For example you could load an 8x8 texture and then call `gDPSetTile()` and write `4` to mask values. With mirror enabled, this would result in coordinates `[0,...,7]` being regularly mapped to texels, `[25,...,31]` mapped mirrored and `[8,...,24]` who knows what will happen.  It isn't very logical to do it as you can acomplish the same with `mask=3`, but I haven't found a reference why it might be disallowed. In fact, I believe from a quick test that ISS64 is doing something like this in the menus, but I didn't check thoroughly, so I could be wrong.

gonetz:
> Done. I think the branch is ready for the testing.

Great!
Anyone, who want to test that branch, test build:
https://ci.appveyor.com/project/gonetz/gliden64/build/artifacts

standard-two-simplex:
I pushed some cleanup. [This branch ](https://github.com/standard-two-simplex/GLideN64/tree/texcoord_refactor_3) has the same commits but squashed into 3. I think the only thing remaining is GLES integration.

gonetz:
>  I think the only thing remaining is GLES integration.

@fzurita could you please test branch
https://github.com/gonetz/GLideN64/tree/texcoord_refactor
on compatibility with GLES2/3 and correct it if necessary?

fzurita:
Sure, I can take a look.

standard-two-simplex:
I didn't write GLES part, so unless you added it, there are missing parts where the code branches (i.e., `glinfo.isGLES()` calls and such).

@fzurita If you want I can add it, but ill probably make sintax mistakes.

gonetz:
> If you want I can add it, but ill probably make sintax mistakes.

@standard-two-simplex Please add GLES code for this branch:
https://github.com/gonetz/GLideN64/tree/texcoord_refactor
It is compacted version of this PR from your repository.
I want to merge it to master, but absence of GLES code stops me.

fzurita:
Yeah, sorry, I would do it... But this whole covid-19 thing has really limited how much time I have due to schedule shift and various other factors.

gonetz:
I understand. I hope that if @standard-two-simplex does preliminary work for GLES it will be easier for you to test and correct it.

standard-two-simplex:
Ok, I will write a preliminary commit. I have two questions though.
- Why is the code branched in the first place? Are there functions don't work in GLES?
- Why are preprocessor defines used in GLSL or avoided in GLES?

loganmc10:
To be clear as well, when we say GLES in this context, I think we mean GLES2. It might be worth considering just dropping support for GLES2, maybe making a new Public Release and saying "PR 5 is the last release that will support GLES2"?. The newest Raspberry Pi supports GLES3, and I'm pretty sure any phone released in the last 4-5 years would support GLES3

fzurita:
I don't think I would get many complaints if we drop GLES2 support.

Most GLES2 devices are too slow to run GLideN64 at am acceptable performance.

standard-two-simplex:
Oh, I didn't realize that GLES3 capable phones use the same path as GLSL. I thought GLES2 path was the one taken by most mobile devices.

Most of the branching happens for GLES2 only. The code would look neater without it, that's for sure. Anyway, I'll write the GLES2 part for this PR, it can be dropped later if it is decided so.

@fzurita Any considerations I should bear in mind for GLES2 side?

fzurita:
I would implement it as you see fit. The main consideration will be making sure that precision is correctly declared everywhere and that most GLES2 devices well be on the slow side.

standard-two-simplex:
@fzurita I pushed a preliminary commit into this [branch](https://github.com/standard-two-simplex/GLideN64/tree/texcoord_refactor). If you need anything just ask.

I hope I didn't forget too much. Check that function `TextureNearest()` is not being defined twice.

gonetz:
I need users feedback for this PR so I merged it to master. GLES issues will be fixed in master.

fzurita:
Unfortunately, this introduced a pretty severe slowdown even in pretty modern devices to the point where my Qualcom SD 821 device can no longer run the goldeneye menus at full speed.

Issue #2197, opened 2020-03-31
FlappyBird HUD elements not displayed


loganmc10:
FlappyBird is a homebrew ROM that runs on an actual N64:

[FlappyBird.zip](https://github.com/gonetz/GLideN64/files/4406749/FlappyBird.zip)

Here is how it looks in Angrylion:
![Screenshot from 2020-03-30 22-26-49](https://user-images.githubusercontent.com/848146/77987036-f27b7880-72d5-11ea-8796-bd82dde01f4d.png)

Here is how it looks in GLideN64:
![Screenshot from 2020-03-30 22-28-05](https://user-images.githubusercontent.com/848146/77987044-fc04e080-72d5-11ea-9967-3123995ffbbc.png)

This game doesn't use the RSP at all (you can run it without an RSP plugin). It does however call ProcessRDPList from the CPU/emulator, it didn't show anything until 3f55f30ea2feab4f9d561aa977aa8d1a67bf2ee5

On mupen64plus, you need to run this game with the Pure Interpereter (R4300Emulator=0). I'm not sure if it boots in Project64, I would suspect not.

I tried various FB Emulation options, as well as the new_lle branch, no luck

loganmc10:
@gonetz , commenting out these lines:
https://github.com/gonetz/GLideN64/blob/master/src/BufferCopy/RDRAMtoColorBuffer.cpp#L305-L306

Fixes the issue, it's unclear to me if that is some kind of optimization? Or is it to make sure data isn't overwritten when it shouldn't be?

gonetz:
These lines were added in b085593d2a
I suppose, you need to check Dr.Mario for regressions.

PR #2214, opened 2020-04-16
FlappyBird64 requires CopyFromRDRAM

loganmc10:
Fixes https://github.com/gonetz/GLideN64/issues/2197

With FBInfo enabled (on mupen64plus) all the display elements are shown correctly, same as Angrylion

### DIFF
--- a/ini/GLideN64.custom.ini
+++ b/ini/GLideN64.custom.ini
@@ -1,6 +1,6 @@
 ; Custom game settings
 [General]
-version=13
+version=14
 
 [TWINE]
 Good_Name=007 - The World Is Not Enough (E)(U)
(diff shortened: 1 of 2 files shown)


loganmc10:
Note that mupen64plus also needed a small modification to its FBInfo implementation: https://github.com/mupen64plus/mupen64plus-core/pull/751 so this won't work with the current master yet

loganmc10:
This is a kind of interesting game for graphics by the way, even though it's a very basic game. It doesn't use the RSP at all, the CPU sends commands directly to the RDP, similar to KI/GoldenEye skies, and then it does some CPU framebuffer effects on top of that, similar to the Dr. Mario pill

loganmc10:
@gonetz I removed those 2 lines in RDRAMtoColorBuffer::copyFromRDRAM, I didn't see any regression in Dr Mario.

This game works with FBInfo or CopyFromRDRAM, I chose to enable CopyFromRDRAM in custom.ini since that works with more emulators, and it's a 2D game, so it won't really benefit from any upscaling

gonetz:
Merged 686fa7d1

PR #2232, opened 2020-04-28
Null free

project64:
fix a possible double free, if destroy was called twice then the free was calling on the old address.

Make sure it is set back to null after being freed


gonetz:
merged: 415cde8c52315

PR #2242, opened 2020-05-05
Update Hybrid Texture Filter

oddMLan:
This version doesn't use texel_fetch, hopefully fixes any slowdowns in mobile devices.
Also it offloads the bilinear filtering to the hardware.
Thanks to DKO for the original code and for the update


oddMLan:
I hope @fzurita has a chance to report if this fixes the slowdowns on mobile

@gonetz I thought about one more possible optimization: only apply hybrid filter if input smaller than output buffer. If input buffer is bigger than output buffer, the filter isn't helpful and it's wasted processing. Native bilinear filtering is enough in this case.
I don't know where is the right part to do this conditional check

gonetz:
> I don't know where is the right part to do this conditional check

Ok, I'll find a good place for it.

gizmo98:
@dankcushions you recently disabled hybrid filter because of performance reasons.
https://github.com/RetroPie/RetroPie-Setup/commit/c8af7a69fe2ce032fd4766a0cc6188ca93632e50

Could you test the new implementation?

dankcushions:
tested on Goldeneye 007:
hybrid filter on gives blank screen during seemingly all 3d scenes. on the menu select screen is visible. it was working before (albeit slowly!)

> @gonetz I thought about one more possible optimization: only apply hybrid filter if input smaller than output buffer. If input buffer is bigger than output buffer, the filter isn't helpful and it's wasted processing. Native bilinear filtering is enough in this case.

i agree with this change but it looks like now bilinear filtering is applied ALWAYS if hybrid filter is disabled. please change that back to nearest-neighbour if input smaller than output. bilinear filtering is blurry and horrible.

dankcushions:
regression occurred here: https://github.com/gonetz/GLideN64/commit/438bcc460f07ad3c9e25b8bd4e72581a443c2a7c

oddMLan:
With commit f5b4151aff99ff374d30c47daaeb8db3f9616fc8 I'm getting a breakpoint error in GoldenEye 007 with PJ64, both qt and wtl uis.

Maybe related?
image

oddMLan:
Now if the regression @dankcushions is talking about is related or not to commit https://github.com/gonetz/GLideN64/commit/f5b4151aff99ff374d30c47daaeb8db3f9616fc8 or https://github.com/gonetz/GLideN64/commit/438bcc460f07ad3c9e25b8bd4e72581a443c2a7c, to find out the crash regression must be fixed first.

With commit 438bcc4 I get proper picture in GoldenEye 007 on desktop.

dankcushions:
Created a debug build of current master. gliden64.log:
```
2020/05/07,14:24:22.979,glsl_Utils.cpp:40,ERROR, "shader_compile error: 0:15(9): error: could not implicitly convert operands to arithmetic operator
0:15(9): error: operands to arithmetic operators must be numeric
0:15(2): error: `return' with wrong type error, in function `norm2denorm' returning vec2
0:20(9): error: could not implicitly convert operands to arithmetic operator
0:20(2): error: `return' with wrong type error, in function `denorm2norm' returning vec2
0:50(2): error: No precision specified in this scope for type `float'
0:57(2): error: No precision specified in this scope for type `float'
"
```
weirdly it also now crashes out after that log. console output: 
```
mupen64plus: /home/pi/RetroPie-Setup/tmp/build/mupen64plus/GLideN64/src/Graphics/OpenGLContext/GLSL/glsl_Utils.cpp:95: static GLuint glsl::Utils::createRectShaderProgram(const char*, const char*): Assertion `checkShaderCompileStatus(fragment_shader_object)' failed.
```
in any case, maybe that gives a clue

oddMLan:
Okay looks like I missed changing `float` to `mediump float` for GLES

oddMLan:
https://github.com/gonetz/GLideN64/pull/2247 should tentatively fix the shader compilation errors in GLSL ES mode

PR #2247, opened 2020-05-07
Fix shader compilation errors in GLSL ES

oddMLan:


oddMLan:
There are still a few errors left
```
2020/05/07,15:56:11.160,glsl_Utils.cpp:40,ERROR, "shader_compile error: 0:15(9): error: could not implicitly convert operands to arithmetic operator
0:15(9): error: operands to arithmetic operators must be numeric
0:15(2): error: `return' with wrong type error, in function `norm2denorm' returning vec2
0:20(9): error: could not implicitly convert operands to arithmetic operator
0:20(2): error: `return' with wrong type error, in function `denorm2norm' returning vec2
```

dankcushions:
works for me on rpi4 / GLES3 now 👍

oddMLan:
@dankcushions reported that this version has no noticeable slowdowns 🍻

PR #2250, opened 2020-05-07
disable threadedVideo by default

### LINKED ISSUE #2249: Threaded video should be off by default
Threaded video is causing the Zelda games (Ocarina of Time + Majora's Mask) to break, either freeze at boot or crash randomly. This happens on Windows and Android and on other platforms as well I assume.

It should be off by default until threaded video is stable.

tim241:
fixes #2249


dankcushions:
is it an OSX issue though? issue doesn't state platform. in any case, i think we should debug the crash in the first instance.

tim241:
Good point, I'm sorry that I didn't properly think this through.

PR #2265, opened 2020-05-12
fix floating point precission issue

standard-two-simplex:


oddMLan:
You were so quick on your feet to fix it, Tysm!

Issue #1630, opened 2017-10-10
Color buffer copy to RDRAM in async mode is partially broken?


Jj0YzL5nvJ:
Color buffer copy to RDRAM in async mode is partially broken, at least in Banjo-Kazooie (the intro effect on the jigsaw puzzle). And the problem is old, is present in bc00985f33c8b1e2e63cebabe53d01f9cf3708a6 too.

 I test this meticulously in deaf61299f12168b775d7e2d448b9eca149c0e7e and in `3cf7377 threaded_GLideN64` (I can't find it now) branch of @fzurita, with and without the changes sugested by @loganmc10 (https://github.com/gonetz/GLideN64/issues/1561#issuecomment-326972673) as part of my research problem in #1561.
 I think you will find my results very interesting.

 The broken jigsaw puzzle effect:
![banjo-kazooie-008](https://user-images.githubusercontent.com/12230570/31359335-cb280cc2-ad06-11e7-9f3a-2b215e83623f.png)

 The correct jigsaw puzzle effect:
![banjo-kazooie-009](https://user-images.githubusercontent.com/12230570/31359355-e7760c80-ad06-11e7-9b8c-6d5df8eb31ca.png)

 The broken jigsaw puzzle effect is always present in this master branch with `EnableCopyColorToRDRAM = 2` and works correcly with `EnableCopyColorToRDRAM = 1`, in HLE and LLE alike.  By another hand, the thing become interesting in the @fzurita `threaded_GLideN64` branch.

 In `threaded_GLideN64` with `ThreadedVideo = False` and  `EnableCopyColorToRDRAM = 2`, all works correctly. But with the `bufferStorage = false` change in the code the thing works exactly like the master branch.
 With `ThreadedVideo = True` the thing become more weirder. In **HLE mode** and the normal code, it works like the master branch. But in **LLE mode** all works again.
 But with `ThreadedVideo = True` and the `bufferStorage = false` in the code, the thing broke like if EnableCopyColorToRDRAM were set to 0 (the classic black puzzle).
 Like in the master branch, `EnableCopyColorToRDRAM = 1` always works "correctly". But I think that `ThreadedVideo = True` is broken with `EnableCopyColorToRDRAM = 1`, the performance is worst that with `ThreadedVideo = False` and uses less CPU... I don't know.

 In general, the`threaded_GLideN64` branch works best on my system. Even with my current lag problems...
[...]

PR #2270, opened 2020-05-15
Add custom setting for 1080 Snowboarding

dmrlawson:
After the title screens the intro is completely corrupted and the game is not playable unless `copyColorToRDRAM` is set to `Async` in mupen64plus_next. I think this is the same setting.

I'm a bit unsure whether the header is important in the ini file and also whether it needs `(U)` etc. I haven't tested but I would guess it needs to be applied to all versions.

### DIFF
--- a/ini/GLideN64.custom.ini
+++ b/ini/GLideN64.custom.ini
@@ -1,10 +1,14 @@
 ; Custom game settings
 [General]
-version=13
+version=14
 
 [TWINE]
 Good_Name=007 - The World Is Not Enough (E)(U)
 frameBufferEmulation\N64DepthCompare=1
+
+[1080%20Snowboarding]
+Good_Name=1080 Snowboarding
+frameBufferEmulation\copyToRDRAM=2
 
 [BIOFREAKS]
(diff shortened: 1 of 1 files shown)


dmrlawson:
Also, I realise this is the default, but it defaults to off for some platforms in mupen64plus. Maybe this isn't the correct place? It is required to make the game work though.

m4xw:
> 
> 
> After the title screens the intro is completely corrupted and the game is not playable unless `copyColorToRDRAM` is set to `Async` in mupen64plus_next. I think this is the same setting.
> 
> I'm a bit unsure whether the header is important in the ini file and also whether it needs `(U)` etc. I haven't tested but I would guess it needs to be applied to all versions.

Its `ctDoubleBuffer` here (= 2).

Jj0YzL5nvJ:
If `*CopyColorToRDRAM = 2` is solving some issue, then it's very likely that the real issue is in `ThreadedVideo` #1630

dmrlawson:
As far as I'm aware it's not affected by ThreadedVideo

PR #2276, opened 2020-05-17
Fix out of bounds texel fetch. Remove Mario Tennis workaround.

standard-two-simplex:
This should work in the same way as the workaround, but with nicer code.


gonetz:
It works. Cool, thanks!

PR #2277, opened 2020-05-18
Fix error introduced by c14ae30

standard-two-simplex:


loganmc10:
@gonetz A lot of people are using the nightly builds by the way, it would be good to either disable binary shader storage until a public release, or increase the shader binary version when the shaders are modified, or people will keep running into bugs

PR #2292, opened 2020-06-13
Fix type cast error for 64bit compilation

paradadf:
`RSP_LoadMatrixX86.cpp:92:28: error: cast from 'u8* {aka unsigned char*}' to 'int' loses precision [-fpermissive]
      : "f"(recip), "S"((int)RDRAM+address), "D"(mtx), "c"(4)`


gonetz:
From one side, RSP_LoadMatrixX86.cpp contains x86 asm code and should not be used for 64bit compilation. RSP_LoadMatrix.cpp should be used instead.
From the other side the fix itself is correct. Merged: 4c8a40c6

PR #2297, opened 2020-06-19
Allow GLideNUI to work with mupen64plus

Rosalie241:
I want to be able to use upstream GLideN64 with GLideNUI on a custom mupen64plus (i.e m64p by Logan).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -4,6 +4,7 @@ option(EGL "Set to ON if targeting an EGL device" ${EGL})
 option(PANDORA "Set to ON if targeting an OpenPandora" ${PANDORA})
 option(ODROID "Set to ON if targeting an Odroid" ${ODROID})
 option(MUPENPLUSAPI "Set to ON for Mupen64Plus plugin" ${MUPENPLUSAPI})
+option(MUPENPLUSAPI_GLIDENUI "Set to ON for GLideNUI for Mupen64Plus" ${MUPENPLUSAPI_GLIDENUI})
 option(MESA "Set to ON to disable Raspberry Pi autodetection" ${MESA})
 option(VERO4K "Set to ON if targeting a Vero4k" ${VERO4K})
 option(ANDROID "Set to ON if targeting an Android device" ${ANDROID})
@@ -16,6 +17,7 @@ add_custom_command(
         COMMAND ${CMAKE_SOURCE_DIR}/getRevision.sh
         COMMENT "Generate Git version"
 )
+set_property(SOURCE ${PATH_REVISION} PROPERTY SKIP_AUTOGEN ON)
 
 project( GLideN64 )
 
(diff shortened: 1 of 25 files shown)


tim241:
I think it's ready for review now

loganmc10:
I'll do some testing on this, it would be nice to have this capability upstream.

One question, what is the purpose of the fullscreen resolution code for Unix/Linux? I don't think the fullscreen resolution, or refresh rate setting, are actually used anywhere for mupen64plus, they are only used in windows_DisplayWindow.cpp (zilmar spec). It kind of looks like dead code, unless I am missing something

tim241:
> I'll do some testing on this, it would be nice to have this capability upstream.
> 
> One question, what is the purpose of the fullscreen resolution code for Unix/Linux? I don't think the fullscreen resolution, or refresh rate setting, are actually used anywhere for mupen64plus, they are only used in windows_DisplayWindow.cpp (zilmar spec). It kind of looks like dead code, unless I am missing something

hmm, yeah, it seems like 'dead code' at the moment, however, I implemented the fullscreen resolution code to make sure I didn't need to change the UI *at all*, anyways, wouldn't it be possible to implement support for it in `mupen64plus_DisplayWindow.cpp`? Maybe it'd be possible to hook it up with xlib for unix platforms?

loganmc10:
@tim241 if you're going to attempt to implement the full screen resolution options in mupen64plus, you should use the builtin API function: ```CoreVideo_ListFullscreenModes``` to determine the fullscreen resolution options.

Documentation for the function is here:
https://github.com/mupen64plus/mupen64plus-core/blob/master/doc/emuwiki-api-doc/Mupen64Plus-v2.0-Core-Video-Extension.mediawiki

Basically you call that function and it provides the full screen resolution options, that way the platform specific code lives in the frontend, not the plugin. This is the way.

Rosalie241:
Considering the mupen64plus PR has been merged, I've modified the commit's description and changed the patch a tiny bit to clean it up, I think it's ready for review now!

gonetz:
@loganmc10 could you please do the review?

loganmc10:
I don't think I'll have time to test but I'll make some inline notes

src/CMakeLists.txt
```diff
@@ -152,13 +180,23 @@ if(MUPENPLUSAPI)
...
 	MupenPlusPluginAPI.cpp
-	mupenplus/Config_mupenplus.cpp
 	mupenplus/CommonAPIImpl_mupenplus.cpp
 	mupenplus/MemoryStatus_mupenplus.cpp
 	mupenplus/MupenPlusAPIImpl.cpp
 	Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
   )
 
+  if(MUPENPLUSAPI_GLIDENUI)
+      add_definitions(-DGLIDENUI)
```
loganmc10, line 190: I would rename this define to something like M64P_GLIDENUI to make it clear it's only related to mupen64plus+GLideNUI

src/Graphics/OpenGLContext/mupen64plus/mupen64plus_DisplayWindow.cpp
```diff
@@ -173,7 +179,32 @@ bool DisplayWindowMupen64plus::_resizeWindow()
 
 void DisplayWindowMupen64plus::_changeWindow()
 {
+#ifdef GLIDENUI
```
loganmc10, line 182: I think it would be nice to have a separate define for GLIDENUI_SUPPORTS_FULLSCREEN or something like that. Full screen resolution detection is pretty platform specific and might not work well on all frontends.

Rosalie241: I think it'd be saner if gliden64 would check if that function returns `M64ERR_UNSUPPORTED` and if it does, fall back to the original code

loganmc10: Yes, in ```DisplayWindowMupen64plus::_start()``` you could try to execute ```FunctionWrapper::CoreVideo_SetVideoModeWithRate```, if that returns ```M64ERR_UNSUPPORTED```, that tells you that the frontend doesn't support full screen resolution setting, and the rest of the code should execute based on that knowledge

Rosalie241: Implemented.

src/CMakeLists.txt
Rosalie241: Fixed

PR #2313, opened 2020-07-28
Correct filtering of video window

### LINKED ISSUE #2311: HD textures show aliasing
Probably introduced by 2f5280c

aliasing

Top: current
Bottom: before

oddMLan:
Fixes #2311

Commit d7d68c46 (current master)

**Native resolution, Hybrid filter ON**
GLideN64_GOLDENEYE_000

This PR
GLideN64_GOLDENEYE_001

**Native resolution, Hybrid filter ON, overscan**
GLideN64_GOLDENEYE_004

**Native resolution, Hybrid filter OFF, overscan**
GLideN64_GOLDENEYE_002

**Native resolution, Hybrid filter OFF**
GLideN64_GOLDENEYE_003

Commit d7d68c46 (current master)
4x native resolution Hybrid filter ON
GLideN64_MarioTennis_000

This PR
GLideN64_MarioTennis_001

### DIFF
--- a/src/GLideNUI-wtl/Settings.cpp
+++ b/src/GLideNUI-wtl/Settings.cpp
@@ -48,6 +48,9 @@ void _loadSettings(GlSettings & settings)
 	config.generalEmulation.enableLOD = settings.value("enableLOD", config.generalEmulation.enableLOD).toInt();
 	config.generalEmulation.enableHWLighting = settings.value("enableHWLighting", config.generalEmulation.enableHWLighting).toInt();
 	config.generalEmulation.enableShadersStorage = settings.value("enableShadersStorage", config.generalEmulation.enableShadersStorage).toInt();
+	config.generalEmulation.enableLegacyBlending = settings.value("enableLegacyBlending", config.generalEmulation.enableLegacyBlending).toInt();			 //ini only
+	config.generalEmulation.enableHybridFilter = settings.value("enableHybridFilter", config.generalEmulation.enableHybridFilter).toInt();					 //ini only
+	config.generalEmulation.enableFragmentDepthWrite = settings.value("enableFragmentDepthWrite", config.generalEmulation.enableFragmentDepthWrite).toInt(); //ini only
 	config.generalEmulation.enableCustomSettings = settings.value("enableCustomSettings", config.generalEmulation.enableCustomSettings).toInt();
 	settings.endGroup();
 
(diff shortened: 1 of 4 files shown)


src/GLideNUI-wtl/Settings.cpp
```diff
@@ -229,6 +232,9 @@ void writeSettings(const char * _strIniFolder)
 		settings.setValue("enableLOD", config.generalEmulation.enableLOD);
 		settings.setValue("enableHWLighting", config.generalEmulation.enableHWLighting);
 		settings.setValue("enableShadersStorage", config.generalEmulation.enableShadersStorage);
+		settings.setValue("enableLegacyBlending", config.generalEmulation.enableLegacyBlending);		 //ini only
+		settings.setValue("enableHybridFilter", config.generalEmulation.enableHybridFilter);			 //ini only
+		settings.setValue("enableFragmentDepthWrite", config.generalEmulation.enableFragmentDepthWrite); //ini only
```
gonetz, line 237: I think this section is useless since these settings have no GUI support and in fact are read-only.
The same is for the second file.

oddMLan: I could remove those sections but then these settings won't be made visible in the ini, you would have to type them manually. Personally I think that's inconvenient as you'd need to remember the exact name/spelling of these settings, check them in the source code. While in the current code they get written with the default value and do no harm. If you change the value in the ini the new value is kept. So I don't understand what you mean by "read-only" in this case

gonetz: > I could remove those sections but then these settings won't be made visible in the ini, you would have to type them manually. 

Yes, you're right.

gonetz:
So, the idea is the following:
* If hybrid filter is enabled, always use FILTER_LINEAR when render frame to screen
* If hybrid filter is disabled, use FILTER_LINEAR for down-scale image and FILTER_NEAREST for upscale.
Did I get it right?

oddMLan:
> So, the idea is the following:
> 
> * If hybrid filter is enabled, always use FILTER_LINEAR when render frame to screen
> * If hybrid filter is disabled, use FILTER_LINEAR for down-scale image and FILTER_NEAREST for upscale.
>   Did I get it right?

Yup, that's correct!
Btw, can you confirm whether the condition here
`blitParams.srcWidth >= blitParams.dstWidth || blitParams.srcHeight >= blitParams.dstHeight;`
should be || instead of &&?
It originally used && but logic dictates if either of the dimensions if downscaled it should be considered downscaling (when non-native res). Maybe you had another specific reason to use &&? Is there any case where you downscale in one dimension but upscale in another? The only game I can think of is Rugrats Scavenger hunt with a wide framebuffer (512x240). If you use 2x native the buffer size will become 1024x480, if you're rendering to, lets say, a 800x600 window, it's considered downscaling. If Hybrid Filter is enabled it doesnt matter since it will always use FILTER_LINEAR, but if it's disabled, it will use... linear as well? Stretching from 480 -> 600 shouldn't cause much blurring. In the old code with && it would've set 'downscale' to false, and resized the picture with FILTER_NEAREST which would cause significant distortion.

gonetz:
> Btw, can you confirm whether the condition here
> `blitParams.srcWidth >= blitParams.dstWidth || blitParams.srcHeight >= blitParams.dstHeight;`
> should be || instead of &&?

Yes, it looks logical to me.

> If you use 2x native the buffer size will become 1024x480, if you're rendering to, lets say, a 800x600 window, it's considered downscaling. 

Without overscan, yes, 1024x480 will be mapped to 800x600
With overscan, it does matter, how many lines VI uses to map 512x240 to TV screen. It may map 512x240 to 640x480, thus 1024x480 will first be mapped to 1280x960 and then cropped part of 1280x960 image downscaled to 800x600.

PR #2326, opened 2020-09-10
Correctly implement FindPluginPath for Linux

Rosalie241:
this uses `/proc/self/maps` to find the library, if `/proc/self/maps` can't be opened for whatever reason/it can't find the library, it'll fall back to the original code which uses `/proc/self/exe`


gonetz:
merged: 229a6bc788

Issue #2324, opened 2020-09-03
Drawing primitives with screen coordinates


standard-two-simplex:
I was investigating the wrong texture coordinates issue and wanted to move the vertex position a little, when I realized it was really hard due to the fact that they don't use a viewport transform in the N64. In order to adapt to OpenGL, current master uses the framebuffer size to normalize vertex position and apply its "compulsory" viewport transform.

Inspired on the investigation I did about the clip ratio, I realized that it is possible to draw without using the buffer size. The viewport transform is normally regarded as a function that maps the `[-1,1]x[-1,1]` rectangle into the `[x_offset, x_offset + width]x[y_offset, y_offset+height]` but it is really an affine transformation, i.e., a multiplication and an addition, given by the formula
```
(x,y) -> ((x+1)*width/2+x_offset, (y+1)*height/2 + y_offset)
```

This allows for "original" viewport transforms such as `glViewport(-1,-1,2,2)` which is in fact... the identity function!
```
(x,y) -> ((x+1)*2/2+ (-1), (y+1)*2/2 + (-1)) = (x,y) 
 ``` 
This would mean that screen coordinates would map directly into screen coordinates with such awkward viewport transform. 

There is an obstacle though: primitives are clipped to the `[-w,w]x[-w,w]` boundary, so the natural `w=1` doesn't work with screen coordinates. In order to overcome this, the following can be done,
- Multiply `w` by a big number such as `1024` at the end of the vertex shader, so that screen coordinates aren't clipped. After perspective division the vertex coordinates will be `(x/1024, y/1024)`.
- Use viewport transform given by `glViewport(-1024,-1024,2048,2048)` which is really a multiplication by 1024.
```
(x,y) ->((x+1)*2048/2 +(-1024), (y+1)*2048/2)+(-1024)) = (1024*x,1024*y)
```
All in all, screen coordinates are mapped to screen coordinates.

PROS:
- No dependence on buffer size (see the simplicity of `updateScreenCoorsViewport2()`).
- No divisions except for the perspective division, which may help avoid precision loss.
- N64-like values in vertex shader, which makes it easier to manipulate them.

CONS:
- Lost OpenGLs built in clipping. There should be ways to overcome this though.

[...]

Issue #2097, opened 2019-10-05
UVs are shifted by about half a texel


mountainflaw:
GLideN64 seems to be pushing my custom models' UVs by about half a texel. This contradicts what real hardware does, where it is properly aligned, and occurs in both HLE and LLE.

GLideN64:
![image](https://user-images.githubusercontent.com/38231712/66262032-5cdbb200-e7a6-11e9-84c4-25e682b4cce3.png)

Angrylion's Plus:
![image](https://user-images.githubusercontent.com/38231712/66262029-564d3a80-e7a6-11e9-93b1-aa653dab6c39.png)

My N64:
![image](https://user-images.githubusercontent.com/38231712/66262023-4a617880-e7a6-11e9-8f7a-f4451ef2dd23.png)

Other plugins (tested by @theboy181). Passed means proper alignment, failed means misalignment:
real hw: passed
angrylions plus: passed
gliden64: failed
glide64mk2: failed
z64gl: passed
1964's video plugin: passed
jabod3d: failed

gonetz:
Hi,

I guess, if you would test with glN64, it also would fail. 

The idea is simple: to fetch texel properly with OpenGL, we need point to its center.
Thus, 0.5 added to texture coordinates.
Glide64 does it, glN64 does it too and GLideN64 inherited it.
I tried to remove that 0.5 shift. Some things became better, some much worse.
There were discussions there about that matter, people with better memory probably will find proper threads.
I think, if we will change these lines in src/Textures.cpp
```
	pCurrent->offsetS = 0.5f;
	pCurrent->offsetT = 0.5f;

```
as
```
	pCurrent->offsetS = 0.0f;
	pCurrent->offsetT = 0.0f;

```
the problem you described will be fixed.
But then we will need to carefully investigate all regression cases.

olivieryuyu:
@gonetz 

This 0.5 does not depends on whether you are on point sample mode or BILERP mode?

mountainflaw:
^ it shouldn't shift in point sample mode, but it should with bilerp

gonetz:
Ok, let's prove it.
I made corresponding fix in fix_2097 branch
https://github.com/gonetz/GLideN64/tree/fix_2097
commit 9b1fec6494

Test build: 
https://drive.google.com/file/d/1stkn3Sf2vJiZqFVj-Q-YOly_Qt9jI1Lk/view?usp=sharing

theboy181:
looking for regressions? you find a case?

[...]

Issue #2073, opened 2019-07-24
Seam issue with sliced textures


JustNoon:
Textures comprised of multiple slices of a larger image have seams when scaled. Native as well but very apparent at higher resolution as there is a consistency problem between graphics plugins. Take for instance this texture (the two disparate textures have been given bounding boxes for illustrative purposes) from Banjo Kazooie.
![BoundingBoxes](https://user-images.githubusercontent.com/45368472/61809408-d6e4d780-ae02-11e9-9c54-9234fe9edf91.png)
There is a 1 px overlap where the same row of pixels is duplicated on both versions of the texture. This is presumably to prevent gaps due to rounding. I would assume that a texture with a 4x scaling factor would have a 4px overlap. That is not the case. 

In GlideN64, the overlap seems to be a subpixel.
![GN64-HDfix7-2](https://user-images.githubusercontent.com/45368472/61813520-7b6b1780-ae0b-11e9-8299-f7fcd00c0cfb.png)

In Glide64 Final and P64, the overlap is about half and is blended.
![G64Final](https://user-images.githubusercontent.com/45368472/61813564-9ccc0380-ae0b-11e9-8515-b967b601f117.png)
![P64-2 2 0 969-39](https://user-images.githubusercontent.com/45368472/61813569-9f2e5d80-ae0b-11e9-885e-2e0a9b4812be.png)

(Jabo is visibly seamed at native res).

I put a white, blue, green, black pattern in the 4px overlap for ease of demonstration. Here is the reference texture.
![Reference](https://user-images.githubusercontent.com/45368472/61814118-f1bc4980-ae0c-11e9-8c4d-6dc58b4dd28b.png)

This does not seem to be an issue in Mario 64's bowling balls which have the expected behavior of a 4px overlap.
![MarioCorrect](https://user-images.githubusercontent.com/45368472/61814210-2c25e680-ae0d-11e9-839c-3633828249e7.png)

Every texture merged in this way in Banjo Kazooie is like this so I've seen.

Also of note, this particular texture seems to drifting horizontally out of alignment as well on the bottom half. I don't know if this is related but it is similar at the very least. I have included all files used here as well as a savestate for v1.0 U of  BK.
[BKSeamIssue.zip](https://github.com/gonetz/GLideN64/files/3428059/BKSeamIssue.zip)

Jj0YzL5nvJ:
angrylion-plus / [VI mode = Filtered](https://0x0.st/zpUs.png)
[...]

Issue #1869, opened 2018-07-24
Resident Evil 2 Texture Alignment Issues


CallistoNTG:
Certain backgrounds have vertical lines. This problem is fairly easy to verify. Simply boot the game and enter the options menu.

![resident_evil_ii-011](https://user-images.githubusercontent.com/41622989/43136918-848bb786-8f8d-11e8-8320-f53243da0a48.png)
![resident_evil_ii-012](https://user-images.githubusercontent.com/41622989/43136920-84d14af8-8f8d-11e8-9711-a8721d9536d3.png)

gonetz:
Fixed in native res with "Use screen coordinates", #2470
Comment from standard-two-simplex: Fixed in native. In order to fix it in higher resolutions measures that may harm other games have to be taken, so I don't recommend it.

PR #2334, opened 2020-09-21
Use screen coords

standard-two-simplex:
I create a pull request to take advantage of the buildbot.

Related to #2324. If found any regressions please report in that issue thread.


standard-two-simplex:
Related issues.
- #1869 Fixed in native
- #1047 Fixed Ogre Battle dialogs, Excitebike position indicator and BAR car select screen and HUD in native.
- #1416 Fixed Zelda and Banjo Tooie
- #2097 Original issue probably fixed (@mountainflaw test please). Banjo-Kazooie, ssb and thps3 issues mentioned fixed in native.
- #2073 Probably fixed in native. Higher resolutions can show seams depending on how the game was programmed.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -502,7 +506,8 @@ class VertexShaderEnd : public ShaderPart
 			}
 		}
 		m_part +=
-			" gl_Position.zw *= vec2(uClipRatio);	 \n"
+			" gl_Position.zw *= vec2(1024.0f);		 \n"
+			// " gl_Position.zw *= vec2(uClipRatio);	\n"
```
gonetz, line 512: Please just remove it.

src/GraphicsDrawer.cpp
```diff
@@ -723,7 +745,8 @@ void GraphicsDrawer::_updateStates(DrawingState _drawingState) const
 		updateScissor(frameBufferList().getCurrent());
 
 	if (gSP.changed & CHANGED_VIEWPORT)
-		_updateViewport();
+		// _updateViewport();
```
gonetz, line 748: Old code can be removed.

src/gSP.cpp
```diff
@@ -1671,10 +1659,12 @@ void gSPModifyVertex( u32 _vtx, u32 _where, u32 _val )
...
 			vtx0.y = _FIXED2FLOAT((s16)_SHIFTR(_val, 0, 16), 2);
 			DebugMsg(DEBUG_NORMAL, "gSPModifyVertex: XY(%02f, %02f);\n", vtx0.x, vtx0.y);
+			/*
 			if ((config.generalEmulation.hacks & hack_ModifyVertexXyInShader) == 0) {
+				
 				vtx0.x = (vtx0.x - gSP.viewport.vtrans[0]) / gSP.viewport.vscale[0];
-				if (gSP.viewport.vscale[0] < 0)
-					vtx0.x = -vtx0.x;
+				//if (gSP.viewport.vscale[0] < 0)
+				//	vtx0.x = -vtx0.x;
```
gonetz, line 1663: Old code can be removed.

src/gSP.cpp
```diff
@@ -1685,9 +1675,11 @@ void gSPModifyVertex( u32 _vtx, u32 _where, u32 _val )
 				}
 
 				vtx0.y = -(vtx0.y - gSP.viewport.vtrans[1]) / gSP.viewport.vscale[1];
-				if (gSP.viewport.vscale[1] < 0)
-					vtx0.y = -vtx0.y;
+				//if (gSP.viewport.vscale[1] < 0)
+				//	vtx0.y = -vtx0.y;
```
gonetz, line 1675: Old code can be removed.

src/gSP.cpp
```diff
@@ -1685,9 +1675,11 @@ void gSPModifyVertex( u32 _vtx, u32 _where, u32 _val )
...
 				vtx0.y *= vtx0.w;
+			*/
+			if (false) {
 			} else {
```
gonetz, line 1679: Please cleanup this part: remove that hack if it is not necessary anymore (remove all the related code) and remove this "if".

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -1302,7 +1306,7 @@ class ShaderFragmentHeaderReadTex : public ShaderPart
 		if (!m_glinfo.isGLES2) {
 
 			if (g_textureConvert.useTextureFiltering()) {
-				shaderPart += "uniform lowp int uTextureFilterMode;								\n";
+				//shaderPart += "uniform lowp int uTextureFilterMode;								\n";
```
gonetz, line 1312: Just remove it.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp
```diff
@@ -216,6 +216,46 @@ class UScreenSpaceTriangleInfo : public UniformGroup
...
+			scale[1] = dwnd().getScaleY();
+		}
+		const float vertexOffset = isNativeRes ? 0.5f : 0.0f;
+		float texCoordOffset[2] = { 0.0f, 0.0f };
+		if (isTexRect && !isNativeRes) {
+			if (gDP.otherMode.textureFilter != G_TF_POINT && gDP.otherMode.cycleType != G_CYC_COPY) {
+				texCoordOffset[0] = -0.5f * gDP.lastTexRectInfo.dsdx;
+				texCoordOffset[1] = -0.5f * gDP.lastTexRectInfo.dtdy;
+			}
+			else {
```
gonetz, line 248: } else {

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp
```diff
@@ -216,6 +216,46 @@ class UScreenSpaceTriangleInfo : public UniformGroup
...
+			scale[0] = scale[1] = static_cast<float>(config.frameBufferEmulation.nativeResFactor);
+		}
+		else {
+			scale[0] = dwnd().getScaleX();
+			scale[1] = dwnd().getScaleY();
+		}
+		const float vertexOffset = isNativeRes ? 0.5f : 0.0f;
+		float texCoordOffset[2] = { 0.0f, 0.0f };
+		if (isTexRect && !isNativeRes) {
+			if (gDP.otherMode.textureFilter != G_TF_POINT && gDP.otherMode.cycleType != G_CYC_COPY) {
```
gonetz, line 244: Please add comments, why you use texCoordOffset in non-native res mode and why is that difference in code between gDP.otherMode.textureFilter modes. You may put a reference to your explanations in the #2324 GitHub task.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp
```diff
@@ -246,14 +250,46 @@ class URasterInfo : public UniformGroup {
 				texCoordOffset[1] = (gDP.lastTexRectInfo.dtdy >= 0.0f ? -0.5f / scale[1] : -1.0f + 0.5f / scale[1]) * gDP.lastTexRectInfo.dtdy;
 			}
 		}
+		float tcbounds[2][4] = {};
+		if (useTexCoordBounds) {
+			f32 s, t, dsdx, dtdy, uls, lrs, ult, lrt;
```
gonetz, line 255: 'dsdx': unreferenced local variable
'dtdy': unreferenced local variable

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp
```diff
@@ -689,8 +689,9 @@ class UViewportInfo : public UniformGroup
 
 	void update(bool _force) override
 	{
+		float ySign = !GBI.isNegativeY() ? 1.0 : -1.0;
```
gonetz, line 692: conversion from double to float.

`float ySign = GBI.isNegativeY() ? -1.0f : 1.0f;`

src/gSP.cpp
```diff
@@ -267,8 +267,8 @@ void gSPViewport( u32 v )
 	gSP.viewport.vtrans[2] = _FIXED2FLOAT( *(s16*)&RDRAM[address + 14], 10 );// * 0.00097847357f;
 	gSP.viewport.vtrans[3] = *(s16*)&RDRAM[address + 12];
 
-	if (gSP.viewport.vscale[1] < 0.0f && !GBI.isNegativeY())
-		gSP.viewport.vscale[1] = -gSP.viewport.vscale[1];
+	//if (gSP.viewport.vscale[1] < 0.0f && !GBI.isNegativeY())
+	//	gSP.viewport.vscale[1] = -gSP.viewport.vscale[1];
```
gonetz, line 271: Just remove the old code.

gonetz (changes requested):
Very good work! Lots of old issues have really gone. What is important, it is not a local hack, but a general approach, based on deep understanding of the problem. Respect!

I noticed several minor issues to clean up. My "native-res texrects" feature is broken, but I'll fix it.

src/GraphicsDrawer.cpp
```diff
@@ -321,6 +321,28 @@ void GraphicsDrawer::_updateScreenCoordsViewport(const FrameBuffer * _pBuffer) c
 	gSP.changed |= CHANGED_VIEWPORT;
 }
 
+void GraphicsDrawer::_updateScreenCoordsViewport2(const FrameBuffer* _pBuffer) const
```
gonetz, line 324: `GraphicsDrawer::_updateViewport()` is unused now and should be removed.

src/GraphicsDrawer.cpp
```diff
@@ -321,6 +321,28 @@ void GraphicsDrawer::_updateScreenCoordsViewport(const FrameBuffer * _pBuffer) c
...
+		scaleX = scaleY = pCurrentBuffer->m_scale;
+	}
+	else {
+		scaleX = dwnd().getScaleX();
+		scaleY = dwnd().getScaleY();
+	}
+	X = roundup(-1024.0f, scaleX);
+	Y = roundup(-1024.0f, scaleY);
+	WIDTH = roundup(2048.0f, scaleX);
+	HEIGHT = roundup(2048.0f, scaleY);
```
gonetz, line 340: It would be good to define constants for these numbers, for example
const f32 viewportOffsetXf = -1024.0f;
const s32 viewportOffsetX = -1024;
const f32 viewportWidthf = 2048.0f;
const s32 viewportWidth = 2048;

src/GraphicsDrawer.cpp
```diff
@@ -321,6 +321,28 @@ void GraphicsDrawer::_updateScreenCoordsViewport(const FrameBuffer * _pBuffer) c
 	gSP.changed |= CHANGED_VIEWPORT;
 }
 
+void GraphicsDrawer::_updateScreenCoordsViewport2(const FrameBuffer* _pBuffer) const
+{
+	s32 X, Y, WIDTH, HEIGHT;
+	const s32 scale = config.frameBufferEmulation.nativeResFactor;
```
gonetz, line 327: scale is not used.

src/GraphicsDrawer.cpp
```diff
@@ -1497,7 +1517,7 @@ void GraphicsDrawer::drawTexturedRect(const TexturedRectParams & _params)
 			return;
 	}
 
-	_updateScreenCoordsViewport(_params.pBuffer);
+	_updateScreenCoordsViewport2(_params.pBuffer);
```
gonetz, line 1504:
```
if (g_debugger.isCaptureMode()) {
		m_rect[0].x = _params.ulx;
		m_rect[0].y = _params.uly;
		m_rect[1].x = _params.lrx;
		m_rect[1].y = _params.uly;
		m_rect[2].x = _params.ulx;
		m_rect[2].y = _params.lry;
		m_rect[3].x = _params.lrx;
		m_rect[3].y = _params.lry;
		g_debugger.addRects(rectParams);
	}
```
Now should be just 

```
if (g_debugger.isCaptureMode()) {
		g_debugger.addRects(rectParams);
	}

```

gonetz:
I fixed Native texrects. It is in https://github.com/gonetz/GLideN64/tree/standard-two-simplex-use_screen_coords branch.

standard-two-simplex:
Thanks for the review. I'll try to take care of the issues during this week.

src/GraphicsDrawer.cpp
standard-two-simplex: Shall I add them in GraphicsDrawer.h?

src/gSP.cpp
standard-two-simplex: I deliberately left this in place because I didn't know what the hack was for. I know the coordinate inversions in the commented part (mentioned in your annotation just above) were giving me trouble, but I can't remember if the rest of the hack was problematic.

Do you know what the hack is for? Should I reenable it or discard it? I'll double check later, but if I remember correctly I saw no issues both with the hack and without it.

src/GraphicsDrawer.cpp
gonetz: I've got another idea. If you change signature of _updateScreenCoordsViewport2 to void `_updateScreenCoordsViewport2(u32 _scale = 0) const;`
then I can call it from texrect drawer with scale 1, so these constants need not be exposed.
If _scale is zero, the function should get the scale as it does it now when _pBuffer = nullptr.

gonetz: And _updateScreenCoordsViewport2  is better to rename to _updateViewport because _updateViewport() is unused now.

src/gSP.cpp
gonetz: The hack was enabled for three games, see RSP_Init().
However, you disabled the main branch, which is without hack.
That is now it works as if the hack is always enabled.
May be it is correct solution now, because previously I applied reversed viewport transformation to the provided screen space coordinate. All games, which use gSPModifyVertex command should be checked. Unfortunately, I don't remember  these games. @olivieryuyu, could you help us with it? You have good memory :)

olivieryuyu: gSPModifyVertex: Tarzan menu options or Jeremy McGrath Supercross 2000 in menu i believe. I think so All Star baseball games as well.

standard-two-simplex: > However, you disabled the main branch, which is without hack.

Mmm, I had not realized this, I must have done it unintentionally. I'll take a closer look at the situation.

src/gSP.cpp
```diff
@@ -1671,10 +1655,12 @@ void gSPModifyVertex( u32 _vtx, u32 _where, u32 _val )
 			vtx0.x = _FIXED2FLOAT((s16)_SHIFTR(_val, 16, 16), 2);
 			vtx0.y = _FIXED2FLOAT((s16)_SHIFTR(_val, 0, 16), 2);
 			DebugMsg(DEBUG_NORMAL, "gSPModifyVertex: XY(%02f, %02f);\n", vtx0.x, vtx0.y);
+			/*
 			if ((config.generalEmulation.hacks & hack_ModifyVertexXyInShader) == 0) {
+
```
Jj0YzL5nvJ, line 1660: Trailing whitespace

src/gDP.cpp
```diff
@@ -865,6 +865,16 @@ void gDPTextureRectangle(f32 ulx, f32 uly, f32 lrx, f32 lry, s32 tile, s16 s, s1
...
 
+	gDP.lastTexRectInfo.ulx = ulx;
+	gDP.lastTexRectInfo.lrx = lrx;
+	gDP.lastTexRectInfo.uly = uly;
+	gDP.lastTexRectInfo.lry = lry;
+	gDP.lastTexRectInfo.s = !flip ? s : t;
+	gDP.lastTexRectInfo.t = !flip ? t : s;
+	gDP.lastTexRectInfo.dsdx = !flip ? dsdx : dtdy;
+	gDP.lastTexRectInfo.dtdy = !flip ? dtdy : dsdx;
+
```
Jj0YzL5nvJ, line 876: Trailing whitespace ...at least with 2334.patch

Jj0YzL5nvJ (changes requested):
$ git apply ../standard-two-simplex-use_screen_coords.patch
```
../standard-two-simplex-use_screen_coords.patch:301: trailing whitespace.
				
../standard-two-simplex-use_screen_coords.patch:579: trailing whitespace.
	
warning: 2 líneas agregan errores de espacios en blanco.
```

Jj0YzL5nvJ:
I can't build...
```
[ 71%] Building CXX object CMakeFiles/mupen64plus-video-GLideN64.dir/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp.o
/tmp/build/GLideN64/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp:258:15: error: 
      use of undeclared identifier 'ceilf'
                        uls = s + (ceilf(gDP.lastTexRectInfo.ulx) - gDP....
                                   ^
/tmp/build/GLideN64/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp:259:15: error: 
      use of undeclared identifier 'ceilf'
                        lrs = s + (ceilf(gDP.lastTexRectInfo.lrx) - gDP....
                                   ^
/tmp/build/GLideN64/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp:260:15: error: 
      use of undeclared identifier 'ceilf'
                        ult = t + (ceilf(gDP.lastTexRectInfo.uly) - gDP....
                                   ^
/tmp/build/GLideN64/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp:261:15: error: 
      use of undeclared identifier 'ceilf'
                        lrt = t + (ceilf(gDP.lastTexRectInfo.lry) - gDP....
                                   ^
/tmp/build/GLideN64/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp:266:24: error: 
      use of undeclared identifier 'fmin'
                                        tcbounds[t][0] = (fmin(uls, lrs...
                                                          ^
/tmp/build/GLideN64/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp:267:24: error: 
      use of undeclared identifier 'fmin'
                                        tcbounds[t][1] = (fmin(ult, lrt...
                                                          ^
/tmp/build/GLideN64/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp:268:24: error: 
      use of undeclared identifier 'fmax'
                                        tcbounds[t][2] = (fmax(uls, lrs...
                                                          ^
/tmp/build/GLideN64/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp:269:24: error: 
      use of undeclared identifier 'fmax'
                                        tcbounds[t][3] = (fmax(ult, lrt...
                                                          ^
8 errors generated.
CMakeFiles/mupen64plus-video-GLideN64.dir/build.make:956: fallo en las instrucciones para el objetivo 'CMakeFiles/mupen64plus-video-GLideN64.dir/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp.o'
make[2]: *** [CMakeFiles/mupen64plus-video-GLideN64.dir/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp.o] Error 1
CMakeFiles/Makefile2:136: fallo en las instrucciones para el objetivo 'CMakeFiles/mupen64plus-video-GLideN64.dir/all'
make[1]: *** [CMakeFiles/mupen64plus-video-GLideN64.dir/all] Error 2
Makefile:102: fallo en las instrucciones para el objetivo 'all'
make: *** [all] Error 2
```

Jj0YzL5nvJ:
WTF, **I can build with gcc-5** but not with `gcc-7 to gcc-9` or `clang-4 to clang-8` ¬¬

fzurita:
Probably just missing this:
```
#include <cmath>
```

src/gSP.cpp
standard-two-simplex: The situation seems a bit complicated. The reverse viewport transformation is no longer needed, so I removed it. 
I am not sure about the `dwnd().isAdjustScreen()` part (seems it was introduced here https://github.com/gonetz/GLideN64/commit/038c74af123f068a36e7e6f29d7fbe28ee79618a).

Tarzan works fine without the hack, but forcing `w` to `1.0f` creates issues because `z` needs to be divided by it. The hack is still needed in Rat Attack for the projector effect in the intro and in Resident Evil 2 for the rotating Nintendo logo. It seems `w` is zero in this games, which creates issues when we multiply by it in the vertex shader and OpenGL performs division by `w`. Thus neither forcing to `1.0f` nor not forcing it seems the correct approach. I will leave the hack there for now, I don't understand the situation well enough to find a good solution.

standard-two-simplex:
I've got most of the fixes ready, I will squash the fixes to the original commits to keep history clean.
I think that if I force push the new commits to GitHub, the reviews will be lost. Shall I push to a new branch?

Jj0YzL5nvJ:
**USE THE FORCE**

src/gSP.cpp
olivieryuyu: Not sure it helps but:

gSPModifyVertex uses the data of a precalculated vertex and includes new screenspace coordinates in it (usually), new RGBA or new S&T.

The W of the vertex remains the very same than the first vertex calculated.

So your code you keep the W of the previous vertex pointed out by the commmand.

gonetz:
>Shall I push to a new branch?

Yes, clean history is more important.

src/gSP.cpp
standard-two-simplex: @olivieryuyu The problem I have is that I don't know at what stage the coordinates are injected. See [OpenGL's vertex post-processing pipeline](https://www.khronos.org/opengl/wiki/Vertex_Post-Processing).

Firstly, primitives must be assembled in order to perform clipping, as one vertex may clip into different position based on its neighbours. This makes it specially difficult to refer to vertices with indices, like gSPModifyVertex does, after clipping has occurred, because new vertices are generated dinamically.

So my GUESS is that the data is modified prior to all this, when only the vertices programmed by the developer exist. However, gSPModifyVertex uses screen space coordinates, so they need no perspective divide nor viewport transform. Assuming the data is injected prior to the vertex post-processing, the RDP should implicitly disable perspective divide and viewport transform. 

The latter should not be too hard to emulate (as it is done in vertex shader now) but the former is complicated. OpenGL enforces perspective divide, and changing w to 1 can be problematic if XY don't need the perspective divide but Z does (which seems to happen in Tarzan). Multiplying XY by w works in this case, as the perspective divide will then restore them to original values, but only because w is not 0. To make it worse, other games (like Rat Attack or Resident Evil 2) use gSPModifyVertex when w=0, so multiplying XY by w doesn't work for those games.

It might be possible to emulate this without hacks if the whole RSP is emulated in software, i.e., if the perspective divide and viewport tranform are also done in software. Then it would be possible to pass `(x_screen, y_screen, z_screen, 1)` to OpenGL, and gSPModifyVertex would only have to replace this x_screen and y_screen values. Anyway, I feel all of this is a bit out of the scope of the pull request and it is better to keep the hack for now, maybe consider trying to implement the idea later.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
standard-two-simplex: I restored it to its original place. I had moved it because I was using it elsewhere at some point, but I am no longer using it.

standard-two-simplex:
I pushed the new code to https://github.com/standard-two-simplex/GLideN64/tree/use_screen_coords_2. I believe all the above reviews have been solved.

Still have to check the HD textures issue.  The texture bound setting is incomplete for PJ64 too.

> #include <cmath>

@Jj0YzL5nvJ  Can you check if this fixes the issue? You may as well try changing the functions to ceil() and min(), max(). I chose the f versions because travis for mac was unable to build without it.

standard-two-simplex:
There are some issues with the widescreen hack. 2D appears in the top left, which causes misalignments.

Jj0YzL5nvJ:
> Can you check if this fixes the issue? 

Adding `cmath` fixes the building issues.

[Trailing whitespace...](https://github.com/standard-two-simplex/GLideN64/blob/44f78dfdc01653d2a318ee2e1455b6f50912ada3/src/gDP.cpp#L877) (use_screen_coords_2)

standard-two-simplex:
> Trailing whitespace... (use_screen_coords_2)

There's one white line where you link, which seems to be normal through all the code. I think I removed all consecutive white lines.

Jj0YzL5nvJ:
<details>
 <summary>What I'm seeing</summary>



</details>

gonetz:
```
	" highp vec2 mTexCoord0 = vTexCoord0 + vec2(0.0001);						\n"
	" highp vec2 mTexCoord1 = vTexCoord1 + vec2(0.0001);						\n"

```
why are these 0.0001?

gonetz:
Regarding whitespaces: I recommend to install one of VisualStudio extensions, which automatically remove trailing whitespaces.
For example https://marketplace.visualstudio.com/items?itemName=Predelnik.RemoveTrailingWhitespaces

standard-two-simplex:
> why are these 0.0001?

Some games are designed so that texture coordinates are integers. If texture filtering is set to nearest neighbor, a small precision error can lead to fetching the wrong neighbour. The ideal design would have been making them half integers, so that a small precision error doesn't affect the texel picked, but for some reason lots of games are susceptible to very little texture coordinate change.

By adding a tiny bias, I avoid such problem. I have not found an issue with it because games don't have texture coordinates with fractional have greater than 0.9999

standard-two-simplex:
I have fixed texrects and vertices coming from gSPModifyVertex(), but I still have this issue with widescreen.
the_legend_of_zelda-000

It seems the heart triangle (the glowing one) aligns correctly, but the blue action button doesn't. I am not sure why. I think it may be due to the viewport transform having non null offset.

This lines seem to affect it in master.
https://github.com/gonetz/GLideN64/blob/92c9bd84086ca36a57f74d0b6b897cad92883069/src/GraphicsDrawer.cpp#L279-L281

gonetz:
Yes, the blue action is transformed by a different viewport, and the widescreen hack needs to take it into account.

standard-two-simplex:
> Yes, the blue action is transformed by a different viewport, and the widescreen hack needs to take it into account.

One last question. Do all the triangles need to be scaled and offseted or do you somehow detect which are 3d and 2d and treat them differently?

gonetz:
All triangles treated the same.
One note: gSPModifyVertex sets screen (2D) coordinates to a vertex.
Before your changes screen coordinate converted back to 3D one and widescreen transformation (multiplication by the scale) applied to x coordinate if widescreen is enabled.
Now you just multiply screen coordinate by the scale. I'm not sure that it is correct.

standard-two-simplex:
> Now you just multiply screen coordinate by the scale. I'm not sure that it is correct.

I have a local change where I add an offset, as with texture rectangles. I'll push to github later.

I am also considering if it is possible to add a scale and offset in VertexShaderEnd for every primitive, rather than in many different places.

standard-two-simplex:
See the lates commit. Aside for the clipping issue, the geometry is in place.

The code needs cleaning, but in essence what I'm doing is
```screen_coordinate*adjustScale + original_screen_width * (1 - adjustScale) / 2```
which should be applicable to all geometry and could be done at the end of the vertex shader.

It has the downside of clipping not happening in the `[-1,1]` x boundaries, but `[-1/adjustScale, 1/adjustScale]`, so clipping code would have to be changed. Do you think it is worth standarizing the widescreen hack for all primitives?

**Comment by gonetz** [comment]
@standard-two-simplex I just checked your latest commit. Yes, all geometry seems to be in place.

> Do you think it is worth standarizing the widescreen hack for all primitives?

What do you mean by standarizing? Do it the same way in vertex shaders? If it is possible, I'm for it.

> It has the downside of clipping not happening in the [-1,1] x boundaries, but [-1/adjustScale, 1/adjustScale], so clipping code would have to be changed. 

I hope that it is possible to adjust the software clipping code. I guess, that it is the only option because we can't change the hardware clipping.
I don't understand one moment: hardware clipping is in fact disabled by your modifications. So, without the software clipping there should be no clipping at all. However, lots of geometry disappears with your latest commit, with or without software clipping. Why?
zelda_majora's_mask-006

**Comment by standard-two-simplex** [comment]
> What do you mean by standarizing? Do it the same way in vertex shaders? If it is possible, I'm for it.

Yes, I mean to do it in the same way for all primitives. I wrote a proof of concept here 73b792cfe6.

The geometry seems in place but many things are missing. I was expecting that clipping would give problems, but something else is.

**Comment by Jj0YzL5nvJ** [comment]
> I don't understand one moment: hardware clipping is in fact disabled by your modifications. So, without the software clipping there should be no clipping at all. However, lots of geometry disappears with your latest commit, with or without software clipping. Why?

That looks more like normal behavior behind the scenes. The real question should be, what **hack** is implemented in the previous implementation to not show such behavior?

Some PSX games discard triangles and / or geometry normally not visible on the sides of the screen to save memory. Those that don't end up freezing the emulator for exceeding the polygon limit that the simulated GPU can display.
https://youtu.be/S8aKUjfpwic?t=1m44s
https://youtu.be/7SxOH6Rjt34?t=2m4s
https://github.com/libretro/beetle-psx-libretro/issues/490#issuecomment-469062488
https://github.com/libretro/beetle-psx-libretro/issues/594#issuecomment-578462228

**Comment by gonetz** [comment]
> That looks more like normal behavior behind the scenes. The real question should be, what **hack** is implemented in the previous implementation to not show such behavior?

Vertex is inside the viewport if 
x/w is inside [-1, 1]
y/w is inside [-1, 1]
z/w is inside [-1, 1]
What the hack does:  it scales x coordinate of all vertices by some constant.
It makes more vertices fit [-1, 1] range for x/w.
That is  x/w was outside of [-1, 1], but x*scale/w is inside, providing the scale is < 1.
Thus. more geometry fits the viewport, **if that geometry actually exists**.
The result of the hack depends on how the game was programmed. 

> Some PSX games discard triangles and / or geometry normally not visible on the sides of the screen to save memory.

Some N64 games do the same, so widescreen hack does not look nice with them. It seems that Zeldas haven't such optimizations, and thus look good with the hack enabled.

**Comment by gonetz** [comment]
> Some geometry is being discarded in `GraphicsDrawer::isClipped()` which discards a triangle if all three vertices are marked as outside.

Yes, the problem is with it. I hacked gSPClipVertex to set vtx.clip = 0 for all vertices and did some other hack to avoid program's crash. It returned missing geometry back. All clipping-related code need to be modified.

**Comment by gonetz** [comment]
@standard-two-simplex I fixed clipping issues with widescreen, commit 0ba5f04e14038d
I found an issue with 2D primitives in widescreen mode:

File: src/GraphicsDrawer.cpp (hunk 2 above)
Comment (standard-two-simplex) [reply src/GraphicsDrawer.cpp:748]: Done.

File: src/gSP.cpp (hunk 5 above)
Comment (standard-two-simplex) [reply src/gSP.cpp:1679]: gSPModifyVertex() was adapted.

File: src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp (hunk 7 above)
Comment (standard-two-simplex) [reply src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp:248]: Done.

File: src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp (hunk 8 above)
Comment (standard-two-simplex) [reply src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp:244]: Done. Please check if it satifactory. I can shorten it if you feel it is too much.

File: src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp (hunk 9 above)
Comment (standard-two-simplex) [reply src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp:255]: Done.

File: src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp (hunk 10 above)
Comment (standard-two-simplex) [reply src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp:692]: Done.

File: src/gSP.cpp (hunk 11 above)
Comment (standard-two-simplex) [reply src/gSP.cpp:271]: Done.

File: src/GraphicsDrawer.cpp (hunk 12 above)
Comment (standard-two-simplex) [reply src/GraphicsDrawer.cpp:324]: Done.

File: src/GraphicsDrawer.cpp (hunk 13 above)
Comment (standard-two-simplex) [reply src/GraphicsDrawer.cpp:340]: > If you change signature of _updateScreenCoordsViewport2 to 
```void _updateScreenCoordsViewport2(f32 _scale = 0.0f) const;```
 then I can call it from texrect drawer with scale 1.0f

Is it fine like it is now?

File: src/GraphicsDrawer.cpp (hunk 14 above)
Comment (standard-two-simplex) [reply src/GraphicsDrawer.cpp:327]: Removed.

File: src/GraphicsDrawer.cpp (hunk 15 above)
Comment (standard-two-simplex) [reply src/GraphicsDrawer.cpp:1504]: Done.

File: src/gSP.cpp (hunk 16 above)
Comment (standard-two-simplex) [reply src/gSP.cpp:1660]: Done.

**Comment by standard-two-simplex** [comment]
I uploaded current WIP code, I will clean up tomorrow.

**Comment by gonetz** [comment]
I checked the current state of your branch. Everything seems to work now. I think it is time to re-arrange the commits again to create a clean chain of commits.

**Comment by gonetz** [comment]
I added GUI control for enableTexCoordBounds: 696b0e1d11

**Comment by standard-two-simplex** [comment]
Most of the things are cleaned up. I put VI.height() and/or gDP.colorImage.width to compute the widescreen offsets. I think gDP.colorImage.width is necessary for primitives with N64 viewport different to zero. Do you know if one should be better than the other?

One thing I notice while cleaning up. SM64 seems to use this code in gSP::processVertex() to render 2D to the screen when the widescreen hack is enabled. If these primitives are not treated differently, the skies don't cover all the screen and create garbage on the sides.
```
			if (gSP.matrix.projection[3][2] == -1.f)
				vtx.w *= adjustScale;
```
It seems it detects whether the matrix is an orthographics projection and applies the widescreen hack differently: instead of adjusting the x coordinates to be closer to the screen center, it adjusts the y coordinates to be further from the center.

I implemented two possibilities,
- 0f460127, streches the textures without keeping aspect ratio. super_mario_64-001
- 2f5582fc, behaves as master (I needed to use VI.real_height to compute the offset, gDP.colorImage.height doesnt work).  super_mario_64-002
Which one should I keep?

**Comment by standard-two-simplex** [comment]
Besides, please check that the last two cleanup commits are correct (no longer necessary code) and I will squash them to the first commit.

**Comment by weinerschnitzel** [comment]
Part of me thinks that should be optional, but I am also in favor of less settings. 

Stretching should probably be preferred as you might lose a needed part of an overlay in other games. Original textures should just work. If people don't like that, a custom texture could replace.

**Comment by gonetz** [comment]
> One thing I notice while cleaning up. SM64 seems to use this code in gSP::processVertex() to render 2D to the screen when the widescreen hack is enabled. If these primitives are not treated differently, the skies don't cover all the screen and create garbage on the sides.

This hack was introduced in commit 4a75d4d35778: 
```
Add hack to fix both mario 64 and mario kart 64 background (blue sky) with widescreen hack enable.
Author: Gillou68310
```
Unfortunately, the commit message does not mention the related issue.

gonetz:
Regarding orthographics projection and the two possibilities. I like the second variant (master-like) more.
You say you need to use VI.real_height to compute the offset. It is not safe.
As you know, N64 does not set color image height, it sets only width. Calculations of gDP.colorImage.height are unreliable, so no surprise it does not work for you. VI.real_height is more reliable, at least it is set. However, it is not actually height of the image, it is the distance between the first and the last drawn scanlines. Usually it can be used as image's height, but not always. May be you can use the height as 
`height = gDP.colorImage.width * 3 / 4` ?
Any solution should be checked with games, which use interlaced mode, for example with PD in hi-res mode.

gonetz:
> Besides, please check that the last two cleanup commits are correct (no longer necessary code) and I will squash them to the first commit.

They look correct for me.

standard-two-simplex:
> height = gDP.colorImage.width * 3 / 4

Yes, this should work if the game uses a 4:3 aspect ratio.

gonetz:
> Yes, this should work if the game uses a 4:3 aspect ratio.

If the game natively supports 'widescreen' mode, it is better to set it and use "stretch" aspect mode.

I'm not sure about the result when game uses 4:3 but frame buffer size is 640x240 or 320x480 interlaced.
`height = gDP.colorImage.width * 3 / 4` still should work in that case too, but it need to be tested.

standard-two-simplex:
I pushed a clean version. If previous conversations are considered resolved, I will push the changes into the branch in this pull request.

gonetz:
Yes, the code looks clean now, thanks!
I made minor changes in the first two commits: I corrected  GraphicsDrawer::_updateViewport and in my texrect drawer fix.
I've uploaded all commits to https://github.com/gonetz/GLideN64/tree/standard-two-simplex-use_screen_coords branch.
I could put it right in master, but I want to test it more first.

standard-two-simplex:
> I could put it right to the master, but I want to test it more first.

Sure, I want to test a bit myself too. I am trying to figure out the Top Gear Overdrive issue, but not much luck so much.

gonetz:
> I am trying to figure out the Top Gear Overdrive issue, but not much luck so much.

Texture issues are always tricky and hard to solve. I'll try to check it too, since that problem bothers me most atm.

fzurita:
I'm starting to test this. I had to clear my shader cache to get any graphics, so I'm guessing the shader cache version needs to be incremented.

standard-two-simplex:
Closed in favor of #2470.

PR #2336, opened 2020-09-22
[travis] enable MUPENPLUSAPI_GLIDENUI

Rosalie241:


Rosalie241:
is <https://github.com/gonetz/GLideN64/blob/master/.travis.yml#L50-L63> used at all? I don't see it being deployed to github releases, so can I remove that part?

gonetz:
> so can I remove that part?

I think, yes.

PR #2344, opened 2020-09-30
getRevision: silence error if Revision.h doesn't exist

Rosalie241:


src/getRevision.sh
```diff
@@ -1,6 +1,6 @@
 SCRIPT_DIRECTORY=`dirname $0`
 rev=\"`git rev-parse --short HEAD`\"
-lastrev=$(head -n 1 $SCRIPT_DIRECTORY/Revision.h | awk -F'PLUGIN_REVISION ' {'print $2'})
+lastrev=$(head -n 1 $SCRIPT_DIRECTORY/Revision.h 2> /dev/null | awk -F'PLUGIN_REVISION ' {'print $2'})
```
Jj0YzL5nvJ, line 3: Since you're at it, change `awk -F'PLUGIN_REVISION ' {'print $2'}` to `cut -d ' ' -f3` too.

Jj0YzL5nvJ (changes requested):

src/getRevision.sh
Rosalie241: isn't that a bit pointless? The only reason I propose this patch is because the error really annoyed me everytime I build from a clean tree, I don't see the point in changing `awk` to `cut`

Jj0YzL5nvJ: Yes... is very pointless. I think is hypotheticaly harder to see:

```
$ ./getRevision.sh
getRevision.sh: line 3: cut: command not found
current revision "ed00993"
last build revision
```

That:

```
$ ./getRevision.sh
getRevision.sh: line 3: awk: command not found
current revision "ed00993"
last build revision
```

Not to mention that `awk` could be troubleshooting locale issues without requiring specs like `LC_ALL=C`:

```
lastrev=$(LC_ALL=C head -n 1 $SCRIPT_DIRECTORY/Revision.h 2> /dev/null | cut -d ' ' -f3)
```

Ultimately the most important result will not change:

```
$ cat Revision.h
#define PLUGIN_REVISION "ed00993"
#define PLUGIN_REVISION_W L"ed00993"
```

PR #2345, opened 2020-09-30
GLideNUI: detect application instance dynamically

Rosalie241:
tested on windows & linux on both project64 and mupen64plus

### DIFF
--- a/src/GLideNUI/GLideNUI.cpp
+++ b/src/GLideNUI/GLideNUI.cpp
@@ -29,27 +29,28 @@ int openConfigDialog(const wchar_t * _strFileName, const char * _romName, bool &
 	if (config.generalEmulation.enableCustomSettings != 0 && _romName != nullptr && strlen(_romName) != 0)
 		loadCustomRomSettings(strIniFileName, _romName);
 
-#ifndef M64P_GLIDENUI
-	int argc = 0;
-	char * argv = 0;
-	QApplication a(argc, &argv);
+	std::unique_ptr<QApplication> pQApp;
+	QCoreApplication* pApp = QCoreApplication::instance();
+
+	if (pApp == nullptr) {
+		int argc = 0;
+		char * argv = 0;
+		pQApp.reset(new QApplication(argc, &argv));
+		pApp = pQApp.get();
+	}
 
 	QTranslator translator;
 	if (translator.load(getTranslationFile(), strIniFileName))
-		a.installTranslator(&translator);
-#endif // M64P_GLIDENUI
+		pApp->installTranslator(&translator);
 
 	ConfigDialog w(Q_NULLPTR, Qt::WindowTitleHint | Qt::WindowSystemMenuHint | Qt::WindowCloseButtonHint);
 
 	w.setIniPath(strIniFileName);
 	w.setRomName(_romName);
 	w.setTitle();
-#ifndef M64P_GLIDENUI
 	w.show();
-	const int res = a.exec();
-#else
-	const int res = w.exec();
-#endif // M64P_GLIDENUI
+
+	int res = pQApp ? pQApp->exec() : w.exec();
 	_accepted = w.isAccepted();
 	return res;
 }


src/GLideNUI/GLideNUI.cpp
```diff
@@ -29,27 +29,32 @@ int openConfigDialog(const wchar_t * _strFileName, const char * _romName, bool &
 	if (config.generalEmulation.enableCustomSettings != 0 && _romName != nullptr && strlen(_romName) != 0)
 		loadCustomRomSettings(strIniFileName, _romName);
 
-#ifndef M64P_GLIDENUI
-	int argc = 0;
-	char * argv = 0;
-	QApplication a(argc, &argv);
+	QCoreApplication* app = QCoreApplication::instance();
```
gonetz, line 32: QCoreApplication* pApp

src/GLideNUI/GLideNUI.cpp
```diff
@@ -29,27 +29,32 @@ int openConfigDialog(const wchar_t * _strFileName, const char * _romName, bool &
...
 
-	QTranslator translator;
-	if (translator.load(getTranslationFile(), strIniFileName))
-		a.installTranslator(&translator);
-#endif // M64P_GLIDENUI
+	if (app == nullptr)
+	{
+		int argc = 0;
+		char * argv = 0;
+		app = new QApplication(argc, &argv);
```
gonetz, line 38: When and how this object destroyed?
I don't see delete app.

src/GLideNUI/GLideNUI.cpp
```diff
@@ -29,27 +29,32 @@ int openConfigDialog(const wchar_t * _strFileName, const char * _romName, bool &
...
-#ifndef M64P_GLIDENUI
 	w.show();
-	const int res = a.exec();
-#else
-	const int res = w.exec();
-#endif // M64P_GLIDENUI
+
+	int res;
+	if (app == nullptr)
+		res = app->exec();
```
gonetz, line 54: ???!!!

src/GLideNUI/GLideNUI.cpp
```diff
@@ -29,27 +29,32 @@ int openConfigDialog(const wchar_t * _strFileName, const char * _romName, bool &
...
-	int argc = 0;
-	char * argv = 0;
-	QApplication a(argc, &argv);
+	QCoreApplication* app = QCoreApplication::instance();
 
-	QTranslator translator;
-	if (translator.load(getTranslationFile(), strIniFileName))
-		a.installTranslator(&translator);
-#endif // M64P_GLIDENUI
+	if (app == nullptr)
```
gonetz, line 34: if (app == nullptr) {  // code formatting style

src/GLideNUI/GLideNUI.cpp
```diff
@@ -29,27 +29,32 @@ int openConfigDialog(const wchar_t * _strFileName, const char * _romName, bool &
...
+	{
```
gonetz, line 35: if (app == nullptr) { // code formatting style

src/GLideNUI/GLideNUI.cpp
Rosalie241: oh, right, this is wrong, will fix!

Rosalie241:
I'll need to test this (on project64), will do when the appveyor build is finished

src/GLideNUI/GLideNUI.cpp
```diff
@@ -29,27 +29,34 @@ int openConfigDialog(const wchar_t * _strFileName, const char * _romName, bool &
...
-#ifndef M64P_GLIDENUI
 	w.show();
-	const int res = a.exec();
-#else
-	const int res = w.exec();
-#endif // M64P_GLIDENUI
+
+	int res;
+	if (!hasApp) {
+		res = pApp->exec();
```
gonetz, line 54: again? !hasApp means pApp is null, so pApp->exec() will creash.

gonetz, line 54: Do I get it correct that you are using QCoreApplication::instance() to find if plugin is in zilmar-spec mode or in mupen64plus one? If QCoreApplication::instance() returns valid pointer, the plugin is in M64P_GLIDENUI mode and thus w.exec() should be called?

src/GLideNUI/GLideNUI.cpp
```diff
@@ -29,27 +29,34 @@ int openConfigDialog(const wchar_t * _strFileName, const char * _romName, bool &
...
-		a.installTranslator(&translator);
-#endif // M64P_GLIDENUI
+	if (!hasApp) {
+		int argc = 0;
+		char * argv = 0;
+		pApp = new QApplication(argc, &argv);
+
+		QTranslator translator;
+		if (translator.load(getTranslationFile(), strIniFileName))
+			pApp->installTranslator(&translator);
```
gonetz, line 42: If hasApp is true, translation will not be applied, will it?

src/GLideNUI/GLideNUI.cpp
```diff
@@ -29,27 +29,34 @@ int openConfigDialog(const wchar_t * _strFileName, const char * _romName, bool &
...
 	w.show();
-	const int res = a.exec();
-#else
-	const int res = w.exec();
-#endif // M64P_GLIDENUI
+
+	int res;
+	if (!hasApp) {
+		res = pApp->exec();
+		delete pApp;
```
gonetz, line 55: I also dislike that manual delete pApp.

I suggest something like this:

```
	std::unique_ptr<QApplication> pQApp;
	QCoreApplication* pApp = QCoreApplication::instance();

	if (pApp == nullptr) {
		int argc = 0;
		char * argv = 0;
		pQApp.reset(new QApplication(argc, &argv));
		pApp = pQApp.get();
	}

	QTranslator translator;
	if (translator.load(getTranslationFile(), strIniFileName))
		pApp->installTranslator(&translator);

	ConfigDialog w(Q_NULLPTR, Qt::WindowTitleHint | Qt::WindowSystemMenuHint | Qt::WindowCloseButtonHint);

	w.setIniPath(strIniFileName);
	w.setRomName(_romName);
	w.setTitle();
	w.show();

	int res = pQApp ? pQApp->exec() : w.exec();
	_accepted = w.isAccepted();
	return res;

```

src/GLideNUI/GLideNUI.cpp
Rosalie241: well, I'm using `QCoreApplication::instance()` to detect if the GUI (Zilmar spec or mupen spec) is Qt or not, if it is we can't create another `QApplication`

src/GLideNUI/GLideNUI.cpp
Rosalie241: that's correct

src/GLideNUI/GLideNUI.cpp
Rosalie241: that works perfectly on both linux & windows, thank you

PR #2352, opened 2020-10-07
DisplayWindow fixes

Rosalie241:
I accidentally broke Logan's GUI's resizing feature, this PR corrects the issue, tested with my GUI & Logan's GUI

### DIFF
--- a/src/DisplayWindow.cpp
+++ b/src/DisplayWindow.cpp
@@ -31,6 +31,7 @@ void DisplayWindow::stop()
 
 void DisplayWindow::restart()
 {
+	_restart();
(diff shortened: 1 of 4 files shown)


loganmc10:
@Rosalie241 There is an AppVeyor error because _restart() needs to exist in the PJ64/Windows implementation as well, even if it's empty

loganmc10:
Beside that it looks good to me, allows me to remove a more custom code from m64p which is good

gonetz:
I combined this PR to the one commit b200b33b849f because some intermediate commits does not compile.

PR #2358, opened 2020-10-12
Ugly script for power users and more...

Jj0YzL5nvJ:
This script has as a minimum requirement to specify an architecture to start working. It will require some external CLI tools like `7z`, `git` and optionally `curl` if `--dlqt` is used.

Run without parameters to see help.

### DIFF
--- a/projects/msvc/GLideN64_builder.cmd
+++ b/projects/msvc/GLideN64_builder.cmd
@@ -1,0 +1,284 @@
+@echo off
+setlocal enableextensions disabledelayedexpansion
+cls
+set DPROJ=GLideN64.sln
+set DMN=REM
+set ERR=if errorlevel 1 goto err
+set MSG=%DPROJ% does not exist in the same directory
+if not exist "%~dp0%DPROJ%" goto err
+
+set ARCH=
+set BOTH=0
+set CONF=Release
+set DLQT=0
+set REB=
+set NOQT=0
+set NOWTL=0
+set NOM64=0
+set ESIM=0
+set EBQ=explorer "."
+taskkill /im msbuild.exe /f 2>nul
+taskkill /im vctip.exe /f 2>nul
+taskkill /im mspdbsrv.exe /f 2>nul
+if "%*"=="" goto help
+for %%P in (%*) do (
+	if /i "%%P"=="--clean" goto clean
+	if /i "%%P"=="--x86" set ARCH=x86
+	if /i "%%P"=="--x64" set ARCH=x64& set BOTH=0
+	if /i "%%P"=="--all" set ARCH=x64& set BOTH=1
+	if /i "%%P"=="--debug" set CONF=Debug
+	if /i "%%P"=="--rebuild" set REB=/t:Rebuild
+	if /i "%%P"=="--noqt" set NOQT=1
+	if /i "%%P"=="--nowtl" set NOWTL=1
+	if /i "%%P"=="--nom64" set NOM64=1
+	if /i "%%P"=="--dlqt" set DLQT=1
+	if /i "%%P"=="--sim" set ESIM=1
+	if /i "%%P"=="--q" set EBQ=REM
+)
+
+if not defined ARCH goto help
+set /a MOD=%NOQT%+%NOWTL%+%NOM64%
+set MSG=All compilation tasks were disabled on request
+if %MOD%==3 goto err
+set MSG=7z was not found in the environment
+7z >nul 2>&1
+%ERR%
+
+set tVSPF=%ProgramFiles(x86)%
+if "%PROCESSOR_ARCHITECTURE%"=="x86" set tVSPF=%ProgramFiles%
+set "tVSPF=%tVSPF%\Microsoft Visual Studio"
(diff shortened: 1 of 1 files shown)


Jj0YzL5nvJ:
@oddMLan, pay attention.

oddMLan:
Looks good to me, but I don't understand why this is necessary. So that is possible to run the cmd file locally and build all projects at once? Skip the configuration step with Qt from Visual Studio?

Jj0YzL5nvJ:
> So that is possible to run the cmd file locally and build all projects at once? Skip the configuration step with Qt from Visual Studio?

Not touch Visual Studio IDE with a ten foot pole, yes. That's true if you have the PATH environment well configured with all the additional CLI programs.
Actually it's not necessary to replace the current Appveyor. But both files share practically the same DNA, but with extended features, so... make sense?

The reason I want you to be aware of this change is because I don't know if the addition of "_x86" will break the package script used by Project64 to get GLideN64's builds, that’s all.

oddMLan:
> The reason I want you to be aware of this change is because I don't know if the addition of "_x86" will break the package script used by Project64 to get GLideN64's builds, that’s all.

It might, but if it does then the Project64 packaging script can be modified accordingly

oddMLan:
What did you change in the lastest two pushes?

Jj0YzL5nvJ:
In the first, a bug that overwrited the QTDIR variable in the second loop by using "`--all`" with `MYQTDIR_x86` and `MYQTDIR_x64`.
In the second, I rearrange the sequence of some sets to remove a `goto`, and add some quotes to avoid an bug with paths with parentheses, that's all.

I think I have fixed most of the potential bugs... any feedback is welcome.

Rosalie241:
IMO this is uglier than it was (batch is god awful to read)

why is `taskkill` needed at all? Seems like an ugly hack
and the `goto`s make the flow incredibly unreadable


```
set MSBUILDQT=msbuild
set MSBUILDWTL=msbuild
set MSBUILDM64P=msbuild
```
why? this is completely unrequired and just makes the code longer and again, uglier to read

```
set VS2013=Microsoft Visual Studio 12.0
set VS2015=Microsoft Visual Studio 14.0
set VS2017=Microsoft Visual Studio\2017
set VS2019=Microsoft Visual Studio\2019
```
why's this needed? if you've launched the dev tools cmd, you don't need this *at all*

also why `curl`, that isn't a program installed by default in windows, is it?

Jj0YzL5nvJ:
Thx!

Jj0YzL5nvJ:
> another 'problem' I have with this is that it's really not difficult to build GLideN64 from source 'manually' using msvc on windows, so this seems unrequired?

Try to run the IDE in a machine with 2GB of RAM or less with any modern AV to recognize what really means "fun".

Rosalie241:
I meant that it was not difficult to build using the developer tools cmd manually

On October 16, 2020 9:12:00 PM GMT+02:00, Jj0YzL5nvJ <notifications@github.com> wrote:
>> another 'problem' I have with this is that it's really not difficult
>to build GLideN64 from source 'manually' using msvc on windows, so this
>seems unrequired?
>
>Try to run the IDE in a machine with 2GB of RAM or less with any modern
>AV to recognize what really means "fun".
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/gonetz/GLideN64/pull/2358#issuecomment-710451239

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

gonetz:
Is it possible to add more informative errors messaging? I run the script and got "ERROR!" And what?

gonetz:
"7z was not found in the environment" - much better!

gonetz:
The script works, but if I run it without parameters it just builds everything.
It is not quite a common behavior for a CLI tool. Would not it be better to print "Usage:" message instead with the list of parameters?
Also, it would be nice to print not only the name of the resulted archive but also the path to it, absolute or relative. The current message is:
"Creating archive: GLideN64-8a5d26db-WTL_for_Project64_x86.7z"
I should then guess that the actual file location is
../../build/GLideN64-8a5d26db-WTL_for_Project64_x86.7z

Jj0YzL5nvJ:
> It is not quite a common behavior for a CLI tool. Would not it be better to print "Usage:" message instead with the list of parameters?

I had thought about adding a "help" parameter, but I think your idea is better.

> Also, it would be nice to print not only the name of the resulted archive but also the path to it, absolute or relative.

I can do two things, show the absolute path and open said path with Windows explorer. Or just do one of those two, personally I prefer to do both...

Other parameters that could be implemented could be to download Qt only if specified (--dlqt). Another would be to clean up the existing build files (--clean).
And with all this it would make more sense if I remove the options to configure functions that are already controlled by parameters...

gonetz:
Yes, it sounds good.

Jj0YzL5nvJ:
I really need feedback from people messing with the parameters and variables.

gonetz:
Just tested - it looks and works as a really good and useful script now. Thanks!

Issue #2062, opened 2019-06-20
Texture Dumping fails with certain roms


gamax92:
Some games have internal rom names that contain invalid characters for filenames on Windows.

Turok: Rage Wars has the internal rom name of "Turok: Rage Wars", the colon cannot appear in filenames, and no textures get dumped. RiceVideo seems to handle this by replacing the colon with a dash.

gamax92:
Any update on this? Just needs to filter out invalid characters and replace them with a dash.

Doommarine23:
I can confirm this is still an issue and I appreciate you mentioning the fix, because I looked into it. I was able to fix this problem by just copy and pasting a bit of code that already exists in GlideN64's codebase and doing exactly as you said, replacing the colon with a dash.

I'm going to take a little bit to test and then I'll throw it up as a pull request.

PR #2368, opened 2020-10-25
Fix: Replace ":" with "-" in Rom Names for textures

Doommarine23:
Replace colons with dashes to fix rom dumping for games like "Turok 3: Shadow of Oblivion". Reuses code from Config_mupenplus.cpp to replace characters in _ident.

I have tested both dumping and loading with Turok 2, Turok 3, and Turok Rage Wars. Everything seems to work fine.

I've kept the replacement quite literal, e.g. "Turok: Rage Wars" becomes "Turok- Rage Wars", so perhaps adding a space would be good.

This should close the following issue: https://github.com/gonetz/GLideN64/issues/2062

### DIFF
--- a/src/GLideNHQ/TxFilter.cpp
+++ b/src/GLideNHQ/TxFilter.cpp
@@ -129,6 +129,11 @@ TxFilter::TxFilter(int maxwidth,
 	if (ident && wcscmp(ident, wst("DEFAULT")) != 0)
 		_ident.assign(ident);
 
+	/* replace : in rom names with -*/
+	std::wstring replace = L"-";
+	for (size_t pos = _ident.find(':'); pos != std::string::npos; pos = _ident.find(':', pos))
+		_ident.replace(pos, 1, replace);
+
 	if (TxMemBuf::getInstance()->init(_maxwidth, _maxheight)) {
 		if (!_tex1)
 			_tex1 = TxMemBuf::getInstance()->get(0);


rez-:
why not just remove the semicolon instead of replacing it with another character?

Doommarine23:
I think that would also be a valid fix, and would be pretty easy to change. I just went with replacing the colon because I was following the suggestion given by gamax92 in their issue report.

ToruTheRedFox:
this would break existing texture packs though

src/GLideNHQ/TxFilter.cpp
```diff
@@ -129,6 +129,11 @@ TxFilter::TxFilter(int maxwidth,
 	if (ident && wcscmp(ident, wst("DEFAULT")) != 0)
 		_ident.assign(ident);
 
+	/* replace : in rom names with -*/
+	std::wstring replace = L"-";
+	for (size_t pos = _ident.find(':'); pos != std::string::npos; pos = _ident.find(':', pos))
+		_ident.replace(pos, 1, replace);
+
```
gonetz, line 136: Use removeColon() instead, see src\GLideNHQ\txWidestringWrapper.h

Doommarine23: Thank you. I will take a look soon.

gonetz:
> Use removeColon() instead, see src\GLideNHQ\txWidestringWrapper.h

Done in 7a38e8044

Issue #1647, opened 2017-11-04
Gauntlets Legends power bars


olivieryuyu:
The power bars are wrong

![31535828-d4609b4e-affc-11e7-937a-8f9efa923853](https://user-images.githubusercontent.com/8987490/32404783-c9798102-c157-11e7-9560-9c019306310e.jpg)

fzurita:
FYI. This is the commit that fixed the bars in this game in angrylion according to LegendOfDragoon:

https://sourceforge.net/p/angrylions-stuff/code/99/tree//trunk/mylittle-nocomment/n64video.cpp?diff=55bb05dc27184618d775bec1:98

LegendOfDragoon:
It was actually r96 that fixed it.

I wrote a bit about the issue here https://github.com/purplemarshmallow/z64/issues/62

gonetz:
As I understand, r96 does not fix the bars, it makes HUD to display.

LegendOfDragoon:
r96 actually did fix the bars in Angrylion. Not sure why it isn't working in GLideN64. One thing I noticed is that GLideN64 doesn't write the HUD to RDRAM, which means that save states generated by GLideN64 won't have the HUD displaying at all. Unless maybe there was some option I was supposed to toggle.

gonetz:
Then it would be useful to find, which lines in r96 fix the bars. It would help me to find, how to fix them in GLideN64

LegendOfDragoon:
It's the alpha compare code that fixed the HUD in Gauntlet (basically the same code that fixed rocks in Mischief Makers).

gonetz:
The HUD and the bars are two different problems.  GLideN64 has the HUD, but bars are broken.  I don't know yet how to debug them.

LegendOfDragoon:
Sorry I meant the bars. I have a bad habit of mixing up words sometimes. Angrylion's used to have the bar issue, until that same commit that fixed the rocks was implemented.

Does GLideN64 have a debugger?

gonetz:
Yes, but it is buggy. It crashes with gauntlet.

standard-two-simplex:
I took a quick glance to r96 commit in Angrylions. At first glance, it looks like two things are done:

- Reimplement YUV textures

- Change condition for dithering based on the flag in other_modes. ``(other_modes.f.getditherlevel )`` in his code.

In the photo above though, it doesn't look like the alpha is dithered. Does the game always show the bars the same way or does it change from frame to frame?

[...]

PR #2372, opened 2020-10-31
Fix shader compilation errors with Adreno GPUs, Implement N64 blending using framebuffer fetch extensions

fzurita:
Adreno GPUs don't like it when sampler2D is passed as a function argument. Also removed comments from what we pass to the shader compiler to reduce string size. Some GPU drivers have limits on shader size.


fzurita:
Don't merge yet. There are still some issues with Adreno with "Hybrid filter" is disabled. It seems to work ok with it enabled for now.

fzurita:
Ok, I think this is ok to be merged. It seems to be a separate issue unrelated to this that is causing Adreno GPUs to crash.

fzurita:
It looks like c48841cfebff118af634135bc0f6a6a02bf4e7d4 broke Adreno devices.

fzurita:
I went ahead and disabled dual source blending for Adreno devices under this same pull request.

oddMLan:
I heard there is an alternative to Dual Source Blending for mobile devices that don't support it...

fzurita:
Yeah, there is. Dolphin implemented it.

fzurita:
https://github.com/dolphin-emu/dolphin/pull/6141

For reference, that is the dolphin implementation that doesn't require the extension.

fzurita:
And it looks like @standard-two-simplex already provided a solution here: https://github.com/gonetz/GLideN64/issues/1647

It's probably worth implementing.

fzurita:
Also, dolphin also had to workaround a Adreno driver bug, it's probably what we are seeing here.

fzurita:
Dolphin workaround doesn't work for us for some reason. The GPU driver still crashes.

gonetz:
@oddMLan could you please review these changes? You are the author of hybrid filter code.

fzurita:
I added the framebuffer fetch implementation of dual source blending work the help of @standard-two-simplex. This works with Adreno and Mali GPUs:
gauntlet_legends-000
pokemon_stadium-000 (1)

fzurita:
@standard-two-simplex Can you make sure that I didn't break dual source blending on PCs?

standard-two-simplex:
Yes, I'll check everything later today.

standard-two-simplex:
The math seems correct and I can run the dual_source_blending part in my computer without issues, so nothing seems broken.

I do not see the need to create special output `realFragColor` and then copy `fragColor` into it. Is it ok if I add some cleanup and abstract all extension specific code to the shader fragment header? This way the body would be the same for all cases.

fzurita:
The reason I have ```realFragColor``` is because of this: https://github.com/dolphin-emu/dolphin/pull/6141/commits/8d68adcaf3442122fff5bb0988c2f06b0128f021

I guess I could try removing it and see if it works correctly without it since I was playing it safe from the start.

Feel free to clean it up and I can test your changes.

standard-two-simplex:
@fzurita I created a PR to you branch.

I think you can remove `ShaderFragmentMainEndSpecial()` as well.

I would also recommend renaming f3ce000 to "Implement N64 blending using framebuffer fetch extensions" as it has nothing to do with dual source blending (which means outputing two colors/alphas from fragment shader and using them in OpenGLs blender).

fzurita:
@standard-two-simplex Thanks. I left you a question in your pull request.

oddMLan:
> @oddMLan could you please review these changes? You are the author of hybrid filter code.

Hybrid filter seems to work fine on my Win10 destkop with AMD gfx and so does dual source blending
GLideN64_POKEMON_STADIUM_000

fzurita:
@standard-two-simplex I had to make a few minor corrections for shader compilation errors. Other than that, it works correctly with Adreno and Mali GPUs.

fzurita:
I updated the pull request with your changes.

src/CMakeLists.txt
```diff
@@ -439,6 +439,10 @@ if(GL_PROFILE)
     )
 endif(GL_PROFILE)
 
+add_definitions(
+        -DGL_DEBUG
+)
+
```
gonetz, line 445: Why?

fzurita: Whoops, I was using that while testing. I'll take it off.

src/GraphicsDrawer.cpp
```diff
@@ -23,6 +23,7 @@
 #include "RSP.h"
 #include "RDP.h"
 #include "VI.h"
+#include "Log.h"
```
gonetz, line 26: Unnecessary change.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -1706,10 +1730,10 @@ class ShaderFragmentRenderTarget : public ShaderPart
 	}
 };
 
-class ShaderFragmentMainEnd : public ShaderPart
+class ShaderFragmentMainEndSpecial : public ShaderPart
```
fzurita, line 1733: @gonetz I forgot to remove this class. This is not needed.

fzurita:
@standard-two-simplex unfortunately, we have some issues with this. Not sure how many games are affected bit at least turok and blast corps are.

Blast corps has flickering textures. Turok text is all messed up, see this:
turok_dinosaur_hunte-000

Any ideas?

PR #2378, opened 2020-11-09
Fix too small threaded OpenGL buffers in some games

fzurita:
This is the issue for which @m4xw found a fix for as reported here:

https://github.com/gonetz/GLideN64/issues/2249

### DIFF
--- a/src/Graphics/OpenGLContext/ThreadedOpenGl/opengl_WrappedFunctions.cpp
+++ b/src/Graphics/OpenGLContext/ThreadedOpenGl/opengl_WrappedFunctions.cpp
@@ -2,8 +2,8 @@
 
 namespace opengl {
 
-	std::vector<char> GlDrawArraysUnbufferedCommand::m_attribsData(2 * 1024 * 1024, 0);
-	std::vector<char> GlDrawElementsUnbufferedCommand::m_attribsData(2 * 1024 * 1024, 0);
+	std::vector<char> GlDrawArraysUnbufferedCommand::m_attribsData(15 * 1024 * 1024, 0);
+	std::vector<char> GlDrawElementsUnbufferedCommand::m_attribsData(15 * 1024 * 1024, 0);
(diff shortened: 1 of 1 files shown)


gonetz:
8x increase buffer size? How did you come to that size? Is this value somehow justified?

m4xw:
Completely forgot about that one.
I just bumped them up a bunch, this was primarily issue with nerrels texture pack so i figured its a adequate amount, but its for sure not tweaked.
I cant say I encountered issues like stated in the original issues (other than the oot new_dynarec mingw issue which isnt RDP) so your mileage might vary there (but i also have lower tex cache etc)

m4xw:
On another note iirc the ringbuffer bump was required as well.
It would mostly choke on generating the cache on lower-end (has been ages tho so fuzzy memory)

fzurita:
The ring buffer already has a 200MB max size.

m4xw:
Oh i see, missed that change!

fzurita:
@gonetz I used a value that @m4xw used thinking that it has gone through enough testing to be found as valid for our use case.

@m4xw how did you end up coming up with that number?

m4xw:
The value was pulled out of thin air, the change wasnt deployed in the wild, this def needs to be revisted on our part!

fzurita:
Oh, ok, I came up with the original number or if thin air as well.

None of the games I tested exceeded that number. I guess ideally it should just have a max size like the ring buffer pool.

Do you have any games that were helped by this?

m4xw:
Majoras with very big texture packs (nerrels)
Would choke on some transitions (leaving clocktown, etc)

fzurita:
Interesting, the two values I changed shouldn't be affected by texture size.

m4xw:
I wonder if it was just the ringbuffer change then.
Also I can't say I heard about the issue lately, spooky stuff for sure

fzurita:
Ok, I will close this pull request unless I hear otherwise. Not hearing much evidence that is actually needed.

PR #2379, opened 2020-11-09
osal_keys_unix: add linux support

Rosalie241:
This PR adds linux support to `osal_keys_unix`


Rosalie241:
@gonetz implementing `osal_virtual_key_to_hid` isn't possible because the keycodes Qt returns (`pEvent->nativeVirtualKey();`) being display server dependent (so the codes would be different for X11 and wayland), would it be a good idea to i.e remove `osal_virtual_key_to_hid` entirely and create a `QtKeyToHID` class (in `src/GLideNUI/`) which converts `Qt::Key` to a keycode from `keycode/keycode.h` ?

Rosalie241:
I also want to propose an osal keys API change, mainly I want to introduce the following functions:

1) `osal_keys_init`
    * this function allows doing some preparations work (i.e getting all the keyboards)
2) `osal_keys_quit`
    * this function allows doing some cleanup work (i.e freeing buffers, closing files)
3) `osal_keys_update_state`
    * this function queries the keyboard' state (i.e for windows we could use `GetKeyboardState` instead) *once a frame* instead of doing it multiple times per frame and store it in a local buffer for `osal_is_key_pressed` to use

and `osal_is_key_pressed` should just use that stored buffer to check if a key has been pressed

gonetz:
> would it be a good idea to i.e remove `osal_virtual_key_to_hid` entirely and create a `QtKeyToHID` function (in `src/GLideNUI/`) which converts `Qt::Key` to a keycode from `keycode/keycode.h` ?

I planned to hide platform-specific stuff in that function, but it is hardly impossible if it is display server dependent.
Ok, lets replace it with `QtKeyToHID` function.

> I also want to propose an osal keys API change, mainly I want to introduce the following functions:

> I also want to propose an osal keys API change, mainly I want to introduce the following functions:

I don't mind, but I'm not sure that osal_keys_update_state based solution will work well. I use GetAsyncKeyState because it allows the plugin to always react on pressed key. Cite GetKeyboardState function description:
"An application can call this function to retrieve the current status of all the virtual keys. The status changes as a thread removes keyboard messages from its message queue. The status does not change as keyboard messages are posted to the thread's message queue, nor does it change as keyboard messages are posted to or retrieved from message queues of other threads. (Exception: Threads that are connected through AttachThreadInput share the same keyboard state.)
...
To retrieve the current state for an individual key regardless of whether the corresponding keyboard message has been retrieved from the message queue, use the GetAsyncKeyState function."

Rosalie241:
I decided not to remove `osal_virtual_key_to_hid` because `Qt::Key` doesn't contain i.e keypad keys, so I added the `QtKeyToHID` function and pass the result of that to a new argument in `osal_virtual_key_to_hid` (so the platform specific code can choose to use it or not.)

here's a quick video showcasing the functionality on linux: https://gfycat.com/immaculatesillydiamondbackrattlesnake

gonetz:
> here's a quick video showcasing the functionality on linux: https://gfycat.com/immaculatesillydiamondbackrattlesnake

Cool!

src/GLideNUI/ConfigDialog.cpp
```diff
@@ -1113,7 +1116,7 @@ void ConfigDialog::on_btn_clicked() {
 			//msgBox.setIcon(QMessageBox::Information);
 			msgBox.exec();
 			if (msgBox.m_nativeVK != 0) {
-				const unsigned int hidCode = osal_virtual_key_to_hid(msgBox.m_nativeVK);
+				const unsigned int hidCode = osal_virtual_key_to_hid(msgBox.m_nativeVK, QtKeyToHID(msgBox.m_Key));
```
gonetz, line 1119: QtKeyToHID makes osal_virtual_key_to_hid redundant. Lets remove it.

src/osal/osal_keys_unix.c
```diff
@@ -1,22 +1,222 @@
 #include "osal_keys.h"
 
+#ifdef OS_LINUX
```
gonetz, line 3: All these #ifdef OS_LINUX look ugly. If this implementation is Linux-specific, it is better to split the file on two parts
#ifdef OS_LINUX
Linux implementation
#else
Dummy implementation
#endif

gonetz (changes requested):

src/GLideNUI/ConfigDialog.cpp
Rosalie241: there's a problem, i.e certain hotkeys wont work due to `Qt::Key` being limited (for example the keypad numkeys wont work on windows if I remove `osal_virtual_key_to_hid`), if you're fine with losing the ability to do such things, then sure I'll remove it.

gonetz: keypad numkeys aren't working on Windows with the current code, at least on my keyboard. If this is the only problem, please remove sal_virtual_key_to_hid.

Rosalie241: pushed

gonetz:
Thanks!

Issue #2048, opened 2019-05-12
GLSL 4.20 is not supported - Pink discoloration and heavy slowdowns on AMD card


HandsMaulman:
My Linux System (Ubuntu 18.04.2) is a Laptop with switchable Intel i7 Sandybridge Mobile and AMD 6630M Graphics.
Using the AMD card and starting e.g. Hybrid Heaven gives the game a pink discoloration in some parts (title screen, ingame-HUD), slowdown and audio issues. On Intel it looks and plays fine. (Playing on Parallel64 core also has no issues.) This is both on standalone Mupen64plus and the Retroarch core. Issues remain after updating Mesa drivers through ubuntu-staic-x ppa.
 
Gliden64.log:
```
shader_compile error: 0:1(10): error: GLSL 4.20 is not supported. Supported versions are: 1.10, 1.20, 1.30, 1.40, 1.50, 3.30, 1.00 ES, 3.00 ES, 3.10 ES, and 3.20 ES

Error in vertex shader#version 420 core											
layout(binding = 0, r16ui) uniform readonly uimage2D uZlutImage;
layout(binding = 1, r16ui) uniform readonly uimage2D uTlutImage;
layout(binding = 0) uniform sampler2D uDepthImage;		
uniform lowp vec4 uFogColor;								
out lowp vec4 fragColor;									
lowp float get_alpha()										
{															
  mediump ivec2 coord = ivec2(gl_FragCoord.xy);			
  highp float bufZ = texelFetch(uDepthImage,coord, 0).r;	
  highp int iZ = bufZ > 0.999 ? 262143 : int(floor(bufZ * 262143.0));
  mediump int y0 = clamp(iZ/512, 0, 511);					
  mediump int x0 = iZ - 512*y0;							
  highp uint iN64z = imageLoad(uZlutImage,ivec2(x0,y0)).r;		
  highp float n64z = clamp(float(iN64z)/65532.0, 0.0, 1.0);
  highp int index = min(255, int(n64z*255.0));				
  highp uint iAlpha = imageLoad(uTlutImage,ivec2(index,0)).r;
  return float(iAlpha>>8)/255.0;							
}															
void main()												
{															
  fragColor = vec4(uFogColor.rgb, get_alpha());			
}										
```					
~$ DRI_PRIME=1 glxinfo | grep OpenGL

```
OpenGL vendor string: X.Org
[...]

PR #2381, opened 2020-11-11
Removed GLES version check for glBufferStorage, allowing for any device that supports it.

dankcushions:
Tested on RPI4 - please see profile comparison of Goldeneye 64:
https://docs.google.com/spreadsheets/d/1kQ15ra9rFnzVx6Plukck8cEPczf446eRlf1rCKm21FI/edit?usp=sharing

When glBufferStorage is permitted calls to `glDrawArraysUnbuffered` and `glDrawElementsUnbuffered` (highlighted in yellow in first 'before' tab) become calls to `glDrawArrays`, `glDrawRangeElementsBaseVertex` and `glBufferStorage` (highlighted in yellow in the second 'after' tab). The total duration of said calls within each profiling interval (ends highlighted in red) are reduced in the 'after' scenario:

1.009887 (before) vs 0.7056322 (after)
0.739989 vs 0.565734

Please verify that I am comparing the right calls! This is based on conversations starting here: https://github.com/gonetz/GLideN64/issues/2329#issuecomment-693515664 - tagging @fzurita, @loganmc10 

I have only checked the intro of Goldeneye as it's a game where there's visible slowdown for the pi4. Please let me know if you want me to profile anything else. There was some concern that bufferStorage could cause slowdown, but I searched through the history and that line was introduced here: https://github.com/gonetz/GLideN64/commit/a625225323c902b614ed9601143df3bc51550fc4, and there's some related comments, but nothing that specifically indicates this will cause slowdown in devices that support it.

Thanks :)

### DIFF
--- a/src/Graphics/OpenGLContext/opengl_ContextImpl.cpp
+++ b/src/Graphics/OpenGLContext/opengl_ContextImpl.cpp
@@ -67,7 +67,7 @@ void ContextImpl::init()
 	}
 
 	{
-		if ((m_glInfo.isGLESX && (m_glInfo.bufferStorage && m_glInfo.majorVersion * 10 + m_glInfo.minorVersion >= 32)) || !m_glInfo.isGLESX)
+		if ((m_glInfo.isGLESX && m_glInfo.bufferStorage) || !m_glInfo.isGLESX)
 			m_graphicsDrawer.reset(new BufferedDrawer(m_glInfo, m_cachedFunctions->getCachedVertexAttribArray(), m_cachedFunctions->getCachedBindBuffer()));
 		else
 			m_graphicsDrawer.reset(new UnbufferedDrawer(m_glInfo, m_cachedFunctions->getCachedVertexAttribArray()));
--- a/src/Graphics/OpenGLContext/opengl_ContextImpl.cpp
+++ b/src/Graphics/OpenGLContext/opengl_ContextImpl.cpp
@@ -67,7 +67,7 @@ void ContextImpl::init()
 	}
 
 	{
-		if ((m_glInfo.isGLESX && m_glInfo.bufferStorage) || !m_glInfo.isGLESX)
+		if ((m_glInfo.isGLESX && (m_glInfo.bufferStorage && m_glInfo.drawElementsBaseVertex)) || !m_glInfo.isGLESX)
 			m_graphicsDrawer.reset(new BufferedDrawer(m_glInfo, m_cachedFunctions->getCachedVertexAttribArray(), m_cachedFunctions->getCachedBindBuffer()));
 		else
 			m_graphicsDrawer.reset(new UnbufferedDrawer(m_glInfo, m_cachedFunctions->getCachedVertexAttribArray()));
--- a/src/Graphics/OpenGLContext/opengl_GLInfo.cpp
+++ b/src/Graphics/OpenGLContext/opengl_GLInfo.cpp
@@ -100,6 +100,8 @@ void GLInfo::init() {
 		config.generalEmulation.enableHybridFilter = 0;
 	}
 
+	drawElementsBaseVertex = !isGLESX || Utils::isExtensionSupported(*this, "GL_EXT_draw_elements_base_vertex");
+
 	bufferStorage = (!isGLESX && (numericVersion >= 44)) || Utils::isExtensionSupported(*this, "GL_ARB_buffer_storage") ||
 			Utils::isExtensionSupported(*this, "GL_EXT_buffer_storage");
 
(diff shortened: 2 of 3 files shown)


loganmc10:
I believe that check is there because ```glDrawRangeElementsBaseVertex``` is only supported on GLES3.2+, the raspberry Pi probably supports ```GL_EXT_draw_elements_base_vertex``` which is am extension that enables it.

So the proper thing to do would be to keep the GLES3.2 check, but and an OR for ```GL_EXT_draw_elements_base_vertex```

dankcushions:
the pi4 supports bufferStorage directly as far as i can tell: https://gitlab.freedesktop.org/mesa/mesa/-/blob/master/docs/features.txt#L397

doesn't `&& m_glInfo.bufferStorage` check for existence of this extension directly? if so, a gles3.2 device would pass this check? maybe i'm misunderstanding the check!

loganmc10:
We are talking about 2 different extensions, ```GL_EXT_draw_elements_base_vertex``` isn't the same as the buffer storage extension, though they are used together.

I'm saying a device could support glBufferStorage, but not support glDrawRangeElementsBaseVertex, in which case we can't use glBufferStorage

dankcushions:
ah I see, so i guess we could add it to `GLInfo` in `opengl_GLInfo.cpp` like:
```
	drawElementsBaseVertex = (!isGLESX && (numericVersion >= 32)) || Utils::isExtensionSupported(*this, "
`GL_ARB_draw_elements_base_vertex`") ||
			Utils::isExtensionSupported(*this, "GL_EXT_draw_elements_base_vertex");
```

and then change the check to
```
if ((m_glInfo.isGLESX && (m_glInfo.bufferStorage && m_glInfo.drawElementsBaseVertex)) || !m_glInfo.isGLESX)
```
?

Jj0YzL5nvJ:
The problem is that currently it's immediately assumed that if "bufferStorage" is available, then the GL/GLES version is _X_ or _higher_. But such logic is totally wrong for Mesa drivers in general. #2048

If said behavior is not changed, this change will create more serious issues than those that are supposed to be fixed.

dankcushions:
Yes I guess the presumption of the availability of GL_ARB_draw_elements_base_vertex is an example of this.

However isn't what [GLInfo.cpp](https://github.com/gonetz/GLideN64/blob/master/src/Graphics/OpenGLContext/opengl_GLInfo.cpp) is supposed to address, by checking for the presence of individual extensions for certain things? Perhaps it should do that for ALL extensions, ratehr than relying on version numbers, but that's a large change!

bufferStorage doesn't seem to be used much in the code so I am hoping that it's not such a huge change for just this one. If I add the check for drawElementsBaseVertex is it not enough?

gonetz:
@loganmc10 , @Jj0YzL5nvJ please review that fix and if it is wrong, please suggest how it can be corrected.

fzurita:
I think if he does what @loganmc10 says, it should be ok.

dankcushions:
I have put the additional check in and everything seems to work for me.

loganmc10:
Given that the minimum OpenGL version for the plugin is 3.3, you don't really need:

```(!isGLESX && (numericVersion >= 32)) || Utils::isExtensionSupported(*this, "GL_ARB_draw_elements_base_vertex")```

It could just be:
```drawElementsBaseVertex = !isGLESX || Utils::isExtensionSupported(*this, "GL_EXT_draw_elements_base_vertex");```

**Comment by loganmc10** [comment]
Looks good to me now!

**Comment by Jj0YzL5nvJ** [comment]
The specific case I referenced only affected ATI/AMD Mesa drivers limited to OpenGL 3.3. [Testing on such hardware is no longer possible ...at least from my side.](https://github.com/gonetz/GLideN64/pull/1738#issuecomment-369848827)
This change looks harmless to PC users.

File: src/Graphics/OpenGLContext/opengl_ContextImpl.cpp [hunk 1]
```diff
@@ -67,7 +67,7 @@ void ContextImpl::init()
 	}
 
 	{
-		if ((m_glInfo.isGLESX && (m_glInfo.bufferStorage && m_glInfo.majorVersion * 10 + m_glInfo.minorVersion >= 32)) || !m_glInfo.isGLESX)
+		if ((m_glInfo.isGLESX && (m_glInfo.bufferStorage && m_glInfo.drawElementsBaseVertex)) || !m_glInfo.isGLESX)
```
gonetz, line 70: The condition looks crazy now. Should not it be
`if (!m_glInfo.isGLESX || (m_glInfo.bufferStorage && m_glInfo.drawElementsBaseVertex))`
?

gonetz:
I corrected the condition and merged this PR: cb76d2f39d

fzurita:
Well, it's never a free lunch. This broke power VR devices that say they support the extension.

PR #2404, opened 2020-12-14
Disable draw_elements_base_vertex with PowerVR

fzurita:
Disable "draw_elements_base_vertex" extension if renderer is PowerVR and GLES is < 3.2.

This will allow the extension to work if the PowerVR GPU supports GLES 3.2.

### DIFF
--- a/src/Graphics/OpenGLContext/opengl_GLInfo.cpp
+++ b/src/Graphics/OpenGLContext/opengl_GLInfo.cpp
@@ -104,7 +104,9 @@ void GLInfo::init() {
 		config.generalEmulation.enableHybridFilter = 0;
 	}
 
-	drawElementsBaseVertex = !isGLESX || Utils::isExtensionSupported(*this, "GL_EXT_draw_elements_base_vertex");
+	drawElementsBaseVertex = !isGLESX ||
+		(Utils::isExtensionSupported(*this, "GL_EXT_draw_elements_base_vertex") && (renderer != Renderer::PowerVR ||
+		numericVersion >= 32));
 
 	bufferStorage = (!isGLESX && (numericVersion >= 44)) || Utils::isExtensionSupported(*this, "GL_ARB_buffer_storage") ||
 			Utils::isExtensionSupported(*this, "GL_EXT_buffer_storage");


src/Graphics/OpenGLContext/opengl_GLInfo.cpp
```diff
@@ -104,7 +104,10 @@ void GLInfo::init() {
 		config.generalEmulation.enableHybridFilter = 0;
 	}
 
-	drawElementsBaseVertex = !isGLESX || Utils::isExtensionSupported(*this, "GL_EXT_draw_elements_base_vertex");
+	drawElementsBaseVertex = !isGLESX || (
+		(Utils::isExtensionSupported(*this, "GL_EXT_draw_elements_base_vertex") && (renderer != Renderer::PowerVR)) ||
+		 (numericVersion >= 32)
+		 );
```
gonetz, line 110: The condition does not correspond to the fix description. It seems that you put parantheses in a wrong place.
Since A || (B || C) == A || B || C
the condition is equvalent to 
```
!isGLESX ||
(Utils::isExtensionSupported(*this, "GL_EXT_draw_elements_base_vertex") && (renderer != Renderer::PowerVR)) ||
(numericVersion >= 32)
```
That is it will b true for any render, which supports GLES version >= 3.2
I suppose it is not what you want. I think, it should be
```
!isGLESX ||
(Utils::isExtensionSupported(*this, "GL_EXT_draw_elements_base_vertex") && (renderer != Renderer::PowerVR ||
numericVersion >= 32))
```
That is it is true if not GLES or if the extension is supported AND either render is not PowerVR or GLES version is >= 3.2. Are you agree?

fzurita: Yes, I agree, an extra set of parenthesis is not needed due to the equivalence you described. To be truthful though, I feel like your modification would achieve the same thing since either GLES 3.2+ or GL 3.2+ will both support the extension.

I do think your modification, would be more explicit in stating that PowerVR is expected to work correctly if GLES is 3.2+.

So I'm ok, with your modification.

fzurita:
I updated it as requested.

PR #2405, opened 2020-12-15
Compute coverage value and use it in the combiner/blender stage

standard-two-simplex:
I am quite satisfied with the current state of the branch. I propose this changes for revision and testing. 

These are the only concerns I have.
- Some primitives in line mode in Turok 2 are gray in AngryLion's plugin. This is because coverage times alpha flag is on. Fixed point arithmetic makes the value smaller than full coverage when multiplying full coverage and full alpha. The floating point arithmetic is doing 1.0 times 1.0 which ends up  as 1.0. I don't know if it is worth trying to emulate such behaviour. I also wonder if real hardware shows those gray polygons.
- In general, many more edges with partial coverage are shown than in AngryLion's plugin. This might be because the polygons are rendered differently by OpenGL or because incorrect blending. I don't know what to do about it.
- Barycentric coordinates are setup in 7c39181. I wonder if there are better places to initialize the vertex attributes.
- 21c549f only adds a comment that can be disabled to check coverage. It can be discarded if it is not wanted.


Jj0YzL5nvJ:
> Some primitives in line mode in Turok 2 are gray in AngryLion's plugin. This is because coverage times alpha flag is on. Fixed point arithmetic makes the value smaller than full coverage when multiplying full coverage and full alpha. The floating point arithmetic is doing 1.0 times 1.0 which ends up as 1.0. I don't know if it is worth trying to emulate such behaviour. I also wonder if real hardware shows those gray polygons.

Turok 2 has two "Pen and Ink Mode" hi-res and low-res: https://youtu.be/n17SDFZSu_o?t=21m58s

gonetz:
I didn't find any major problems. I pushed the code in https://github.com/gonetz/GLideN64/tree/standard-two-simplex-cvg  branch because it is not completed yet. It is implemented for bufferd drawer only. I also think that this feature should be optional (controlled by a setting) because it uses quite heavy shader calculations, which may have bad impact on performance.

standard-two-simplex:
What is the difference between the buffered and unbuffered drawer? How do I run the plugin in each mode?

gonetz:
Buffered drawer uses VBO and VAO to pass vertices data, unbuffered one does not.
Unbuffered drawer is used for GL ES renderes, which do not support buffer_storage extensiion.
You may force use the particular buffer in ContextImpl::init(), but probably it is not enough for correct work. I'll check it.

gonetz:
It is quite easy to enable unbuffered drawer for Windows build;
* force unbuffered drawer usage in ContextImpl::init()
```
		//if (!m_glInfo.isGLESX || (m_glInfo.bufferStorage && m_glInfo.drawElementsBaseVertex))
		//	m_graphicsDrawer.reset(new BufferedDrawer(m_glInfo, m_cachedFunctions->getCachedVertexAttribArray(), m_cachedFunctions->getCachedBindBuffer()));
		//else
			m_graphicsDrawer.reset(new UnbufferedDrawer(m_glInfo, m_cachedFunctions->getCachedVertexAttribArray()));

```
* Since GL 3.2 and above requires VBO, unbuffered drawer is incompatible with 3.2 core profile. Thus we need to ask for the compatible profile in WindowsWGL::start()
```
			const int attribList[] =
			{
				WGL_CONTEXT_MAJOR_VERSION_ARB, majorVersion,
				WGL_CONTEXT_MINOR_VERSION_ARB, minorVersion,
				WGL_CONTEXT_PROFILE_MASK_ARB, WGL_CONTEXT_COMPATIBILITY_PROFILE_BIT_ARB,
				0        //End
			};

```
Patch:
[0001-Enable-unbuffered-drawer.patch.zip](https://github.com/gonetz/GLideN64/files/5734466/0001-Enable-unbuffered-drawer.patch.zip)

gonetz:
I corrected unbuffered drawer: e689d77

However, now number of attributes is 10. AFAIR, some GLES devices support only up to 8 vertex attributes. @fzurita could you confirm it?  If it is so, this feature must be automatically disabled for such devices. Also, I still need to make it optional, at least for GLES.

fzurita:
I can check. Although, I did a quick search and found that the maximum number of vertex attributes can be retrieved by ```glGetIntegerv(GL_MAX_VERTEX_ATTRIBS, &n)```

gonetz:
Yes, I can probe for maximum number of vertex attributes and disable coverage if the max is less than 10.
There is one issue: index of the attribute triangleAttrib::barycoords is currently 5. If there are devices with GL_MAX_VERTEX_ATTRIBS less than 10, then triangleAttrib::barycoords should be 8, what is a bit uncomely.

standard-two-simplex:
> I corrected unbuffered drawer: e689d77

Thanks

> AFAIR, some GLES devices support only up to 8 vertex attributes

That's a bummer. Since triangles aren't rectangles and viceversa, wouldn't it be possible to only use one set of attributtes and somehow repeat indices? Or perhaps standarize names and upload information differently based on which vertex shader will be run (for example texcoord can use two channels of a vec4 for a triangle and four channels of a vec4 for rects (for texcoord0 and texcoord1)?

fzurita:
I checked several devices. Even the lowest end one supported 12 vertex attributes. Most support 16 vertex attributes. And Adreno shorts the most at 32 vertex attributes.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -2731,6 +2731,12 @@ CombinerInputs CombinerProgramBuilder::compileCombiner(const CombinerKey & _key,
 	}
 
 
+	if (g_cycleType <= G_CYC_2CYCLE) {
+		ssShader << "if (uCvgXAlpha != 0) cvg *= clampedColor.a;" << std::endl;
+		ssShader << "if (uAlphaCvgSel != 0) clampedColor.a = cvg; " << std::endl;
+	}
```
gonetz, line 2737: duplication. I already moved these lines to the if() clause above.

standard-two-simplex:
I added a fixup commit for `update blender to use coverage` so that it correctly uses coverage now.

>  I also think that this feature should be optional (controlled by a setting) because it uses quite heavy shader calculations, which may have bad impact on performance.

I think this can be implemented in a simple way by computing boolean `bUseCoverage` and doing,
```
	if (g_cycleType <= G_CYC_2CYCLE && bUseCoverage)
		m_shaderCoverage->write(ssShader);
	else
		ssShader << "cvg = 1.0f; \n" << std::endl;
```

This way if `CvgXAlpha` is selected cvg will be `clampedColor.a` and otherwise 1 (which shouldn't be worse than `clampedColor.a` if CvgXAlpha is off.

gonetz:
How to compute bUseCoverage ?

standard-two-simplex:
> How to compute bUseCoverage ?

It should be the config option by the user.

Coverage is used by color blender if `alpha_cvg_select` is picked and by the alpha blender always (if it enabled by dual_source_blending or framebuffer_fetch that is). Since the former is used mainly for N64 style like antialiasing (which is not implemented) I wonder if it is worth forcing it to false if the extensions aren't available. I find it likely that a device without support for the extensions will not benefit from the calculations and will suffer the performance overhead, but this is of course especulation.

gonetz:
> It should be the config option by the user.

You said that "this can be implemented in a simple way by computing boolean bUseCoverage", that is bUseCoverage should be computed. If you meant that it should be config option, ok, I'll add it and use as you suggested.

standard-two-simplex:
> If you meant that it should be config option, ok, I'll add it and use as you suggested.

I used those words because I considered the possibility of doing something like, `bUseCoverage = config.generalEmulation.enableCoverage && (gl_info.dual_source_blending || gl_info.ext_fetch || gl_info.ext_fetch_arm)`.

But it is also possible to simply use the config option and tell users they can disable it if they need performace.

gonetz:
> I used those words because I considered the possibility of doing something like, bUseCoverage = config.generalEmulation.enableCoverage && (gl_info.dual_source_blending || gl_info.ext_fetch || gl_info.ext_fetch_arm).

Ok. If bUseCoverage  is defined like this, user still can disable that feature for performance sake.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
standard-two-simplex: Ah, sorry, I missed it. I guess I duplicated it when rebasing.

gonetz:
Done in 32dbfe1d1f. @standard-two-simplex please check my commit.

@oddMLan I added new checkbox to the Emulation tab of the configuration dialog. Please do the same for GLideNUI-wtl.

standard-two-simplex:
> Done in 32dbfe1. @standard-two-simplex please check my commit.

Seems to be working as intended.

The fixup commit I made is supposed to be squashed with `update blender to show coverage` commit. I didn't squash them myself not to mess with commit history and prevent it from going unnoticed. So you might squash them into one commit.

> I checked several devices. Even the lowest end one (GLES 2.0) supported 12 vertex attributes. Most support 16 vertex attributes. And Adreno supports the most at 32 vertex attributes.

About this, do we leave 10 vertex attributtes and see if someone raises an issue report? Or is it worth taking action just in case?

Besides, I have one cleanup commit on the way, so hold the branch for now, please.

fzurita:
I think we should only fix it if someone raises an issue.

fzurita:
Here is a lookup, there are indeed some devices that only support 8, but they are all GL ES 2.0. All GL ES 3.0 devices support at least 16:

https://opengles.gpuinfo.org/displaycapability.php?name=GL_MAX_VERTEX_ATTRIBS

standard-two-simplex:
> Besides, I have one cleanup commit on the way, so hold the branch for now, please.

I added two commits. The first one sets up two barycentric coordinates as vertex attributes to match the number rectangles use. Now the third is computed in the vertex shader for triangles (plus a dummy positive coordinate in the fourth channel). The third and fourth were being computed in the vertex shader for rectangles as well. This way, it should be easier to unify vertex attributtes for triangles and rectangle if necessary.

The second one cleans the computation of coverage value, making it clearer how it works. However, there is a for loop inside of a for loop, which I had avoided previously. I'm hoping that compilers will be smart enough to unroll those loops, but not using the commit is also an option.

gonetz:
> Here is a lookup, there are indeed some devices that only support 8, but they are all GL ES 2.0.
So the easy answer, disable this for GLES 2.0 and we should be all set?

Thanks for the info. No, disable this for GLES 2.0 is not enough. Currently rectAttrib::texcoord1 has index 8 and thus will not work for such devices even with coverage disabled. The simple solution is to set triangleAttrib::barycoords to 8 and disable coverage for GLES 2.0. It is also necessary to disable barycoords attributes initialization and activation for GLES 2.0 to avoid GL errors.

gonetz:
> The fixup commit I made is supposed to be squashed with update blender to show coverage commit. I didn't squash them myself not to mess with commit history and prevent it from going unnoticed. So you might squash them into one commit.

Ok, or you may re-arrange commits as necessary and make another PR.


> I added two commits.

I merged them.

> The second one cleans the computation of coverage value, making it clearer how it works. However, there is a for loop inside of a for loop, which I had avoided previously. I'm hoping that compilers will be smart enough to unroll those loops, but not using the commit is also an option.

Yes, compilers should unroll these loops.

gonetz:
I implemented checks for max vertex attributes. The task looks finished for me.
@fzurita could you build standard-two-simplex-cvg branch and test it on Android devices? I'd like to know that nothing is broken before merging it to master.

fzurita:
Sure, I will try a few of my devices.

fzurita:
I tried various GLES 3.0 devices (Power VR, Intel, Mali, Adreno) and they were all able to compile the shader successfully. I tried the pen and ink mode in Goldeneye in all those devices and it displayed as I have seen in screenshots.

fzurita:
Also, I forgot to mention, it's a 30 to 40% hit in performance.

oddMLan:
> Also, I forgot to mention, it's a 30 to 40% hit in performance.

It should be a per-game option then. 99% of the library doesn't use this mode.

fzurita:
It's already optional, I'm leaving it disabled unless explicitly enabled.

gonetz:
@fzurita Thanks! So, it can be merged now.

@oddMLan yes, it is optional. I added new checkbox to the Emulation tab of the configuration dialog. Please do the same for GLideNUI-wtl.

oddMLan:
> @oddMLan yes, it is optional. I added new checkbox to the Emulation tab of the configuration dialog. Please do the same for GLideNUI-WTL.

I opened a Pull request for it last night.

I guess it needs to be enabled in the ini for turok ii, south park and other titles that make real use of coverage.

Besides the wireframe effect, does coverage have any other uses?

gonetz:
> I opened a Pull request for it last night.

Ops! I did not notice that and just removed the branch :(

gonetz:
@oddMLan It seems to be a bit incomplete. Copy-paste?

gonetz:
wtl fix works fine, thanks!

> I guess it needs to be enabled in the ini for turok ii, south park and other titles that make real use of coverage.

It is arguable. Users hardly will be happy to get 30 to 40% drop in performance in these games all of sudden.

standard-two-simplex:
> It's a pretty darn good approximation, but it looks like the lines are slightly thicker in angrylion and some textures should look gray-ish (the leaves) whereas in your algo they look white.

@oddMLan Is it correct to assume you rendered the scene in native resolution?

There are a few nuances that are hard to emulate accurately.
-  When writing to memory, the N64 writes `cvg - 1` in  u0.3 fixed point format. Then, when reading from memory `cvg + 1` is computed. The second part is difficult to do with OpenGL's blending, so I ignored both the substraction and addition.
- There is a cvg blend mode that allows coverage overflow. Again, this is not supported by OpenGL's blending.
- The gray pixels on the second picture are polygons with `cvgXAlpha` enabled. I thought this was due to fixed point arithmetic, but I'm not so sure anymore, so I will investigate a little further.

standard-two-simplex:
> The gray pixels on the second picture are polygons with cvgXAlpha enabled. I thought this was due to fixed point arithmetic, but I'm not so sure anymore, so I will investigate a little further.

I did some research and it is indeed due to fixed point arithmetic nuances. Those polygons are in 2 cycle mode. In the first cycle the combiner does `(texel_alpha - 0.0) * primitive_alpha + 0.0`. Primitive color holds `0x7F7F7FFF` so the alpha channel is `0xFF`. The texel alph is also `0xFF` so the result ends up being `(0xFF - 0x00) * 0xFF + 0x00) >> 8 = 0xFE`. The second cycle just selects the output of the first cycle.

Afterwards, pixel alpha is multiplied by coverage. The interiors of those primitives have full coverage `0x8` so cvg times alpha becomes `(0xFE * 0x8) >> 3 = 0xFE` where coverage is padded to the three most significant bytes. Angrylion's code truncates this value (rather than round) and thus coverage becomes `0xFE >> 5 = 0x7`, which ends up as slightly grey.

If someone could confirm that this is the correct behaviour on a N64 it would be great.

In GLideN64, the 0xFF are cast to 1.0 so the end value after multiplications becomes `1.0` instead of `7.0/8.0`.

There are two possibilities to reproduce this behaviour.
- The most obvious would be performing shader operations in integers. This would require to load all textures as integer data, since texel fetches should return integer colors, which will then be used by the combiner. 
- The other option would be casting integers to floats as `x -> x/256.0` (for 8bit channels) and loading such values to the texture, instead of the more standard `x->x/255.0`. Then `0xFF * 0xFF` would become `(255.0*255.0)/(256.0*256.0) = 7.93/8.0` which would still have to be truncated to 7.0.

A bit complicated, but at least all combiner operations would be in general more accurate.

gonetz:
@standard-two-simplex I just notice one thing. Barycentric coordinates are set in GraphicsDrawer::addTriangle as
```
	for (u32 i = firstIndex; i < triangles.num; ++i) {
		SPVertex& vtx = triangles.vertices[triangles.elements[i]];
		vtx.bc0 = i - firstIndex == 0 ? 1.0f : 0.0f;
		vtx.bc1 = i - firstIndex == 1 ? 1.0f : 0.0f;
	}

```
Suppose, we have 4 vertices V0 .. V3 and 2 triangles T1(V0, V1, V2), T2(V1, V2, V3)
There will be two calls of addTriangle :
addTriangle(V0, V1, V2)
addTriangle(V1, V2, V3)

V1 will get bc0 0.0, bc1 1.0 after the first addTriangle,
but next addTriangle will change it to
bc0 1.0, bc1 0.0

thus coverage result for the first triangle will be incorrect. Am I right?

standard-two-simplex:
> thus coverage result for the first triangle will be incorrect. Am I right?

I think coverage value should still be correct. The important thing is that each vertex in a triangle has one barycentric coordinate as 1.0 and the others 0.0. It doesn't matter which coordinate is one, and which are zero.

For example, if we initialize
v0 with bc0 = 1, bc1 = 0, bc2 = 0
v1 with bc0 = 0, bc1 = 1, bc2 = 0
v2 with bc0 = 0, bc1 = 0, bc2 = 0
coverage will be computed correctly, but if we initialize
v0 with bc0 = 0, bc1 = 1, bc2 = 0
v1 with bc0 = 0, bc1 = 0, bc2 = 1
v2 with bc0 = 1, bc1 = 0, bc2 = 0
coverage will also be correct.

Note, bc2 is computed in vertex shader as 1 - bc0 - bc1.

So in your example, both triangles will have a valid setup.

standard-two-simplex:
There's one thing I'm not so sure about though.
If you send those two calls, addTriangle(V0, V1, V2) and addTriangle(V1, V2, V3), will 6 vertices be added to triangles.vertices?
If 6 are loaded (which is what I understand) there should be no problem, but if 4 are loaded and rendered as a triangle strip then it should be,
```
	for (u32 i = firstIndex; i < triangles.num; ++i) {
		SPVertex& vtx = triangles.vertices[triangles.elements[i]];
		vtx.bc0 = (i - firstIndex) % 3 == 0 ? 1.0f : 0.0f;
		vtx.bc1 = (i - firstIndex) % 3 == 1 ? 1.0f : 0.0f;
	}
```

gonetz:
> If you send those two calls, addTriangle(V0, V1, V2) and addTriangle(V1, V2, V3), will 6 vertices be added to triangles.vertices?

No. Display list first loads fixed number of vertices. Following triangle commands use these vertices. addTriangle function adds indices of vertices to the list of elements to draw. Your fix will not help, because if a vertex is shared by several triangles, its barycentric coordinates will be valid only for the last triangle it belongs to. In my example with 4 vertices we have
V0(1, 0) V1(0, 1),  V2(0, 0) 
after the first addTriangle, but
V1(1, 0) V2(0, 1), V3(0, 0)
after the second one
so, the first triangle will get coordinates V0(1, 0), V1(1, 0) V2(0, 1) and the result will be wrong

standard-two-simplex:
You're right, I was overwriting previous data before sending it.

I can't see any better way than duplicating vertices, which is a bit of a pity.

When applying clipping you have two choices. One possibility is to write ones and zeros to the new vertices. Another option would be interpolating the original values so that the barycentric coordinates are calculated as if it were unclipped.

gonetz:
> I can't see any better way than duplicating vertices, which is a bit of a pity.

yep



> When applying clipping you have two choices. One possibility is to write ones and zeros to the new vertices. Another option would be interpolating the original values so that the barycentric coordinates are calculated as if it were unclipped.

I think I'll use the first approach. I'm not sure how to interpolate the original values to the clipped ones to get the correct result. The result of triangle clipping is a set of new triangles. If I interpolate clipped barycentric coordinates the same way as other coordinates, will the coverage calculation still work correctly?

standard-two-simplex:
> If I interpolate clipped barycentric coordinates the same way as other coordinates, will the coverage calculation still work correctly?

It will compute coverage as if it weren't clipped. I. e, the new edges will be treated as an interior.

This [extension ](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_fragment_shader_barycentric.txt)uses a similar method, but it isn't widely supported. You may check the second issue at the bottom for information. 

I am not saying it is necessarily better, just that it is an option. You may use the one you find easier to implement.

gonetz:
> It will compute coverage as if it weren't clipped. I. e, the new edges will be treated as an interior.

Ok, thanks.

> This [extension ](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_fragment_shader_barycentric.txt)uses a similar method, but it isn't widely supported. You may check the second issue at the bottom for information.

Wow! I did not know that it is supported as an extension!

Narann:
> This extension uses a similar method, but it isn't widely supported.

Rational here: https://opengl.gpuinfo.org/listreports.php?extension=GL_NV_fragment_shader_barycentric

Coverage: 2.71%

PR #2407, opened 2020-12-19
Fix missing check for frame buffer fetch in unemulated blending modes

fzurita:
This fixes Turok text as seen in this screenshot:
turok_dinosaur_hunte-000

There is still a problem with blast corps. It has flickering textures. @standard-two-simplex can you check if blast corps has flickering textures in PC with dual source blending? I can't find any other problems compared to the dual source blending implementation.

The flickering goes away in blast corps if I force "ordinary blending"

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp
@@ -358,7 +358,7 @@ public:
 		const int forceBlend2 = gDP.otherMode.forceBlender;
 		uForceBlendCycle2.set(forceBlend2, _force);
 
-		if (!graphics::Context::DualSourceBlending || dwnd().getDrawer().isTexrectDrawerMode()) {
+		if (!(graphics::Context::DualSourceBlending || graphics::Context::FramebufferFetchColor) || dwnd().getDrawer().isTexrectDrawerMode()) {
 			// Modes, which shader blender can't emulate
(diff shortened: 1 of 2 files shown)


fzurita:
I'm not seeing any textured rects in the flickering portions of blast corps, so no opportunities for the updated blending to be disabled. I'm at a loss with blast corps...

fzurita:
I enabled the blast corps hack for frame buffer fetch blending in the pull request.

standard-two-simplex:
The first commit looks correct to me.
The second one is working, but it is calling ordinary blending function, then following the path of the hack. It would be cleaner to do
```
if (Context::FramebufferFetchColor && !isTexrectDrawerMode()) {
    gfxContext.enable(enable::BLEND, false);
    if (config.generalEmulation.hacks & hack_blastCorps) != 0 &&  gSP.texture.on == 0 && currentCombiner()->usesTexture()) {
        gfxContext.enable(enable::BLEND, true);
        gfxContext.setBlending(blend::ZERO, blend::ONE);
    }
    return;
}
```

Another possibility would be converting the ifs in setBlendMode() to an if-else chain, remove the returns and override the blending mode before exiting setBlendMode() if the hack condition is met. This way there would be no need to check for the hack in every blending path. I don't know if it is much cleaner though.

fzurita:
I made an update to make the hack usage cleaner and common between all blending modes. Can you check that it works with dual source blending? I don't have a way access to a PC.

standard-two-simplex:
Dual source blending and ordinary blending are working well on my PC.

fzurita:
ok, then I think this should be good to merge.

PR #2411, opened 2020-12-21
GLideNUI: correct macro in saveCustomRomSettings

Rosalie241:
This fixes per-game ROM settings texture directories:

to reproduce the issue:
1) delete `GLideN64.ini`
2) open a game
3) enable texture pack for said game & configure directories
4) re-open config dialog

image


gonetz:
yep, a stupid mistake. Thanks!

PR #2412, opened 2020-12-21
GLideNUI: attempt to mkdir texture pack directories

Rosalie241:
This improves the user experience on i.e the first run, instead of telling the user the directories don't exist, it'll attempt to create them.


Rosalie241:
I modified the commit to also allow relative paths instead of GLideN64 turning relative paths into absolute paths,

this also improves the user experience because then you can have an actual portable installation.

gonetz:
Looks good, thanks!

PR #2414, opened 2020-12-23
Implement cross-platform max MSAA level detection

Rosalie241:
This PR implement cross-platform max MSAA level detection

### DIFF
--- a/src/DisplayWindow.cpp
+++ b/src/DisplayWindow.cpp
@@ -19,6 +19,11 @@ bool DisplayWindow::start()
 	m_drawer._initData();
 	m_buffersSwapCount = 0;
 
+	// only query max MSAA when needed
+	if (m_maxMsaa == 0) {
+		m_maxMsaa = gfxContext.getMaxMSAALevel();
+	}
+
 	return true;
 }
 
@@ -193,5 +198,5 @@ void DisplayWindow::readScreen2(void * _dest, int * _width, int * _height, int _
 
 u32 DisplayWindow::maxMSAALevel() const
 {
-	return _maxMSAALevel();
+	return m_maxMsaa;
 }
--- a/src/GLideNUI-wtl/Settings.cpp
+++ b/src/GLideNUI-wtl/Settings.cpp
@@ -28,6 +28,7 @@ void _loadSettings(GlSettings & settings)
 	config.video.windowedHeight = settings.value("windowedHeight", config.video.windowedHeight).toInt();
 	config.video.fullscreenRefresh = settings.value("fullscreenRefresh", config.video.fullscreenRefresh).toInt();
 	config.video.multisampling = settings.value("multisampling", config.video.multisampling).toInt();
+	config.video.maxMultiSampling = settings.value("maxMultiSampling", config.video.maxMultiSampling).toInt();
 	config.video.fxaa= settings.value("fxaa", config.video.fxaa).toInt();
 	config.video.verticalSync = settings.value("verticalSync", config.video.verticalSync).toInt();
 	config.video.threadedVideo = settings.value("threadedVideo", config.video.threadedVideo).toInt();
@@ -210,6 +211,7 @@ void writeSettings(const char * _strIniFolder)
 		settings.setValue("windowedHeight", config.video.windowedHeight);
 		settings.setValue("fullscreenRefresh", config.video.fullscreenRefresh);
 		settings.setValue("multisampling", config.video.multisampling);
+		settings.setValue("maxMultiSampling", config.video.maxMultiSampling);
 		settings.setValue("fxaa", config.video.fxaa);
 		settings.setValue("verticalSync", config.video.verticalSync);
 		settings.setValue("threadedVideo", config.video.threadedVideo);
--- a/src/GLideNUI/ConfigDialog.cpp
+++ b/src/GLideNUI/ConfigDialog.cpp
@@ -167,9 +167,21 @@ void ConfigDialog::_init(bool reInit, bool blockCustomSettings)
 	ui->fullScreenResolutionComboBox->setCurrentIndex(fullscreenMode);
 	ui->fullScreenRefreshRateComboBox->setCurrentIndex(fullscreenRate);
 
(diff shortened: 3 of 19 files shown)


src/DisplayWindow.cpp
```diff
@@ -8,6 +8,8 @@
 #include "PluginAPI.h"
 #include "FrameBuffer.h"
 
+#include <Graphics/OpenGLContext/ThreadedOpenGl/opengl_Wrapper.h>
+
```
gonetz, line 12: Top level code must not use GL code directly, never.
This code does not know about graphics backend. It can be Vulcan or DX backends, top level code should work the same. All access to graphics must be done via graphics::Context from src\Graphics\Context.h.
You need to extend it with a new method. Check getMaxLineWidth() as an example. You may add one specialized method, like `getMaxMSAALevel()` or write a general one like
`int getIntegerParameter(Parameter _name);`
and add new `SpecialParameterClass` in Parameters.h with its proper implementation for GL_MAX_SAMPLES parameter. The second option needs more efforts and it is not evident yet that this general function is necessary.

src/GLideNUI/ConfigDialog.cpp
```diff
@@ -167,9 +167,21 @@ void ConfigDialog::_init(bool reInit, bool blockCustomSettings)
 	ui->fullScreenResolutionComboBox->setCurrentIndex(fullscreenMode);
 	ui->fullScreenRefreshRateComboBox->setCurrentIndex(fullscreenRate);
 
+	if (m_maxMSAA == 0 && config.video.maxMultiSampling == 0) {
```
gonetz, line 170: As I understand, we can't get valid config.video.maxMultiSampling until the first run of a game. It's not nice.

src/DisplayWindow.cpp
```diff
@@ -193,5 +200,5 @@ void DisplayWindow::readScreen2(void * _dest, int * _width, int * _height, int _
 
 u32 DisplayWindow::maxMSAALevel() const
 {
-	return _maxMSAALevel();
+	return m_maxMsaa;
```
gonetz, line 203: WindowsWGL::maxMSAALevel() becomes useless and should be removed.

gonetz (changes requested):

src/GLideNUI/ConfigDialog.cpp
Rosalie241: That's correct, sadly that's a limitation I can't easily fix (to get it working on i.e mupen64plus), so yes, to get the correct max level of msaa, you need to run a game at least once and save at least once to get it 'cached' into the config file, how would ypu propose I fix that limitation then?

src/DisplayWindow.cpp
Rosalie241: this should be fixed, I hope I put the `getMaxMSAALevel()` in the right place now

src/DisplayWindow.cpp
Rosalie241: fixed

src/GLideNUI/ConfigDialog.cpp
gonetz: I don't see anything better than my approach from WindowsWGL::maxMSAALevel()
that is create GL context, probe max msaa level, close the context.

src/DisplayWindow.cpp
gonetz: Thanks! I'll check it a bit later. I need to finish with #2405

src/GLideNUI/ConfigDialog.cpp
Rosalie241: That's also not ideal because it causes a delay and that doesn't work on mupen64plus (nor linux)

src/GLideNUI-wtl/Settings.cpp
```diff
@@ -412,6 +414,7 @@ void saveCustomRomSettings(const char * _strIniFolder, const char * _strRomName)
 	WriteCustomSetting(video, windowedHeight);
 	WriteCustomSetting(video, fullscreenRefresh);
 	WriteCustomSetting(video, multisampling);
+	WriteCustomSetting(video, maxMultiSampling);
```
gonetz, line 417: I don't think that this is necessary. maxMultiSampling can't be per-game setting.

src/GLideNUI/ConfigDialog.cpp
gonetz: I agree, it is not very nice too, but small delay is better than incorrect max msaa level.
Ok, let it be as this. I'll probably modify it for Windows GUI.

src/GLideNUI/Settings.cpp
```diff
@@ -411,6 +413,7 @@ void saveCustomRomSettings(const QString & _strIniFolder, const char * _strRomNa
 	WriteCustomSetting(video, windowedHeight);
 	WriteCustomSetting(video, fullscreenRefresh);
 	WriteCustomSetting(video, multisampling);
+	WriteCustomSetting(video, maxMultiSampling);
```
gonetz, line 416: I don't think that this is necessary. maxMultiSampling can't be per-game setting.

src/Graphics/OpenGLContext/opengl_ContextImpl.cpp
```diff
@@ -527,6 +527,11 @@ bool ContextImpl::isSupported(graphics::SpecialFeatures _feature) const
 	return false;
 }
 
+u32 ContextImpl::getMaxMSAALevel()
+{
+	return FunctionWrapper::maxMSAALevel();
```
gonetz, line 532: Why do we need special FunctionWrapper::maxMSAALevel() here?
Why not just call glGetIntegerv directly:
```
s32 ContextImpl::getMaxMSAALevel() const
{
	GLint maxMSAALevel;
	glGetIntegerv(GL_MAX_SAMPLES, &maxMSAALevel);
	return maxMSAALevel;
}

```

gonetz (changes requested):

gonetz:
Looks good now, thanks! Merged.

PR #2415, opened 2020-12-26
Add pixel coverage checkbox to GLideN64-wtl

oddMLan:


gonetz:
It looks like this for me:
gui-wtl

Issue #2418, opened 2020-12-26
Shader compilation from keys issue when hardware lighting is enabled


fzurita:
When changing the bilinear mode setting, shaders should be recompiled from the shader cache keys, but this is not happning. This is causing odd graphical glitches where mostly they appear the wrong color.

This may also be an issue with the HW lighting setting.

gonetz:
I can't reproduce it. ShaderStorage::_loadFromCombinerKeys is called every time I change shaders-related settings, including  bilinear mode.  I tried to change bilinear mode along with HW lighting - no problems. Do you have that problem for every game?

fzurita:
I was seeing this with battle zone.

fzurita:
Yeah, I just checked, it seems like only battlezone is affected by this. Strange.

fzurita:
And I think I see the problem, ```GBI.isHWLSupported()``` needs to be unique to each key and not global for all keys saved in the keys file.

gonetz:
I implemented your idea to add new bit for HW lighting state to CombinerKey: 955c2f3b17451
Please test https://github.com/gonetz/GLideN64/tree/add_hwlight_bit_to_key branch.

gonetz:
One note: I changed keys file format.
I decided that users can be upset to loose their keys file, so read from a file in the old format is still supported. It will be converted to the new format on next save. I now understand, that reading old key file will lead to the problem we are trying to solve.  However, the key file will be updated right after that, so on the next run or load from savestate the issue should gone.

fzurita:
I tested your branch and it worked as well as I what I had. I'll close my pull request.

gonetz:
Can this issue be closed?

fzurita:
Sure, but it's not merged to master yet.

gonetz:
Yes, I forgot about it.

PR #2422, opened 2020-12-27
Fix shader storage when hardware lighting is enabled

fzurita:
This is a fix for https://github.com/gonetz/GLideN64/issues/2418.

As currently implemented, shader storage does not seem to compile keys correctly if the microcode changes in the middle of a game when the previous keys were generated.

### DIFF
--- a/src/Combiner.h
+++ b/src/Combiner.h
@@ -91,7 +91,6 @@
 #define G_GCI_ZERO				20
 #define G_GCI_HALF				21
 #define G_GCI_HW_LIGHT			22
-#define G_GCI_HW_LIGHT			22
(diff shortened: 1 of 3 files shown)


fzurita:
Now, the bigger question, the shader lookup key are based on combiner inputs. Can you have two different microcodes in a game with the same combiner inputs while at the same time one supports hardware lighting and the other doesn't.

This could possibly explain the glitch with HW lighting in smash brothers.

gonetz:
> The only way I could think of how to fix that would be to make the HW lighting state be part of the shader key in addition to the combiner inputs.

It can be done. There is a spare bit in the shader key for that.

PR #2424, opened 2020-12-31
Fix recently introduced GLES 2.0 GLSL compilation errors

fzurita:
Looks like the coverage implementation introduced some GLES 2.0 GLSL compilation errors. I went ahead and fixed them.


src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -243,6 +243,7 @@ class VertexShaderHeader : public ShaderPart
 				"# define OUT out			\n"
 				"#else						\n"
 				"# define IN attribute		\n"
+				"# define highp mediump		\n"
```
gonetz, line 246: Why?

fzurita: GLSL compiler is giving me a warning that it doesn't support highp precision and it will use mediump instead. I added that to silence the warning.

See this: https://www.khronos.org/registry/OpenGL/specs/es/2.0/GLSL_ES_Specification_1.00.pdf

Section 4.5.4. Highp is not available by default unless supported by extension. To check if highp is available in GLSL version 100, we need to check if ```GL_FRAGMENT_PRECISION_HIGH``` is defined. Ideally, I could replace that line with this:
```
#ifndef GL_FRAGMENT_PRECISION_HIGH
#define highp mediump
#endif
```

Would you like me to do that instead?

gonetz: Yes, please. Otherwise it looks as a hack.

fzurita: Ok, done.

PR #2431, opened 2021-01-08
Compile shader from keys if shader storage binary is invalid

fzurita:
In one of my low end devices, probably due to some driver bug, the binary shaders being produced can't be loaded. I can see when this happens when the shader fails to link.

I had to take the shader error checks out of the GL_DEBUG macro since I can't check for link errors otherwise. I don't see any worse stuttering when a shader is compiling due to checking for the errors.

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_Utils.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_Utils.cpp
@@ -55,6 +55,13 @@ bool Utils::checkShaderCompileStatus(GLuint obj)
 bool Utils::checkProgramLinkStatus(GLuint obj)
 {
 #ifdef GL_DEBUG
+	checkProgramLinkStatusRequired(obj);
+#endif
+	return true;
+}
+
+bool Utils::checkProgramLinkStatusRequired(GLuint obj)
+{
 	GLint status;
 	glGetProgramiv(obj, GL_LINK_STATUS, &status);
 	if (status == GL_FALSE) {
@@ -64,7 +71,6 @@ bool Utils::checkProgramLinkStatus(GLuint obj)
 		LOG(LOG_ERROR, "shader_link error: %s", shader_log);
 		return false;
 	}
-#endif
 	return true;
 }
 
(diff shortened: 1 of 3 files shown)


fzurita:
So I put this in the play store build and people are complaining about worse shader stutter during gameplay. So this may not be a good idea after after the initial shader compilation.

So, it's probably better to only perform error checking when loading from keys or binary files. I'll see if I can limit it to only that.

fzurita:
Ok, now I only check shader link status and only at startup.

src/Graphics/OpenGLContext/GLSL/glsl_Utils.cpp
```diff
@@ -55,6 +55,13 @@ bool Utils::checkShaderCompileStatus(GLuint obj)
 bool Utils::checkProgramLinkStatus(GLuint obj)
 {
 #ifdef GL_DEBUG
+	checkProgramLinkStatusRequired(obj);
+#endif
+	return true;
+}
+
+bool Utils::checkProgramLinkStatusRequired(GLuint obj)
+{
```
gonetz, line 64: What is the purpose of that modification?

fzurita: Well, it adds a ```checkProgramLinkStatusRequired``` method. If that is used instead of ```checkProgramLinkStatus``` then it will check if a GLSL program linked successfully even if the  ```GL_DEBUG``` macro is disabled.

If you are referring to me calling ```checkProgramLinkStatusRequired``` within ```checkProgramLinkStatus``` then or was simply to keep common code.

gonetz: Yep, sorry, it was a stupid question.
I understand what the code does, but I dislike that solution with the new method. Btw, it has an error - there should be
`return checkProgramLinkStatusRequired(obj);`
I corrected your code a little bit - I added `_force` parameter to `checkProgramLinkStatu` - e129d66c22

fzurita: Ok, thanks.

Issue #2046, opened 2019-04-15
Pokemon Snap Red dot doesn't work Android


fzurita:
The Pokemon snap red dot doesn't seem to work in Android. I tried copying these settings and it still doesn't work:
frameBufferEmulation\copyAuxToRDRAM=1
frameBufferEmulation\copyToRDRAM=1

In addition, I enabled copy color buffer from RDRAM.

Is there some other setting I'm missing to get it to work?

gonetz:
Try to enable FBInfo.

fzurita:
It doesn't work in android since the ARM new dynarec doesn't support it.

gonetz:
project 64 does not support FBInfo, but mupen64plus for windows does. I'm not sure that FBInfo fixes all the issues with Pokemon Snap, and it could ad some new as well. Anyway, FBInfo is the best approach to support features such as Red Dot. Other methods are even less reliable.

CallistoNTG:
>frameBufferEmulation\copyToRDRAM=1
Shouldn't that be frameBufferEmulation\copyToRDRAM=2? I thought Snap required the software depth buffer.

fzurita:
@gonetz I meant the red dot works fine in project 64 without FB info.

fzurita:
@CallistoNTG I do have that enabled and the red dot still doesn't work. The configuration you quoted is for copying color buffer to RDRAM.

DonelBueno:
https://github.com/gonetz/GLideN64/issues/380

I think this is the reason for it not to work:

The red dot will appear with copyFromRDRAM=1, but this setting also breaks Oak's review. So we can have the red dot if we break the game.

It's a tradeoff between playability and a comestic feature

Jj0YzL5nvJ:
@DonelBueno, can you test using `copyFromRDRAM=1` and ThreadedVideo disabled?
See #1630

DonelBueno:
@Jj0YzL5nvJ The game runs apparently at native resolution with these settings. The red dot appears, but oak's checks don't work, exactly as described on #380.

standard-two-simplex:
I can get the red do effect to work by disabling the memsets introduced in 538674da5

fzurita:
Yeah, I don't think that hack I introduced is needed any more since we have the FB info API in mupen64plus. The only downside of that API is that it requires interpreter instead of dynarec.

standard-two-simplex:
[...]

PR #2436, opened 2021-01-17
Fix incorrect width used for sync color buffer reads with EGLImage

fzurita:
I found this while doing some unrelated testing.

It fixes: https://github.com/gonetz/GLideN64/issues/2046


rez-:
Just curious, why isnt the height coming from m_pTexture as well?

fzurita:
Because we don't need that much. The width has to come from the texture because that's how big the texture used for the FBO is. If you use a smaller value, you get out of alignment. If you use a smaller value for the height, you don't get out of alignment because the stride is still the same.

Issue #2427, opened 2021-01-03
Software clipping and near plane clipping


gonetz:
I'm working on software polygons clipping. It is in soft_clip branch.
The result is disappointing. It works, but it works exactly as the hardware one.
The main goal is to fix the issue with near plane clipping on GLES devices, on which I can't disable it.
If I remove usage of graphics::ClampMode::NoNearPlaneClipping I get this:
![GLideN64_THE_LEGEND_OF_ZELDA_003](https://user-images.githubusercontent.com/8983092/103481971-b18a3480-4e10-11eb-94ae-262d4b339e2e.jpg)
Glide64 somehow avoids this issue, but I don't understand how. I know why it worked on 3dfx hardware - glide3x worked with screen space coordinates and had no hardware clipping. The plugin skips near plane clipping. If some part of polygon is outside the screen because of it, the scissor cuts it. However, the glide3x wrapper uses OpenGL. It converts screen space vertex coordinates into homogeneous ones before sending them to GPU. OpenGL should clip them but everything looks fine:
![Glide64_THE_LEGEND_OF_ZELDA_01](https://user-images.githubusercontent.com/8983092/103482307-bbad3280-4e12-11eb-8110-9928a533e15c.jpg)
Any idea how it works?

standard-two-simplex:
> Any idea how it works?

I don't know about Glide64, but maybe you can try dividing z of all vertices by a big number (say 1024.0) at the end of the vertex shader and multiplying it back at the beggining of the fragment shader.

This way the near plane test will pass more easily. I think that if you divide all vertices by the same constant, attribute interpolation will still be correct.

Jj0YzL5nvJ:
> Any idea how it works?

[Bitmask?](https://en.wikipedia.org/wiki/Mask_(computing)#Image_masks)

gonetz:
> > Any idea how it works?
> 
> [Bitmask?

Please explain, how bitmask can help here?

> but maybe you can try dividing z of all vertices by a big number (say 1024.0) at the end of the vertex shader and multiplying it back at the beggining of the fragment shader.

Hmm. I removed usage of graphics::ClampMode::NoNearPlaneClipping in GBI.cpp and added 
	" gl_Position.z /= 10.0;	 \n"
to VertexShaderEnd. It fixed the issue with slingshot! I did the same in master and got the same result.

How does it work? Negative z coordinate remained negative after the division and thus the polygon still should be clipped.
[...]

Issue #2382, opened 2020-11-12
Zelda OOT issue with fragment based depth


fzurita:
With fragment based depth enabled I get this issue when zelda OOT at the very beginning of the game:
![the_legend_of_zelda-000](https://user-images.githubusercontent.com/11581687/98878484-1f3d6400-2451-11eb-811b-0475fecc7cd7.png)

Not sure when the problem was introduced. I tried to go back pretty far, but once CMake stops being used for android, it makes it really hard to compile.

This is in an Adreno GPU.

fzurita:
Ok, this will only happen with Adreno GPUs because it's the only GLES GPU that also supports GL_NV_shader_noperspective_interpolation. What we are doing with this extension is causing this problem.

gonetz:
As I see, GL_NV_shader_noperspective_interpolatio was introduced by Logan in 2e227cef531c1
It was April 2018.
@fzurita as I understand, it is hard for you to build the project from the sources that old. Can you revert that commit from the current master and check, is it really the source of the problem? I have a device with Adreno 420, but I can't reproduce that issue with it. Probably it does not support that extension.

fzurita:
@gonetz if I force the extension to disabled, the problem goes away. Although, it would be good to find out if the problem existed when it was first introduced.

If you are using the Mupen64plus AE or FZ app, you should be able to reproduce it using the GLideN64-Very-Accurate profile.

gonetz:
Yes, I use Mupen64plus FZ, ver 251. I can't reproduce that issue on my device with  GLideN64-Very-Accurate profile.



> Although, it would be good to find out if the problem existed when it was first introduced.

Yes. As I understand, it is enough to enable only "write fragment based depth" option to reproduce that error. N64 depth compare can be off. If so, I suppose these changes in glsl_CombinerProgramBuilder.cpp  can be responsible:

```
			ss << "# define IN in" << std::endl << "# define OUT out" << std::endl;
-			if (_glinfo.noPerspective)
-				ss << "noperspective OUT highp float vZCoord;" << std::endl << "uniform lowp int uClampMode;" << std::endl;
+			if (_glinfo.noPerspective) {
+				ss << "#extension GL_NV_shader_noperspective_interpolation : enable" << std::endl
+					<< "noperspective OUT highp float vZCoord;" << std::endl << "uniform lowp int uClampMode;" << std::endl;
+			}
			m_part = ss.str();
```

[...]

PR #2439, opened 2021-01-21
GLideN64 soft vertex clipping performance improvement for Adreno GPUs

fzurita:
As stated in https://github.com/gonetz/GLideN64/issues/2427, the software vertex clipping has some really bad performance penalties in Adreno GPUs. For example it drops Banjo-Tooie from around 50fps in the opening area to around 27fps. This change restores some of the performance but not all. It's now reaching around 40fps in the same area.

This change essentially restores some of the old code that was used when ```GL_NV_shader_noperspective_interpolation``` minus anything that discards pixels since that is now being done in software.

I tested the usual cases like the zelda slingshot and https://github.com/gonetz/GLideN64/issues/2382 still remains fixed.

I'm not sure why this doesn't have the same performance as the old method that used to discard pixels. Disabling fragment depth write completely brings back the old performance. So it's possible that the extra discards that were happening in the shader were preventing extra depth writes from happening.

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
@@ -255,7 +255,8 @@ public:
 			ss << "#version " << Utils::to_string(_glinfo.majorVersion) << Utils::to_string(_glinfo.minorVersion) << "0 es " << std::endl;
 			ss << "# define IN in" << std::endl << "# define OUT out" << std::endl;
 			if (_glinfo.noPerspective) {
-				ss << "#extension GL_NV_shader_noperspective_interpolation : enable" << std::endl;
+				ss << "#extension GL_NV_shader_noperspective_interpolation : enable" << std::endl
+				   << "noperspective OUT highp float vZCoord;" << std::endl;
 			}
 			m_part = ss.str();
 		}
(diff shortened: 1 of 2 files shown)


src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -1938,13 +1958,24 @@ class ShaderWriteDepth : public ShaderPart
...
 				} else {
 					if (_glinfo.isGLESX && config.generalEmulation.enableClipping != 0) {
-						m_part =
-							"highp float writeDepth()						        										\n"
-							"{																								\n"
-							"  if (uDepthSource != 0) return uPrimDepth;													\n"
-							"  return clamp(8.0 * (gl_FragCoord.z * 2.0 - 1.0) * uDepthScale.s + uDepthScale.t, 0.0, 1.0);	\n"
-							"}																								\n"
-						;
+						if (_glinfo.isGLESX && _glinfo.noPerspective) {
```
gonetz, line 1961: Just 
`if (_glinfo.noPerspective)`
`_glinfo.isGLESX` is checked above

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -1918,10 +1931,17 @@ class ShaderWriteDepth : public ShaderPart
 						"{																								\n"
 						;
 					if (_glinfo.isGLESX && config.generalEmulation.enableClipping != 0) {
-						m_part +=
-							"  highp float FragDepth = (uDepthSource != 0) ? uPrimDepth :								\n"
-							"      clamp(8.0 * (gl_FragCoord.z * 2.0 - 1.0) * uDepthScale.s + uDepthScale.t, 0.0, 1.0);	\n"
-							;
+						if (_glinfo.isGLESX && _glinfo.noPerspective) {
```
gonetz, line 1934: Just
if (_glinfo.noPerspective)
_glinfo.isGLESX is checked above

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
fzurita: Tep, thanks.

fzurita:
From reading, it seems like writing ```gl_fragDepth``` disables the early z test that allows a shader not to run if it's not needed.

Maybe there is a way to determine if writing gl_fragDepth will change the output, and if not, then don't write it.

fzurita:
Hmm, there is an extension we can potentially use:
https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_conservative_depth.txt

It would allow the early z to still run when writing gl_fragDeprh.

PR #2441, opened 2021-01-30
GLideNHQ: add wildcard support

Rosalie241:
This patch implements wildcard support.

The problem before this patch is that some games have textures which change `chksum` but have the same `palchksum` (based on in-game location or other factors, note that some games have the reverse problem and only differ in `palchksum`), this causes pain for texture pack creators who need to copy textures only to change the filename.

As an example, take Resident Evil 2 which has `165` different `chksum`s for the same texture while having the exact same `palchksum`.
image

This is not ideal of course so to resolve the issue this patch allows texture pack creators to replace either `chksum` or `palchksum` in the filename with `$` making everyone's life a bit easier!

image

I'm not sure if you like this implementation but it seems like the simplest way to add this functionality.

### DIFF
--- a/src/GLideNHQ/TxHiResCache.cpp
+++ b/src/GLideNHQ/TxHiResCache.cpp
@@ -237,6 +237,7 @@ TxHiResCache::LoadResult TxHiResCache::loadHiResTextures(const wchar_t * dir_pat
 		/* Rice hi-res textures: begin
 		 */
 		uint32 chksum = 0, fmt = 0, siz = 0, palchksum = 0;
+		bool hasWildcard = false;
 		char *pfname = nullptr, fname[MAX_PATH];
 		std::string ident;
 		FILE *fp = nullptr;
@@ -255,6 +256,7 @@ TxHiResCache::LoadResult TxHiResCache::loadHiResTextures(const wchar_t * dir_pat
 
 		/* read in Rice's file naming convention */
 #define CRCFMTSIZ_LEN 13
+#define CRCWILDCARD_LEN 15
 #define PALCRC_LEN 9
 		wcstombs(fname, foundfilename, MAX_PATH);
 		/* XXX case sensitivity fiasco!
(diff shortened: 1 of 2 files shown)


src/GLideNHQ/TxHiResCache.cpp
```diff
@@ -281,13 +283,39 @@ TxHiResCache::LoadResult TxHiResCache::loadHiResTextures(const wchar_t * dir_pat
...
 		if (pfname != fname) pfname = 0;
 		if (pfname) {
-			if (sscanf(pfname + ident.size(), "#%08X#%01X#%01X#%08X", &chksum, &fmt, &siz, &palchksum) == 4)
-				pfname += (ident.size() + CRCFMTSIZ_LEN + PALCRC_LEN);
-			else if (sscanf(pfname + ident.size(), "#%08X#%01X#%01X", &chksum, &fmt, &siz) == 3)
-				pfname += (ident.size() + CRCFMTSIZ_LEN);
-			else
+			uint32_t length = 0;
+
+			if (sscanf(pfname + ident.size(), "#%08X#%01X#%01X#%08X", &chksum, &fmt, &siz, &palchksum) == 4) {
```
gonetz, line 288: pfname + ident.size() is used in several places. It's better to use a local variable, like:
```
const char * strName = pfname + ident.size();
if (sscanf(strName, "#%08X#%01X#%01X#%08X", &chksum, &fmt, &siz, &palchksum) == 4) {
```

src/GLideNHQ/TxHiResCache.cpp
```diff
@@ -281,13 +283,39 @@ TxHiResCache::LoadResult TxHiResCache::loadHiResTextures(const wchar_t * dir_pat
...
+			} else if (sscanf(pfname + ident.size(), "#%08X#%01X#%01X", &chksum, &fmt, &siz) == 3) {
+				length = (ident.size() + CRCFMTSIZ_LEN);
+			}
+
+			/* sscanf seems to not care about things that come after the input
+			 * so to combat sscanf's behavior, check if there's '$' in the str
+			 * and re-try sscanf,
+			 * this is needed for wildcard support
+			 */
+			if (strchr(pfname + ident.size(), '$')) {
```
gonetz, line 299: Since we do this check anyway, it is better to put the code above in the else clause:
```
if (strchr(pfname + ident.size(), '$')) {
...
} else {
	if (sscanf(pfname + ident.size(), "#%08X#%01X#%01X#%08X", &chksum, &fmt, &siz, &palchksum) == 4) {
		length = (ident.size() + CRCFMTSIZ_LEN + PALCRC_LEN);
	} else if (sscanf(pfname + ident.size(), "#%08X#%01X#%01X", &chksum, &fmt, &siz) == 3) {
		length = (ident.size() + CRCFMTSIZ_LEN);
	}
}

```

src/GLideNHQ/TxHiResCache.cpp
```diff
@@ -310,8 +338,11 @@ TxHiResCache::LoadResult TxHiResCache::loadHiResTextures(const wchar_t * dir_pat
 		/* check if we already have it in hires texture cache */
 		if (!replace) {
 			uint64 chksum64 = (uint64)palchksum;
-			chksum64 <<= 32;
-			chksum64 |= (uint64)chksum;
+			if (chksum)
+			{
```
gonetz, line 342: if (chksum) {

src/GLideNHQ/TxHiResCache.cpp
```diff
@@ -734,8 +765,11 @@ TxHiResCache::LoadResult TxHiResCache::loadHiResTextures(const wchar_t * dir_pat
 		/* load it into hires texture cache. */
 	{
 		uint64 chksum64 = (uint64)palchksum;
-		chksum64 <<= 32;
-		chksum64 |= (uint64)chksum;
+		if (chksum)
+		{
```
gonetz, line 769: if (chksum) {

gonetz (changes requested):
The changes should be harmless for existing hi-res packs, so I have nothing against them. However I have a few small comments about the implementation.

SergeyLipskiy-Hexagon (approved):

gonetz (approved):

Nes-player4Life:
Some textures use a common hash however they are visually not the same causing improper overwrite.
image

Rosalie241:
Hmm? If you have a wildcard in one of the hashes and then you have a texture without wildcard (with a 'common' hash), the texture without wildcard has priority

Nes-player4Life:
image
the issue then becomes one texture is wildcarded with all others staying pesky.

Rosalie241:
I'm sorry but I don't understand what you mean, can you explain? Note that texture dumping hasn't changed by the way.

Nes-player4Life:
sorry, wildcard works however certain images share the hash that would be replaced with $ thus rendering errors once loaded to screen.

Rosalie241:
That shouldn't happen, can you open discord & check discord64? It's easier to discuss this that way

GhostlyDark:
As mentioned by Rosalie, normally named HD textures have priority over wildcarded ones. So you can try to wildcard a texture and cover any exceptions by using textures with their full name alongside it.

In your case however, the textures seem to be visually different, so you can't use a wildcard in that case.

PR #2444, opened 2021-02-03
Improve the draw elemenets base vertex check

fzurita:
No extension is required to use ```glDrawRangeElementsBaseVertex``` for GLES 3.2. In fact, my Adreno devices say that they don't support the extension and have GLES 3.2 and the call works just fine.

See https://www.khronos.org/registry/OpenGL-Refpages/es3/html/glDrawRangeElementsBaseVertex.xhtml

It's part of GL ES 3.2.

### DIFF
--- a/src/Graphics/OpenGLContext/opengl_GLInfo.cpp
+++ b/src/Graphics/OpenGLContext/opengl_GLInfo.cpp
@@ -105,8 +105,8 @@ void GLInfo::init() {
 	}
 
 	drawElementsBaseVertex = !isGLESX ||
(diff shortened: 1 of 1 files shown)


fzurita:
Also, I think that the reason that the function ```glDrawRangeElementsBaseVertex``` doesn't work in PowerVR is because we are supposed to be using ```glDrawRangeElementsBaseVertexEXT``` instead of ```glDrawRangeElementsBaseVertex``` when using the extension. I will have to prove that out though.

PR #2453, opened 2021-02-10
Small correction and new options

Jj0YzL5nvJ:
Prevent the script from being enslaved by becoming the parent process of `explorer.exe` if such isn't running when invoked. And new features.


gonetz:
I corrected commit message a bit:
e6c3797726

PR #2454, opened 2021-02-12
Optimize coverage computation algorithm

standard-two-simplex:
Fixes a small mistake with one of the biases and optimizes the per sample loop to perform all comparisons with one step function call.


fzurita:
The shader storage format version should probably be incremented, although, keep in mind that I just incremented it in another pull request...

standard-two-simplex:
> The shader storage format version should probably be incremented, although, keep in mind that I just incremented it in another pull request...

You're right. If I add it here it will confict with your PR, so maybe it's better to wait for now.

gonetz (approved):

Issue #2449, opened 2021-02-07
Cruis'n world invisible car


fzurita:
This only seems to happen with fragment based depth enabled.

This used to work ok recently, I'll try to find out which commit broke it.

![cruis'n_world-000](https://user-images.githubusercontent.com/11581687/107160145-770d3d80-6962-11eb-9edb-c23d56dd16d9.png)

fzurita:
Looks like it was broken by 402cf6f8b6e5a105ced42010965a1d72ff3f7f43

That's the software vertex clipping.

fzurita:
It seems to only be broken in Adreno, it's not broken in Mali.

fzurita:
@gonetz I think the software vertex clipping is clipping something that it shouldn't be clipping.

fzurita:
@gonetz changing this line fixes it and I'm not seeing any negative effects anywhere:

https://github.com/gonetz/GLideN64/blob/master/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp#L1966

Changing it to:
```								"  return clamp((vZCoord - uPolygonOffset) * uDepthScale.s + uDepthScale.t - 0.5, 0.0, 1.0);		\n"```

See the additional ```- 0.5```

The thing is that I have no idea why it fixes and if it's the right thing to do.

All I can think of is that ```gl_Position.z``` is off by 0.5 for some unknown reason.

**Comment by gonetz** [comment]
> I think the software vertex clipping is clipping something that it shouldn't be clipping.

Software clipping works the same on all platforms. Windows build has not that issue. I suppose, the problem is in GLES-specific shaders.

**Comment by gonetz** [comment]
I see one logical mistake.
I implemented software clipping in commit 402cf6f8b6e5a1.
I replaced near-plane clipping code 
```
-		} else if (config.generalEmulation.enableFragmentDepthWrite != 0 && _glinfo.noPerspective) {
+		} else if (config.generalEmulation.enableClipping != 0) {
+			// Move the near plane towards the camera.
+			// It helps to avoid issues with near-plane clipping in games, which do not use it.
+			// Z must be scaled back in fragment shader.
 				m_part =
-					"  vZCoord = gl_Position.z / gl_Position.w;	\n"
-					"  if (uClampMode > 0)	\n"
-					"    gl_Position.z = 0.0;	\n"
-					;
+					"  gl_Position.z /= 8.0;	\n"
+				;
 		}
```
		
[...]

PR #2456, opened 2021-02-12
Make software vertex clipping config be part of shader storage config

fzurita:
I discovered while trying to debug https://github.com/gonetz/GLideN64/issues/2449 software vertex clipping is not part of the shader storage saved configuration. This adds it.


gonetz:
I want to make software vertex clipping always enabled.
I made it optional to simplify detection of regressions caused by it.
#2449 is probably the last such regression. I'd prefer to fix it instead of adding clipping to shader storage config.

PR #2458, opened 2021-02-13
Fix some issues in previous commit

fzurita:


src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilder.cpp
```diff
@@ -489,19 +490,17 @@ class VertexShaderEnd : public ShaderPart
...
 			m_part =
 				"  vZCoord = gl_Position.z / gl_Position.w;	\n"
 				"  if (uClampMode > 0)	\n"
 				"    gl_Position.z = 0.0;	\n"
 				;
+		} else if (config.generalEmulation.enableClipping != 0) {
+			// Move the near plane towards the camera.
+			// It helps to avoid issues with near-plane clipping in games, which do not use it.
+			// Z must be scaled back in fragment shader.
+			m_part = "  gl_Position.z /= 8.0;	\n";
```
gonetz, line 503: My idea was to use "Move the near plane towards the camera" method if clipping is enabled.
It should work better than the old method 
"  vZCoord = gl_Position.z / gl_Position.w;	\n"
				"  if (uClampMode > 0)	\n"
				"    gl_Position.z = 0.0;	\n"
				
I suppose it did not work for Adreno just because I forgot to add uClampMode support.
Could you try to add only uClampMode support and left the other code untouched?

gonetz:
Two commits are squashed and merged: 442d4ed894f

fzurita:
Great, thanks!

Issue #2459, opened 2021-02-13
N64 Depth compare broken in Adreno


fzurita:
I'm Extreme-G only the HUD is rendered work Adreno. See screenshot:
![extremeg-000](https://user-images.githubusercontent.com/11581687/107861907-f6918580-6e16-11eb-81a0-9f3debd60889.png)

Not sure at what point it broke.

Edit 1: Disabling software vertex clipping does not fix it.

fzurita:
Ok, it looks like N64 depth compare is broken with Adreno currently.

I went back pretty far back and it was not broken by any recent changes. @loganmc10 Any ideas since you implemented that?

fzurita:
Ok, I went back further. I bisected to this commit: 2600a6bc90f7f367f3a9f2ee3ee86b1efef90af5 That's when N64 depth compare first became broken on Adreno.

fzurita:
@gonetz It seems like the "Fast N64 depth compare" is broken. Is this supported by anything but Adreno?

loganmc10:
I believe it works well on most Intel and Nvidia GPUs, I don't know about mobile devices though

fzurita:
Ok, I found the original pull request and it seems like you got it working pretty well with Adreno here:

https://github.com/gonetz/GLideN64/pull/1792

I'm going to try to see what the difference is to the current implementation in a bit...

loganmc10:
It could be something related to a driver update/newer Android version since that was first written. As I recall it really depended on how those particular GL extensions were implemented by the vendors.

fzurita:
Ok, I think I figured it out, at least partially. Everyhing works right in Adreno if I force imageTextures to off.

fzurita:
Based on what I'm seeing so far, I see no reason for Mali GPUs to not support the fast N64 depth compare.

They have a similar extension to framebuffer fetch that is used for Adreno.

Panderner:
International superstar soccer also affected this issue
![Screenshot_2021-02-14-21-12-33-918_org mupen64plusae v3 fzurita](https://user-images.githubusercontent.com/37503397/107877787-6f710a00-6f09-11eb-8090-2bb372173c22.jpg)

fzurita:
Yeah, I have a fix here for Adreno: https://github.com/mupen64plus-ae/mupen64plus-ae/tree/fix_n64_depthcompare


I'm trying to come with a way to get it working on Mali too.

fzurita:
[...]

PR #2460, opened 2021-02-14
Fix N64 Depth compare with Adreno

fzurita:
Mainly check for FB fetch extension to be present before image textures. This should be tried in PC to make sure I didn't break anything there.

This is a fix for https://github.com/gonetz/GLideN64/issues/2459


gonetz:
> This should be tried in PC to make sure I didn't break anything there.

Yes, I checked it. N64 depth compare works.

PR #2470, opened 2021-03-20
Use screen coordinates

standard-two-simplex:
Primitives are rendered using their screen coordinates in the N64. Clean version of  #2334.


gonetz (approved):

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp
```diff
@@ -216,6 +216,50 @@ class UScreenSpaceTriangleInfo : public UniformGroup
...
 
+class URasterInfo : public UniformGroup {
+public:
+	URasterInfo(GLuint _program) {
+		LocateUniform(uVertexOffset);
+		LocateUniform(uTexCoordOffset);
+	}
+
+	void update(bool _force) override {
+		const bool isNativeRes = config.frameBufferEmulation.nativeResFactor == 1 && config.video.multisampling == 0;
```
gonetz, line 231: Why multisampling should be off? It is serious limitation for native-res mode.

standard-two-simplex: The idea of the vertex shift is to move the vertices so that the only sample of the fragments falls in the same relative position as in the N64. In multisampling mode, there are several samples across the fragment, so moving the vertices doesn't help. I am not sure if the texture coordinate shift should be applied in this case either, I guess that the last time I tested it gave better results.

If you remember, there was this issue with it in Excitebike and it is probably not the only issue.
excitebike64-000

standard-two-simplex:
If the last commit is fine I will squash it with 653b6d9

gonetz:
> If the last commit is fine I will squash it with 653b6d9

It's fine.

gonetz:
Merged.
I tested it as much as I could.
Various glitches can be overcome by toggling "enableTexCoordBounds" option.
It can be done with custom settings.
There are games, where toggling "enableTexCoordBounds" option fixes one issue but causes another one, for example Paper Mario. But Paper Mario was not perfect before that feature.

PR #2476, opened 2021-04-05
Per pixel shift offset

standard-two-simplex:


standard-two-simplex:
Closing this. Some framebuffer effects are missplaced.
If I manage to improve the situation in the future, I'll reopen it.

Issue #1617, opened 2017-10-02
1x native resolution factor breaks pause menu in CBFD (Adreno only)


fzurita:
Using 1x native tradition factor breaks the pause menu in CBFD. Example:

![conker_bfd-002](https://user-images.githubusercontent.com/11581687/31084125-9704af7c-a761-11e7-8d22-802e94519777.png)

My guess is that it has to do with the resolution the game runs at since for example, Mario kart effects work correctly at 1x (monitor in first course).

This pause menu works correctly at 2x native resolution.

Frank-74:
With latest deaf612 build, I'm not seeing any problems with pause menu on nativeResFactor=1.
First time I ran, I had inverse green/blue colours on pause menu.  I restored defaults, set native resolution and restarted.  Now it's fine.

fzurita:
Ok, must be Android only.

loganmc10:
Tested with mupen64plus+Windows, no issues (tested during the opening cutscene and in-game)

fzurita:
I did further testing and this only affect Adreno.... not sure why though.

fzurita:
Fixed by #2480

PR #2480, opened 2021-04-17
Stop using predefined texture sizes introduced for Android

fzurita:
For Android we use EGL image for async copies and for sync copies and devices that don't support EGL image, the fixed texture sizes don't seem to improve performance.

Also, this fixes the red dot not working on Adreno devices.

### DIFF
--- a/src/BufferCopy/ColorBufferToRDRAM.cpp
+++ b/src/BufferCopy/ColorBufferToRDRAM.cpp
@@ -12,17 +12,6 @@
 #include "Log.h"
 #include "MemoryStatus.h"
 
-/*
-#include "ColorBufferToRDRAM_GL.h"
-#include "ColorBufferToRDRAM_BufferStorageExt.h"
-#elif defined(OS_ANDROID) && defined (GLES2)
-#include "ColorBufferToRDRAM_GL.h"
-#include "ColorBufferToRDRAM_GLES.h"
-#else
-#include "ColorBufferToRDRAMStub.h"
-#endif
-*/
-
 #include <Graphics/Context.h>
 #include <Graphics/Parameters.h>
 #include <Graphics/ColorBufferReader.h>
@@ -165,12 +154,12 @@ bool ColorBufferToRDRAM::_prepareCopy(u32& _startAddress)
 		return false;
 
 	if(m_pTexture == nullptr ||
-		m_pTexture->width != _getRealWidth(pBuffer->m_width) ||
-		m_pTexture->height != VI_GetMaxBufferHeight(_getRealWidth(pBuffer->m_width)))
+		m_pTexture->width != pBuffer->m_width ||
+		m_pTexture->height != VI_GetMaxBufferHeight(pBuffer->m_width))
 	{
 		_destroyFBTexure();
 
-		m_lastBufferWidth = _getRealWidth(pBuffer->m_width);
+		m_lastBufferWidth = pBuffer->m_width;
 		_initFBTexture();
 	}
 
@@ -339,18 +328,6 @@ void ColorBufferToRDRAM::_copy(u32 _startAddress, u32 _endAddress, bool _sync)
 	m_bufferReader->cleanUp();
 
 	gDP.changed |= CHANGED_SCISSOR;
-}
-
-u32 ColorBufferToRDRAM::_getRealWidth(u32 _viWidth)
-{
-	u32 index = 0;
-	const u32 maxIndex = static_cast<u32>(m_allowedRealWidths.size()) - 1;
-	while (index < maxIndex && _viWidth > m_allowedRealWidths[index])
-	{
-		++index;
-	}
-
-	return m_allowedRealWidths[index];
 }
 
 void ColorBufferToRDRAM::copyToRDRAM(u32 _address, bool _sync)
--- a/src/BufferCopy/ColorBufferToRDRAM.cpp
+++ b/src/BufferCopy/ColorBufferToRDRAM.cpp
@@ -87,47 +87,43 @@ bool ColorBufferToRDRAM::_prepareCopy(u32& _startAddress)
 		readBuffer = m_pCurFrameBuffer->m_FBO;
 	}
 
-	if (m_pCurFrameBuffer->m_scale != 1.0f) {
-		u32 x0 = 0;
-		u32 width;
-		if (config.frameBufferEmulation.nativeResFactor == 0) {
-			const u32 screenWidth = wnd.getWidth();
-			width = screenWidth;
-			if (wnd.isAdjustScreen()) {
-				width = static_cast<u32>(screenWidth*wnd.getAdjustScale());
-				x0 = (screenWidth - width) / 2;
-			}
-		} else {
-			width = m_pCurFrameBuffer->m_pTexture->width;
+	u32 x0 = 0;
+	u32 width;
+	if (config.frameBufferEmulation.nativeResFactor == 0) {
+		const u32 screenWidth = wnd.getWidth();
+		width = screenWidth;
+		if (wnd.isAdjustScreen()) {
+			width = static_cast<u32>(screenWidth*wnd.getAdjustScale());
+			x0 = (screenWidth - width) / 2;
 		}
-		u32 height = (u32)(bufferHeight * m_pCurFrameBuffer->m_scale);
-
-		CachedTexture * pInputTexture = m_pCurFrameBuffer->m_pTexture;
-		GraphicsDrawer::BlitOrCopyRectParams blitParams;
-		blitParams.srcX0 = x0;
-		blitParams.srcY0 = 0;
-		blitParams.srcX1 = x0 + width;
-		blitParams.srcY1 = height;
-		blitParams.srcWidth = pInputTexture->width;
-		blitParams.srcHeight = pInputTexture->height;
-		blitParams.dstX0 = 0;
-		blitParams.dstY0 = 0;
-		blitParams.dstX1 = m_pCurFrameBuffer->m_width;
-		blitParams.dstY1 = bufferHeight;
-		blitParams.dstWidth = colorBufferTexture->width;
-		blitParams.dstHeight = colorBufferTexture->height;
-		blitParams.filter = textureParameters::FILTER_LINEAR;
-		blitParams.tex[0] = pInputTexture;
-		blitParams.combiner = CombinerInfo::get().getTexrectDownscaleCopyProgram();
-		blitParams.readBuffer = readBuffer;
-		blitParams.drawBuffer = pBuffer->getColorFbFbo();
-		blitParams.mask = blitMask::COLOR_BUFFER;
-		wnd.getDrawer().blitOrCopyTexturedRect(blitParams);
-
-		gfxContext.bindFramebuffer(bufferTarget::READ_FRAMEBUFFER, pBuffer->getColorFbFbo());
 	} else {
-		gfxContext.bindFramebuffer(bufferTarget::READ_FRAMEBUFFER, readBuffer);
+		width = m_pCurFrameBuffer->m_pTexture->width;
 	}
+	u32 height = (u32)(bufferHeight * m_pCurFrameBuffer->m_scale);
+
+	CachedTexture * pInputTexture = m_pCurFrameBuffer->m_pTexture;
+	GraphicsDrawer::BlitOrCopyRectParams blitParams;
+	blitParams.srcX0 = x0;
+	blitParams.srcY0 = 0;
+	blitParams.srcX1 = x0 + width;
+	blitParams.srcY1 = height;
+	blitParams.srcWidth = pInputTexture->width;
+	blitParams.srcHeight = pInputTexture->height;
+	blitParams.dstX0 = 0;
+	blitParams.dstY0 = 0;
+	blitParams.dstX1 = m_pCurFrameBuffer->m_width;
+	blitParams.dstY1 = bufferHeight;
+	blitParams.dstWidth = colorBufferTexture->width;
+	blitParams.dstHeight = colorBufferTexture->height;
+	blitParams.filter = m_pCurFrameBuffer->m_scale == 1.0f ? textureParameters::FILTER_NEAREST : textureParameters::FILTER_LINEAR;
+	blitParams.tex[0] = pInputTexture;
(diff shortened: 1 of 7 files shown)


fzurita:
For some reason, this breaks the CBFD pause menu but only when "copy auxiliary buffers to RDRAM" and "copy color to RDRARM (async)" is enabled. I have threading disabled just in case it was that.

The CBFD pause menu works fine with either disabled. I don't quiet understand why. @gonetz Can you check if this breaks the CBFD pause menu on PC? It theoretically shouldn't break anything since this is how it was before.

fzurita:
Ok, I see the problem, and it seems like it has been working by luck all this time...

https://github.com/gonetz/GLideN64/blob/master/src/BufferCopy/ColorBufferToRDRAM.cpp#L167-L169

The ```getRealWidth``` call I removed is returning the same size for the auxiliary and color buffer textures. Once I removed that call and async color buffer reads are enabled, the color buffer texture is being destroyed before we have a chance to read the texture asynchronously.

fzurita:
I implemented such a solution, @gonetz let me know what you think.

fzurita:
And I see a weird issue with CBFD. With auxiliary frame buffer copies enabled, Conker's shadow seems to "float" away from when he stops moving and reappears when movement starts again.

fzurita:
This fixes this https://github.com/gonetz/GLideN64/issues/1617

Also, only with sync color buffer copies. With 1x resolution factor and async copies, nothing seems to get copied to RDRAM. Trying to investigate why...

fzurita:
Ok, I see the issue, that blit that doesn't happen at 1x native resolution seems to be necessary for EGL image to work correctly, is there a specific issue that check is there? I removed it in that last commit.

src/FrameBuffer.cpp
```diff
@@ -529,6 +534,112 @@ CachedTexture * FrameBuffer::getTextureBG()
...
+	m_pColorBufferTexture->size = 2;
+	m_pColorBufferTexture->clampS = 1;
+	m_pColorBufferTexture->clampT = 1;
+	m_pColorBufferTexture->frameBufferTexture = CachedTexture::fbOneSample;
+	m_pColorBufferTexture->maskS = 0;
+	m_pColorBufferTexture->maskT = 0;
+	m_pColorBufferTexture->mirrorS = 0;
+	m_pColorBufferTexture->mirrorT = 0;
+	//The actual VI width is not used for texture width because most texture widths
+	//cause slowdowns in the glReadPixels call, at least on Android
```
gonetz, line 582: This comment looks outdated.

src/BufferCopy/ColorBufferToRDRAM.cpp
```diff
@@ -87,47 +87,43 @@ bool ColorBufferToRDRAM::_prepareCopy(u32& _startAddress)
...
+	blitParams.srcY1 = height;
+	blitParams.srcWidth = pInputTexture->width;
+	blitParams.srcHeight = pInputTexture->height;
+	blitParams.dstX0 = 0;
+	blitParams.dstY0 = 0;
+	blitParams.dstX1 = m_pCurFrameBuffer->m_width;
+	blitParams.dstY1 = bufferHeight;
+	blitParams.dstWidth = colorBufferTexture->width;
+	blitParams.dstHeight = colorBufferTexture->height;
+	blitParams.filter = textureParameters::FILTER_LINEAR;
```
gonetz, line 118: May be it is better to set FILTER_NEAREST for "m_pCurFrameBuffer->m_scale == 1.0f" case to avoid blurring.

gonetz (approved):

gonetz:
> is there a specific issue that check is there?

No, it was an optimization: don't blit if the buffer already of necessary size.

gonetz:
The changes are more global that I could expect from that local Android issue.
I see no better solutions for the problems you found. Approved.

gonetz:
> And I see a weird issue with CBFD. With auxiliary frame buffer copies enabled, Conker's shadow seems to "float" away from him when he stops moving and reappears when movement starts again.

Conker's shadow does not work at all with auxiliary frame buffer copies enabled on PC.

fzurita:
Ok, I can make those fixes. Have you tried to check the CBFD shadow after these changes with copy auxiliary buffers to RDRAM enabled?

gonetz:
> Have you tried to check the CBFD shadow after these changes with copy auxiliary buffers to RDRAM enabled?

Yes. I thought it does not work, but actually the problem is the same as on Android: "Conker's shadow seems to "float" away from him when he stops moving and reappears when movement starts again."

fzurita:
For the shadow, it seems like a texture offset problem, the texture is refreshed and reused when it doesn't change. But it's refreshed at a slightly wrong offset causing it to shift over time.

gonetz:
> For the shadow, it seems like a texture offset problem, the texture is refreshed and reused when it doesn't change. But it's refreshed at a slightly wrong offset causing it to shift over time.

Could you point me, where it is?

fzurita:
I mean, that's my guess based on what I'm visually seeing, I couldn't find a place in the code where this is being done wrong.

fzurita:
@gonetz I made the changes that you asked.

gonetz:
Thanks! Merged.

gonetz:
PC build crashes on Zelda OOT pause screen when "copy aux buffer" enabled. 
Broken in commit 43a5d16905c0

fzurita:
It happens in Android too.

fzurita:
FYI backtrace:
```
Build fingerprint: 'samsung/p3quew/p3q:11/RP1A.200720.012/G998U1UEU2AUC8:user/release-keys'
#00 0x0000000000180458 libmupen64plus-video-GLideN64.so (FrameBuffer::readPixels(int, int, unsigned int, unsigned int, unsigned int, bool)+8) (BuildId: 3d89e27f3f30013f0784499663abaa64d05fdd4a)
                                   FrameBuffer::readPixels(int, int, unsigned int, unsigned int, unsigned int, bool)
                                   /home/fzurita/git_projects/mupen64plus/mupen64plus-ae/mupen64plus-video-gliden64/upstream/src/FrameBuffer.cpp:557:25
#01 0x00000000001a1414 libmupen64plus-video-GLideN64.so (ColorBufferToRDRAM::_copy(unsigned int, unsigned int, bool)+204) (BuildId: 3d89e27f3f30013f0784499663abaa64d05fdd4a)
                                   ColorBufferToRDRAM::_copy(unsigned int, unsigned int, bool)
                                   /home/fzurita/git_projects/mupen64plus/mupen64plus-ae/mupen64plus-video-gliden64/upstream/src/BufferCopy/ColorBufferToRDRAM.cpp:208:41
#02 0x00000000001a1ae4 libmupen64plus-video-GLideN64.so (ColorBufferToRDRAM::copyToRDRAM(unsigned int, bool)+160) (BuildId: 3d89e27f3f30013f0784499663abaa64d05fdd4a)
                                   ColorBufferToRDRAM::copyToRDRAM(unsigned int, bool)
                                   /home/fzurita/git_projects/mupen64plus/mupen64plus-ae/mupen64plus-video-gliden64/upstream/src/BufferCopy/ColorBufferToRDRAM.cpp:247:2
#03 0x0000000000180d44 libmupen64plus-video-GLideN64.so (FrameBufferList::saveBuffer(unsigned int, unsigned short, unsigned short, unsigned short, bool)+208) (BuildId: 3d89e27f3f30013f0784499663abaa64d05fdd4a)
                                   FrameBuffer_CopyToRDRAM(unsigned int, bool)
                                   /home/fzurita/git_projects/mupen64plus/mupen64plus-ae/mupen64plus-video-gliden64/upstream/src/FrameBuffer.cpp:1754:28
                                   FrameBufferList::saveBuffer(unsigned int, unsigned short, unsigned short, unsigned short, bool)
                                   /home/fzurita/git_projects/mupen64plus/mupen64plus-ae/mupen64plus-video-gliden64/upstream/src/FrameBuffer.cpp:794:4
#04 0x000000000019594c libmupen64plus-video-GLideN64.so (RSP_ProcessDList()+672) (BuildId: 3d89e27f3f30013f0784499663abaa64d05fdd4a)
                                   _ProcessDList()
                                   /home/fzurita/git_projects/mupen64plus/mupen64plus-ae/mupen64plus-video-gliden64/upstream/src/RSP.cpp:59:3
                                   RSP_ProcessDList()
                                   /home/fzurita/git_projects/mupen64plus/mupen64plus-ae/mupen64plus-video-gliden64/upstream/src/RSP.cpp:182:3
#05 0x000000000000a284 libmupen64plus-rsp-hle.so (BuildId: 9e77e4d37a06dad0cd5420045340df9da8856dc7)
#06 0x0000000000010dfc libmupen64plus-rsp-hle.so (DoRspCycles+132) (BuildId: 9e77e4d37a06dad0cd5420045340df9da8856dc7)
#07 0x00000000000a2620 libmupen64plus-core.so (BuildId: 1d42afdaf9d94dec27e3d5878e77571745693733)
#08 0x0000000000068e30 libmupen64plus-core.so (BuildId: 1d42afdaf9d94dec27e3d5878e77571745693733)
#09 0x000000000009d508 libmupen64plus-core.so (BuildId: 1d42afdaf9d94dec27e3d5878e77571745693733)
#10 0x000000000002f738 [anon:.bss]

```

gonetz:
On my PC it crashes at line 55 of
src\Graphics\OpenGLContext\opengl_ColorBufferReaderWithBufferStorage.cpp
`m_bindBuffer->bind(Parameter(GL_PIXEL_PACK_BUFFER), ObjectHandle(m_PBO[m_curIndex]));`

fzurita:
Interesting, it's probably the same root cause.

gonetz:
Yes. Call stack is the same.

fzurita:
Yikes! We are trying to read from a frame buffer that has already been deleted. I can't see how this wasn't a problem before.

gonetz:
> We are trying to read from a N64 frame buffer that has already been deleted.

Is it? 
```
		if (m_pCurrent->isAuxiliary()) {
			FrameBuffer_CopyToRDRAM(m_pCurrent->m_startAddress, true);
			removeBuffer(m_pCurrent->m_startAddress);
		}

```
The buffer is alive when CopyToRDRAM is called.

fzurita:
```removeBuffer``` is deleting a frame buffer and then we are trying to read from it after it has been deleted.

**Comment by gonetz** [comment]
> On my PC it crashes at line 55 of
src\Graphics\OpenGLContext\opengl_ColorBufferReaderWithBufferStorage.cpp
m_bindBuffer->bind(Parameter(GL_PIXEL_PACK_BUFFER), ObjectHandle(m_PBO[m_curIndex]));

Actually, it crashes here because I made some changes for debug purposes.
Without these changes the copy works ok, but then emulator crashes. Most likely the copy corrupts RDRAM somehow.

**Comment by fzurita** [comment]
It seems like the game is creating two auxiliary frame buffers. We read from the first then delete it. When we are trying to read from the second, we somehow end up reading from the first that was already deleted.

**Comment by fzurita** [comment]
This case is happening in _prepareCopy:
```
	if (m_frameCount == curFrame && pBuffer == m_pCurFrameBuffer && m_startAddress != _startAddress) {
		return true;
	}
```

When that happens, the current frame buffer never gets updated because we exit the function too early.

**Comment by gonetz** [comment]
Good catch!
Yes, no crash with these lines removed.
It is an optimization.

**Comment by fzurita** [comment]
I'm not quiet understanding why that code is wrong though. ```pBuffer == m_pCurFrameBuffer``` should mean that we don't need to updated the pointer.

**Comment by gonetz** [comment]
The buffer could be removed and new one created at the same address.

**Comment by gonetz** [comment]
I checked:
```
	if (m_frameCount == curFrame && pBuffer == m_pCurFrameBuffer && m_startAddress != _startAddress)
	{
		pBuffer->getColorFbTexture();
		return true;
	}

```
also fixes the crash.

**Comment by fzurita** [comment]
Yeah, that makes sense. We need to initialize that texture.

**Comment by fzurita** [comment]
I think the fix is to remove the optimization. I believe you would want the blit to still happen in this case.

**Comment by gonetz** [comment]
It fixes the crash only for my hacked version :(
43a5d16 still crashes with it.

**Comment by gonetz** [comment]
Yes, the only working fix is removal of that optimization at all.

**Comment by fzurita** [comment]
Do you wish me to submit a new pull request or can you fix it directly in master?

**Comment by gonetz** [comment]
Removal of these lines fixes the pause screen only for 43a5d16905
The next your commit, 1d9061078, does not crash without these lines, but the pause screen is wrong - Link is missing.

**Comment by fzurita** [comment]
It behaves like that before as well. I checked before my changes and link appears incorrectly when copy auxiliary buffers to RDRAM is enabled.

**Comment by gonetz** [comment]
No, I mean that Link is missing with "aux buffer copy" enabled.
He appears with 43a5d16 with this option enabled and with the optimization removed.

**Comment by gonetz** [comment]
I use non-native resolution.
I checked commit 7a6289b27, which is right before your fixes.
Zelda pause screen crashes with "aux buffer copy" enabled, but works ok with the optimization removed.
So, the problem existed before your fixes.

With commit 1d90610782 removal of the optimization does not help - Link is missing.
May be it is because aux buffers also have scale 1.0 when "aux buffer copy" is enabled.

**Comment by fzurita** [comment]
Hmm, copy aux buffers to RDRAM on at 1x native resolution:
the_legend_of_zelda-000 (3)


Copy aux buffers to RDRAM on and native resolution off:
the_legend_of_zelda-000 (2)

**Comment by gonetz** [comment]
Yes, the same on PC

**Comment by gonetz** [comment]
Fix

```
bool ColorBufferToRDRAM::_prepareCopy(u32& _startAddress)
{
...
	if (config.frameBufferEmulation.nativeResFactor == 0 && m_pCurFrameBuffer->m_scale != 1.0f) {
```

fzurita:
I see, these seem like pre-existing issues.

gonetz:
Yes.

fzurita:
Got it.

gonetz:
I'll fix it in master, no need to create new PR.
Thanks for finding the source of the problem!

fzurita:
No problem!

gonetz:
Done: e85e0f4a

fzurita:
For some reason I'm back to crashing zelda subscreen with auxiliary frame buffer copies enabled and I have no idea why...

gonetz:
Is it fixed with your #2493?
If not, please open new issue.

fzurita:
Yeah, that fixed it for me.

Issue #2482, opened 2021-04-22
SM64 regression with GLideN64 rev.380f116e - Garbage lines around HUD textures


IronYoshi64:
Hey. When using the current latest master build of GLideN64, which right now is Rev. 380f116e, there is some pretty noticeable garbage lines around the Hud elements like lives, stars, and some numbers. There is also a thin weird line at the very bottom of the screen as well.

Here is an image of what it looks like, marked with red circles:
![GLideN64_SUPER_MARIO_64_000](https://user-images.githubusercontent.com/57850880/115733807-d4ecd700-a34e-11eb-8e81-d95be33046aa.png)

standard-two-simplex:
What rrendering resolution are you using? Multiple of native or same as output? If the latter, whats the outpit resolution?

IronYoshi64:
> What rrendering resolution are you using? Multiple of native or same as output? If the latter, whats the outpit resolution?

For windowed resolution, I use 1152x864. For Full screen resolution, I use 1920x1080 (16:9) 60 Hz.
Internal resolution is set to Same as output resolution. Happens on both QT and WTL versions.

oddMLan:
Same as output resolution should not be default... No other emulator makes it the default... However I got an alternative suggestion #2121

IronYoshi64:
> Same as output resolution should not be default... No other emulator makes it the default... However I got an alternative suggestion #2121

Same as output resolution has always been the default for as long as I can remember when using GLideN64. My sizes I listed are defiantly not default though.

GhostlyDark:
Same thing with the Zelda Debug text. I assume this is a result of the screen coordinate changes.

texcoords bounds off:
![image](https://user-images.githubusercontent.com/31742919/115976135-2e315200-a56b-11eb-9a25-91d3a71e93e5.png)

texcoords bounds on:
![image](https://user-images.githubusercontent.com/31742919/115976183-9c761480-a56b-11eb-8686-dc315e66c18a.png)

standard-two-simplex:
> For windowed resolution, I use 1152x864. For Full screen resolution, I use 1920x1080 (16:9) 60 Hz.

The behaviour of the games is unpredictable in such resolutions. The best I can do is to disable the heuristic if such rendering resolutions are used. Check if 0293323301fc7b5ac43fa8edbf679215014efad2 is satisfactory.

> Same thing with the Zelda Debug text.

How do you access the debug menu?

GhostlyDark:
[...]

PR #2486, opened 2021-04-27
Fix regressions introduced by recent changes

standard-two-simplex:
Aims to fix regressions introduced by recent changes.


gonetz:
#2482 is fixed

PR #2493, opened 2021-05-01
Fix other android issues

fzurita:
I discovered a bunch of other issues while doing some testing and debugging issues reported by users.

* First issue, we are not destroying an EGL image once we are done with it, this results in a resource leak and eventually we crash due to that.
* Next issue, we try to initialize a hardware buffer even if the EGL image we are trying to associate with it failed to initialize.
* Final issue, for some reason the blit of auxiliary frame buffers is failing in Android, so I put back the old method only for auxiliary frame buffers which are sync copies anyways and don't use EGL image. They are also always scaled to 1x.
* Also fix recently introduced GLES 2.0 GLSL shader compilation error.

### DIFF
--- a/src/BufferCopy/ColorBufferToRDRAM.cpp
+++ b/src/BufferCopy/ColorBufferToRDRAM.cpp
@@ -22,8 +22,6 @@ using namespace graphics;
 
 ColorBufferToRDRAM::ColorBufferToRDRAM()
 	: m_pCurFrameBuffer(nullptr)
-	, m_frameCount(-1)
-	, m_startAddress(-1)
 {
 }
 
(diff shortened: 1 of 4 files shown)


src/BufferCopy/ColorBufferToRDRAM.cpp
```diff
@@ -84,46 +82,48 @@ bool ColorBufferToRDRAM::_prepareCopy(u32& _startAddress)
...
+		blitParams.combiner = CombinerInfo::get().getTexrectDownscaleCopyProgram();
+		blitParams.readBuffer = readBuffer;
+		blitParams.drawBuffer = pBuffer->getColorFbFbo();
+		blitParams.mask = blitMask::COLOR_BUFFER;
+		wnd.getDrawer().blitOrCopyTexturedRect(blitParams);
+
+		gfxContext.bindFramebuffer(bufferTarget::READ_FRAMEBUFFER, pBuffer->getColorFbFbo());
 	} else {
-		width = m_pCurFrameBuffer->m_pTexture->width;
+		gfxContext.bindFramebuffer(bufferTarget::READ_FRAMEBUFFER, readBuffer);
```
gonetz, line 124: If I understand you correctly, you skip blit for aux buffers.
Why? Aux buffer will be in non-native resolution if frameBufferEmulation.copyAuxToRDRAM is off, and thus it needs blit to native-res buffer too.

gonetz (changes requested):

src/BufferCopy/ColorBufferToRDRAM.cpp
fzurita: I thought this code only get executed for auxiliary frame buffers if copy aux to RDRAM is enabled?

gonetz: You're right.

gonetz (approved):

Issue #2484, opened 2021-04-23
Starcraft 64 blurry at 1x


fzurita:
Starcraft 64 is extremely blurry at 1x after recent changes: 

![starcraft_64-000](https://user-images.githubusercontent.com/11581687/115902153-1acd9c00-a430-11eb-80a0-357454756e2f.png)

standard-two-simplex:
This is how I see it. I haven't managed to reproduce the issue in your picture, despite trying different settings. Can you try to find out which setting is provoking it?
 
![starcraft_64-001](https://user-images.githubusercontent.com/12202580/116290277-9a8e9a00-a793-11eb-8101-89dcee68e33c.png)

fzurita:
It took me a while to find the setting causing it. It's the stripped background mode. Using precise = no blurry text, fast = blurry text.

This was working fine before.

gonetz:
It looks incorrect with "one piece" background mode at higher resolutions as well.
Something is broken here indeed.

standard-two-simplex:
Removing this fixes the issue. Any idea why it is there? Perhaps worth mentioning, it used to have a different condition in the past.

https://github.com/gonetz/GLideN64/blob/4049f7d1ff2489429788be8d1e41232502f53784/src/uCodes/S2DEX.cpp#L505-L512

standard-two-simplex:
So the issue reported by fzurita seems to be happening because the vertex shift affects both one piece and stripped modes. After the changes the vertex shift provokes the texture coordinates being 0.5 smaller.

Before the changes, stripped mode rendered the primitive via drawTextureRect() where the 0.5 subtraction was happening. The recent changes remove this subtraction, therefore, the same result is reached before and after the changes.

In one piece mode, however, the code goes through the piece I quoted above and then it is rendered via drawScreenSpaceTriangle(). No subtraction was removed by recent changes, therefore, the texture coordinates are now 0.5 smaller, causing the blurryness.

If the above code were all subtractions, I would say simply remove it. But I see that the lower right coordinates are added instead, increasing the boundary.  It looks a bit suspicious to me, does the microcode really do such a thing?

gonetz:
Yes, initially it was like this:
```
	if ((gSP.objRendermode&G_OBJRM_BILERP) != 0) {
		uls -= 0.5f;
		ult -= 0.5f;
		lrs -= 0.5f;
		lrt -= 0.5f;
	}

```
that is all subtractions. Then @gizmo98 changed it to 
```
[...]

Issue #2304, opened 2020-07-10
(Switch) Yoshi Story Background issue regression


HappehLemons:
Example here (https://github.com/libretro/mupen64plus-libretro-nx/issues/97)

Happening in 363c47a(https://github.com/libretro/mupen64plus-libretro-nx/commit/363c47aaa950b40a8d702518f4274ef83de4f3b5) on Nintendo Switch. bgMode set to OnePiece.

Doesn't happen with stripped, but game is unplayable with stripped as FPS drop to about 1 frame per second when moving.

Issue is not present in Lakkas May build on Switch with matching settings, but I can't find a build version on the core.

Also doesn't happen on PC.

m4xw:
This might've already been fixed between the time i made the release and current master.
Will confirm soon-ish

m4xw:
Still a issue for me

fzurita:
This is broken in Android too. It would probably be good to find it which commit broke it.
![yoshi_story-000](https://user-images.githubusercontent.com/11581687/98612100-6e548f00-22c1-11eb-8208-89e291f1034b.png)

fzurita:
@standard-two-simplex This is not specific to switch. It happens in Android as well. Issues with background in that game were initially introduced by the updated texture coordinate calculation: 4834359ec1ad031210fb6ade6c9d50e953b20d0e
In that commit, the background is completely blank. A follow up commit improved it slightly to what we see in the screenshots. I'm trying to find it.

fzurita:
@gonetz So you are aware, be2f2244a2a5d9238b6fa6f9cd03166dc2692454 Improved the situation but not fully fixed it. It leads to the screenshots above.

gonetz:
Ok

standard-two-simplex:
> Still a issue for me on Switch (aarch64) with Mesa + GL

The original issue happens when backgrounds are set to render in one piece. It seems we are clamping such textures in wrong boundaries. It is probably best to disable all clamping wrapping for special textures.

213b00cf503d8fadc06fbd7ca9cddf54d7f4a1d1 fixes the issue. I also initialized wrap bounds to 1024.0. Even if they are meant to be unused due to the flag wrapEn being 0, it is necessary to avoid divisions by 0.

So far I left the previous code for framebuffer textures, but perhaps it would be best to avoid clamping them too. @gonetz Should I PR as is or should I change the behaviour of framebuffer textures?

standard-two-simplex:
[...]

PR #2497, opened 2021-05-02
Fix S2DEX one-piece backgrounds

standard-two-simplex:
Should be tested first

### DIFF
--- a/src/uCodes/S2DEX.cpp
+++ b/src/uCodes/S2DEX.cpp
@@ -476,10 +476,10 @@ struct ObjCoordinates
 
 		f32 frameW = _FIXED2FLOAT(_pObjScaleBg->frameW, 2);
 		f32 frameH = _FIXED2FLOAT(_pObjScaleBg->frameH, 2);
-		f32 imageW = (f32)(_pObjScaleBg->imageW >> 2);
-		f32 imageH = (f32)(_pObjScaleBg->imageH >> 2);
-		//		const f32 imageW = (f32)gSP.bgImage.width;
-		//		const f32 imageH = (f32)gSP.bgImage.height;
+		//f32 imageW = (f32)(_pObjScaleBg->imageW >> 2);
+		//f32 imageH = (f32)(_pObjScaleBg->imageH >> 2);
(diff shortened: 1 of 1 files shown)


gonetz:
I'm testing it now.

standard-two-simplex:
It still has the issues discussed in #2484. I remember trying to apply a texture coordinate offset, in a similar way to texture rectangles, but there was still an issue with the last line, so I wasn't satisfied.

I will upload the test tomorrow for reference.

gonetz:
The fix works, but "one-piece backgrounds" are incompatible with "native-res texrects" - heavy glitches when both options are enabled.

> It still has the issues discussed in #2484. 

Yes, it has :(

standard-two-simplex:
The last commit adjusts the one piece background behaviour with regarding the texture coordinte offset to that of texture rectangles.

There is still the issue of the last line in higher resolutions. This type of issue seems to be common to many games. When rendering in non-native, bigger texture coordinates are reached. This causes fetching outside the intended texture.

In my opinion, the best possible solution is using texture coordinate bounds, as with texture rectangles. I believe it should be possible to compute good bounds. However, the current implementation seems to be failing in some aspects.

I tried computing bounds for one-piece backgrounds too. I had trouble with finding good bounds. It seems that the texture coordinates computed by GLideN64 are offset (not sure why), so the clamping bound fails. This causes issues like #2304 when I try to clamp the texture coordinates.

gonetz:
I'd like to actualize that fix. Current issue: regression in native res mode.

gonetz:
I see that your commit https://github.com/standard-two-simplex/GLideN64/commit/51b9a3449175143a8b940cad42ceecc40ba7328c
is not included into this PR. Is it on purpose?

standard-two-simplex:
> Is it on purpose?

Yes, it is unfinished. I wanted to use the boolean for validity instead of forcing the bounds for background textures.

This PR fixes the blur, which is the regression. The other commits try to address the extra lines at the bottom. I will try to finish it with the hint you left on the issue thread.

standard-two-simplex:
I realize now that these commits were merged. I will close the PR, but leave the issue open until the discussion there is resolved.

Issue #2491, opened 2021-05-01
Kirby: Krystal Shards black level selection


fzurita:
With recent changes, with copy color to RDRAM disabled, Kirby Krystal Shards produces a black screen in the level selection screen.

fzurita:
Actually, this has been broken for a long time.

fzurita:
The odd thing is if I make a save state at the black screen, then exit the emulator and start back up then load the save state, the level selection screen works correctly.

PR #2498, opened 2021-05-02
Fix Kirby 64 broken level selection

fzurita:
Without this fix, the level selection is broken when copy color to RDRAM is disabled.

This fixes https://github.com/gonetz/GLideN64/issues/2491

@gonetz Do you think this is the right fix?

### DIFF
--- a/src/VI.cpp
+++ b/src/VI.cpp
@@ -256,8 +256,10 @@ void VI_UpdateScreen()
 				 (config.generalEmulation.hacks & hack_RE2) == 0 &&
 				 !pBuffer->isValid(true)) {
 			gDP.changed |= CHANGED_CPU_FB_WRITE;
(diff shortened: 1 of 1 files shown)


gonetz:
> @gonetz Do you think this is the right fix?

Yes.  `m_cleared `means that the RDRAM buffer contains only fillcolor.
It should be false after `copyRdram()` since from now the buffer validity check should work with `m_RdramCopy`.
Actually, `m_cleared `should be set to false in `copyRdram().` I don't understand why it is not so.
I moved `m_cleared = false;` to copyRdram(), see 7abf593da3
It should fix #2491 too. Please check it.

fzurita:
@gonetz You commit worked as well, I'll close the pull request and issue.

gonetz:
Thanks!

Issue #2208, opened 2020-04-11
starcraft shakes


olivieryuyu:
run in the intro, the screen shakes

loganmc10:
Are you talking about the intro video? In mupen64plus it doesn't shake, but it looks like this:

![starcraft_64-000](https://user-images.githubusercontent.com/848146/79058421-37cf6c80-7c2b-11ea-9527-05aa00056a10.png)

However, it seems to work fine when running in 1x native res

fzurita:
@loganmc10 I think that video needs "stripped" background render mode.

loganmc10:
It's weird, sometimes it seems to do it, sometimes it doesn't even with stripped mode, but in either case I don't see any shaking

gonetz:
> I don't see any shaking

I also can't reproduce the issue. I fixed an issue with the intro in "OnePiece" mode.
@olivieryuyu please test again. If the issue can be reproduced, please give me your GLideN64.ini, or try to reset the settings.

Searinox:
I think the lack of details in the main post here is being a problem. The intro - expansion pak detected, N64 logo, Blizz entertainment and Mass Media logos - are shaking very rapidly - I'd gather once per frame - from left to right. Happening 100% on my end.

fzurita:
Hmmm, ok... Probably have to fix this.

fzurita:
Can I get a video of it since I don't see it on my end?

Searinox:
https://drive.google.com/file/d/12lKR2lEWmdqQ-Rh-8NO_l4xbrjA79POf/view

Note that since the video isn't 60 FPS the effect sometimes doesn't sync up with the frames in the video, and it appears to go away for a while. It persisted right up to the title screen.

fzurita:
This seems to only happen with NVidia. I tried a few Android devices and AMD Radeon and it doesn't happen there. It did happen in my AMD graphics card though.

fzurita:
@gonetz Can you re-open this?

fzurita:
For reference, these are the commands executed on each shake:
```
2021/05/22,21:43:17.134,opengl_Command.cpp:31,ERROR, "Command: glActiveTexture"
2021/05/22,21:43:17.135,opengl_Command.cpp:31,ERROR, "Command: glBindTexture"
2021/05/22,21:43:17.135,opengl_Command.cpp:31,ERROR, "Command: glTexSubImage2D"
[...]

PR #2505, opened 2021-05-23
Temporary fix for NVidia shaking on Windows

fzurita:
Fix for #2208.

There was one race condition that reallocated a buffer while it was still being used, but that didn't fix the root cause. As a temporary workaround I'm just increasing the initialize memory pool size. Performing no reallocations of the memory pool allows it to work correctly.

The root cause seems to be something deeper with memory buffers being used across multiple threads that affects NVidia only. I may need some sort of memory barrier.


fzurita:
I see another issue. When reallocating the memory pool, the consumer thread still has a pointer to the old memory which is not done using yet. When increasing the memory pool size, I will have to wait until the memory pool is empty to avoid that. Hopefully that fixes it.

Issue #2496, opened 2021-05-02
EnableTexCoordBounds breaks pokemon puzzle league video


fzurita:
When EnableTexCoordBounds is enabled, the Pokemon Puzzle League videos stops working. No need for screenshot, it's just a black screen.

fzurita:
It also breaks starcraft.

fzurita:
@standard-two-simplex In case you haven't seen.

standard-two-simplex:
I can't seem to be able to reproduce this issue, with or without texcoord bounds. No black screen in Pokemon Puzzle League or StarCraft.

fzurita:
Here is what is common between Starcraft 64 and Pokemon Puzzel league:

```glsl_CombinerProgramUniformFactory.cpp:305,ERROR, "bound: tile=0 bounds=(0.000000, 0.000000, 0.000000, 0.000000)"```

In project 64, update is not being called at all somehow.

fzurita:
@standard-two-simplex @gonetz I figured it out why there are differences in behavior between PC and Android. In Android, I don't use the OSD function. When we draw the OSD, we set the drawing state returned by ```dwnd().getDrawer().getDrawingState()``` to ```Non```. This essentially disables the logic ```URasterInfo```.

If you disable the OSD in Project 64, then you will get the same result as Android. This looks like a bug to me.

fzurita:
Fixed by  66e8ff3c001ad4139ce5de2c79c65792f60d363e

PR #2509, opened 2021-05-31
Fix tex coord bounds

fzurita:
Fix for https://github.com/gonetz/GLideN64/issues/2496. Hopefully it's the correct approach.

### DIFF
--- a/src/GraphicsDrawer.cpp
+++ b/src/GraphicsDrawer.cpp
@@ -1467,7 +1467,6 @@ void GraphicsDrawer::correctTexturedRectParams(TexturedRectParams & _params)
 
 void GraphicsDrawer::drawText(const char *_pText, float x, float y)
 {
-	m_drawingState = DrawingState::Non;
 	g_textDrawer.drawText(_pText, x, y);
 }
 
--- a/src/GraphicsDrawer.cpp
+++ b/src/GraphicsDrawer.cpp
@@ -1685,7 +1685,7 @@ bool GraphicsDrawer::isRejected(u32 _v0, u32 _v1, u32 _v2) const
 
 void GraphicsDrawer::copyTexturedRect(const CopyRectParams & _params)
 {
-	m_drawingState = DrawingState::TexRect;
+	m_drawingState = DrawingState::TexRectCopy;
 
 	const float scaleX = 1.0f / _params.dstWidth;
 	const float scaleY = 1.0f / _params.dstHeight;
--- a/src/GraphicsDrawer.h
+++ b/src/GraphicsDrawer.h
@@ -29,7 +29,8 @@ enum class DrawingState
 	Triangle,
 	ScreenSpaceTriangle,
 	Rect,
-	TexRect
+	TexRect,
+	TexRectCopy
 };
 
 struct RectVertex
--- a/src/gDP.cpp
+++ b/src/gDP.cpp
@@ -904,6 +904,7 @@ void gDPFullSync()
 	}
 
 	dwnd().getDrawer().flush();
+	dwnd().getDrawer().dropRenderState();
 
 	frameBufferList().updateCurrentBufferEndAddress();
 


standard-two-simplex:
The redefinition of `useTexCoordBounds` in 13a3dbf only hides the real issue. `useTexCoordBounds` is meant to be true for texrects, when last texrect data has been set for it. 
If tcbounds are zero, then last texCoordData is incorrect. This might be happening because non-texrect (gDPTextureRectangle() to be precise) have drawing state set to DrawingState::Texrect. I will check if I can find such an issue.

fzurita:
In the starcraft case '''gDPTextureRectangle''' is never called at start up before we set '''useTexCoordBounds'''.

It does later get called, but several parameters of last texCoordData are zero even though the dimensions of 640x480 are correct.

standard-two-simplex:
`GraphicsDrawer::copyTexturedRect()` seems to set the drawing state to TexRect. So my detecting a texrect in 
```
isTexRect = dwnd().getDrawer().getDrawingState() == DrawingState::TexRect;
```
is giving unintended false positives.

fzurita:
Maybe the answer here is to have another drawing state? ```DrawingState::TexRectCopy```?

standard-two-simplex:
> Maybe the answer here is to have another drawing state? DrawingState::TexRectCopy?

It will fix this issue, I believe.

I will eventually try to rewrite the texCoordBounds implementation. It started as a simple idea and got complicated as I had to support different situations (texcoord shifts, native rects, etc.). Unfortunately, I expect it to be time consuming, and I don't have plenty right now, so I don't know when this will happen.

fzurita:
I tried adding ```DrawingState::TexRectCopy``` that is only set when copying a text rect and not used anywhere. It improves it, but it doesn't completely fix it. There are still some cases where ```useTexCoordBounds``` is coming out to true when it should not be, which causes flickering in the starcraft intro when it transitions from picture to picture.

gonetz:
> For now I think we should probably keep the sanity check until you can improve your heuristics.

Or find all cases, which cause wrong drawing state.


> which causes flickering in the starcraft intro when it transitions from picture to picture.

May be we should set drawing state to non in gDPFullSync()

gonetz:
> `GraphicsDrawer::copyTexturedRect()` seems to set the drawing state to TexRect.

GraphicsDrawer::copyTexturedRect is an auxiliary function , which is not related to a RDP command.
It can set the drawing state to texrect, but I think it should restore it on exit:
`	ValueKeeper<DrawingState> otherMode(m_drawingState, DrawingState::TexRect);`
instead of 
`	m_drawingState = DrawingState::TexRect;
`

gonetz:
> May be we should set drawing state to non in gDPFullSync()

This fixes the flickering in the starcraft intro:
```
 	dwnd().getDrawer().flush();
+	dwnd().getDrawer().dropRenderState();
```

fzurita:
@gonetz I can close this pull request, it seems you have the correct solution.

gonetz:
Please check that it fixes the issues, which you tried to fix with your commits.

fzurita:
@gonetz I tried what you suggested and it doesn't seem to work, it behaves like before. Just to make sure, in ```GraphicsDrawer::copyTexturedRect``` replace ```m_drawingState = DrawingState::TexRectCopy;``` with ```ValueKeeper<DrawingState> otherMode(m_drawingState, DrawingState::TexRect);```

In ```gDPFullSync``` add ```
	dwnd().getDrawer().dropRenderState();``` after ```dwnd().getDrawer().flush();```

fzurita:
@gonetz Please see my last commit, that seems to work correctly, for some reason the ```ValueKeeper``` doesn't seem to work right.

gonetz:
> Please see my last commit, that seems to work correctly, for some reason the `ValueKeeper` doesn't seem to work right.

I changed only gDPFullSync() to fix the starcraft flickering, in master, not in your branch. Yes, only `dwnd().getDrawer().dropRenderState();` added.
GraphicsDrawer::copyTexturedRect modification is not necessary for that issue, but probably is useful for other scenarios.
I hope that the drop drawing states in gDPFullSync resolves other similar issues, but it should be proved. You mentioned only starcraft, so I don't know where else to look.

fzurita:
Pokémon puzzle league also has issues without this fix.

fzurita:
Also, please turn off the OSD to see the issues in master, the issues only show when the OSD is disabled. My first commit in this pull request allows the issue to be seen with the OSD enabled.

gonetz:
> Also, please turn off the OSD to see the issues in master,

Master, OSD off:
No intro screens in starcraft. Nothing until the video.
With 7e0abbe498, no intro screens even with OSD enabled.
I'll try to find what is going on here. Hopefully, tomorrow.

gonetz:
@fzurita I made another attempt to fix the problem:
https://github.com/gonetz/GLideN64/tree/drop_render_state
please test it.

The idea is the following: render state is valid for currently drawn primitive. It should not be a state variable.
That is, render state should be reset to Non after drawing of a primitive is completed. It was not done until now because
it was not really necessary - each primitive changes the state at the beginning of drawing process. There are special cases:
drawText and copyTexturedRect, which draw auxiliary primitives. TexCoordBounds heuristics should not be applied to them, so I set drawing state here to Non. I hope that "false positive for useTexCoordBounds" problem is fixed now.

fzurita:
Ok, I will try it.

fzurita:
@gonetz Should you also drop the render state when drawing text? Otherwise GLideN64 has slightly different behavior when the OSD is enabled.

fzurita:
@gonetz Everything works correct in your branch. I will close this pull request in favor of your solution.

gonetz:
Good, thanks!

PR #2511, opened 2021-06-01
Updated translations

Felipefpl:
Please, take a look at how the translations look in all tabs/sections once compiled, apparently Glide doesnt support the expansion of the GUI for translated words (which are larger in other languages than english)


gonetz:
> apparently Glide doesnt support the expansion of the GUI for translated words

It is possible. Most of translations are outdated and the current GUI layout is tested only with English.

gonetz:
I checked your work. GUI with Portuguese translation is wider than English one:
GUI_translated
It looks fine. Merged: 189afd3996
Thanks!

Felipefpl:
Thanks for checking, any idea when a new release will come?

gonetz:
It was planned yet the last year, but still there are regressions, which must be fixed before the release.

Felipefpl:
I see, good to know it's coming.

PR #2512, opened 2021-06-03
Render non-filtered rectangles in native resolution too, when nativeResTexrects are enabled

### LINKED ISSUE #2506: [Regression] Mario Kart 64 - Glitchy gfx in Title screen and other menus
On GLideN64 rev. be851a3, the version that comes with Project64 3.0, has horrible glitchy graphics in all the Mario Kart 64 title and menus. I using all default settings that are provided with the version, besides the windowed mode and fullscreen mode sizes.

Here are some pictures of what they look like:
GLideN64_MARIOKART64_000
GLideN64_MARIOKART64_001
GLideN64_MARIOKART64_002
GLideN64_MARIOKART64_003

standard-two-simplex:
Fixes #2506.

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp
@@ -242,7 +242,7 @@ public:
 		float texCoordOffset[2][2] = { 0.0f, 0.0f };
 		if (isTexRect && !isNativeRes) {
 			float scale[2] = { 0.0f, 0.0f };
-			if (config.graphics2D.enableNativeResTexrects != 0) {
+			if (config.graphics2D.enableNativeResTexrects != 0 && gDP.otherMode.textureFilter != G_TF_POINT) {
 				scale[0] = scale[1] = 1.0f;
 			} else {
 				scale[0] = scale[1] = static_cast<float>(config.frameBufferEmulation.nativeResFactor);
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp
@@ -257,6 +257,10 @@ public:
 					} else {
 						texCoordOffset[t][0] = (gDP.lastTexRectInfo.dsdx >= 0.0f ? 0.0f : -1.0f) * gDP.lastTexRectInfo.dsdx * _pTexture->hdRatioS;
 						texCoordOffset[t][1] = (gDP.lastTexRectInfo.dtdy >= 0.0f ? 0.0f : -1.0f) * gDP.lastTexRectInfo.dtdy * _pTexture->hdRatioT;
+						if (gDP.otherMode.textureFilter != G_TF_POINT) {
+							texCoordOffset[t][0] -= 0.5f;
+							texCoordOffset[t][1] -= 0.5f;
+						}
 					}
 				}
 			}


src/GraphicsDrawer.cpp
```diff
@@ -1233,7 +1233,7 @@ void GraphicsDrawer::drawTexturedRect(const TexturedRectParams & _params)
 	TextureCache & cache = textureCache();
 	const bool bUseBilinear = gDP.otherMode.textureFilter != 0;
 	const bool bUseTexrectDrawer = m_bBGMode || ((config.graphics2D.enableNativeResTexrects != 0)
-		&& bUseBilinear
+		//	&& bUseBilinear
```
gonetz, line 1236: Why? Native res texrects were invented to fix issues with bilinear filtering.
Unfiltered texrects usually worked fine. 
What is changed?

standard-two-simplex: The problem is a conflict with this. Scale is computed equally for all texrects.
https://github.com/gonetz/GLideN64/blob/c8a25914bcc47a176aad12cb5e9c2ffdc3ff3847/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp#L245-L249
It is also possible to change this condition to `config.graphics2D.enableNativeResTexrects != 0 && gDP.otherMode.textureFilter = Texture`.

> Unfiltered texrects usually worked fine.

The Ogre Battle case is one where unfiltered rects don't render correctly. It is possible to render correctly in native, but not in higher resolutions, due to dsdx being too big. Unfortunately, the scene is not correct with the texrect drawer enabled either, not sure why.

> What is changed?

Mario Kart press start menu with enableNativeResTexrects enabled has an issue after the latest changes, where an incorrect texture coordinate bias is being applied.

standard-two-simplex:
@gonetz I added 2ce9c13 as well. If you prefer this approach, I can clean things up.

gonetz:
> The Ogre Battle case is one where unfiltered rects don't render correctly. It is possible to render correctly in native, but not in higher resolutions, due to dsdx being too big. Unfortunately, the scene is not correct with the texrect drawer enabled either, not sure why.

Do you mean that texrect drawer was enabled for all texrects, including unfiltered ones, but it did not help?

> I added [2ce9c13](https://github.com/gonetz/GLideN64/commit/2ce9c13c2c53ba55cd87a7d84203a40492834456) as well. If you prefer this approach, I can clean things up.

Yes, I prefer this one, since it keeps previous behavior. 

May be I need to add an option to use texrect drawer for all texrects. ParaLLEl RDP renders rects in native-res when in upscale mode.:  [how-the-upscaled-rendering-works](https://www.libretro.com/index.php/parallel-rdp-how-the-upscaled-rendering-works/)
Cite: "Many games render sprites with TEX_RECT with the expectation that textures are rendered 1:1 with input texels to output texels. When we start upscaling, the game might have forgot to disable bilinear filtering, and we start filtering outside the texture boundaries, i.e., against garbage, which shows up as ugly seams in the image. The simple workaround is to render TEX_RECT primitives as if they are not upscaled. This is necessary anyways for the COPY pipe, since the COPY pipe only updates the varying interpolator every 8th framebuffer byte. We cannot safely upscale these kinds of primitives either way."

standard-two-simplex:
Done.

> Do you mean that texrect drawer was enabled for all texrects, including unfiltered ones, but it did not help?

Yes, I enabled it in 0bd9ad014558bda, but the scene in Ogre Battle in 1x is different from the scene in 2x with native res rects on.

> When we start upscaling, the game might have forgot to disable bilinear filtering, and we start filtering outside the texture boundaries, i.e., against garbage, which shows up as ugly seams in the image

Yes, this happens in lots of games. But I found out a similar issue happens with nearest neighbour too, because texrects that are meant to be rendered 1:1 don't always use `dsdx = 1`.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp
```diff
@@ -242,7 +242,7 @@ class URasterInfo : public UniformGroup {
 		float texCoordOffset[2][2] = { 0.0f, 0.0f };
 		if (isTexRect && !isNativeRes) {
 			float scale[2] = { 0.0f, 0.0f };
-			if (config.graphics2D.enableNativeResTexrects != 0) {
+			if (config.graphics2D.enableNativeResTexrects != 0 && gDP.otherMode.textureFilter == G_TF_BILERP) {
```
gonetz, line 245: GraphicsDrawer::drawTexturedRect uses
const bool bUseBilinear = gDP.otherMode.textureFilter != 0;
0 is G_TF_POINT
I think it is better to use that approach here as well:
`if (config.graphics2D.enableNativeResTexrects != 0 && gDP.otherMode.textureFilter != G_TF_POINT) {
`

gonetz:
> Yes, I enabled it in 0bd9ad0, but the scene in Ogre Battle in 1x is different from the scene in 2x with native res rects on.

Yes, it's strange. I guess, there is a code, which executed for 1x but not for native res rects.

> I found out a similar issue happens with nearest neighbour too, because texrects that are meant to be rendered 1:1 don't always use `dsdx = 1`

Well, it's an argument, thanks. The main problem with native res rects is performance.

Btw, the fact that G_TF_AVERAGE is treated as G_TF_BILERP may also cause texture issues, I suppose. I don't remember the difference between average and bilinear, but I remember that they are different. Since we implemented bilinear in shaders, we can do it for average too.

standard-two-simplex:
> I don't remember the difference between average and bilinear, but I remember that they are different. 

AngryLion's code perfoms the following but in integer arithmetic: if the fractional part of the texture coordinates is exactly (0.5,0.5) average all neighbouring colours, else apply three point bilinear filter. Note that averaging all four colours is equivalent to applying a four point filter for those texture coordinates.

It can be implemented, but I don't think it will make any noticeable difference. The one thing pushing me back is that there are already lots of texture filters (nearest neighbour, bilinear standard, bilinear standard with color bleeding, bilinear three point, bilinear three point with color bleeding) and their GLES2 counterparts.

src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactory.cpp
standard-two-simplex: Done

gonetz:
> Note that averaging all four colours is equivalent to applying a four point filter for those texture coordinates.
It can be implemented, but I don't think it will make any noticeable difference. 

In that case yes, it hardly matters.

gonetz:
I checked the current fix. Unfortunately, it fixes nothing for me.
MK64 title screen works correct with nativeResTexrects only when internal resolution is multiple of native.
If internal resolution is the same as output one, MK64 is glitchy with and without nativeResTexrects , and especially glitchy with TexCoordBounds enabled.

standard-two-simplex:
The case of nativeResFactor=0 are difficult to handle. I restored the old behaviour with equivalent code for this case. Mario Kart is fine, and games that used to be should also be. I would still not recommend using this setting.

PR #2526, opened 2021-06-29
GLideNUI: fix multisampling causing odd UI behavior

### LINKED ISSUE #2524: Anti-Aliasing settings not saving properly
I'm having a strange issue with the MSAA setting, where no matter what I do, even returning to default settings, it will max out the MSAA setting to 16x. Setting it to 0, or turning it off completely, and saving causes it to just jump straight back to 16x, and it's causing noticeable lag in the games that I play

Rosalie241:
When a user sets MSAA to 0x, it should stay that way and not revert to the maximum possible value.

Fixes #2524


Squall-Leonhart:
Verified fix, good for merging.

Issue #2457, opened 2021-02-13
Space Station Silicon Valley colors from texture not being picked correctly


JappaWakkaP3D:
In Space Station Silicon Valley, Dan Danger appears to be fully pink. The game should pick the colors from the colored squares at the top, but when using a hires texture, for some reason the game doesn't do that.
![image](https://user-images.githubusercontent.com/66885565/107855172-1e85e680-6e21-11eb-8e90-dcb7d52ed1de.png)
This is the original texture:
![Silicon Valley#888B8774#0#2_all](https://user-images.githubusercontent.com/66885565/107855293-a4099680-6e21-11eb-8b13-5f2f532012b6.png)

and this is mine:
![Silicon Valley#888B8774#0#2_all](https://user-images.githubusercontent.com/66885565/107855300-b257b280-6e21-11eb-9b28-d539003f0c3a.png)

Am I doing something wrong?

Jj0YzL5nvJ:
<details>
  <summary>sha256sum 107855*</summary>

```
64baefaf274776f4c97f2c4d72f20bdc6b24593d69aae66bbfb7efc3689f54a9  107855293-a4099680-6e21-11eb-8b13-5f2f532012b6.png
de41d0bca73ef3dfce6fa4e135bf148f0b6652f544da3a18a08bc6cdff18586b  107855300-b257b280-6e21-11eb-9b28-d539003f0c3a.png
```

</details>

<details>
  <summary>mediainfo 107855*</summary>

```
General
Complete name                            : 107855293-a4099680-6e21-11eb-8b13-5f2f532012b6.png
Format                                   : PNG
Format/Info                              : Portable Network Graphic
File size                                : 724 Bytes

Image
Format                                   : PNG
Format/Info                              : Portable Network Graphic
Width                                    : 32 pixels
Height                                   : 32 pixels
Bit depth                                : 32 bits
Compression mode                         : Lossless
Stream size                              : 724 Bytes (100%)

General
Complete name                            : 107855300-b257b280-6e21-11eb-9b28-d539003f0c3a.png
Format                                   : PNG
Format/Info                              : Portable Network Graphic
File size                                : 5.31 KiB

Image
Format                                   : PNG
Format/Info                              : Portable Network Graphic
Width                                    : 256 pixels
Height                                   : 256 pixels
Bit depth                                : 24 bits
Compression mode                         : Lossless
Stream size                              : 5.31 KiB (100%)
```

</details>

#2287

[...]

Issue #1971, opened 2018-12-23
S2DEX fixes. Feedback.


gonetz:
Current HLE implementation for backgrounds commands is too high level. Actual commands are very complex. The microcode is documented, but the documentation does not cover some tiny details in internal mechanics. Thus, existing implementation does not always work right. olivieryuyu decided to not rely on documentation and decode the microcode itself. The goal of that new decoding project was to implement commands in HLE as close as possible to LLE and obtain the result, which is as good as with LLE or better.

It was very long and hard project, which took circa six months. It is not finished yet, but we are close. Both background commands implemented and they produce exactly the same result as in LLE mode on all games we tested.

You may run current versions of the plugin in LLE mode and find numerous graphics glitches in 2D games. Some of glitches caused by errors in texture loading code. Backgrounds commands do texture loading in quite weird manner, which "normal" game code never uses. I fixed several hard issues in plugin's texturing subsystem during development of that project. However, many glitches remained after these fixes. These glitches can be called "the curse of hardware rendering". There are two main groups of such glitches:

* gaps between polygons. Background commands can not draw background by one piece because of limited texture memory. Background image becomes split on narrow strips. Each strip is loaded in texture memory and rendered by separate rectangle. There are no gaps between rectangles, but gaps between images in them may appear when rendering resolution is higher than game's native resolution. It usually happens when size of the image is not exactly the same as size of area where it will be rendered. Texture scaling in higher resolution may cause fetching texels out of texture image bounds, thus cause artifacts, which look as gaps between polygons.

* garbage between polygons. Looks like a line of colored garbage between two polygons. Is is seldom kind of glitches, but it does not fixed by switching to native resolution. I did not find yet, how it can be cured.

So, new implementation has issues, but most of them appears only in non-native resolutions. GLideN64 already has tool to eliminate issues with 2D graphics in high resolutions. You may enable "Render 2D elements in native resolution" option to activate it. I adopted that tool to work with new background commands. StarCraft backgrounds have glitches in high-res, but "Render 2D elements in native resolution" eliminates them. However, the tool is not bug-free by itself and in some games its use may cause disappearance of part of graphics. Besides, sometimes it is slow.

[...]

PR #2528, opened 2021-07-17
Per pixel shift offset

standard-two-simplex:
Performs per pixel shift and offset, to solve the problem in paper mario described  in https://github.com/gonetz/GLideN64/issues/2324#issuecomment-804192065

<table><tr><td>Before</td><td>After</td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

The refactor made me realize how coordinates should be upscaled for HD textures.  https://github.com/gonetz/GLideN64/issues/2324#issuecomment-815054148 should be working correctly now. @GhostlyDark, please test it is working.

This is probably also fixed. https://github.com/gonetz/GLideN64/issues/2457. @JappaWakkaP3D please check.

Any tests woul be appreciated.

Written on top of https://github.com/gonetz/GLideN64/pull/2521. Some cleanup still needed.


GhostlyDark:
Minimaps now align perfectly. Things like the rupee icon, the magic bar gauge and the ammo numbers at the top right have misalignments. This matches however with how the SD textures looked since a while now, so in that sense it is accurate.

SD
image

HD
image

HD (old)
image

SD
image

HD
image

HD (old)
image

JappaWakka:
HD
HD

SD
SD

It works!!!!! He's back in full color!

standard-two-simplex:
@GhostlyDark I restored the initial heuristic for texture coordinates correction. I checked the screen coordinates issue discussion, and the only complaints about it were about HD textures. I only modified the internal resolution = multiple of native (2x, 4x, 8x) etc, so make sure to check those. The SD textures look better after this changes, and I believe HD textures will look like the SD ones now.

GhostlyDark:
That is much better. The UI looks good now in both SD and HD, including HD Debug text.
I am not sure what's up with the floor in East Clock Town. It is not caused by the heuristic change, as Nerrel noticed this before that change was pushed.

SD
image

HD
image

HD (old)
image

standard-two-simplex:
What's wrong with the floor?

GhostlyDark:
I think the floor consists of two textures, one colored and one alpha grayscale. At least one of them is misaligned, causing the floor to appear blurry.

This issue existed at one point, disappeared and is now back again. @Nerrel might know more about this.

Nerrel:
I mentioned the problem back in the S2DEX fixes issue here: https://github.com/gonetz/GLideN64/issues/1971#issuecomment-452570892

Whenever a large grayscale and small color texture are combined by the game, like with the floor tiles posted above, the pre-S2DEX builds wouldn't align the custom textures correctly.  I had to offset one of them by maybe 8 pixels on each axis, depending on the size, in order to get them to properly match up and avoid blurring.  

I also mentioned custom textures being shifted incorrectly in older builds, which I think may have been the same issue as the grayscale mismatch: https://github.com/gonetz/GLideN64/issues/2097#issuecomment-566208022

Recent builds of GlideN64 fixed these issues and textures have been mapping/aligning as expected for a while, it seems like this merge reintroduced those issues again.  I don't fully understand what glitches this merge solved but hopefully there's a way to keep those fixes while also undoing the misalignment issues that have been brought back.

standard-two-simplex:
> I can upload a set of these grayscale+color textures if that would be helpful

Yes, please, I will need to find out which lines of code affect it and why.

> Recent builds of GlideN64 no longer had these issues and textures have been mapping/aligning as expected for a while

How recent are we talking about (more or less)?

oddMLan:
Hey @standard-two-simplex are you on Discord by any chance? It might suited more for realtime chat & testing.
We have a #gliden64 channel. Nerrel and gonetz are there too.
http://discord.com/invite/FrPyDtr

Nerrel:
"Recent" was a very loose statement, judging from the dates I made those comments it's been about two years since those issues were resolved.  It's very problematic (for me, at least) that the misalignment is back since I'm on the verge of releasing this texture pack- in fact I'm about to upload a beta so that people can test for issues.  If a lot of textures are being broken at the plugin level then that's going to dominate the comments and make it hard to get feedback about anything else.

My solution two years ago was to intentionally offset/misalign the textures by a few pixels to compensate for the misalignments made by the plugin.  After the changes around the S2DEX merge, where the alignments were fixed, I was able to go through the pack and re-export all of those textures without the offsets.  I believe that the correct behavior should be that the textures align by default without needing to be offset, especially since Dolphin handles them that way.  

Here are two texture pairs, the floor in east clock town and the door to the clock tower.  Both easily visible in clock town, you shouldn't need a savestate or anything as long as you have a save file that's past the introduction.  Another example of this type of texture pairing is the treeline in termina field near the swamp exit.  
https://drive.google.com/file/d/1Q_Z6DdJ1dCj4iFnkLM6OAtY7Wml90V6J/view?usp=sharing

Nerrel:
I also just confirmed that other texture types are also misaligned, not just these blended textures.  A huge portion of the standard textures aren't tiling correctly and are creating seams that shouldn't be there; left is the newest build from appveyor, right is the newest build before this change was merged:
GLideN64_ZELDA_MAJORA'S_MASK_027

GLideN64_ZELDA_MAJORA'S_MASK_028

There's also this weird seam that appears in north clock town when using my textures: 
GLideN64_ZELDA_MAJORA'S_MASK_025

It's basically the exact same set of problems I was commenting on in #1971 and #2097 again, which I think were fixed by the builds being posted in #2097

standard-two-simplex:
@oddMLan I don't have a discord account, and I wouldn't hang out much there anyway. I am okay with conversation via GitHub, I don't need instant responses.

@Nerrel I tested the floor textures you gave me. In the SD case, both old and new builds look the same. In the HD case, the old texture doesn't align with the SD one, while the new build does. 

This is how I see them. They are taken with a still camera, after loading the savestate.
OLD SD
old_2x
OLD HD
old_hd
NEW SD
new_2x
NEW HD
new_hd

You can see that the OLD HD is shifted to the left, its easily noticeable on the left of the screen, where one of the floor tiles is barely cut in three pictures. Maybe the new one is a bit blurrier than the old one, but I can't say the game didn't intend to draw them like that.

Can you give me grass textures where the visual artifact is more obvious?

Nerrel:
I grabbed some screenshots of another texture that I know has this issue in the mayor's office.  Newest build:
GLideN64_ZELDA_MAJORA'S_MASK_021

Pre-merge build: 
GLideN64_ZELDA_MAJORA'S_MASK_033

Between these two the second shot is obviously the correct one in terms of what you'd expect to see when looking at the original texture: 
ZELDA MAJORA'S MASK#8ACD2634#0#2_all

It doesn't make sense for a tiny strip of the wallpaper's top to appear at the bottom of the wall; the texture wasn't drawn that way and I don't know why it would be mapped that way.  It's harder to see but the panels below are doing the same thing, tiling from a few pixels inward from the edge of the texture rather than tiling from the actual edge and thereby creating a visible seam:
GLideN64_ZELDA_MAJORA'S_MASK_017


But to my surprise the original textures actually seem to be tiling the first way, in both the newest build and the pre-merge build: 
GLideN64_ZELDA_MAJORA'S_MASK_022

I checked using angrylion and it seems to do the same:  
ZELDA MAJORA S MASK_0000

So at this point I have no idea what's going on, I don't have the time to test more cases to see if they're all the same.  Maybe it _is_ more accurate to map things the new way but it would make texture packs substantially harder to create because it requires editing textures on an individual basis to offset and move the seams away from the edges of textures by a specific number of pixels; to fix the mayor's office texture I'd have to edit it to have a precisely sized strip of the _bottom_ at the top.

My pack has around 5,000 textures and the amount of time it would take to determine the specific pixel offset needed for each affected texture and to then do that edit would be totally impractical.  In this case I'd have to argue that the previous behavior is far better for texture packs because, even if it gets confirmed to be inaccurate, it makes more sense to tile the textures from the edges of an object rather than a few rows of pixels inward from the edge.  Doing things that way makes more sense and eliminates a ton of tedious work, and I doubt any HD texture artists would want their textures tiling in such a weird way.  

Here's another example from great bay using the newest build: 
GLideN64_ZELDA_MAJORA'S_MASK_003

Even if this is the accurate way, the texture really "shouldn't" be tiling with the edge of the texture misaligned with the edge of the walkway.  It's very hard to tell due to the blurriness, but I think in this case the original texture may also be working this way with the seam in the wrong place: 
GLideN64_ZELDA_MAJORA'S_MASK_002

The pre-merge build once again does not create this seam, it tiles the texture from the very edge which is what I intended the HD texture to do.  I think testing these out in Dolphin and comparing how things work there would be interesting, but again I don't have the time to try it now. 

Here's a folder with the textures I've shown as well as savestates for those areas.  The brick texture I showed in the previous comment is on the ground in south clock town, no save state is needed to see it:
https://drive.google.com/file/d/1xaIdjED1LMLCCFGiukUNURooJzHyuCFL/view?usp=sharing

Nerrel:
Yeah, I don't think it would make sense to hold back on a plugin improvement just for one game.  Maybe I'll have to try to get users of the pack to grab that specific build, but I _could_ fix all the textures if I could be absolutely sure that the plugin wouldn't change again and undo that work.  Do you expect this new behavior to be pretty much permanent going forward? 

Admentus tested things out in Dolphin for me and it's actually the same as the new GlideN64 build, I had it totally wrong: 
NARE01_2021-07-20_12-26-06

standard-two-simplex:
>  Do you expect this new behavior to be pretty much permanent going forward?

I would say this behaviour should be expected, and not looking like SD should be considered a bug. I would still keep a backup of the current textures, just in case something new is discovered.

Nerrel:
One thing I still don't understand is why the previous build tiled everything "correctly" (inaccurate but better looking) if the game had bad texture mappings... do you know what was going on there?  I'm not questioning that the new behavior is better overall, it just seems weird to me that the plugin was somehow putting all of these sloppily mapped textures into just the right places in every case.  

Knowing the way the plugin was behaving before vs now might help me make the adjustments, since I may be able to figure out the offset needed for each texture in advance without having to experiment in-game by editing and reloading them over and over.

standard-two-simplex:
> do you know what was going on there?

Old formula to compute HD coordinates: `original * hdRatio`
New formula to compute HD coordinates: if nearest neighbour or copy `original * hdRatio`, if bilinear `(original + 0.5)*hdRatio -0.5`

The new formula takes into account that in bilinear mode texel centers have integer coordinates and maps the texture coordinates accoringly (-0.5 -> -0.5 and SDWidth-0.5 -> HDWidth-0.5).

If you perform NEW - OLD subtraction and simplify, you get `0.5*(hdRatio-1)`. So a 32x native texture will be offset by 15.5 texels with respect to the previous build. The 0.5 is likely to create the blur. If you want to adapt the texture, you probably want to offset it 16 pixels (i.e. half of the HdRatio), **but only if it is bilinearly filtered**.

Nerrel:
Thanks for the explanation, that should help.  I don't know if the blur is intended on those blended textures, I thought the small color+large grayscale technique was just for cutting down file sizes compared to using one large color texture, but I don't really know.  Is there any simple way of identifying the textures that are meant to have bilinear filtering applied, or does it just have to be eyeballed?  I don't think the filenames indicate that

standard-two-simplex:
Unless I missed some type of primitive, this build should disable primitives in copy mode or nearest neighbour mode. It seems almost everything in MM is using bilinear filtering.
[build.zip](https://github.com/gonetz/GLideN64/files/6850783/build.zip)

gonetz:
All builds for this PR (standard-two-simplex-per_pixel_shift_offset branch): 
https://ci.appveyor.com/project/gonetz/gliden64/builds/40077667/artifacts

gonetz:
I tested that branch. It works really good. I didn't find any regressions. 
"paper mario in the dark" problem is really fixed. I have no notices for the code too.

gonetz:
@standard-two-simplex please resolve conflicts. I think it can be merged after that.

standard-two-simplex:
@gonetz I was doing some cleanup. 

```
"mediump vec2 calcTexCoord(in vec2 texCoord, in int idx)		\n"
"{																\n"
"    vec2 texCoordOut = texCoord;								\n"
"    if (uTextureFilterMode != 0 && uCacheFrameBuffer[idx] != 0) \n"	/* Workaround for framebuffer textures. */
"      texCoordOut -= vec2(0.0,1.0);							\n"		/* They contain garbage at the bottom.  */
"    return texCoordOut;										\n"
"}																\n"
```
These lines only apply a hack at the moment. It was introduced for Conker's shadow and Bob-omb garbage in Mario Tennis, if I recall correctly. If they were removed, it would be possible to use only one vertex attribute for texture coordinates, as vTexCoord0 and vTexCoord1 now both contain the original texture coordinate. It is the fragment shader that separates the coordinate from cycle 0 and cycle 1.

gonetz:
Yes, Conker's shadow looks awful now. Any idea how to fix it? May be with a hack?

standard-two-simplex:
It should be possible with a hack. I don't know why the framebuffer textures seem offseted.

standard-two-simplex:
I pushed two options. The first one is similar to the previous one, the other one moves it to another place so the vertex shader can be cleaned up.

fzurita:
Shader cache version needs to be incremented.

fzurita:
No issues noted in Android with an Adreno GPU other than having to manually clear the shader cache.

standard-two-simplex:
So I propose option two, because it cleans the vertex shader code considerably. If it is okay, I will clean up the PR and increase the shader cache version.

gonetz:
> So I propose option two, because it cleans the vertex shader code considerably. If it is okay, I will clean up the PR and increase the shader cache version.

I tested the option two. Conker's shadow is fixed. I also found no new issues. Good work! Please proceed with it.

PR #2537, opened 2021-07-27
Change LOD calculation formula. Use the maximum value of the change o…

standard-two-simplex:
…f S and T per change in X and Y.


standard-two-simplex:
Shader storage version has to be increased too.

gonetz:
> Shader storage version has to be increased too.

Lets increase it once in #2528

PR #2539, opened 2021-07-28
GLideNUI: introduce HIDKeyToName

Rosalie241:
Removes the need for `osal_keycode_name`

PR #2541, opened 2021-07-28
GLideNUI: introduce config.hotkeys.enabledKeys

Rosalie241:
depends on the changes in #2539, will rebase as needed.

this allows someone to set a key but save it without having it enabled:
image

Before this patch this was impossible to do and each disabled key wouldn't save and reset to having no value, that behavior is unexpected in my opinion.


oddMLan:
Oh thank god
This will make it easier to port to WTL :D

gonetz:
LGTM. Yes, hotkeys work more naturally now, thanks!

PR #2544, opened 2021-07-29
GLideNUI: introduce Config{Init,Cleanup}

Rosalie241:
This should hopefully make opening the GLideNUI window slightly faster on first launch after launching the emulator.

### DIFF
--- a/src/GLideNUI/GLideNUI.cpp
+++ b/src/GLideNUI/GLideNUI.cpp
@@ -19,29 +19,21 @@ Q_IMPORT_PLUGIN(QICOPlugin)
 inline void initMyResource() { Q_INIT_RESOURCE(icon); }
 inline void cleanMyResource() { Q_CLEANUP_RESOURCE(icon); }
 
+static std::unique_ptr<QApplication> s_pQApp;
+static QCoreApplication* s_pAppInstance;
+
 static
 int openConfigDialog(const wchar_t * _strFileName, const char * _romName, unsigned int _maxMSAALevel, float _maxAnisotropy, bool & _accepted)
 {
-	cleanMyResource();
-	initMyResource();
 	QString strIniFileName = QString::fromWCharArray(_strFileName);
 	loadSettings(strIniFileName);
 	if (config.generalEmulation.enableCustomSettings != 0 && _romName != nullptr && strlen(_romName) != 0)
 		loadCustomRomSettings(strIniFileName, _romName);
 
-	std::unique_ptr<QApplication> pQApp;
-	QCoreApplication* pApp = QCoreApplication::instance();
-
-	if (pApp == nullptr) {
-		int argc = 0;
-		char * argv = 0;
-		pQApp.reset(new QApplication(argc, &argv));
-		pApp = pQApp.get();
-	}
-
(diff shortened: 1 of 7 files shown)


src/GLideNUI/GLideNUI.cpp
```diff
@@ -19,29 +19,21 @@ Q_IMPORT_PLUGIN(QICOPlugin)
 inline void initMyResource() { Q_INIT_RESOURCE(icon); }
 inline void cleanMyResource() { Q_CLEANUP_RESOURCE(icon); }
 
+static std::unique_ptr<QApplication> pQApp;
+static QCoreApplication* pAppInstance;
```
gonetz, line 23: Please use s_ prefix for static variables: s_pQApp, s_pAppInstance

src/GLideNUI/GLideNUI.cpp
```diff
@@ -96,6 +81,25 @@ int runAboutThread(const wchar_t * _strFileName) {
...
+		int argc = 0;
+		char * argv = 0;
+		pQApp.reset(new QApplication(argc, &argv));
+		pAppInstance = pQApp.get();
+	}
+}
+
+EXPORT void CALL ConfigCleanup(void)
+{
+	cleanMyResource();
```
gonetz, line 100: Application crashes on exit. now
I think, pQApp should be released here.

gonetz (changes requested):

src/GLideNUI/GLideNUI.cpp
Rosalie241: can you re-test with the latest push? I personally haven't experienced any crashes myself.

src/GLideNUI/GLideNUI.cpp
```diff
@@ -96,6 +81,28 @@ int runAboutThread(const wchar_t * _strFileName) {
...
+		char * argv = 0;
+		s_pQApp.reset(new QApplication(argc, &argv));
+		s_pAppInstance = s_pQApp.get();
+	}
+}
+
+EXPORT void CALL ConfigCleanup(void)
+{
+	cleanMyResource();
+	if (s_pQApp) {
```
gonetz, line 101: Redundant check. You can safely release an empty smart pointer.

src/GLideNUI/GLideNUI.cpp
```diff
@@ -96,6 +81,28 @@ int runAboutThread(const wchar_t * _strFileName) {
...
+		s_pQApp.release();
```
gonetz, line 102: Yes, this fixes the crash on exit.

gonetz (changes requested):

src/GLideNUI/GLideNUI.cpp
Rosalie241: Removed the check, thank you!

gonetz (approved):

src/GLideNUI/GLideNUI.cpp
Squall-Leonhart: This has fixed the Pj64 plugins screen crashing too when not the selected plugin.

PR #2549, opened 2021-08-06
GLideNHQ: Introduce TxNoCache

Rosalie241:
This introduces a no cache mode where (mainly) texture pack creators can reload textures quickly instead of having to regenerate a cache each time they update a texture (which is time consuming and it strains the disk with bigger texture packs, i.e Nerrel's Majora's Mask HD), the no cache mode does have more stutter but that's expected and fine for texture pack development purposes.

### DIFF
--- a/src/GLideNHQ/TxHiResCache.cpp
+++ b/src/GLideNHQ/TxHiResCache.cpp
@@ -28,17 +28,10 @@
 #pragma warning(disable: 4786)
 #endif
 
-/* use power of 2 texture size
- * (0:disable, 1:enable, 2:3dfx) */
-#define POW2_TEXTURES 0
-
-/* use aggressive format assumption for quantization
- * (0:disable, 1:enable, 2:extreme) */
-#define AGGRESSIVE_QUANTIZATION 1
-
 #include "TxHiResCache.h"
 #include "TxDbg.h"
 #include <osal_files.h>
+#include <osal_keys.h>
 #include <zlib.h>
 #include <math.h>
 #include <stdlib.h>
@@ -59,16 +52,10 @@ TxHiResCache::TxHiResCache(int maxwidth,
 						   const wchar_t *ident,
 						   dispInfoFuncExt callback)
 							 : TxCache((options & ~(GZ_TEXCACHE | FILE_TEXCACHE)), 0, cachePath, ident, callback)
-							 , _maxwidth(maxwidth)
-							 , _maxheight(maxheight)
-							 , _maxbpp(maxbpp)
+							 , TxHiResLoader(maxwidth, maxheight, maxbpp, options)
 							 , _abortLoad(false)
 							 , _cacheDumped(false)
-							 , _txImage(new TxImage())
-							 , _txQuantize(new TxQuantize())
-							 , _txReSample(new TxReSample())
 {
-
 	if (texPackPath)
 		_texPackPath.assign(texPackPath);
 
@@ -125,7 +112,7 @@ boolean TxHiResCache::_HiResTexPackPathExists() const
 	return osal_path_existsW(dir_path.c_str());
 }
 
-bool TxHiResCache::load(boolean replace) /* 0 : reload, 1 : replace partial */
+bool TxHiResCache::_load(boolean replace) /* 0 : reload, 1 : replace partial */
 {
 	if (_texPackPath.empty() || _ident.empty())
 		return false;
@@ -151,7 +138,7 @@ bool TxHiResCache::load(boolean replace) /* 0 : reload, 1 : replace partial */
 		dir_path += OSAL_DIR_SEPARATOR_STR;
 		dir_path += _ident;
 
-		const LoadResult res = loadHiResTextures(dir_path.c_str(), replace);
+		const LoadResult res = _loadHiResTextures(dir_path.c_str(), replace);
 		if (res == resError) {
 			if (_callback) (*_callback)(wst("Texture pack load failed. Clear hiresolution texture cache.\n"));
 			INFO(80, wst("Texture pack load failed. Clear hiresolution texture cache.\n"));
@@ -162,7 +149,12 @@ bool TxHiResCache::load(boolean replace) /* 0 : reload, 1 : replace partial */
 	return false;
 }
 
-TxHiResCache::LoadResult TxHiResCache::loadHiResTextures(const wchar_t * dir_path, boolean replace)
+bool TxHiResCache::reload()
+{
+	return _load(0) && !TxCache::empty();
+}
+
+TxHiResCache::LoadResult TxHiResCache::_loadHiResTextures(const wchar_t * dir_path, boolean replace)
 {
 	DBG_INFO(80, wst("-----\n"));
 	DBG_INFO(80, wst("path: %ls\n"), dir_path);
@@ -193,8 +185,8 @@ TxHiResCache::LoadResult TxHiResCache::loadHiResTextures(const wchar_t * dir_pat
 	tx_wstring texturefilename;
 
 	do {
-
-		if (KBHIT(0x1B)) {
+		osal_keys_update_state();
+		if (osal_is_key_pressed(KEY_Escape, 0x0001)) {
 			_abortLoad = true;
 			if (_callback) (*_callback)(wst("Aborted loading hiresolution texture!\n"));
 			INFO(80, wst("Error: aborted loading hiresolution texture!\n"));
@@ -210,13 +202,14 @@ TxHiResCache::LoadResult TxHiResCache::loadHiResTextures(const wchar_t * dir_pat
 		if (wccmp(foundfilename, wst(".")))
 			// These files we don't need
 			continue;
+
 		texturefilename.assign(dir_path);
 		texturefilename += OSAL_DIR_SEPARATOR_STR;
 		texturefilename += foundfilename;
 
 		/* recursive read into sub-directory */
 		if (osal_is_directory(texturefilename.c_str())) {
-			result = loadHiResTextures(texturefilename.c_str(), replace);
+			result = _loadHiResTextures(texturefilename.c_str(), replace);
 			if (result == resOk)
 				continue;
 			else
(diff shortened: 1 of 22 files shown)


Rosalie241:
@gonetz I wanted to ask you about how I'd go about adding a mask/flag to `GLideNHQ` to enable `TxNoCache` with a configuration option?

I've tried adding a flag and enabling it in `TextureFilterHandler.cpp` per the config but I'm not sure how to check for the mask correctly in `TxCache::TxCache`, it seems like it's an inverted check:
```cpp
if ((options & FILE_CACHE_MASK) == 0)
		_pImpl.reset(new TxMemoryCache(options, cachesize, _callback));
	else
		_pImpl.reset(new TxFileStorage(options, cachePath, _callback));
```


I also see that flags get removed [here](https://github.com/gonetz/GLideN64/blob/master/src/GLideNHQ/TxHiResCache.cpp#L61):
```cpp
TxCache((options & ~(GZ_TEXCACHE | FILE_TEXCACHE)), 0, cachePath, ident, callback)
```

which looks odd and I'm uncertain where to go from here.

i.e do I add a flag and check for the mask in `TxCache::TxCache` like this?
```cpp
if ((options & FILE_NOCACHE_MASK)) {
        _pImpl.reset(...);
} else if ((options & FILE_CACHE_MASK) == 0) {
	_pImpl.reset(new TxMemoryCache(options, cachesize, _callback));
 } else {
	_pImpl.reset(new TxFileStorage(options, cachePath, _callback));
}
```

Rosalie241:
I think it's ready for review now 💕

gonetz (changes requested):
The idea to add no cache mode for hi-res textures is great and it really should help texture artists. The implementation is also mostly well done. However, the idea to pass pointer to TxHiResCache into TxCache ctor is weird.
Something is wrong with the architecture if you need to do smth like this.
What is wrong? TxCache is a cache. If you need no-cache implementation, you should not use it. I suggest to resolve the problem as follow:

- Create parent class for cached and non-cached hires-texture modes:

```
class TxHiResLoader
{
protected:
	uint32_t checkFileName(char* ident, char* fname, uint32_t* pChksum, uint32_t* pPalchksum, uint32_t* pFmt, uint32_t* pSiz);
	uint8_t* loadFileInfoTex(char* fname, int siz, int* pWidth, int* pHeight, uint32_t fmt, ColorFormat* pFormat);
 
	some_members // members, necessary for checkFileName and loadFileInfoTex

public:
	TxHiResLoader(...);
	virtual ~TxHiResLoader(){}

	virtual bool empty() const = 0;
	virtual bool add(Checksum checksum, GHQTexInfo *info, int dataSize = 0) = 0;
	virtual bool get(Checksum checksum, GHQTexInfo *info) = 0;
	virtual bool reload() = 0;
	virtual void dump() = 0;
}

```

- Create TxHiResCache and TxHiResNoCache from TxHiResLoader

```
class TxHiResCache : TxCache, public TxHiResLoader
{
...

public:
	TxHiResCache(...);
	bool empty() const override;
	bool add(Checksum checksum, GHQTexInfo *info, int dataSize = 0) override;
	bool get(Checksum checksum, GHQTexInfo *info) override;
	bool reload() override;
	void dump() override;
}

```
```
class TxHiResNoCache : public TxHiResLoader
{
...

public:
	TxHiResNoCache(...);
	bool empty() const override;
	bool add(Checksum checksum, GHQTexInfo *info, int dataSize = 0) override;
	bool get(Checksum checksum, GHQTexInfo *info) override;
	bool reload() override;
	void dump() override;
}

```

- Use TxHiResLoader as member in TxFilter

```
class TxFilter
{
private:
...
//  TxHiResCache *_txHiResCache;
  TxHiResLoader *_txHiResLoader;
}

```
and create necessary loader depending on the options:
```
TxFilter::TxFilter(...)
{
...
//	_txHiResCache = new TxHiResCache(...);
	if ((options & FILE_NOTEXCACHE) == FILE_NOTEXCACHE)
		_txHiResLoader = new TxHiResNoCache(...);
	else
		_txHiResLoader = new TxHiResCache(...);
...
}
```

Rosalie241:
> The idea to add no cache mode for hi-res textures is great and it really should help texture artists. The implementation is also mostly well done. However, the idea to pass pointer to TxHiResCache into TxCache ctor is weird.

Sure, I agree it's not perfect but it was the cleanest way to do it without a major rework of that part of the code.

> Something is wrong with the architecture if you need to do smth like this.
> What is wrong? TxCache is a cache. If you need no-cache implementation, you should not use it. 

why does that matter though? in the end the code will be largely the same regardless of naming/structure here, and I wanted to touch the glidenhq codebase as little as possible to do this.

> I suggest to resolve the problem as follow:

Sure, I'll get on that, thank you.

gonetz:
> why does that matter though? in the end the code will be largely the same regardless of naming/structure here, and I wanted to touch the glidenhq codebase as little as possible to do this.

Bad architecture leads to ugly solutions and tangled code.
I also prefer not to touch glidenhq codebase, but it is a good reason to do that.

> Sure, I'll get on that, thank you.

Thanks

src/GLideNHQ/TxCache.cpp
```diff
@@ -854,6 +861,269 @@ bool TxFileStorage::isCached(Checksum checksum)
...
+	void setOptions(uint32 options) override { _options = options; }
+
+private:
+	typedef struct fileIndexEntry
+	{
+		char fname[MAX_PATH];
+		tx_wstring directory;
+		uint32 siz;
+		uint32 fmt;
+	} fileIndexEntry_t;
```
gonetz, line 896: Why use C-style here?

src/GLideNHQ/TxCache.cpp
```diff
@@ -854,6 +861,269 @@ bool TxFileStorage::isCached(Checksum checksum)
...
+	, _txHiresCache(txHiResCache)
+	, _ident(ident)
+	, _callback(callback)
+{
+	/* store this for _createFileIndexInDir */
+	wcstombs(_identc, _ident.c_str(), MAX_PATH);
+
+#ifdef OS_WINDOWS
+	/* lowercase on windows */
+	for (uint32 i = 0; i < strlen(_identc); i++) _identc[i] = tolower(_identc[i]);
```
gonetz, line 923: This code is duplicated at least twice. Better to make a static function or static method for it.

src/GLideNHQ/TxCache.cpp
```diff
@@ -854,6 +861,269 @@ bool TxFileStorage::isCached(Checksum checksum)
...
+#endif
+
+	/* loop over each file from the index and try to match it with checksum */
+	auto indexEntry = _filesIndex.find(checksum);
+	if (indexEntry == _filesIndex.end()) {
+		DBG_INFO(80, wst("TxNoCache::get: chksum:%08X %08X not found\n"), chksum, palchksum);
+		return false;
+	}
+
+	entry = indexEntry->second;
```
gonetz, line 954: better 
const fileIndexEntry_t& entry = indexEntry->second;
to avoid copy of the entry.

src/GLideNHQ/TxCache.cpp
```diff
@@ -854,6 +861,269 @@ bool TxFileStorage::isCached(Checksum checksum)
...
+	clear();
+}
+
+bool TxNoCache::get(Checksum checksum, GHQTexInfo *info)
+{
+	if (!checksum) {
+		return false;
+	}
+
+	fileIndexEntry_t entry;
```
gonetz, line 940: better to init it where it is used.

src/GLideNHQ/TxCache.cpp
```diff
@@ -854,6 +861,269 @@ bool TxFileStorage::isCached(Checksum checksum)
...
+		DBG_INFO(80, wst("TxNoCache::get: cached chksum:%08X %08X found\n"), chksum, palchksum);
+		*info = loadedTexMap->second;
+		return true;
+	}
+
+	DBG_INFO(80, wst("TxNoCache::get: loading chksum:%08X %08X\n"), chksum, palchksum);
+
+	int width = 0, height = 0;
+	ColorFormat format;
+	uint8_t* tex = nullptr;
```
gonetz, line 968: better declare these variables where are they used, that is above 
_txHiresCache->loadFileInfoTex
call

src/GLideNHQ/TxCache.cpp
```diff
@@ -854,6 +861,269 @@ bool TxFileStorage::isCached(Checksum checksum)
...
+
+		/* recursive read into sub-directory */
+		if (osal_is_directory(texturefilename.c_str())) {
+			result = _createFileIndexInDir(texturefilename.c_str(), update);
+			if (result) {
+				continue;
+			}
+			else {
+				break;
+			}
```
gonetz, line 1076:
```
if (result)
  continue;
else
  break;
```
or even

```
if (result)
  continue;

break;
```

src/GLideNHQ/TxCache.cpp
Rosalie241: would a macro be fine too?

gonetz: macro 
```
#ifdef OS_WINDOWS
#define correctString(str)  for (uint32 i = 0; i < strlen(str); i++) str[i] = tolower(str[i])
#else
#define correctString(str)
#
```
or a function
```
void correctString(char * str)
{
#ifdef OS_WINDOWS
	/* lowercase on windows */
	for (uint32 i = 0; i < strlen(str); i++) str[i] = tolower(str[i]);
#endif
}
```

src/GLideNHQ/TxCache.cpp
Rosalie241: I can't make it const, else I get the following error:
```
/home/rosalie/dev/GLideN64/src/GLideNHQ/TxHiResNoCache.cpp: In member function ‘virtual bool TxHiResNoCache::get(Checksum, GHQTexInfo*)’:
/home/rosalie/dev/GLideN64/src/GLideNHQ/TxHiResNoCache.cpp:101:61: error: invalid conversion from ‘const char*’ to ‘char*’ [-fpermissive]
  101 |         uint8_t* tex = TxHiResLoader::loadFileInfoTex(entry.fname, entry.siz, &width, &height, entry.fmt, &format);
      |                                                       ~~~~~~^~~~~
      |                                                             |
      |                                                             const char*
```

Rosalie241:
pushed & tested the changes on windows and linux, everything seems to work fine

src/GLideNHQ/TxHiResNoCache.cpp
```diff
@@ -0,0 +1,225 @@
...
+#include "Ext_TxFilter.h"
+#include <osal_files.h>
+
+TxHiResNoCache::TxHiResNoCache(int maxwidth,
+			   int maxheight,
+			   int maxbpp,
+			   int options,
+			   const wchar_t *cachePath,
+			   const wchar_t *texPackPath,
+        	   const wchar_t *fullTexPath,
```
gonetz, line 13: formatting

src/GLideNHQ/TxTexCache.cpp
```diff
@@ -36,7 +36,7 @@ TxTexCache::~TxTexCache()
 {
 }
 
-TxTexCache::TxTexCache(int options, int cachesize, const wchar_t *cachePath, const wchar_t *ident,
+TxTexCache::TxTexCache(int options, int cachesize, const wchar_t *cachePath, const wchar_t *fullTexPath, const wchar_t *ident,
```
gonetz, line 39: TxTexCache does not need fullTexPath

src/GLideNHQ/TxTexCache.h
```diff
@@ -36,7 +36,7 @@ class TxTexCache : public TxCache
 
 public:
   ~TxTexCache();
-  TxTexCache(int options, int cachesize, const wchar_t *cachePath, const wchar_t *ident, dispInfoFuncExt callback);
+  TxTexCache(int options, int cachesize, const wchar_t *cachePath, const wchar_t *fullTexPath, const wchar_t *ident, dispInfoFuncExt callback);
```
gonetz, line 39: TxTexCache does not need fullTexPath.
All changes in TxTexCache.h  and TxTexCache.cpp can be dropped.

gonetz (changes requested):
It looks way better imo, but some polishing is still needed.

src/GLideNHQ/TxHiResNoCache.cpp
Rosalie241: should be fixed now?

src/GLideNHQ/TxHiResCache.cpp
```diff
@@ -59,16 +52,10 @@ TxHiResCache::TxHiResCache(int maxwidth,
 						   const wchar_t *ident,
 						   dispInfoFuncExt callback)
 							 : TxCache((options & ~(GZ_TEXCACHE | FILE_TEXCACHE)), 0, cachePath, ident, callback)
-							 , _maxwidth(maxwidth)
-							 , _maxheight(maxheight)
-							 , _maxbpp(maxbpp)
+							 , TxHiResLoader(maxwidth, maxheight, maxbpp, options, callback)
```
gonetz, line 55: TxHiResLoader does not use callback.

src/GLideNHQ/TxHiResLoader.cpp
```diff
@@ -0,0 +1,586 @@
...
+
+/* use aggressive format assumption for quantization
+ * (0:disable, 1:enable, 2:extreme) */
+#define AGGRESSIVE_QUANTIZATION 1
+
+TxHiResLoader::TxHiResLoader(int maxwidth,
+			   int maxheight,
+			   int maxbpp,
+			   int options,
+			   dispInfoFuncExt callback)
```
gonetz, line 22: TxHiResLoader does not use callback, so this parameter is useless

src/GLideNHQ/TxHiResLoader.h
```diff
@@ -0,0 +1,51 @@
...
+	int _maxheight;
+	int _maxbpp;
+	int _options;
+
+public:
+	TxHiResLoader(int maxwidth,
+			   int maxheight,
+			   int maxbpp,
+			   int options,
+			   dispInfoFuncExt callback);
```
gonetz, line 41: TxHiResLoader does not use callback, so this parameter is useless

src/GLideNHQ/TxHiResNoCache.cpp
```diff
@@ -0,0 +1,225 @@
...
+			   const wchar_t *ident,
+			   dispInfoFuncExt callback)
+	: TxHiResLoader(maxwidth, maxheight, maxbpp, options, callback)
```
gonetz, line 16: TxHiResLoader does not use callback

gonetz (changes requested):
Almost there. Please remove redundant callback parameter for txHiResLoader.

Rosalie241:
Fixed & pushed, hopefully that was everything

gonetz (approved):
LGTM, thanks!

PR #2552, opened 2021-08-10
ScreenShot: move to common windows implementation

Rosalie241:
There's no real good reason to have `SaveScreenshot` be part of GLideNUI, it makes more sense to have a 'common' implementation.

the only thing TODO is jpeg support, however I'm uncertain if that's needed at all (if you want it added, I will implement it).


gonetz:
> There's no real good reason to have `SaveScreenshot` be part of GLideNUI

There is one reason: it is easy to support multiple screen shot formats with Qt.
However, I don't mind to drop jpeg support.
Please add a commit, which removes screen shot format from settings and GUI.

Rosalie241:
Pushed.

I'm not sure if you saw this but I finished https://github.com/gonetz/GLideN64/pull/2549 a few days ago, could you review that?

gonetz:
Thanks!

Issue #2556, opened 2021-08-22
Beetle  Adventure Racing. Fog is broken.


gonetz:
Regression. Broken in acfb569dea commit. @standard-two-simplex please check.

standard-two-simplex:
This [PR](https://github.com/gonetz/GLideN64/pull/2557) seems to fix, but halves the max resolution supported.

I do not understand why one implementation is working and the other isn't, so I haven't been able to find a better solution. Vertex X and Y end up being the same, so I fail to see why one is fogged and the other isn't.

gonetz:
Need to check the fog shader: ShadowMapFragmentShader 
May be it needs corrections.

standard-two-simplex:
A special shader is used to perform fog effects? Doesn't the N64 do fog effects as part of the blending stage?

gonetz:
BAR fog is special. It works as post-processing and uses depth buffer as the texture source.

standard-two-simplex:
Ah, I see.

standard-two-simplex:
I found out why this is failing. As you said, the gray rectangles use a special shader. The special vertex shader doesnt apply the changes in VertexShaderEnd() as it uses the special VertexShaderRectNoColor(), but the texture rectangle call updates the viewport as if it did. Thus, the texture rectangle is misplaced.

gonetz:
Thanks!
I'll think how to fix it.

PR #2557, opened 2021-08-23
Change the way screen size coordinates are normalized.

standard-two-simplex:
Halves supported maximum resolution but fixes fog issues #2556


gonetz:
Closed because another way to fix #2556 is found.

PR #2565, opened 2021-09-01
GLideNUI: allow loading default settings from GLideN64.default.ini

Rosalie241:
This allows a ''distribution'' of Project64 with GLideN64 (i.e something my installer installs) to ship a file named `GLideN64.default.ini` which'll be used whenever a user clicks on `Restore Defaults`.

### DIFF
--- a/src/GLideNUI/ConfigDialog.cpp
+++ b/src/GLideNUI/ConfigDialog.cpp
@@ -799,7 +799,7 @@ void ConfigDialog::on_buttonBox_clicked(QAbstractButton *button)
 		msgBox.setButtonText(QMessageBox::Cancel, tr("Cancel"));
 		if (msgBox.exec() == QMessageBox::RestoreDefaults) {
 			const u32 enableCustomSettings = config.generalEmulation.enableCustomSettings;
-			config.resetToDefaults();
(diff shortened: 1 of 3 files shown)


Rosalie241:
Build failure should be fixed when #2564 is merged.

gonetz:
> file named `GLideN64.default.ini` which'll be used whenever a user clicks on `Restore Defaults`.

Why it is necessary? What is wrong with config.resetToDefaults()?

Rosalie241:
> > file named `GLideN64.default.ini` which'll be used whenever a user clicks on `Restore Defaults`.
> 
> Why it is necessary? What is wrong with config.resetToDefaults()?

my installer installs Project64 with GLideN64, texture packs and a configuration file to make the entire setup automatic, some users accidentally click on `Restore Defaults` after experimenting with the settings themselves, causing the custom settings to be removed and texture pack(s) will be disabled then, which is annoying because then they ask me why the texture pack isn't loading, this PR makes it possible to let my installer ship a 'default' config file which is used whenever a user clicks on `Restore Defaults`, solving that issue for users and improving the user experience.

gonetz:
Ok. As I see, the old behavior is preserved if GLideN64.default.ini is not present.

PR #2566, opened 2021-09-05
GLideNHQ: remove unsupported dds code

Rosalie241:
GLideNHQ doesn't actually support loading DDS files so we shouldn't pretend like we do, this removes the dead code in `TxImage` and removes the `dds` bits in `TxHiResLoader`.


GhostlyDark:
The code for loading BMP textures is a leftover too, isn't it? GLideN64 doesn't dump textures as BMP and only loads PNG textures from my testing.

Rosalie241:
Technically it should be able to load BMP images, it's very picky about header etc though, see https://github.com/gonetz/GLideN64/blob/master/src/GLideNHQ/TxImage.cpp#L398

gonetz:
BMP textures support was added to properly support palettized CI textures, that is use HD textures with original N64 palettes.
It could  be especially useful for Paper Mario, where palette changes dynamically. Alas, this possibility is never used by texture artists, afaik.

PR #2568, opened 2021-09-05
Fix some shader compilation errors

fzurita:
* The device I was testing on doesn't support Anisotropic filtering
* GLSL 100 doesn't support texture, only texture2d
* GLSL 100 doesn't seem to support ```min``` and ```clamp``` with integers.

After making those fixes, it mostly works, it's extremely slow in my GLES 2.0 device though, about a 50% performance loss in addition to the other recent performance losses. The plugin is no longer playable even at lowest settings.

Also, there is some strange issues with lod:
goldeneye-002


gonetz:
merged: 21eb341b8d

PR #2571, opened 2021-09-11
TxHiResNoCache: add memory management

Rosalie241:
This patch adds some very basic form of memory managment to TxHiResNoCache, this is needed because otherwise project64 will eventually crash due to running out of memory.

### DIFF
--- a/src/GLideNHQ/TxHiResNoCache.cpp
+++ b/src/GLideNHQ/TxHiResNoCache.cpp
@@ -4,6 +4,8 @@
 #include "Ext_TxFilter.h"
 #include <osal_files.h>
 
+#define MAX_TEX_MEMORY 1073741824 /* 1 GiB */
+
 TxHiResNoCache::TxHiResNoCache(int maxwidth,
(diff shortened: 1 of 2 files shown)


gonetz (changes requested):
I suggest to use the same approach to memory management, as in Texures.h/.cpp
That is, use 
```
using Textures = std::list<GHQTexInfo>;
using Texture_Locations = std::unordered_map<uint64, Textures::iterator>;
Textures _loadedTex;
Texture_Locations _loadedTexLocations;
```

TxHiResNoCache::get should work similar to TextureCache::update,
that is find texture in the cache using _loadedTexLocations.
If texture is found, relocate it to the top of _loadedTex
_loadedTex.splice(_loadedTex.begin(), _loadedTex, iter);
otherwise create new texture on the top of the list.
If texture size limit is exceeded, remove textures from the back of _loadedTex list, as in TextureCache::_checkCacheSize()
I think it is faster and more natural than finding oldest texture.

PR #2572, opened 2021-09-11
Introduce txHiResUploadLimit 

Rosalie241:
This makes the maximum amount of textures loaded in `Textures.cpp` configurable by the user instead of hardcoding it.

on systems with little VRAM/RAM lowering this value may help to prevent crashes, it will however cause some stutter when we reach the limit (due to the constant removal of textures, however that seems to stabilize after a little while), but that's better than crashing.

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -104,6 +104,8 @@ void Config::resetToDefaults()
 	textureFilter.txEnhancedTextureFileStorage = 0;
 	textureFilter.txHiresTextureFileStorage = 0;
 	textureFilter.txNoTextureFileStorage = 0;
+	
+	textureFilter.txHiResUploadLimit = 2000;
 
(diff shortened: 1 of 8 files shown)


gonetz:
Thanks!
I have several notes, but it was faster to fix them than to write a review :)
My fixes:
* hi-res tex limit should be 0 by default.
* You placed control for txHiResUploadLimit setting to Hi Res tab.
It is right. However, your code removes any texture from the cache when the limit is exceeded.
It should be applied only for hi-res textures.
* You removed old check for texture cache size. It is wrong. This check is still necessary, especially when HD pack is not used or incomplete.

I uploaded your PR and my fix to branch https://github.com/gonetz/GLideN64/tree/Rosalie241-txMaximumAmountInVram
Please check my fixes. May be I missed smth.

Issue #2442, opened 2021-01-30
GlideN64 not working on second monitor (tv) when going into full screen move


kamikazilucas:
i want to have project 64 be in full screen on my tv but even when i put the p64 box on my tv it just goes full screen on the primary monitor anyway, how do i fix this

Jj0YzL5nvJ:
GLideN64, not Glide64. #1852

gonetz:
@kamikazilucas You may choose to duplicate image on your monitors or display it only on the second one (tv). In this case full screen works.

kamikazilucas:
> @kamikazilucas You may choose to duplicate image on your monitors or display it only on the second one (tv). In this case full screen works.

yes but that is super distracting to me

gonetz:
Set TV as the main display then:
![DisplaySetting](https://user-images.githubusercontent.com/8983092/106380359-4d916680-63e4-11eb-986c-1abeaf9cea8d.png)

It works for me. Sorry, I don't know what else can be done.

kamikazilucas:
> Set TV as the main display then:
> ![DisplaySetting](https://user-images.githubusercontent.com/8983092/106380359-4d916680-63e4-11eb-986c-1abeaf9cea8d.png)
> 
> It works for me. Sorry, I don't know what else can be done.

yeah but when i do that all my settings like desktop shortcuts go to the other screen, with jabo when you put the window on the other screen and then fullscreen it goes to the other screen anyway

gonetz:
I don't know how to do that with OpenGL. I'll try to find a solution, but can't promise anything.

kamikazilucas:
> I don't know how to do that with OpenGL. I'll try to find a solution, but can't promise anything.

ok thank you, another way is to have an option in graphics that lets you choose which monitor

Jj0YzL5nvJ:
Borderless fullscreen windowed, Non-exclusive fullscreen mode, Fake fullscreen, etc, call it whatever you want. 
https://www.pcgamingwiki.com/wiki/Glossary:Borderless_fullscreen_windowed

gonetz:
@Jj0YzL5nvJ Thanks! @kamikazilucas have you tried these workarounds? "Windowed Borderless Gaming" looks as a good solution.

kamikazilucas:
> @Jj0YzL5nvJ Thanks! @kamikazilucas have you tried these workarounds? "Windowed Borderless Gaming" looks as a good solution.

[...]

Issue #447, opened 2015-05-01
Fullscreen only uses primary monitor


MofoMan2000:
When switching to fullscreen, the plugin always begins rendering to the primary monitor (the one with the Start menu). There is no setting to change this, and it doesn't matter what monitor the emulator is on before switching to fullscreen.

Jabo's plugin goes to fullscreen on the monitor which the emulator window is currently on. This is probably the best behavior.

MofoMan2000:
Probably should mention, I'm using the Zilmar spec version of the plugin on PJ64 2.2.0.3. I have an NVIDIA GTX550Ti.

PR #2578, opened 2021-09-17
Add multi-monitor support

Rosalie241:
This adds the ability to choose a fullscreen monitor, this is handy when someone wants to make GLideN64 fullscreen on a different monitor than their primary one.

Fixes https://github.com/gonetz/GLideN64/issues/447
Fixes https://github.com/gonetz/GLideN64/issues/2442

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -23,6 +23,7 @@ void Config::resetToDefaults()
 #endif
 	video.fullscreenHeight = video.windowedHeight = 480;
 	video.fullscreenRefresh = 60;
+	video.fullscreenMonitor = 0;
 	video.fxaa = 0;
 	video.multisampling = 0;
 	video.maxMultiSampling = 0;
--- a/src/Config.h
+++ b/src/Config.h
@@ -25,6 +25,7 @@ struct Config
 		u32 fullscreen;
 		u32 windowedWidth, windowedHeight;
 		u32 fullscreenWidth, fullscreenHeight, fullscreenRefresh;
+		u32 fullscreenMonitor;
 		u32 fxaa;
 		u32 multisampling, maxMultiSampling;
 		u32 verticalSync;
--- a/src/GLideNUI/ConfigDialog.cpp
+++ b/src/GLideNUI/ConfigDialog.cpp
@@ -165,9 +165,16 @@ void ConfigDialog::_init(bool reInit, bool blockCustomSettings)
 	ui->overscanPalTopSpinBox->setValue(config.frameBufferEmulation.overscanPAL.top);
 	ui->overscanPalBottomSpinBox->setValue(config.frameBufferEmulation.overscanPAL.bottom);
 
+	QStringList fullscreenMonitorList;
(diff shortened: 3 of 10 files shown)


gonetz:
I tested that PR on my laptop connected to my main monitor.
I use "extend screens" mode.
In that case the plugin correctly detects that I have two monitors, and provides correct list of full-screen resolution for each of them. However, no matter which monitor I select in the config dialog, full-screen always opens on the primary one.

Rosalie241:
that shouldn't be possible, are you sure `config.video.fullscreenMonitor` is saving correctly?

Rosalie241:
Actually, can you try it again with this patch applied on top of this PR?

```patch
diff --git a/src/Graphics/OpenGLContext/windows/windows_DisplayWindow.cpp b/src/Graphics/OpenGLContext/windows/windows_DisplayWindow.cpp
index 71e3ac24..090ac38d 100644
--- a/src/Graphics/OpenGLContext/windows/windows_DisplayWindow.cpp
+++ b/src/Graphics/OpenGLContext/windows/windows_DisplayWindow.cpp
@@ -187,7 +187,7 @@ bool DisplayWindowWindows::_resizeWindow()
                m_heightOffset = 0;
                _setBufferSize();
 
-               return (SetWindowPos(hWnd, NULL, m_monitorInfo.rcWork.left, m_monitorInfo.rcWork.top, m_screenWidth, m_screenHeight, SWP_NOACTIVATE | SWP_NOZORDER | SWP_SHOWWINDOW) == TRUE);
+               return (SetWindowPos(hWnd, NULL, m_monitorInfo.rcMonitor.left, m_monitorInfo.rcMonitor.top, m_screenWidth, m_screenHeight, SWP_NOACTIVATE | SWP_NOZORDER | SWP_SHOWWINDOW) == TRUE);
        } else {
                m_screenWidth = m_width = config.video.windowedWidth;
                m_screenHeight = config.video.windowedHeight;
```

gonetz:
> are you sure `config.video.fullscreenMonitor` is saving correctly?

Yes. When I reopen the dialog, "Monitor" setting is correct.

> can you try it again with this patch applied on top of this PR?

Done. Nothing changed.

As I remember, creating GL context on non-primary display is a problem, at least for Windows. I tried to solve it but did not find a working solution.

gonetz:
I checked it again on my desktop with AMD CPU and GPU.
The result is the same - plugin correctly detects that I have two monitors, but full-screen always opens on the same display regardless of settings.

Rosalie241:
Hrm, I'm really not sure what causes that

gonetz:
Does it work on your current hardware?

Rosalie241:
I tested it on my sibling's laptop connected to a monitor a week ago (or so?) and it seemed to work fine, will re-test when I have the chance with my other sibling's laptop and a monitor and investigate the issues you've reported.

Rosalie241:
I'm gonna close this PR because I don't have the motivation to work on it anymore, sorry.

PR #2580, opened 2021-09-20
TextureCache: add better handling when at the cache size limit

Rosalie241:
This makes sure the texture that is removed is one which hasn't been used in the longest time.

### DIFF
--- a/src/Textures.cpp
+++ b/src/Textures.cpp
@@ -519,10 +519,29 @@ void TextureCache::destroy()
 void TextureCache::_checkCacheSize()
 {
 	if (m_textures.size() >= m_maxCacheSize) {
-		CachedTexture& clsTex = m_textures.back();
-		gfxContext.deleteTexture(clsTex.name);
-		m_lruTextureLocations.erase(clsTex.crc);
-		m_textures.pop_back();
+    	u64 oldestNum = 0;
+    	CachedTexture* oldestTex = nullptr;
+    	Textures::iterator oldestTexIt;
+    	for (auto it = m_textures.begin(); it != m_textures.end(); it++) {
+    	  if (oldestNum == 0 || it->counterNum < oldestNum) {
+    	    oldestNum = it->counterNum;
+    	    oldestTex = &(*it);
+    	    oldestTexIt = it;
+    	  }
+    	}
+
+    	// remove oldest texture when something was found
+    	if (oldestTex != nullptr) {
+    	  gfxContext.deleteTexture(oldestTex->name);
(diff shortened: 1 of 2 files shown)


gonetz:
I don't understand the point of these manipulations.
We already have this code in TextureCache::update

```
Texture_Locations::iterator locations_iter = m_lruTextureLocations.find(crc);
if (locations_iter != m_lruTextureLocations.end()) {
...
	m_textures.splice(m_textures.begin(), m_textures, iter);

```

That is most recently used textures will be on top of m_textures list. 
TextureCache::_checkCacheSize() removes textures from the back of the list, that is the least used textures.

Rosalie241:
Oh, didn't see that, closing

jackson1331:
Is their a way to change the texture size cache on Mupenfz(Android), as in the number of textures? My problem lies here, Im trying to load a Gaming Revived texture HTC file from gaming revived(downloadable from the discord) Banjo Tooie a 1.2gb HTC. The problem is that the zipped file cache stutters like frigging mad on initial load areas every time(the ram stores things for one session but they dissapear on shutdown), apparently this was fixed on the pc version of GlideN64 beta 5.0 project64, but on android there are still problems. 
Now i managed to extract the HTC file and create my own cache, the zipped file problems still exist, even using mupenfz as the host. However…. If I create the HTC cache as an UNZIPPED cache, textures load flawlessly. The problem lies in that my system eventually slows down and craps itself after 10 minutes. Creating an HTS has a similar effect, in that playing the game loading areas, things become smoother every time despite starting poorly. I can even quit Mupenfz, then on reopening things are still solid. As soon as shutdown/sleep though its back to square 1.

jackson1331:
Still had zero luck even with trying to add strings to the emulation.cfg using root.
0F863A3B38A5FC840038BB55362FE10F is the zip file

https://mega.nz/folder/KNs2DQRD#0BSj4gmo3RzP2oVAKJ_hgA

is the current link if anyone wishes to try it, however they only go up on fridays. 
Its a shame this works poorly on android, because as far as i know its the only complete banjo tooie hires texture pack their is. Creating an unzipped cache doesnt even crash mupenfz. The whole system just freezes forcing a reboot after a couple of minutes.

jackson1331:
Ok managed to get it to work on android, 
Ive converted to Retroarch Mupen Next Core for now, 
It loads the uncompressed cache without crashing,
& also supports Vulkan backend for me which is great.
Ive got a feeling the audio is a touch better aswell.
Only downside is it requires more CPU for the same performance, it is not crashing though.

Issue #1284, opened 2016-12-04
GLideN64 skips frames when running at high multiple of N64 resolution


oddMLan:
If you use 1x or 2x game looks smooth, no skipped frames.
When the internal resolution chosen is higher than the output window, it starts skipping frames.

retrobenny:
On Mupen64Plus FZ Edition (Android) with scale factor disabled (for highest quality?),I have been seeing choppier looking framerates again,even with color buffer disabled on legacy mode for top speed,but it runs too slow for me on Sync mode to guage if it also skips/drops frames.
Normally,I get slight drops with 16x Anisotropy,but not as bad as it is now.
Usually,both Sync mode and disabled color buffer are smooth while Async is known to cause framedrops whenever performance is challenged,the same spots Sync slows down in are usually where Async gets framedrops.

Maybe there should be an advanced option like frame-by-frame to prevent dropped/skipped frames by the plugin,if possible without horrible issues.

The behaviour matches Async mode when color buffer is disabled,it can run faster (90-120VI/s on fast forward) when the framedrops occur,defeating the purpose of having the headroom for a smoother framerate experience.

fzurita:
Async mode is supposed to be faster than sync mode. How is performance if you turn off completely color buffer to RDRAM?

retrobenny:
Async is faster,the faster speed is the cause of it dropping frames in Async mode while Sync mode doesn't allow frame drops but makes it much slower.
Turning it off made it faster AND smooth until this difference occurred.

fzurita:
So turning off copy computer buffer to RDRAM no longer makes it faster?

I wonder if the accuracy improvements in the plugin are slowing things down.

retrobenny:
Sorry,faster still,but no longer smooth output with it now being rather choppy while still showing 60/60fps on the fps counter.
It looks like skipped frames but it is not skipping them,it is dropping frames for whatever reason while still running at the same speed.

Still faster than Sync at least,it is either the same speed or slightly faster than the Async method.
Framedrops are different than slowdowns,its something in one end or the other discarding frames for some reason or something else ridiculous.

I do have Anisotropy on at 16x plus 6xBRZ De-posterized and fully Native scale instead of a 1x-4x scale along with legacy mode so it doesn't run at 1fps from the accuracy bloat of modern rendering alone.
Anything in multi-threading for video is likely the culprit of framedrops.

[...]

PR #2583, opened 2021-09-24
WindowsWGL: use adaptive vsync when supported

Rosalie241:
From https://www.khronos.org/opengl/wiki/Swap_Interval:

> if the renderer takes slightly longer than the v-blank intervals to render, say 18ms, then a different problem can result. It will effectively take two full v-blank intervals to display an image to the user, turning a 60fps program into a 30fps program.

I think this behavior is undesirable for end-users, so this patch makes `WindowsWGL` use adaptive vsync instead (when supported), which works in a different way:

> Adaptive vsync enables v-blank synchronisation when the frame rate is higher than the sync rate, but disables synchronisation when the frame rate drops below the sync rate. Disabling the synchronisation on low frame rates prevents the common problem where the frame rate syncs to a integer fraction of the screen's refresh rate in a complex scene. 

I personally feel like such behavior is more desirable, because before this patch, some users were complaining that sometimes their game would halve in speed for some time when vsync was enabled due to some slowdown (i.e vsync would force the game at 30 Hz instead of 60 Hz).

### DIFF
--- a/src/Graphics/OpenGLContext/windows/WindowsWGL.cpp
+++ b/src/Graphics/OpenGLContext/windows/WindowsWGL.cpp
@@ -102,7 +102,15 @@ bool WindowsWGL::start()
 
 		if (strstr(wglextensions, "WGL_EXT_swap_control") != nullptr) {
 			PFNWGLSWAPINTERVALEXTPROC wglSwapIntervalEXT = (PFNWGLSWAPINTERVALEXTPROC)wglGetProcAddress("wglSwapIntervalEXT");
-			wglSwapIntervalEXT(config.video.verticalSync);
+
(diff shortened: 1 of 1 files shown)


oddMLan:
This might actually solve #1284 #1534, this part sounds suspicious

> if the renderer takes slightly longer than the v-blank intervals to render, say 18ms, then a different problem can result. It will effectively take two full v-blank intervals to display an image to the user, turning a 60fps program into a 30fps program.

Because this is the effect it had on me while playing something like Mario Tennis using like 5x-6x resolution factor, the framerate would sometimes stutter/halve. I gotta try this later, see if it had any effect.

Issue #2547, opened 2021-08-05
Goldeneye Fog HLE issues


fzurita:
This was first reported here: https://github.com/mupen64plus-ae/mupen64plus-ae/issues/926

Mainly, fog is not working correctly in goldeneye in HLE, but it works in LLE.

HLE
![Screenshot_20210805-193755_M64Plus AE](https://user-images.githubusercontent.com/11581687/128434808-7c7f9dba-41fe-496c-a539-6afff38f04b4.jpg)

LLE
![Screenshot_20210805-193723_M64Plus AE](https://user-images.githubusercontent.com/11581687/128434826-4fcc7cbb-32b6-4822-a6f2-9d1897e81147.jpg)

gonetz:
It seems to be a problem with mobile version.
It looks like this on my PC with cdx4 hle:
![goldeneye-003](https://user-images.githubusercontent.com/8983092/131216726-f79543d4-be67-4d40-b88f-386efd01d7d0.jpg)

fzurita:
Interesting, I went through all the settings and I found nothing that made the issue go away.

Blending seems to not be working correctly for the mountain texture. I know that for mobile we are doing something special here.

@standard-two-simplex any ideas? Do you think the blending changes I did here with your help are breaking this?

https://github.com/gonetz/GLideN64/pull/2372

standard-two-simplex:
It's difficult to tell just by looking at the code.

Can you try forcing ordinary or legacy blending in `GraphicsDrawer::setBlendMode()`? If those are working, then it is a problem of the framebuffer fetch implementation.

Also, is it a regression or has it always been broken in android?

fzurita:
Yeah, you are probably right. It really sounds like a ucode implementation issue.

fzurita:
Glide64mk2 works correct, I'll compare the ucode implementation and see if there is anything significant there.

weinerschnitzel:
I'm also curious about this being a regression. Glide64 differences may also be difficult leads. Does pr4 show proper result on mobile? If that much is narrowed, it would be easier to bisect and pinpoint when the regression happened for mobile and have a solid lead.

fzurita:
Going back to ordinary blending doesn't fix it.

fzurita:
@gonetz Any clues? How is fog different between HLE and LLE in this instance?

Nejsplitter:
[...]

PR #2584, opened 2021-09-26
Remove CLIP control hack for GLES

fzurita:
It's not longer needed due to software vertex clipping.

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -394,11 +394,6 @@ CachedTexture * FrameBuffer::_getSubTexture(u32 _t)
 		copyHeight = m_pTexture->height - y0;
 
 	ObjectHandle readFBO = m_FBO;
-	if (Context::WeakBlitFramebuffer &&
-			m_pTexture->frameBufferTexture == CachedTexture::fbMultiSample) {
-		resolveMultisampledTexture(true);
-		readFBO = m_resolveFBO;
-	}
 
 	Context::BlitFramebuffersParams blitParams;
 	blitParams.readBuffer = readFBO;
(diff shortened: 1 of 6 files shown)


fzurita:
Also, disable software vertex clipping for all GLES versions.

fzurita:
I also did a separate commit to remove WeakBlitFramebuffer, I'm not sure I'm seeing the need for it. Do you remember why you added it @gonetz ?

Jj0YzL5nvJ:
https://github.com/gonetz/GLideN64/commit/f147927af92a018e5b9792464ca19db009f51dcf

gonetz:
> I also did a separate commit to remove WeakBlitFramebuffer, I'm not sure I'm seeing the need for it. Do you remember why you added it @gonetz ?

cite https://github.com/gonetz/GLideN64/issues/2547#issuecomment-927327508:
"Regarding WeakBlitFramebuffer. It is unrelated to this issue.
As I remember, GLES can't do blit frame buffer if source and destination buffers have different sizes.
Plugin works a bit less optimal with WeakBlitFramebuffer."

I would not recommend you to remove it, as it may cause crash on GLES devices.

gonetz:
I merged first two commits.

gonetz:
> I would not recommend you to remove it, as it may cause crash on GLES devices.

> as I know, GLES 3.0 does allow blitting using different size buffers.

It should be tested. If GLES 3.0 can do it, the condition for WeakBlitFramebuffer can be changed from 
`	return m_glInfo.isGLESX;
`
to
`	return !m_glInfo.isGLES2;
`
until GLES2 support is not completely dropped.

fzurita:
I checked the choice where WeakBlitFramebuffer is checked. It seems to be related to multisampling which is not exercised by GLES 2.0 as far as I know.

fzurita:
Also, I did try it in a GLES 3.2 device and I encountered no issues. I'll try it in a GLES 2.0 device just to make sure.

gonetz:
GLES 2.0 does not support glBlitFramebuffer
https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glBlitFramebuffer.xhtml

Cite:
GL_INVALID_OPERATION is generated if GL_SAMPLE_BUFFERS for the read buffer is greater than zero and the formats of draw and read buffers are not identical, or the source and destination rectangles are not defined with the same (X0, Y0) and (X1, Y1) bounds.

I don't quite understand that error. Does it mean
Error if ("GL_SAMPLE_BUFFERS for the read buffer is greater than zero" AND "the formats of draw and read buffers are not identical") OR ("the source and destination rectangles are not defined with the same (X0, Y0) and (X1, Y1) bounds")

or

Error if ("GL_SAMPLE_BUFFERS for the read buffer is greater than zero") AND ("the formats of draw and read buffers are not identical" OR "the source and destination rectangles are not defined with the same (X0, Y0) and (X1, Y1) bounds")

If the first is the case, then "the source and destination rectangles are not defined with the same (X0, Y0) and (X1, Y1) bounds" will cause GL_INVALID_OPERATION. WeakBlitFramebuffer is used exactly to avoid that situation. As I remember, WeakBlitFramebuffer was introduced because of crashes in glBlitFramebuffer when source and destination buffers have different sizes. Probably, it was a bug in drivers and now the situation is changed.

fzurita:
Ok, I see, it makes sense now. If multisampling is enabled, the source and destination buffers need to be the same size. At least that's how I interpret that error.

fzurita:
So based on the spec, we should keep that check around, but maybe document it a little better. I did not try enabling multisampling, but I have the feeling it will cause issues without that check.

PR #2607, opened 2021-11-23
Duplicated texture coordinates

standard-two-simplex:
Removes duplicated texture coordinate bounds and offsets.

DO NOT MERGE. Doesn't work with fast shaders. `ShaderFragmentCorrectTexCoords()` has to be split in two for it to work.

Besides, it should be possible to rasterize only one texture coordinates. `vTexCoord0` and `vTexCoord1` contain the same texture coordinates. They only differ after the texture engine applies tile specific shift, offset, clamp, mirror and wrap. Again, this only applies to accurate texture coordinates, so it may be tangled to implement atm.

Is the plan to keep the fast texture coordinates path forever or just until Public Release 5 is accomplished? It is a pain in the arse to write everything twice and in two different ways.


fzurita:
I hope we keep the fast path for as long as there are mobile devices. I do believe the @gonetz plans to remove GLES 2.0 support after the next public release.

gonetz:
> Is the plan to keep the fast texture coordinates path forever or just until Public Release 5 is accomplished? It is a pain in the arse to write everything twice and in two different ways.

>I hope we keep the fast path for as long as there are mobile devices.

Yes, I think that we should keep "fast path" code, but I don't plan to add new fixes to it. There are common parts between "fast" and "accurate". As soon as some of that part should be modified, current part goes to "fast" and modified to "accurate".

gonetz:
> As soon as some of that part should be modified, current part goes to "fast" and modified to "accurate".

Done in https://github.com/gonetz/GLideN64/tree/standard-two-simplex-duplicated_texture_coordinates branch,
commit 3b881ecc89

gonetz:
> Besides, it should be possible to rasterize only one texture coordinates. `vTexCoord0` and `vTexCoord1` contain the same texture coordinates. They only differ after the texture engine applies tile specific shift, offset, clamp, mirror and wrap. Again, this only applies to accurate texture coordinates, so it may be tangled to implement atm.

Done in 83840372d970

standard-two-simplex:
> Done in 8384037

Looks good to me.

gonetz:
Close this PR since it is merged as 4e0a4f700c.

PR #2613, opened 2021-11-28
cmake: Set OpenGL_GL_PREFERENCE

orbea:
Silences a cmake warning for Policy CMP0072 with `cmake-3.21.4`.

Please confirm this doesn't cause any problems for you, it seems to work as well as before here.
```
CMake Warning (dev) at /usr/share/cmake/Modules/FindOpenGL.cmake:315 (message):
  Policy CMP0072 is not set: FindOpenGL prefers GLVND by default when
  available.  Run "cmake --help-policy CMP0072" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  FindOpenGL found both a legacy GL library:

    OPENGL_gl_LIBRARY: /usr/lib64/libGL.so

  and GLVND libraries for OpenGL and GLX:

    OPENGL_opengl_LIBRARY: /usr/lib64/libOpenGL.so
    OPENGL_glx_LIBRARY: /usr/lib64/libGLX.so

  OpenGL_GL_PREFERENCE has not been set to "GLVND" or "LEGACY", so for
  compatibility with CMake 3.10 and below the legacy GL library will be used.
Call Stack (most recent call first):
  CMakeLists.txt:497 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
```
```
$ cmake --help-policy CMP0072
CMP0072
-------

``FindOpenGL`` prefers GLVND by default when available.

The ``FindOpenGL`` module provides an ``OpenGL::GL`` target and an
``OPENGL_LIBRARIES`` variable for projects to use for legacy GL interfaces.
When both a legacy GL library (e.g. ``libGL.so``) and GLVND libraries
for OpenGL and GLX (e.g. ``libOpenGL.so`` and ``libGLX.so``) are available,
the module must choose between them.  It documents an ``OpenGL_GL_PREFERENCE``
variable that can be used to specify an explicit preference.  When no such
preference is set, the module must choose a default preference.

CMake 3.11 and above prefer to choose GLVND libraries.  This policy provides
compatibility with projects that expect the legacy GL library to be used.

The ``OLD`` behavior for this policy is to set ``OpenGL_GL_PREFERENCE`` to
``LEGACY``.  The ``NEW`` behavior for this policy is to set
``OpenGL_GL_PREFERENCE`` to ``GLVND``.

This policy was introduced in CMake version 3.11.  CMake version
3.21.4 warns when the policy is not set and uses ``OLD`` behavior.
Use the ``cmake_policy()`` command to set it to ``OLD`` or ``NEW``
explicitly.

.. note::
  The ``OLD`` behavior of a policy is
  ``deprecated by definition``
  and may be removed in a future version of CMake.
```


gonetz:
Merge this one since all checks have passed. @fzurita please check Android build after all these cmake-related modifications.

fzurita:
Everything still builds fine in Android.

gonetz:
Thanks!

PR #2618, opened 2021-12-06
Correct mip-map shader for the case when texture level is 0 and detai…

gonetz:
…led texture enabled


gonetz:
@standard-two-simplex please check.

standard-two-simplex:
It seems it is working fine.

gonetz:
> What was the reason for this? I believe uMaxTile should be gSP.texture.level always.

I agree. It is probably a rudiment from the older mipmap implementation.

gonetz:
Fix updated. Previous variant caused issues with Zelda MM HD pack.

PR #2623, opened 2021-12-17
Fix build with CMake on Windows

fzurita:
What the title says.

It appears that "env" is required when using CMake in windows to declare an environment variable.


gonetz:
@orbea please review this PR.

orbea:
@gonetz Sorry for the slow reply. Frankly I haven't used windows in years, but if it works there it seems fine. env(1) is part of the posix spec and its used correctly.

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/env.html

Issue #2628, opened 2022-01-02
Indiana Jones crash


fzurita:
I'm often getting a crash report when Indiana Jones is running, I don't have a reproducible case, but I think it's good to document what we have so far. The crash is a segmentation fault and it happens here:

https://github.com/gonetz/GLideN64/blob/8343fd05c9a01ce2d73f7e32bb8355036759a05b/src/uCodes/F5Indi_Naboo.cpp#L213

Stack trace:
```
********** Crash dump: **********
#00 0x000000000019801e     lib/arm/libmupen64plus-video-GLideN64.so
    F5INDI_Lighting_Overlay1(unsigned int, unsigned int)
    ./src/uCodes/F5Indi_Naboo.cpp:213:19
    F5INDI_MoveMem(unsigned int, unsigned int)
    ./src/uCodes/F5Indi_Naboo.cpp:577:6
#00 0x0000000000130e39     lib/arm/libmupen64plus-video-GLideN64.so (RSP_ProcessDList()+704)
    _ProcessDListFactor5()
    ./src/RSP.cpp:92:3
    RSP_ProcessDList()
    ./src/RSP.cpp:179:3
#00 0x0000000000006a60     lib/arm/libmupen64plus-rsp-hle.so
#00 0x0000000000009e58     lib/arm/libmupen64plus-rsp-hle.so (DoRspCycles+3760)
#00 0x0000000000091c64     lib/arm/libmupen64plus-core.so
#00 0x00000000000644d0     lib/arm/libmupen64plus-core.so
#00 0x000000000008b95c     lib/arm/libmupen64plus-core.so
#00 0x00000000000b11ca [anon:.bss]
```

fzurita:
DMEM size is this is the mupen64plus core:
```
enum { SP_MEM_SIZE = 0x2000 };
```

So we could probably just bound the pointer to that size. Not sure what to do when this happens though? Just ignore the command?

fzurita:
@loganmc10 You may know this, how should out of bounds reads be handled? What should they return?

loganmc10:
As far as I remember, if it's a DMA type access, the out of bounds portion would return all 0's. If some of the address space was valid, that data would be returned, just the out of bounds portion would be 0.

If it's CPU access (like a word or dword), then it does this odd thing based on the address being accessed, see https://github.com/mupen64plus/mupen64plus-core/blob/6860d134afd16f40262fe335cfee04967265fbac/src/device/device.c#L39

[...]

PR #2631, opened 2022-01-07
Apply mask on DMEM address to avoid segmentation fault in Indiana Jones,

gonetz:
see #2628


gonetz:
@fzurita I can't get the segmentation fault reported in #2628. Without it I don't know where to dig.
So, I just added a mask to DMEM address to avoid out of bounds read in that particular place.
So, user will get a wrong color instead of the crash.
The plugin also will show an error message when this situation will happen for a first time.
I hope that somebody will give us more details about that problem. Does it look ok to you?

fzurita:
Yeah, that looks reasonable to me, with that mask it should be impossible to go above 0x2000 address offset. And hopefully users report the issue.

gonetz:
TMEM, IMEM, DMEM have size of 4kb. It is unclear to me, why DMEM have 0x2000 limit in mupen64plus.

PR #2632, opened 2022-01-08
Fix texrects in native resolution for fast shader path

fzurita:
My understanding here is limited, so this solution is definitely up for discussion. See here:

https://github.com/gonetz/GLideN64/issues/2506

The real fix in this is always setting the ```vertexOffset``` to zero in the fast path.

The other change is something I noticed while in this file that is in the accurate path but appears to be missing from the fast path.

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactoryFast.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramUniformFactoryFast.cpp
@@ -30,13 +30,6 @@ public:
 		const bool isNativeRes = config.frameBufferEmulation.nativeResFactor == 1 && config.video.multisampling == 0;
 		const bool isTexRect = dwnd().getDrawer().getDrawingState() == DrawingState::TexRect;
 		const bool useTexCoordBounds = isTexRect && !isNativeRes && config.graphics2D.enableTexCoordBounds;
-		/* At rasterization stage, the N64 places samples on the top left of the fragment while OpenGL		*/
-		/* places them in the fragment center. As a result, a normal approach results in shifted texture	*/
-		/* coordinates. In native resolution, this difference can be negated by shifting vertices by 0.5.	*/
-		/* In higher resolutions, there	are more samples than the game intends, so shifting is not very		*/
-		/* effective. Still, an heuristic is applied to render texture rectangles as correctly as possible  */
-		/* in higher resolutions too. See issue #2324 for details. 											*/
-		const float vertexOffset = isNativeRes ? 0.5f : 0.0f;
 		float texCoordOffset[2][2] = { 0.0f, 0.0f };
 		if (isTexRect && !isNativeRes) {
 			float scale[2] = { 0.0f, 0.0f };
@@ -109,7 +102,7 @@ public:
 			}
 		}
 
-		uVertexOffset.set(vertexOffset, vertexOffset, _force);
+		uVertexOffset.set(0.0, 0.0, _force);
 		uTexCoordOffset[0].set(texCoordOffset[0][0], texCoordOffset[0][1], _force);
 		uTexCoordOffset[1].set(texCoordOffset[1][0], texCoordOffset[1][1], _force);
 		uUseTexCoordBounds.set(useTexCoordBounds ? 1 : 0, _force);


fzurita:
@standard-two-simplex Could you please look at this?

fzurita:
@standard-two-simplex to answer your question here:

https://github.com/gonetz/GLideN64/issues/2506#issuecomment-1008160608

Zelda in-game menus look correct with the ```vertexOffset``` set to zero in the fast path.

With ```vertexOffset``` set to 0.5, zelda HUD and in-game menus look blurry.

Ogre battle 64 dialogs look wrong with ```vertexOffset``` set to zero, it adds some gaps that don't belong.

fzurita:
Ok, this produces the correct result for both situations:

```const float vertexOffset = isNativeRes && gDP.otherMode.textureFilter == 0 ? 0.5f : 0.0f;```

**Comment by standard-two-simplex** [comment]
Maybe you can try to adjust the texture filter. 
I think it should go,
```
"  mediump vec2 offset = fract(texCoord*texSize);			\n"
```
to compute the correct fractional part.

And 
```
"#define TEX_OFFSET(off) texture2D(tex, texCoord - (off + 0.5)/texSize)			\n"
```
to make sure that `texture2D()` is called on a normalized half integer texture coordinate.

I believe this changes will work with the vertex offset.

**Comment by fzurita** [comment]
I'm a little confused, are you saying that I should do that in addition to 
```
const float vertexOffset = isNativeRes && gDP.otherMode.textureFilter == 0 ? 0.5f : 0.0f;
```
or that the above is wrong and I should do what you suggest in the shader instead?

**Comment by standard-two-simplex** [comment]
You should do the above changes to the texture filter and a vertex offset of 0.5 always (in native) for the correct behaviour.

**Comment by fzurita** [comment]
Ok, I did that. Legend of Zelda and Ogre battle 64 still look ok. Unfortunately, Mario Kart 64 got broken by that.

Did I do it as you expected?

https://github.com/mupen64plus-ae/mupen64plus-ae/commit/7743923c64d1c08667b00136b8e2367f2673de34

**Comment by standard-two-simplex** [comment]
I realize it should be
```
"#define TEX_OFFSET(off) texture2D(tex, texCoord - (off + vec2(0.5))/texSize)			\n"
```
but unless you have shader compilation errors, it shouldn't matter. 

Zelda is fixed but Mario Kart isn't? You're testing in native I guess?

**Comment by fzurita** [comment]
I am testing in native, I'll try that.

**Comment by fzurita** [comment]
Everything looks correct after that change.

**Comment by gonetz** [comment]
> Everything looks correct after that change.

Unfortunately, it does not. Mario Kart is not completely fixed in native, copyright looks wrong:
MK_(C)

What is much worse, non-native resolutions now look completely wrong even with native-res texrects enabled. It was correct before the fix.

**Comment by fzurita** [comment]
Ok, I've been looking at a low resolution screen, see that is hard to tell for me.

How does it look with my original change?

```
const float vertexOffset = isNativeRes && gDP.otherMode.textureFilter == 0 ? 0.5f : 0.0f;
```

**Comment by gonetz** [comment]
MK64 title screen, fast path
* `const float vertexOffset = 0.0f`
No issues with native res, no additional issues in hi-res. The screen looks good in non-native with native-res texrects enabled.

* `const float vertexOffset = isNativeRes && gDP.otherMode.textureFilter == 0 ? 0.5f : 0.0f;`
Copyright looks incorrect in native res.

It seems that the vertexOffset is not applicable for the fast path.

**Comment by fzurita** [comment]
Ogre battle 64 dialogs look wrong with vertexOffset set to zero, it adds some gaps that don't belong. With the offset set to 0.5, they look correct.

**Comment by gonetz** [comment]
Yes, OB64 is a special case. It works correct with accurate path, so I think we can ignore its issues with fast path.

**Comment by standard-two-simplex** [comment]
> It seems that the vertexOffset is not applicable for the fast path.

I have been checking the situation a bit. I had the same issue with the accurate path.

The game generates integer texture coordinates. For textures using nearest neighbour filtering mode, if the texture coordinates were slightly smaller, an incorrect texel would be fetched. While approximating this behaviour with OpenGL, some precision errors are made. In order to avoid this, I introduced a small bias in `ShaderFragmentCorrectTexCoords()`.
```
			" highp vec2 mTexCoord0 = vTexCoord0 + vec2(0.0001);						\n"
			" highp vec2 mTexCoord1 = vTexCoord1 + vec2(0.0001);						\n"
```

This prevents the previous precision issue, while it is very unlikely that a texture coordinate intended to have fractional part 0.9999 overflows to the next integer because of this. 

However, in the fast path all the code in `ShaderFragmentCorrectTexCoord()` is being completely ignored. Which means neither the small bias, nor the texture offsets passed via `uTextureOffset()` are being used at all. On top of this, I realize that the values computed for `uTextureOffset()` are not being normalized, so those offsets should be divided by texture size. Otherwise, they could be computed exactly as in the accurate path.

From my understanding, the inaccurate path doesn't emulate clamp wrap and mirror correctly on texture edges (remember the floor on the graveyard in MM) or the per pixel shift and offset. This should be edge cases, so I believe that most of the time it is possible to achieve an accurate image.

gonetz:
> From my understanding, the inaccurate path doesn't emulate clamp wrap and mirror correctly on texture edges (remember the floor on the graveyard in MM) or the per pixel shift and offset.

Yes. Fast path does less per-pixel calculations, that's why it is fast :)

> This should be edge cases, so I believe that most of the time it is possible to achieve an accurate image.

Could you make a proper fix then?

standard-two-simplex:
> Could you make a proper fix then?

I will try. However, it might be a bit complicated to get it all right. The accurate path had lots of regressions that I had to debug and fix.

fzurita:
It may not be worth it, the accurate path is the path forward, so any fixes in the fast path have a limited life span. We could just leave the fast path with the bugs and accept that it's not as good.

Issue #2108, opened 2019-10-22
 PowerVR ge8320 (mediatek Helio p22) corrupted graphics


fzurita:
There seem to be some issues with the GPU drivers for PowerVR ge8320 on a  Realme C2. See the attached screenshot.

![65404379-cc4abc80-de0a-11e9-9eb9-9a6023f9a93e](https://user-images.githubusercontent.com/11581687/67296140-0f558980-f4b6-11e9-891c-cf25f009bcba.gif)

This will be hard to debug without having the device, the best hope is that the phone gets an update with a non-bugged GPU driver.

fzurita:
@Panderner In Mupen64plus FZ, go to settings, then display. Scroll down to the "Flicker Reduction" option. Please try all the options it gives you and see if any of them make the issue go away.

Panderner:
Yeah, I tried all flicker reduction profiles but it didn't work

fzurita:
This will be hard to solve without having the device unfortunately.

Panderner:
And it's happened for other devices like Samsung Galaxy a10s, redmi 6, Xperia l3, lg k50s and vivo y91i

Panderner:
Enabling Fragment based depth and N64 depth compare to fix corruptions

fzurita:
Do you have those other devices? How do you know that they have the exact same problem?

Panderner:
I don't have any devices but it's corrupted for PowerVR ge8320 devices including mediatek Helio P22, a22 and p35 on most games

Panderner:
Enabling N64 depth compare to fix corruptions but it has problems for some games like tiles in FIFA 99 and Conker's bad fur day shadows

Panderner:
> This will be hard to solve without having the device unfortunately.

Have you tried other PowerVR GE8320 Devices?

fzurita:
I don't have any to try it on.

Panderner:
In glideN64-medium there are almost no polygons are drawn.

Tested 5 Games:
Super Mario 64
![Screenshot_2020-02-02-12-51-15-39_9784a853058524697766614cb28dab3d](https://user-images.githubusercontent.com/37503397/73603318-e0be7300-45bb-11ea-9ef0-974e598f1857.png)
The Legend of Zelda Ocarina of Time
[...]

PR #2633, opened 2022-01-08
Workaround for PowerVR issues

fzurita:
Newer PowerVR devices need depth to be cleared every frame, otherwise most geometry is rendered behind the background.

It also needs fragment based depth to be disabled, otherwise geometry has various depth glitches.

I have not been able to figure where the driver is going wrong with either of these issues. For the 2nd issue, it almost looks like the driver is discarding fragments when they should not be discarded.

Another tidbit, if fragment based depth is enabled, then we don't have to force depth buffer clear, but at the same time, there are randomly geometry depth issues.

@gonetz What games do you think will break when we force depth buffer clear? Also, what games break without fragment based depth? I know that Pilotwings and RE2 are two of them.

See this issue for an idea of the glitches: https://github.com/gonetz/GLideN64/issues/2108

I'm fixing them now because I was given an affected piece of hardware.


fzurita:
If I put a ```glFinish``` right after this line:
https://github.com/gonetz/GLideN64/blob/8343fd05c9a01ce2d73f7e32bb8355036759a05b/src/Graphics/OpenGLContext/opengl_ContextImpl.cpp#L213

It fixes the first issue at a 30% performance loss. Given that a ```glFinish``` fixes the issue, it tells me that it's most likely a driver bug.

fzurita:
A ```glFlush``` also seems to fix it and the performance penalty is not nearly as bad.

fzurita:
From a lot of debugging, I have been able to find that newer PowerVR devices have problems reliably reading from ```gl_FragCoord.z```. It almost seems like the shader does not have the latest depth data in the framebuffer set by ```GraphicsDrawer::_updateStates```.

PR #2636, opened 2022-01-17
README Formatting

ElectronicsArchiver:


oddMLan:
Extraneous html breaks, emphasis on random words and questionable rewording in some places

> A next generation,* ***N64*** *emulator*, ***Graphics Plugin** 

does not sound right.

ElectronicsArchiver:
That isn't `html` though, it's **Markdown**,
while it's possible to use html in markdown,
that type of formatting is a standard asset
of Githubs toolbox.

---

>  emphasis on random words

I don't see how topic / context indicating 
words can be considered random.

CI -> Topic
features, fixes, bugs, imcomplete translations -> Subjects of CI

Workflow / Release -> Place of download origin

32-bit / 64-bit -> The users choice

etc.

Even if they were random, they'd
be better than no emphasis at all
considering that while reading text,
landmarks give the reader a reference
to where they currently are in the
context of the content.

---

I will adjust the headline.

dankcushions:
> *A next generation* ***Graphics Plugin*** *for the* ***N64*** *emulator.*
still doesn't sound right. what's 'the' n64 emulator?

i would say:
> *A next generation* ***Graphics Plugin*** *for* ***N64*** *emulators.*

ElectronicsArchiver:
Sure~

gonetz:
Ok, lets give it a try. @ElectronicsArchiver, thanks!
@oddMLan @dankcushions thanks for review and corrections.

Issue #2617, opened 2021-12-06
RE2 bottom of the screen is being cut off


fzurita:
This seems to be relatively new. With RE2, the bottom of the screen is getting cut off:

![Resident Evil 2 (U) (V1 1)  !](https://user-images.githubusercontent.com/11581687/144783436-e7f2ce34-bfc8-4cbb-a76b-3a51b466a17c.png)

standard-two-simplex:
Bisected to 865e9b443e665730c3d24005b2c71acacc59ce2d

gonetz:
> Bisected to [865e9b4](https://github.com/gonetz/GLideN64/commit/865e9b443e665730c3d24005b2c71acacc59ce2d)

Any idea, what is wrong?

standard-two-simplex:
The hack condition is no longer triggered after I changed the commented lines for starcraft. I changed it because the other one was working, I do not know what the correct behaviour is.
Current master
```
                //f32 imageW = (f32)(_pObjScaleBg->imageW >> 2);
		//f32 imageH = (f32)(_pObjScaleBg->imageH >> 2);
		f32 imageW = (f32)gSP.bgImage.width;
		f32 imageH = (f32)gSP.bgImage.height;
```

gonetz:
The correct behavior is to load data from uObjScaleBg structure.
However, for all games but RE2 result should be the same:
```
	f32 imageW = (f32)(_pObjScaleBg->imageW >> 2);
	f32 imageH = (f32)(_pObjScaleBg->imageH >> 2);

```
vs
```
	const u32 imageW = _pBgInfo->imageW >> 2;
	const u32 imageH = _pBgInfo->imageH >> 2;
	gSP.bgImage.width = imageW - imageW % 2;
	gSP.bgImage.height = imageH - imageH % 2;

```

The only difference is that gSP.bgImage.width and gSP.bgImage.height are forced to be even.
Can you try this fix:
```
		f32 imageW = (f32)((_pObjScaleBg->imageW >> 2) & 0xFFFFFFFE);
		f32 imageH = (f32)((_pObjScaleBg->imageH >> 2) & 0xFFFFFFFE);

```
?
It makes width and height even. It may be wrong too if some game uses an image with odd sizes.

I also noticed a regression in C&C with one-piece bg:
![GLideN64_Command Conquer_000](https://user-images.githubusercontent.com/8983092/145246765-ec31fa46-408d-4ec9-91aa-5d00ffccc80a.jpg)

[...]

PR #2638, opened 2022-01-17
Do not apply shift scale to background textures.

standard-two-simplex:
Fixes issue with Command & Conquer described in #2617


gonetz:
Thanks!

Issue #2569, opened 2021-09-09
[Feature Request] Option to respect texture format and size


GhostlyDark:
Same texture, different texture format. First dump is a water bubble, second one is a blood effect during the fight against Barinade. Current behaviour retextures both instances with one and the same HD texture.

I'd like to suggest a checkbox to control whether or not texture format and size information is ignored. Current behaviour should remain the default as it would otherwise break tons of texture packs. Such an option can fix instances like these and can help in uncovering wrong texture names.

**Current behaviour:**
![image](https://user-images.githubusercontent.com/31742919/132851106-4e8c5632-acb3-4135-af6e-219c3a963789.png)

**What is should look like:**
![image](https://user-images.githubusercontent.com/31742919/132851128-7da2bdd3-8b1f-425b-bdb8-ee1ee1fd55a2.png)

**Bubble texture (the one with more transparency):**
![image](https://user-images.githubusercontent.com/31742919/132851175-6446e3f5-6361-4171-a40d-7c53ee0efd15.png)

**Vanilla textures:**
THE LEGEND OF ZELDA#AE749099#3#1_all.png
![THE LEGEND OF ZELDA#AE749099#3#1_all](https://user-images.githubusercontent.com/31742919/132669511-d43ffbf2-7566-403b-af0e-57c4dad38a8a.png)

THE LEGEND OF ZELDA#AE749099#4#1_all.png
![THE LEGEND OF ZELDA#AE749099#4#1_all](https://user-images.githubusercontent.com/31742919/132669527-8a4d60e6-b4cc-46ea-8717-cb3e2c88e71f.png)

**Ocarina of Time 1.0 (USA) Savestates:**
[bubble.pj.zip](https://github.com/gonetz/GLideN64/files/7143599/bubble.pj.zip)
[blood.pj.zip](https://github.com/gonetz/GLideN64/files/7143600/blood.pj.zip)

**HD textures:**
[bubble-textures.zip](https://github.com/gonetz/GLideN64/files/7143602/bubble-textures.zip)

gonetz:
Thanks for detailed FR!

>I'd like to suggest a checkbox to control whether or not texture format and size information is ignored.

The texture size and format information is not ignored.
As you noticed: "Same texture, different texture formats".
The texture is indeed the same, with the same checksum.
Current tex cache code distinguishes textures only by checksum.
Textures with the same checksum treated as duplicates, duplicate is removed.
If you use only one of these two textures, it loads correctly.
[...]

PR #2640, opened 2022-01-22
Add support for HD textures, which have the same checksum but differrent format

gonetz:
See #2569 for details.


gonetz:
New version, which compares textures by checksum, format and size.

GhostlyDark:
Works as intended. Both texture format and size are respected.

PR #2643, opened 2022-01-25
GLideNUI: copy config files to user config path when needed

Rosalie241:
I've started packaging RMG for linux (it's in the AUR & flatpak) but GLideN64 can't save settings in the current codebase because `-DMUPENPLUSAPI_GLIDENUI=ON` is enabled, to fix this, this patch checks if the plugin path is writeable, if it isn't, copy the config files over to the user config directory & use that path instead for the configuration files.

Another thing this patch fixes is the uneeded call to `_cutLastPathSeparator` in `_getWSPath` for some mupen64plus functions which don't need it (like `ConfigGetUserDataPath()`, `ConfigGetUserCachePath()`, `ConfigGetUserConfigPath()`), because RMG doesn't end those paths with a trailing slash, causing GLideN64 to use the wrong path without this fix.


gonetz (approved):
LGTM

PR #2645, opened 2022-01-31
Add inaccurate texture coordinates hotkey

Rosalie241:
Requested by @GhostlyDark because Nerrel's Majora's Mask HD requires changes to look correct without the "inaccurate texture coordinates" option enabled, to help with this, a hotkey is added by this patch which makes seeing the changes & comparing them faster and easier.


gonetz:
LGTM!

PR #2650, opened 2022-02-12
Tex load refactor

gonetz:
Refactor texture load: apply proper mask to tmem read address to prevent out of bounds reads.


GhostlyDark:
Ocarina of Time and AeroFighters Assault look fine. GoldenEye still shows black textures.

PR #2658, opened 2022-02-19
Detect the Angle renderer

fzurita:
Also, in Android disable EGL image if using the angle renderer.

In the windows 11 subsystem for Android, EGL image does not seem to work.


gonetz (approved):
LGTM

PR #2667, opened 2022-03-06
GLideNUI: allow resizing

Rosalie241:
This patch allows the config window to be resized freely by the user instead of it being a fixed size.

### DIFF
--- a/src/GLideNUI/configDialog.ui
+++ b/src/GLideNUI/configDialog.ui
@@ -9,8 +9,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>747</width>
-    <height>715</height>
+    <width>597</width>
+    <height>612</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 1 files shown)


gonetz:
> allows the config window to be resized freely

Why? Which problem does it fix?

Rosalie241:
Some users in the RMG discord complained the gliden64 settings window was too big for their screen, so it didn't fully fit, this patch allows resizing so that should make it fit.

oddMLan:
> Some users in the RMG discord complained the gliden64 settings window was too big for their screen, so it didn't fully fit, this patch allows resizing so that should make it fit.

Can confirm the GLideNUI config doesn't fit fully in my laptop screen.

Jj0YzL5nvJ:
Yep, old 16:8 displays over here =P

gonetz:
> Some users in the RMG discord complained the gliden64 settings window was too big for their screen, so it didn't fully fit, this patch allows resizing so that should make it fit.

Ok, thanks.

PR #2670, opened 2022-03-13
Fix #2663: Transparent textures in Densha De Go

### LINKED ISSUE #2663: Transparent textures in Densha De Go
Out of the box, running Densha De Go in Mupen64Plus-Next in Retroarch with Gliden64 plugin, almost all of the textures in the game are transparent. Disabling legacy blending seems to fix this.

Can we add `generalEmulation\enableLegacyBlending=0` to GLideN64.custom.ini for this game? I'm happy to open a PR if this is the right approach.

danepowell:
I'm not sure that these are the right ROM names. I understand that for Japanese roms, the CRC is used as the name. But when I checked all of the other Japanese ROM names in this file, none of them matched any official CRCs that I could find in an actual ROM or db.

Additionally, a full CRC is 8 bytes, but the names in the ini file are 4 byte strings. I'm not sure where those 4 bytes come from?

### DIFF
--- a/ini/GLideN64.custom.ini
+++ b/ini/GLideN64.custom.ini
@@ -19,6 +19,14 @@ frameBufferEmulation\copyFromRDRAM=1
 frameBufferEmulation\copyToRDRAM=0
 frameBufferEmulation\copyDepthToRDRAM=0
 
+[362D06B6]
+Good_Name=Densha de Go! 64 (J)
+generalEmulation\enableLegacyBlending=0
+
+[68D128AE]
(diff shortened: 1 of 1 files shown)


gonetz:
> I'm not sure that these are the right ROM names. I understand that for Japanese roms, the CRC is used as the name. But when I checked all of the other Japanese ROM names in this file, none of them matched any official CRCs that I could find in an actual ROM or db.

Japanese ROM name can't be used directly because Japanese symbols are obviously not ASCII, and reading such symbols is locale-dependent. Thus, plugin calculates Adler32 checksum of ROM name bytes. This checksum is the rom name.

Issue #2647, opened 2022-02-01
Lens of truth depth issues


fzurita:
It seems like the lens of truth effect in Majora's mask is broken in some situations. Here is an example:
![Legend of Zelda, The - Majora's Mask (U)  !](https://user-images.githubusercontent.com/11581687/151992846-9a79fa47-2edc-4b53-b38b-e74e09738670.png)

The ice is rendering in front of the ghost.

Here is a save state for the USA version of the game for Mupen64plus.
[Legend of Zelda, The - Majora's Mask (U) [!].zip](https://github.com/gonetz/GLideN64/files/7979399/Legend.of.Zelda.The.-.Majora.s.Mask.U.zip)

Jj0YzL5nvJ:
#2074

fzurita:
Yeah, there are some elements in the environment have are rendering in front of everything else when the lens of truth is enabled. Depth is doing something weird...

gonetz:
I tried to fix #2074, but did not find the source of  those problems. I'll again if I have time.

gonetz:
I've advanced with that problem. I need more saves to test.

gonetz:
I opened a PR with the fix: #2674
@fzurita could you test it on Android devices?

fzurita:
Sure, will do.

gonetz:
PR updated.

Jj0YzL5nvJ:
Small issue with the shadows outside the area of effect.
https://imgur.com/a/shTf6uW

gonetz:
> Small issue with the shadows outside the area of effect.

PR with the fix is updated, #2674

Issue #2074, opened 2019-07-27
Majora's Mask Lens of Truth Depth Issue


Pingler:
The Lens of Truth (LoT) effect breaks visuals in GlideN64, some objects become rendered at the wrong distance. This issue occurs the same way in both HLE and LLE mode, and no combination of settings fixes it.

Fairy Bubbles:
Correct behavior on hardware:
![LoT Real N64](https://user-images.githubusercontent.com/47263169/61998795-6ef0f400-b083-11e9-8214-abd611d89feb.jpg)

GlideN64 behavior:
![LoT GlideN HLE 1](https://user-images.githubusercontent.com/47263169/61998812-9e076580-b083-11e9-9f70-9a06c2588369.png)
Fairies render in front of terrain and bridges

Correct behavior in Angrylion RDP Plus (also console verified):
![LoT AL RDP Plus 1](https://user-images.githubusercontent.com/47263169/61998857-33a2f500-b084-11e9-9870-77c9ef2db430.png)
![LoT AL RDP Plus 2](https://user-images.githubusercontent.com/47263169/61998861-37367c00-b084-11e9-85ff-307ab8e6b714.png)
LoT off, Fairy Visible, LoT on, hidden wall blocks fairy

GlideN64 behavior:
![LoT GlideN LLE](https://user-images.githubusercontent.com/47263169/61998932-14f12e00-b085-11e9-998a-cf2977b87036.png)
LoT off, same as AL+, LoT on, fairy renders in front of hidden wall

Ice:
GlideN64 behavior:
![LoT GlideN HLE 2](https://user-images.githubusercontent.com/47263169/61998961-c3956e80-b085-11e9-9d85-c05036ef38c6.png)
![LoT GlideN HLE 3](https://user-images.githubusercontent.com/47263169/61998964-c728f580-b085-11e9-8722-67125adb5dc5.png)
Correct behavior would be that the ice block shouldn't be seen in both cases as they are obscured by Link and the terrain.

Tested on console (PAL), Project64 Nightlies, with GlideN64 3.0, 4.0, and HD texture fix 7, AL RDP Plus (all 3 speed presets) built from source recently (all emulation NTSC-U)

gonetz:
Could you give me save states to test it?

Pingler:
[ZELDA MAJORA'S MASK.zip](https://github.com/gonetz/GLideN64/files/3490139/ZELDA.MAJORA.S.MASK.zip)
Snowhead Temple Bridge

[...]

PR #2674, opened 2022-03-19
Zelda MM LoT fixes.

gonetz:
Fixed issues from tasks:
Majora's Mask Lens of Truth Depth Issue #2074
Lens of truth depth issues #2647

Both regular and N64 depth compare work.

### DIFF
--- a/src/FrameBuffer.cpp
+++ b/src/FrameBuffer.cpp
@@ -1006,7 +1006,7 @@ void FrameBufferList::attachDepthBuffer()
 	if (pCurrent == nullptr)
 		return;
 
-	DepthBuffer * pDepthBuffer = depthBufferList().getCurrent();
+	DepthBuffer * pDepthBuffer = pCurrent->m_isDepthBuffer ? depthBufferList().findBuffer(pCurrent->m_startAddress) : depthBufferList().getCurrent();
 
 	if (pCurrent->m_FBO.isNotNull() && pDepthBuffer != nullptr) {
 		pDepthBuffer->initDepthImageTexture(pCurrent);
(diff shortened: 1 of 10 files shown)


fzurita:
Is the reason that the new code is only activated for MM is because you are not sure if other games will break? Or do other games break for sure?

gonetz:
The mechanism of LoT in Zelda MM is unique for this game. My previous solution for Zelda MM LoT worked only for Zelda MM. My new solution is also very hackish and Zelda MM specific. I don't know how to make a general solution. Since I never saw smth like this in other games, it's ok to have a specialized solution for a specific problem. Note, that LoT works only in HLE mode with "One piece" BG mode. It was like that before, nothing new here.

Jj0YzL5nvJ (commented):
Small issue with the shadows outside the area of effect.
https://i.imgur.com/xc7EYbf.mp4

Savestate: http://0x0.st/oN2A.zip

fzurita:
I tested the one scenario with the ghost and I didn't see any issues and it appears fixed. I'll checked the shadow issue mentioned in the issue I made.

fzurita:
I do see the issue with the shadow. Shadows seem to disappear when the LoT is active, but only in the area outside the lens of truth effect as shown in the other image posted.

weinerschnitzel:
Would understanding how the actor flags are used in oot help to solve this? I dont think the code would be too different in mm, but some actors may have different flags.

Example case:

https://github.com/zeldaret/oot/blob/4f6967b02778e61ed0e9fa471f9ba7d67524c908/src/overlays/actors/ovl_En_Test/z_en_test.c#L1810

gonetz:
> Small issue with the shadows outside the area of effect. https://i.imgur.com/xc7EYbf.mp4
> 
> Savestate: http://0x0.st/oN2A.zip

Emulator crashes on loading of that savestate.

gonetz:
As I see, the shadow issue happens only with N64 depth compare enabled.
It is because only z-buffer image texture is copied, while delta-z image texture is just cleared.
delta-z is important for correct display of decal surfaces, like shadows.
I need to introduce 2 more texture inputs in shaders to copy not only z-buffer, but also delta-z one.

gonetz:
PR updated. Please disable/clear your shader cache.

gonetz:
Another update.

Jj0YzL5nvJ:
Fixed! It also fixes an unreported problem with footprints =P

PR #2679, opened 2022-04-08
Fix more issues found with Address Sanetizer

fzurita:
It seems like there are still some issues being detected by the Address Sanetizer. I took a shot at fixing them. I'm still not too sure about this call:
```
Rasterize(vdraw, 3, dzdx);
```

That expects 3 vertices always.


gonetz:
@fzurita You fix is wrong but thanks to your commit I finally noticed the error in this code.
Proper fix: 47b0de1ec1a

fzurita:
Ok, thanks!

PR #2722, opened 2022-11-20
Add qt5 dependencies

bladeoner:
It seems qt5-default is obsolete for some time now: https://www.mail-archive.com/ubuntu-bugs@lists.ubuntu.com/msg5902890.html

Replacing it with the qt5 dependencies seems to work which gives us time to see if we can use a newer version. I got a succesful build:
https://github.com/bladeoner/GLideN64/actions/runs/3507337878


gonetz:
Good, thanks!

PR #2723, opened 2022-11-20
Update msbuild

bladeoner:
This PR updates msbuild and removes another Node.js 12 deprecation warning.


bladeoner:
This fails because #2722 should be added first.

PR #2728, opened 2022-12-02
Check correct variable

mjunix:


gonetz (approved):
Ops, copy-paste bug. Thanks!

PR #2734, opened 2022-12-16
GLideNUI: remove fullscreen UI elements as needed

Rosalie241:
This is handy for front-ends which i.e don't support the vidext functions to retrieve the resolutions/refresh rates and disables the fullscreen frame as needed.


Rosalie241:
CI build failure is unrelated to this change, I'm not sure what's causing it, the URL looks correct and works fine on my end.

gonetz:
> CI build failure is unrelated to this change, I'm not sure what's causing it, the URL looks correct and works fine on my end.

Yes, the changes look correct. CI build succeeded after restart.

Issue #2729, opened 2022-12-05
Feature request: Texture folder devloper options


Nes-player4Life:
I was thinking the other day about how the plugin reads folders but not zips and how I often zip a folder up so the contents aren't loaded during editing.
Question is this; can a wildcard be made so any folder contents within a folder name are bypassed during load//reload of a pack.
Example: highres_/yourgamename/pac/foldernotreadname
This could be assigned to a simple key such as (~) or possibly userdefined.

gonetz:
I'd like to avoid adding a new user option. I think a key like "~" is too simple. I suggest to use "~!~" as the escape prefix. It is unlikely that this prefix is used by any existing HD pack.
It is done in PR #2735. Add "~!~" to a folder name inside your HD pack and this folder will be skipped.

Nes-player4Life:
Great news Thank you so much Gonetz. 
Can you belive GlideN64 is almost 10 years old now?

gonetz:
> Great news Thank you so much Gonetz.

You welcome :)

> Can you belive GlideN64 is almost 10 years old now?

Yep. And it still has so many issues :(

gonetz:
PR landed.

PR #2735, opened 2022-12-18
Skip folders with HD textures, if folder name has special prefix "~!~"

gonetz:
related to FR #2729


Nes-player4Life:
Thank you kindly

Nes-player4Life:
works absolutely wonderfully! 

Note to users:
By adding ~!~ to a texture folder will cause GlideN64 to not read folder. 
Useful in development with HighResEaser where multiple files of the same hash are used or in pack where user wish to temporarily isolate files from loading without removing them or zipping them up.

PR #2742, opened 2023-01-02
GLideNHQ: remove usage of CHDIR and GETCWD

Rosalie241:


Rosalie241:
I forgot that windows is annoying and requires `wchar_t` for paths to support unicode, will update this PR to use `_wfopen` on windows today.

gonetz (approved):

Issue #2236, opened 2020-04-30
Usage of cryptographically secure RNG for MinGW in NoiseTextures


m4xw:
Hey @gonetz , rather a old Issue that mudlord fixed on my fork like a 3/4 year ago.
I just started pulling in some updates and saw that enableNoise is gone.
![image](https://user-images.githubusercontent.com/13141469/80697994-1a7b1900-8ada-11ea-80a8-519e0c1f6b7c.png)
This may cause delays up to 20 seconds due to a lack of entropy without our fix under some circumstances.
Probably something you want to deal with upstream?
Reference replaces it with a simple xorshift (https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/GLideN64/src/NoiseTexture.cpp#L38-L44) and there is virtually no visual difference.
I'd be fine doing a PR too.

mudlord:
Problem was:

There was TLS issues with MinGWs implementation of srand and rand(). They use thread local storage which can interfere with how libretro's port of mupen works, thanks to its usage of libco.

Not using anything that relies on threads fixes any stalls. This seems to give enough entropy.

loganmc10:
Yeah I've never noticed any delay in upstream mupen64plus, or heard any complaints, and all the stuff I distribute for Windows uses MinGW

gonetz:
> Reference replaces it with a simple xorshift (https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/GLideN64/src/NoiseTexture.cpp#L38-L44) and there is virtually no visual difference.

![yoda](https://user-images.githubusercontent.com/8983092/80794793-03652580-8bc5-11ea-9e71-89214fcbac44.jpg)

gonetz:
> I'd be fine doing a PR too.

Please do a PR.

m4xw:
@gonetz Is it AL snippet tho?
If anything speaks against it I can implement it from scratch too.
Would you prefer a RNG class that takes in different types or just have it in NoiseTextures like we do for now?

mudlord:
Yes it is.

mudlord:
> Yeah I've never noticed any delay in upstream mupen64plus, or heard any complaints, and all the stuff I distribute for Windows uses MinGW

Prob because it doesn't rely on cothreads or how libco does cothreads.

gonetz:
> If anything speaks against it I can implement it from scratch too.

No, I don't think that it is a problem.

m4xw:
[...]

PR #2757, opened 2023-02-13
Improve Rand() in NoiseTexture.cpp

Rosalie241:
This patch is pulled from #2236 to improve the initializing noise texture speed, this patch makes it pretty much instant.


m4xw:
Should be obsolete if i ever get to #2236 for all intents and purposes but lets see

mudlord:
Thanks for this @Rosalie241 I forgot I even did this.

gonetz (approved):

PR #2759, opened 2023-02-19
Fix Display Bug in Settings UI

RetraCarteR:
The Settings UI has an issue where, when switching to the "Emulation" tab, the "Fix black lines between 2D elements" and "Background rendering mode (HLE only)" options appear squashed and garbled at the default window size. Attempting to resize the window temporarily fixes the issue by automatically expanding the window so that the options have enough space to display correctly, but switching to another tab allows the user to shrink the window back down again.

I assume that the expected behavior should be for the window's minimum height to be automatically set to the height of the largest tab, and for the window to disallow users from setting it smaller than that no matter what tab is open. This pull request implements this behavior by ~~changing the vertical Size Policy of tabWidget from "Expanding" (its default value) to "Minimum Expanding"~~ removing the erroneous minimum size property that was causing this issue.

An image showing the bug:
GLideN64_bug

An image showing the same screen after the fix is applied:
GLideN64_fix

### DIFF
--- a/src/GLideNUI/configDialog.ui
+++ b/src/GLideNUI/configDialog.ui
@@ -1445,9 +1445,6 @@
             <string>2D elements</string>
            </property>
            <layout class="QVBoxLayout" name="verticalLayout_48">
-            <property name="spacing">
-             <number>18</number>
-            </property>
             <item>
              <layout class="QVBoxLayout" name="verticalLayout_31">
               <property name="spacing">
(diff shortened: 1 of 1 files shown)


RetraCarteR:
Here's a picture of what the "Emulation" tab looks like after my latest commit. I was actually wrong about the window only being eight pixels larger than before I started, but it's still much smaller than it was before my latest commit.

GLideN64_fix_01

RetraCarteR:
Picture of the Settings UI window as of my latest commit (it is now exactly the same size as it was before the bugfix was applied):
GLideN64_fix_02

gonetz:
@RetraCarteR I have no issues with current GUI on my monitor, but if you had some and have fixed them, it's good. I have two notices:
* I don't like the new radio buttons placement for "Fix black lines between 2D elements". It is more compact, but it is easy now to overlook the first option. It looks like a choice between "Always" and "Never"
* The panel has too much free space, at least on my monitor, see the screenshot.
GUI

RetraCarteR:
I should have clarified that this issue occurs specifically in @Rosalie241's RMG emulator. All screenshots of the settings UI window that I've posted in this PR have been taken while it was running through RMG. I now suspect that it might not happen in other emulators, which is why this issue has flown under the radar for so long. Rosalie said on Discord that this has been an issue for her for a while now and that she's tried to fix it on her own in the past.

On RMG, this fix increases the default height of the settings UI window. After I showed Rosalie my initial fix, she expressed concern over the new default size of the window, believing that it was too large, and requested that I shrink it. Apparently, you are of the opposite mindset, so I've made one more commit to try and split the difference between the two of you. The radio button layout for "Fix black lines between 2D elements" has been reverted to its original vertical orientation and the spacing between options in the "Emulation" tab has been slightly increased, though it's still less than it was initially.

I've included a screenshot showing what the "Emulation" tab looks like as of my latest commit. If you're still unsatisfied with my layout changes, that's fine. You can just merge this PR up to my second commit, which fixes the bug but makes no changes to the layout. I would like to ask that you at least do that though, as the minimum size property that was causing this bug is clearly a remnant of an old layout from seven years ago that shouldn't exist in GLideN64 today, even if it is only causing problems with one emulator.

Thank you for your time, and I hope to hear back from you soon.

Picture of the Settings UI window as of my latest commit:
GLideN64

Rosalie241:
The main reason it doesn't happen in other emulators (i.e project64) is because RMG uses GLideN64 with Qt6 while the CI builds and project64 builds use Qt5.

gonetz:
Sorry for the late reply.
Current version of the PR looks better, but there is still an unused space at the bottom:
GUI
Can it be fixed? I tried, but failed. I don't understand, for what that blank space is allocated.
It's not a big deal, since total height of the dialog is not increased, but it's a kind of regression anyway.

Issue #2701, opened 2022-08-27
Asynchronous shader compilation


Nejsplitter:
Would it be possible to add an asynchronous(skip drawing) shader compilation method similar to dolphin emulator? (removes shader compilation stutters at the cost of slightly late rendering).

gonetz:
@Nejsplitter and anyone else with shader compilation stuttering problem: could you compare build from #2775 with regular build from master? Does it help with the problem anyhow? The fix works only if "threaded video" option is enabled, so you may turn it on/off to see the difference.

Nejsplitter:
hi @gonetz , sorry for taking so long to test this but i was about to do it now and realised the artifacts have expired, is there a chance i could get a download for it i use rosalies mupen gui

Nejsplitter:
Hi @gonetz , thank you but i did not notice a difference with this feature, i guess it doesnt matter anyway since i only wanted to test it on a custom mupen ANGLE branch that has no shader cache, thanks again and you might as well delete the pull request #2775 too since its not needed.

Issue #2693, opened 2022-07-05
Texture dump doesn't happen (Detailed explanation)


md500:
**Version information:**

 - Platform: Platform: Windows 10 Enterprise 1809 64bit build 17763.1012
 - Plugin version of Emulator: Project 64
 - GPU: Intel HD Graphics 620 
 - GPU Driver version: 26.20.100.8141
 - GPU Gen: 7th
 - CPU: Intel Core i5-7200U 2.50Ghz (3.10Ghz on turbo)
 - Architecture of Emulator: 32bit
 - Architecture of GlideN64: 32bit
 - Build/commit of Emulator: Dev-4.0.0.5825-3913fb5
 - Build/commit of GlideN64: a2f1660 (using GLideN64-a2f1660-Windows-Project64-WTL-x86.zip on Releases)

**Describe the bug**

Latest plugin version doesnt dump textures in Donkey Kong 64 (not tested if happens in other games). Tried setting File Storage instead of memory cache in N64 textures, Use File Storage istead of memory cache, and Save Enhanced texture cache to hard disk. Only hires_texturesS folder is loaded but texture_dumpS it seems that it does not see the folder and does not dump anything (although the program creates the folder if it is deleted). However Hi Res texture loading works.

![DUMP](https://user-images.githubusercontent.com/3111925/177421891-ac94c889-95f3-4492-837d-4ad7465b180b.png)

Folder permissions is in Everyone (no permissions error)

![PERMS](https://user-images.githubusercontent.com/3111925/177422254-cc8d6335-0f2c-4045-8fd1-be3203522563.png)

In Glide64 Final ( May 8 2012) when "Texture dumping/editing mode"  and "save texture cache to hard disk" is ticked, textures are dumped if i press "D" Button toggle, on keyboard (controls are configured to prevent conflict with Glide64) and hires_texture folder is loaded if there are texture files texture files there if (ej. hires_texture\DONKEY KONG 64\GlideHQ) exists and texture dumping happens in (ej. texture_dump\DONKEY KONG 64\GlideHQ), and Press "R" keyboard button to reload hires_texture regenerating its hires cache related files. In other words, you can dump textures and/or load hires textures at same time with option reloading with "R" button.

![glide64options](https://user-images.githubusercontent.com/3111925/177422147-2ce03058-e00e-4c66-81b0-912f814ec870.png)

GLideN64 finds hires_textures files without GlideHQ (i think to prevent colission if both gfx plugin texture location points to same folder).

[...]

PR #2775, opened 2023-04-22
Attempt to compile shaders asynchronously

gonetz:


Squall-Leonhart:
First reaction is, "Why?"

blackhand1001:
> First reaction is, "Why?"

Games like Banjo Tooie are stutter fests because they have thousands of shaders.

roioros:
I'm interested in this branch, how could I try it?

Jj0YzL5nvJ:
You can try my fork: https://github.com/Jj0YzL5nvJ/GLideN64/actions/runs/12737353428

Keep in mind:

- [Async mode is for speed, it corrupts some effects...](https://github.com/gonetz/GLideN64/issues/2693#issuecomment-1175794763)
- [This branch requires the use of "threaded video"](https://github.com/gonetz/GLideN64/issues/2701#issuecomment-1518674246)
- As far as I know, the current implementation of "threaded video" only works correctly with mupen64plus on Linux, it does not work on Windows

roioros:
Oh, I didn't remember that threaded video only works on Linux, I use Windows 10.

Thanks, I already have it downloaded anyway, do you know if it's possible to get threaded video working on Windows?

Thank you!

gonetz:
> * As far as I know, the current implementation of "threaded video" only works correctly with mupen64plus on Linux, it does not work on Windows

I can't say for sure if it works on Windows or not. The code itself has nothing Linux-specific. Some older versions of PJ64 can crash when this option is enabled. Newer versions do not crash. Mupen64plus based emus work too. I don't see any difference in performance though.

roioros:
From what I've been testing, in Project64 activating "threaded video" does not crash but I don't notice any difference when it is activated or deactivated, so I don't know if that option works well.

I still have stuttering problems, as I mentioned before, one of the most notable is the case of "Beetle Adventure Racing", you don't need to enter to play, just by watching the intro you can see the stuttering.

Once the intro is over (and the shaders are cached), if I restart the game the stuttering disappears completely.

Jj0YzL5nvJ:
> I can't say for sure if it works on Windows or not. The code itself has nothing Linux-specific. Some older versions of PJ64 can crash when this option is enabled. Newer versions do not crash. Mupen64plus based emus work too. I don't see any difference in performance though.

mupen64plus/mupen64plus-core#769

I have only tested it with `ui-console`, I don't know if it works with other frontends... or PJ64.

Issue #2776, opened 2023-04-23
frame duplication with threaded renderer


Nejsplitter:
would it be possible to have a frame duplication option in gliden64 for scenarios where a user has framebuffer emulation off+threaded renderer on? currently the frame duplication option in retroarch mupen nx only works if gliden64 threaded renderer is off.

frame duplication gives more smooth frame pacing for 30fps games on a 60hz monitor, it is possible to get frame duplication in all scenarios by enabling framebuffer emulation but not every user wants to deal with the performance penalty and lag of having framebuffer emulation always enabled just to get frame duplication/better frame pacing

gonetz:
Regarding "frame duplication option" - there is no such an option in GLideN64.
I may guess that you are talking about "Frame buffer swap" option, which controls frame buffer swap.
If it set to "On VI update", frame buffer updates 60 times per second for NTSC games and 50 for PAL ones.
It indeed works only if frame buffer emulation is enabled. It's not hard to implement similar logic when fb emulation is disabled: PR #2777

Nejsplitter:
Thank you gonetz, this is exactly what i had in mind, works well and i will close this issue as completed. Hopefully you might consider https://github.com/gonetz/GLideN64/issues/2794 at some point too and implement a feature that allows nearest filtering for all textures.

PR #2777, opened 2023-05-07
Correct fb swap when fb emulation is off, issue #2776

gonetz:


Nejsplitter1:
hi @gonetz is it possible to not use buffer swap at all, and render fbo right into front buffer?

Nejsplitter1:
hi @gonetz  is it possible to not use buffer swap at all, and render fbo right into front buffer? i want to test some things on my fork and wondering if i can achieve this

Nejsplitter1:
hi @gonetz , in vi.cpp what does this line do?

bNeedSwap = bCFB ? true : (gDP.colorImage.changed != 0);

and this one:

bNeedRender = (gDP.changed & CHANGED_COLORBUFFER) != 0;

also, is it possible to not use buffer swap at all like gln64 did?

PR #2788, opened 2023-07-10
Fix WTL MSAA slider

Frank-74:
Fixes the MSAA slider going 1000's of times past 8x MSAA.  Increased tic count to 4 for 16x MSAA max setting.


Frank-74:
Needs to get maxMSAA setting for dialog resource instead.  As the resource file expects max 8x MSAA, as it's hard coded currently, when it's set different elsewhere, there is no limit to maxMSAA.

PR #2812, opened 2024-01-21
Add "Try to adjust 16:9 game to fit" option (ultrawidescreen hack)

Calinou:
This is the 16:9 counterpart of the existing widescreen hack, which is designed for 4:3-native games. This one is designed for games that natively support 16:9 and are configured to run this way, or games that have 16:9 widescreen cheats available.

When targeting ultrawide aspect ratios like 21:9, adjusting from 16:9 results in fewer culling issues compared to adjusting from 4:3. The downside is that in some games, HUD elements will be stretched up to 16:9 (this depends on the game's widescreen implementation).

You can usually get good results with aspect ratios up to 21:9, although wider aspect ratios will start exhibiting more and more culling issues inherent to each game.

The current setting is preserved in existing configurations; "Try to adjust game to fit" was renamed to "Try to adjust 4:3 game to fit" for clarity.

Screenshot_20240121_203132

## Preview

*21:9 aspect ratio:*

[ultrawide10.webm](https://github.com/gonetz/GLideN64/assets/180032/84746794-309c-44ad-ac5c-256386206878)

*Very wide aspect ratios like 48:9 aren't viable for gameplay due to frequent culling errors, but sometimes, stars will align and there won't be any culling issues in a specific spot* :slightly_smiling_face:

Screenshot_20240121_202222 webp
Screenshot_20240121_203440 webp

### DIFF
--- a/src/DisplayWindow.cpp
+++ b/src/DisplayWindow.cpp
@@ -174,11 +174,20 @@ void DisplayWindow::_setBufferSize()
 			m_height = m_screenHeight;
 		}
 		break;
-	case Config::aAdjust: // adjust
+	case Config::aAdjust43: // adjust
 		m_width = m_screenWidth;
 		m_height = m_screenHeight;
 		if (m_screenWidth * 3 / 4 > m_screenHeight) {
 			f32 width43 = m_screenHeight * 4.0f / 3.0f;
+			m_adjustScale = width43 / m_screenWidth;
+			m_bAdjustScreen = true;
+		}
+		break;
+	case Config::aAdjust169: // adjust
+		m_width = m_screenWidth;
+		m_height = m_screenHeight;
+		if (m_screenWidth * 9 / 16 > m_screenHeight) {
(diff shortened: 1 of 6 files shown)


src/DisplayWindow.cpp
```diff
@@ -183,6 +183,15 @@ void DisplayWindow::_setBufferSize()
 			m_bAdjustScreen = true;
 		}
 		break;
+	case Config::aAdjust169: // adjust
+		m_width = m_screenWidth;
+		m_height = m_screenHeight;
+		if (m_screenWidth * 9 / 16 > m_screenHeight) {
+			f32 width43 = m_screenHeight * 16.0f / 9.0f;
```
gonetz, line 190: f32 width43 => f32 width169

gonetz (commented):
Looks good, I have just a minor remark.

gonetz:
merged to master: d1d5308b2414

Issue #2806, opened 2023-12-30
[HD] Turok 3: Cerebral Bore UI texture bug


GhostlyDark:
The red, center parts of the Cerebral Bore crosshair are supposed to dump as 4 textures. When dumping, only 3 textures appear however. 

**Leaving them out:**
![image](https://github.com/gonetz/GLideN64/assets/31742919/0171c2aa-4f7f-48ff-8faf-1829b35121cb)

**Adding the 3 dumps unedited:**
![image](https://github.com/gonetz/GLideN64/assets/31742919/512f31c7-498a-45bd-a629-2e615e57440c)

**Useful files:**
[Dumps.zip](https://github.com/gonetz/GLideN64/files/13798957/Dumps.zip)
[PJ64_Savestate.zip](https://github.com/gonetz/GLideN64/files/13798967/PJ64_Savestate.zip)

gonetz:
@GhostlyDark  I checked that problem. GLideN64 dumps only 3 of 4 textures because it calculates the same Rice CRC for two textures. GLideN64 texture CRC is different for them, so native textures work correctly. It can be a bug in GLideN64 implementation of Rice CRC calculation. I tried to dump textures with Rice plugin to check that. Unfortunately, I can't find these textures among dumped ones. Could you try to dump these textures with Rice and if it makes correct dump, attach it to the task?

GhostlyDark:
Rice exhibits the exact same issues.

![image](https://github.com/gonetz/GLideN64/assets/31742919/7a57c600-1c69-46dc-abc4-277ca3d5a7dd)

![image](https://github.com/gonetz/GLideN64/assets/31742919/2c4094cf-92aa-442c-9d5b-79654c073505)

![image](https://github.com/gonetz/GLideN64/assets/31742919/697cd007-e7b6-408a-aeab-e029d0fcfd6b)

gonetz:
@GhostlyDark , thanks! So, it's not a problem with implementation of Rice CRC, it is problem with Rice CRC itself.
The only way to solve it is to somehow modify CRC calculation.
I can suggest the following solution: 
* add new dump mode: "dump with strong hash".
* if some textures dumped incorrectly with vanilla RiceCRC, texture artist uses alternative dump mode.
Texture format remains the same, only checksum in texture name will be different.
* During textures load, graphics plugin will first try to load HD texture as usual, with RiceCRC.
If load fails, the plugin calculates alternative CRC and tries to load HD texture with it.

[...]

PR #2820, opened 2024-02-25
Add "Use strong CRC for texture dump" option.

gonetz:
Problem: RiceCRC fails to calculate different CRC for some textures, see #2806

Solution:
add new dump mode: "Use strong CRC for texture dump".
If some textures fail to dump with RiceCRC, texture artist should use alternative dump mode.
Texture format remains the same, only checksum in texture name is different.
During textures load, graphics plugin will first try to load HD texture as usual, with RiceCRC.
If load fails, the plugin calculates alternative CRC and tries to load HD texture with it.


GhostlyDark:
I am stumbling into crashing issues with these changes. Most easily reproducible is running GoldenEye with HD textures on and firing a gun. This does not happen with latest master it seems.

gonetz:
> I am stumbling into crashing issues with these changes. Most easily reproducible is running GoldenEye with HD textures on and firing a gun. This does not happen with latest master it seems.

I'll check it, thanks!

gonetz:
> > I am stumbling into crashing issues with these changes. Most easily reproducible is running GoldenEye with HD textures on and firing a gun. This does not happen with latest master it seems.
> 
> I'll check it, thanks!

Fixed.

PR #2826, opened 2024-03-26
Mip Map support for HD textures

gonetz:


gonetz:
Draft.

Jhair007:
Hey, but I have this problem. Apparently, it is not compatible with Texture Packs
Bug

PR #2832, opened 2024-04-22
Cleanup unused code in NoiseTexture.cpp

scurest:
Cleans up some unused stuff in NoiseTexture.cpp left over from before #2757.

Should have no change in behavior.


scurest:
Note that this PR is mutually exclusive with #2833 which completely deletes NoiseTexture.cpp. If you want to merge #2833, you can close this one.

Issue #1474, opened 2017-04-28
Is it possible to re-add glN64's noise at PC resolution option?


Aquatakat:
Some games use the noise effect to add a film grain effect, like the Rush series. When running games like that at high resolution the effect looks blocky.

gonetz:
* I don't know what is glN64's noise.
* It is surely possible to calculate noise with higher precision, up to per-pixel. It will take more texture memory, but technically nothing hard. However, I don't have time for that atm.

Aquatakat:
@gonetz GLideN64 used to do actually do noise at PC resolution at some point a long time ago. At some point it changed to match N64's resolution instead. If you don't have time you can mark this issue as low priority, it's just a thought I had while playing Rush 2049.

gonetz:
Yes, I used noise shader, which generated pseudo-random output.

Ok, I'll mark it as low priority.

mudlord:
Funny, I remember doing such BS............................
https://github.com/gonetz/GLideN64/issues/710

Not sure if its worth the bother resurrecting it.

fzurita:
@mudlord Your solution is not bad at all:
http://mudlord.info/blog/2017/05/21/gpu-based-noise-emulation-take-2/

```
static const char* fragment_shader_noise =
"highp float rand(vec2 co) {\n"
"return fract(sin(mod(dot(co.xy,vec2(12.9898,78.233)),3.14) * 43758.5453));\n"
"}\n"
"highp float snoise()									\n"
"{														\n"
"  mediump vec2 coord = floor(gl_FragCoord.xy/uScreenScale);		\n"
"  return rand(vec2(rand(coord*12.9898),rand(coord*78.233)));	\n"
"}														\n"
;
```
It would really simplify a lot of code, I wonder how performance would be in low end devices.

mudlord:
you will still need a timer uniform to see the rng.

oddMLan:
I wonder if mudlords solution can be looked at again. It worked superbly in SM64 last time I saw.

PR #2833, opened 2024-04-25
Replace noise textures with hashing

scurest:
Shaders currently get noise by sampling a 640x580 noise texture with the fragment location. Variation over time is provided by keeping a pool of 30 noise textures and binding a different one every frame.

This PR replaces this with hashing the fragment location and a frame-varying seed value, obviating the need for noise textures.

```glsl
// Before
uniform sampler2D uTexNoise;
noise = texture2D(uTexNoise, coord).r;

// After
uniform float uNoiseSeed;
noise = hash(vec3(uNoiseSeed, coord));
```

### Noise seed

uNoiseSeed is updated every frame with the low byte of the frame count, ie. it's a simple counter with period 256. This is unlike the noise textures, which would cycle in a random order with long period (but never repeating on two consecutive frames). I don't think this matters, but I can easily change it to the old behavior if you like.

The seed changes automatically every frame and does not need an explicit `update()` when noise is used.

### Hash function

The hash functions used are Dave Hoskins's [Hashing without Sine](https://www.shadertoy.com/view/4djSRW) GLSL functions that take 1-4 integer-spaced float inputs and produce 1-4 float outputs in [0,1).

### Noise resolution

One noise value is generated per native-res pixel, ie. as by

```glsl
ivec2 coord = ivec2(gl_FragCoord.xy/uScreenScale);
noise = texelFetch(uTexNoise, coord, 0).r;
```

To match this, the hash function is passed the native-res pixel location

```glsl
vec2 coord = floor(gl_FragCoord.xy/uScreenScale);
```

Note that, as pointed out in #1474, generating high-res noise would be as easy as switching to `vec2 coord = gl_FragCoord.xy`. But that is not proposed in this PR.

### Correlations between noise functions

Previously `snoise()`, `snoiseRGB().r`, and `snoiseA()` were always the same, because they sampled the noise texture at the same location. Does this property need to be maintained? `snoiseRGB().r` will now be different than the others, because it uses a different hash function.

### Font atlas

All noise texture code has been removed, except the `noiseFormat` etc. variables. These are also used for the font atlas, so they've been renamed `fontFormat`, etc. to reflect their current use.

### Testing

`snoise()` function works in the Kirby 64 intro.

**I have not tested the `snoiseRGB()` or `snoiseA()` functions, in particular the `enableHiresNoiseDithering` option**. I'm happy to do so if you can tell me how/where to test it.

Thanks for considering.


mudl0rd:
Thanks for having a proper look at what I did back then. :)

gonetz (approved):
LGTM!

gonetz:
@scurest  Very accurate work. The result is really good.
I tested several cases where noise is used and didn't find any regressions.
The noise looks pretty random, as good as with noise textures. Thanks!

> Previously snoise(), snoiseRGB().r, and snoiseA() were always the same, because they sampled the noise texture at the same location. Does this property need to be maintained?

I think, no. We need a random input for each color channel, that's all.

> snoiseRGB().r will now be different than the others, because it uses a different hash function.

ok.

> I have not tested the snoiseRGB() or snoiseA() functions, in particular the enableHiresNoiseDithering option. I'm happy to do so if you can tell me how/where to test it.

As I remember, snoiseA() is used in SM64 with Vanish Cap (invisibility hat).
snoiseRGB() is used for noise color dithering. It sometimes is used in cut-scenes.
As I remember, it is used in Zelda MM intro, when Link falls into the pit, and in "remembering Zelda" black-and-white cut-scene. I checked these scenes, no problems found.

@mudl0rd 
> Thanks for having a proper look at what I did back then. :)

Yes, the same idea, but seems to work properly this time.

scurest:
@gonetz 
So this broke shader compilation on GLES (#2837) because

```glsl
uniform float uNoiseSeed;
```

needs to have a precision qualifier

```glsl
uniform mediump float uNoiseSeed;
```

which is an easy fix.

However now that I think about precision, I'm pretty sure the hash functions requires `highp` for `p3`; if `mediump` is a float16, the noise will always be 0. This creates a problem for old GLES implementations that do not have `highp`. Can you confirm if these are supported? If they are, this will need to reverted, or the hashing constant will need to be tweaked somehow.

It looks to me like GLideN64 does try to support them (by replacing `highp` with `mediump`)

https://github.com/gonetz/GLideN64/blob/b021d8ee437266cfdd7251daf8c23203578b02b6/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilderCommon.cpp#L32-L34

PR #2835, opened 2024-04-30
Change copyToRDRAM default when building with M64P_GLIDENUI

Rosalie241:
Should fix all corruption when resizing the window for all games, because some users were still experiencing this issue

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -73,7 +73,11 @@ void Config::resetToDefaults()
 	frameBufferEmulation.copyDepthToRDRAM = cdSoftwareRender;
 	frameBufferEmulation.copyFromRDRAM = 0;
 	frameBufferEmulation.copyAuxToRDRAM = 0;
+#ifdef M64P_GLIDENUI
+	frameBufferEmulation.copyToRDRAM = ctSync;
+#else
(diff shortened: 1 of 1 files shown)


GhostlyDark:
Should this really be specific to `M64P_GLIDENUI` instead of being the default in general? Some games need it in the first place (which comes into play when per-game settings are turned off), though I have no idea how much it would impact performance out of the box.

Rosalie241:
> Should this really be specific to `M64P_GLIDENUI` instead of being the default in general? Some games need it in the first place (which comes into play when per-game settings are turned off), though I have no idea how much it would impact performance out of the box.

I'm not sure, but the performance impact question is why I limited it to `M64P_GLIDENUI`

gonetz:
> I have no idea how much it would impact performance out of the box.

The impact is noticeable for games running in Hires mode, that is with internal resolution above 320x240.
I'd prefer to enable sync mode via custom game ini for safety. However, if Rosalie thinks that it should be default mode for M64P_GLIDENUI, ok.

Rosalie241:
if users will start complaining about worse performance when I release a new version, I'll consider reverting it, but we'll see for now, thanks for merging it! :heart:

PR #2856, opened 2024-07-07
Simplify _readScreen2 for non-mobile builds

Morilli:
Removes an unnecessary allocation while keeping the codepath that was supposedly necessary on android (see #1003).


gonetz (approved):
Yes, it looks better that way, thanks!

PR #2890, opened 2024-12-18
Implement mge_get_video_size

Aurumaker72:
Required for capture on legacy Mupen64, as it needs a way to request the video size in a thread and wgl context-agnostic way.

### DIFF
--- a/src/ZilmarGFX_1_3.h
+++ b/src/ZilmarGFX_1_3.h
@@ -20,6 +20,7 @@ the plugin
 #define _GFX_H_INCLUDED__
 
 #if defined(__cplusplus)
+#include <cstdint>
 extern "C" {
 #endif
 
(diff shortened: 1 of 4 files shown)


src/ZilmarGFX_1_3.h
```diff
@@ -285,6 +286,15 @@ EXPORT void CALL ReadScreen (void **dest, long *width, long *height);
...
 EXPORT void CALL DllCrtFree(void* addr);
 
+/******************************************************************
+  Function: mge_get_video_size
+  Purpose:  Gets the current video size.
+  Input:    width - Pointer receiving the video width. Can be null.
+		    height - Pointer receiving the video height. Can be null.
+  Output:   none
+ ******************************************************************/
+EXPORT void CALL mge_get_video_size(int32_t* width, int32_t* height);
```
gonetz, line 296: Is it essential to use int32_t? Why not just int?

Aurumaker72: It's not essential, but since x64 support for legacy Mupen64 is planned, the Zilmar spec headers have had their types indiscriminately changed to the C++ constant-width ones to avoid miscellaneous annoyances during porting of plugins.

This specific change is free to be reverted.

gonetz: ok

PR #2897, opened 2025-01-12
CI: Add missing dependencies and other minor changes

Jj0YzL5nvJ:
See: https://github.com/gonetz/GLideN64/pull/2895#issuecomment-2585845545

PR #2900, opened 2025-01-24
CMake: expose option to disable OSD text drawing (TextDrawerStub.cpp)

Morilli:
This allows using `TextDrawerStub.cpp` instead of `TextDrawer.cpp` in the compilation to disable text drawing, which also removes the freetype dependency.


gonetz (approved):
Looks ok.
I think it should be rebased to the current master.
I merged https://github.com/gonetz/GLideN64/pull/2897
so Linux build should be fixed after that.

PR #2908, opened 2025-02-16
GLideNHQ: fix build with gcc 15

orbea:
Adds a missing include to fix a build failure with gcc 15 as reported for the Gentoo issue tracker.

Gentoo Issue: https://bugs.gentoo.org/949732


orbea:
@gonetz If you have time I would really appreciate if you could review this PR.

gonetz (approved):

PR #2911, opened 2025-03-23
Add bounds check to gSPCullVertices()

Rosalie241:
This fixes a crash that only happens with a decompressed OOT ROM when trying to open the pause menu.

The issue is that for some reason `vn` is way higher than the amount of `triangles.vertices`, `vn` is in some cases `2204` or even higher.

Obviously this is a game bug, but I do think a crash from GLideN64's side should be fixed, I'm not entirely sure if a `std::min()` is preferred or an early return, but I'd rather stay on the safe side because I'm not sure how other games behave.

I added some debug logging which you can see:
```
vn is too high = 2204
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x0E"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x0A"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x0D"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x2D"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x27"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x17"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x0D"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x12"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0xC1"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x50"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x24"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x18"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x37"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x36"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x25"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0xBA"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0xAA"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x37"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x09"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0xAB"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x7D"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x0D"
vn is too high = 2204
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x0E"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x0A"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x0D"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x2D"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x27"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x17"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x0D"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x12"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0xC1"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x50"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x24"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x18"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x37"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x36"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x25"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0xBA"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0xAA"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x37"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x09"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0xAB"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x7D"
[GFX]   GBI.cpp:159, "UNKNOWN GBI COMMAND 0x0D"
vn is too high = 2204
```

### DIFF
--- a/src/gSP.cpp
+++ b/src/gSP.cpp
@@ -1568,7 +1568,7 @@ bool gSPCullVertices( u32 v0, u32 vn )
 	}
 	u32 clip = 0;
 	GraphicsDrawer & drawer = dwnd().getDrawer();
-	for (u32 i = v0; i <= vn; ++i) {
+	for (u32 i = v0; i <= std::min(vn, (VERTBUFF_SIZE - 1)); ++i) {
 		clip |= (~drawer.getVertex(i).clip) & CLIP_ALL;
 		if (clip == CLIP_ALL)
 			return false;


gonetz:
There couldn't be "UNKNOWN GBI COMMAND" on OOT. If it happens, it means that memory was broken or some other core issue. So, your fix only hides the real problem. I may check it if you tell me how it can be reproduced.

Admentus64:
A decompressed OoT ROM (64 MB, either decompress with ndec or use the already decompressed debug ROM) can crash in two different ways.

1. Press Start to open the Pause Screen. If there's a minimap active it will most likely crash. That would be inside dungeons or overworld areas. Not every overworld area crashes the game though like when in Kakariko Village. The most likely cullprit in C code is in the Minimap_Draw function (z_map_exp.c file using the decomp) where the code condition fires a bit too late causing some kind of race condition. Opening the Pause Screen goes through a value cycle from 0 to 6 (where on 6 it is loaded) and the minimap code applies when it's value 3 or lower. In other words once it reaches value 4 when opening the pause screen it stops drawing.

2. Dying as Adult Link. Dying in the first Deku Tree dungeon should work here. And do Save & Continue (Continue mostly) and it will freeze after Continue is selected. From what I understand it seems to be some kind of camera pose problem for Adult Link. You can use the Debug ROM and the Debug Map Select to easily warp to any location with either age.

gonetz:
I checked this scenario with Project64 and got a similar issue, but not the same.
It also started as a sequence of "UNKNOWN GBI COMMAND " but crushed in gSPModifyVertex, not in the gSPCullVertices() as in your case. It seems that an arbitrary command can be called when it happens. Thus, your fix won't solve the problem in general case. We need to validate all parameters in all commands, which is redundant in general. I made a simpler and fast fix, which solves the out of bounds read for the vertices array, see the attached patch. Please check that it helps. It doesn't help for Project64 though, as it hangs somewhere else anyway when that problem happens.
[0001-Fix-vertices-array-out-of-bounds-read.patch.zip](https://github.com/user-attachments/files/19517304/0001-Fix-vertices-array-out-of-bounds-read.patch.zip)

GhostlyDark:
> I checked this scenario with Project64 and got a similar issue, but not the same. It also started as a sequence of "UNKNOWN GBI COMMAND " but crushed in gSPModifyVertex, not in the gSPCullVertices() as in your case. It seems that an arbitrary command can be called when it happens. Thus, your fix won't solve the problem in general case. We need to validate all parameters in all commands, which is redundant in general. I made a simpler and fast fix, which solves the out of bounds read for the vertices array, see the attached patch. Please check that it helps. It doesn't help for Project64 though, as it hangs somewhere else anyway when that problem happens. [0001-Fix-vertices-array-out-of-bounds-read.patch.zip](https://github.com/user-attachments/files/19517304/0001-Fix-vertices-array-out-of-bounds-read.patch.zip)

I tried this patch but it doesn't solve anything. Instead, `UNKNOWN GBI COMMAND` is logged indefinitely now, which causes an `Invalid string error` in my launcher as the log causes an overflow rather quickly.

gonetz:
Interesting side effect. I'll try to investigate it.

gonetz:
> I guess gSPCullVertices returns different result now and that bad display list never ends.

Yes, it is the reason. New patch
[patches.zip](https://github.com/user-attachments/files/19521679/patches.zip)

GhostlyDark:
> > I guess gSPCullVertices returns different result now and that bad display list never ends.
> 
> Yes, it is the reason. New patch [patches.zip](https://github.com/user-attachments/files/19521679/patches.zip)

That entirely fixes the crashes, both instances that Admentus described. We can now sometimes see the (yellow) player cursor being distorted when pausing, which could possibly be the reason for otherwise crashing the game.

image

Admentus64:
It's caused by Minimap_DrawCompassIcons (z_map_exp.c, line 351) which draws the cursors for Link's position and last entrance position, specifically either `gSPDisplayList(OVERLAY_DISP++, gCompassArrowDL);` call.

gCompassArrowDL consists out of this:

```Gfx gCompassArrowDL[] = {
    gsSPTexture(0xFFFF, 0xFFFF, 0, G_TX_RENDERTILE, G_OFF),
    gsSPClearGeometryMode(G_CULL_BACK),
    gsSPVertex(gameplay_keepVtx_00C848, 3, 0),
    gsSP1Triangle(0, 1, 2, 0),
    gsSPEndDisplayList(),
};```

Minimap_DrawCompassIcons ends it's function with `CLOSE_DISPS(play->state.gfxCtx, "../z_map_exp.c", 607);`

gonetz:
> That entirely fixes the crashes, both instances that Admentus described.

Good. Thanks for testing. I pushed the fixes. @Rosalie241 thanks for the initial fix!

PR #2912, opened 2025-03-30
F3DEX3: Fixed Strip and Fan commands flushing

aglab2:
Fixed Strip and Fan commands flushing. They behave similarly to SPTriangleX commands so I have adjusted to code to handle flushing appropriately.


gonetz (approved):
Looks good

PR #2924, opened 2025-05-28
Update video spec

oddMLan:
Depends on https://github.com/project64/project64/pull/2483

The video spec was recently updated to allow access to RSP registers.
This can potentially improve support for games like Gauntlet Legends, World Driver Championship, and a few other games to work properly in HLE.

As a bonus, I added the functionality from the newer spec, which passes a status string to the plugin while in full screen mode, for an improved OSD (eg. messages that would normally show in the status bar, like selecting a save slot through the 1-9 keys, or loading/making a savestate, will show up on full screen now).

GLideN64_OSD_PJ64_status

Considerations:
- This **will** break backwards compatibility with ~~Project64 1.6~~ ALL old versions of Project64 including 2.x and 3.0.1 (will be compatible with 3.0.2 once it releases), and 1964, og Mupen64 and Nemu as well (depends how they have implemented the 1.3 zilmar gfx spec). However, no one should be using those old emulators anymore, especially PJ64 1.6, which had a [few vulnerabilities discovered a while back](https://gist.github.com/aglab2/6bb351c4bd0ac6e13c931e03aa8f9949).
- I still put the old 1.3 gfx spec headers in a subfolder just in case. Keep in mind, the older spec used windows types such as HWND, and the newer spec uses void types which allows multiplatform support. I really don't think it's worth adding more #ifdefs than there currently are just to keep support for obsolete emus available.


Fanatic-64:
> The video spec was recently updated to allow access to RSP registers.
Exactly how recently was that update made? Is it present in PJ64 3.0, or only in 4.0 nighty builds? In the latter case, Gonetz might want to wait until 4.0 gets a stable release (since it's in a rather unfinished state currently) before breaking compatibility with all previous versions.

gonetz:
Merged

oddMLan:
> This PR broke the WTL plugin on Pj64 4.x

No it didn't 
It was broken way before this but I couldn't find the root cause.
Thanks in advance for #2937, which hopefully should fix the real issue.

PR #2954, opened 2025-12-04
GLideNUI: hide translations UI when no translations are found

Rosalie241:
This patch makes the translations UI hidden when no translations have been found, but keeps the same behavior for any build configuration that isn't `M64P_GLIDENUI`.

It achieves this behavior by adding a spacer, which is deleted when translations have been found or GLideNUI is built without `M64P_GLIDENUI`, else, when no translations have been found, it hides the translations groupbox instead.

### DIFF
--- a/src/GLideNUI/ConfigDialog.cpp
+++ b/src/GLideNUI/ConfigDialog.cpp
@@ -499,6 +499,15 @@ void ConfigDialog::setIniPath(const QString & _strIniPath, const QString & _strS
 	QStringList translationFiles;
 	_getTranslations(translationFiles);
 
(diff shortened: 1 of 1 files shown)


Rosalie241:
Should be fixed with the latest changes, thanks for letting me know!

gonetz:
> Should be fixed with the latest changes,

Well, it works as before with the current patch of course, because you guarded all the changes by #ifdef M64P_GLIDENUI.
I tried to check, if it works with Zilmar build too. Your code compiled fine, but then I got several link errors:
image
I don't want to fight with QT libs again. If your fix works for RMG, ok.

Issue #2950, opened 2025-11-18
Missing Geometry On Rogue Squadron


Nioh8:
Hi I am using RMG 0.8.8 with gliden64 

I’m running windows 11 and my cpu is a AMD Ryzen 7 8745HS 

The game is missing a lot of its geometry as you can see in the video I will link below! 

Any ideas on a solution thanks 

https://limewire.com/d/WSCS7#XvroZp0SCW

Nioh8:
https://github.com/user-attachments/assets/dc8b0e4c-fe5a-41a7-96f5-d033c64df848

@gonetz managed to upload it here

gonetz:
> [@Rosalie241](https://github.com/Rosalie241), this only happens with your **Windows** distribution... or with NVIDIA GeForce RTX 4060 _(driver 581.80)_...

I installed RMG 0.8.8 and got the same problem with missing polygons as on Nioh8's screenshots/video
I built the current version of GLideN64 and tested it with Project64 3.0.1
The game works properly, with no missing polygons.
Then I tested the current version of GLideN64 with old versions of mupen64plus-gui, which support GLideN64.
Both 32bit and 64bit versions work well.
So, it is not a problem with Windows or drivers. Something is broken on RMG side.

Nioh8:
@Rosalie241 i can confirm on my windows machine, with the build you have linked, the issue has been resolved for me

Jj0YzL5nvJ:
That build works... As part of the diagnostic, it's fine. But I don't like it as _hotfix_. Perhaps `-flto=thin` would be a better alternative?

Rosalie241:
Fix has been included in RMG v0.8.9

PR #2956, opened 2025-12-04
cmake: add USE_IPO option

Rosalie241:
This allows people to disable LTO/IPO during the build.


gonetz (approved):
So, it makes IPO/LTO optional.
I guess it should fix #2950, if USE_IPO is set to OFF
Ok.

Issue #2798, opened 2023-09-15
Banjo-Kazooie: Some textures from texture pack not loading in Mupen64Plus


TheEarthrise:
Some textures in Banjo-Kazooie are not loading with Mupen64Plus GlideN64, but they do in Project64 GlideN64. Settings are all the same. I dumped the textures to see what's going on, just to find out that all textures that are not loading have the same name and therefore get overwritten while dumping them.

Mupen64Plus:
![banjo-kazooie-000](https://github.com/gonetz/GLideN64/assets/87270139/3bcd5940-2206-48af-9553-a8db5da8635c)

Project64:
![banjo-kazooie-001](https://github.com/gonetz/GLideN64/assets/87270139/33fc9526-c50c-451b-913a-a3a6d30fa79a)

Another texture pack on Mupen64Plus, same problem:
![banjo-kazooie-002](https://github.com/gonetz/GLideN64/assets/87270139/37ca57f1-8dd4-4090-a897-5a747690205f)

Jj0YzL5nvJ:
mupen64plus/mupen64plus-user-issues#730

weinerschnitzel:
Interesting... Those are jpeg textures? I recall behavior with textures missing with neutered rsp plugins without the jpeg decoding (zelda bgs) but not the text in banjo...

I guess performing bsmiles modification to HLERSP would tell the tale. 

Or you could try one of the HLERSP plugin ports to pj64 spec and see if it shows the same issue as mupen64plus. I did one before but I think rosalies is most recent.

That would confirm the same HLERSP bug.

Or try cx64 rsp with both. Theory would be it is correct in both emulators with cx64.

If this does not pan out, has to be a different issue.

TheEarthrise:
I have done some further testing as suggested. First, I tried RMG with all the RSP plugins it ships with (cxd4, parallel and hle). All show the same result as shown above. Then I tried cxd4 and hle in Project64, again everything remained the same, means in PJ64 all textures were loaded fine, regardless of the RSP plugin chosen.

GhostlyDark:
Banjo-Kazooie does not use JPEG backgrounds. This has nothing to do with the RSP plugin.

[...]

PR #2959, opened 2026-01-05
GLideNHQ: Fix Banjo-Kazooie 0x0 texture dumps

GhostlyDark:
Fixes the dumped texture names for `#0#0` textures. This seems to match Rice behaviour.

Banjo-Kazooie is to my knowledge the only game that benefits from this. Affected textures include (mostly Pause Screen):

- Eggs
- Gold Feathers
- **Jinjos (changing color by using different palettes!)**
- Joker (Grunty's Furnace Fun)
- Mumbo Token
- Propeller (special timer in Rusty Bucket Bay)
- Red Feathers

Sidenote: GLideN64 can load either `#0#0#[...]_ciByRGBA` or `#0#0_all` as HD textures and this change solely affects the naming of such textures when dumped.

See: https://github.com/gonetz/GLideN64/issues/2798#issuecomment-1987193853


gonetz (approved):
Cool!
I forgot about that issue.
Thanks for the fix!

PR #2963, opened 2026-02-01
Switch to vcpkg package manager.

gonetz:
Delete builds and headers of external libraries.
vcpkg QT package is not added yet - external static build is still required.


gonetz:
This PR adds use of vcpkg package manager to manage external libraries.
I added it only for Visual Studio project files.
vcpkg works with CMake, but I haven't time to fix CMake scripts yet.
To install and setup vcpkg:
https://learn.microsoft.com/en-us/vcpkg/get_started/get-started-msbuild?pivots=shell-powershell

I use Microsoft Visual Studio 2026 Community.
I checked 32bit builds for Project64 and 32/64bit builds for mupen64plus.

Other platforms need to be checked.

gonetz:
@GhostlyDark @Rosalie241 @project64 @fzurita

Rosalie241:
> vcpkg works with CMake, but I haven't time to fix CMake scripts yet.

there's no real need to add it to cmake because on Linux and even for Windows through msys2, packages are mamaged seperately by the system's package manager, if you do add it to cmake, please make it optional.

PR #2968, opened 2026-06-07
Switch to vcpkg for MSVC

GhostlyDark:
Supersedes #2963.

Does not modify anything non-MSVC, simply requires `vcpkg integrate install` from the user so that `MSBuild` can do its magic.


gonetz:
@GhostlyDark  Thanks for the updated PR. I was going to test and merge it, but my VS 2026 asked me for update.
After the update I found that I can't build QT versions of GLideNUI anymore with our static QT libs. The libs are not compatible with the latest VS, lots of compile errors. I downloaded latest sources for QT 5.15 and made new static builds. Luckily our wiki instruction is not completely outdated yet. New builds solved the problem. Now I need to upload these builds and correct the build scripts. After that we can go further.

PR #2977, opened 2026-07-16
Custom anisotropic filtering for in-shader texture filtering

### LINKED ISSUE #2967: Anisotropic filtering doesn't work
The anisotropic filtering slider has no effect on the current development version of GLideN64; all settings look the same as 0x. Tested on a clean installation on AMD, Intel and Nvidia GPUs.

With 0x anisotropic filtering:
Image

With 16x anisotropic filtering:
Image

gonetz:
* The accurate combiner (texelFetch) and the fast combiner (per-texel texture()/textureGrad) do their own bilinear/3-point filtering, so the sampler's GL_TEXTURE_MAX_ANISOTROPY does not apply. Emulate AF in the shaders.
* Accurate path: supersample the custom filter along the major axis of the pixel's texture-space footprint (from screen-space derivatives); for mip-mapped textures also bias the LOD toward the minor axis. Non-LOD and mipmap paths, LOD emulation on/off, desktop and GLES2. Driven by a uMaxAnisotropy uniform wired through UTextureFetchMode and UMipmap.
* Fast path: for textures with a mip chain, keep hardware AF but sample with textureGrad using the continuous pre-wrap gradient (dFdx(vTexCoord)*uTexScale) to avoid seams at texture-repeat boundaries. For textures without a mip chain (HD textures, detected via uTextureHD; and the fake-mipmap path), hardware AF cannot work, so supersample manually at the minor-axis LOD. Original (non-HD) textures are unaffected.
* Bump shader storage format version to force shader regeneration.

### DIFF
--- a/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilderAccurate.cpp
+++ b/src/Graphics/OpenGLContext/GLSL/glsl_CombinerProgramBuilderAccurate.cpp
@@ -323,6 +323,7 @@ public:
 
 			if (CombinerProgramBuilder::s_textureConvert.useTextureFiltering()) {
 				shaderPart += "uniform lowp int uTextureFilterMode;								\n";
+				shaderPart += "uniform lowp int uMaxAnisotropy;									\n";
 				shaderPart += "#define TEX_NEAREST(name, tex, tcData)							\\\n"
 					"{																			\\\n"
 					" name = texelFetch(tex, ivec2(tcData[0]), 0); \\\n"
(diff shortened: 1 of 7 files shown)


gonetz:
I tested that commit on my desktop only.
I tested only Zelda MM, Termina fields.
I tested both fast and accurate texture path, with mip-mapping emulation on/off.
I tested it with original textures and with Nerrel's HD pack.
It doesn't work perfectly in all cases, sometimes mip-mapped textures still shake on a distance.
Anyway, it now looks way better with AF enabled.
The code is not perfect. If somebody can improve it, welcome.
The code is probably is not fast enough to work on mobile platforms.
It probably won't work on mobile platforms at all. It should be tested.

@fzurita can you test it on gles3 and gles2 devices?
gles2 code has been added too, but no guarantee that it will work somehow.

PR #2978, opened 2026-07-18
Compile combiner shader programs asynchronously to avoid stuttering.

gonetz:
If GL_ARB/KHR_parallel_shader_compile is supported, compilation of a new combiner program is issued without waiting for the result: the driver compiles it in background threads while the current program is used as a substitute. Ready programs are added to the combiners map when the display list is finished, see gDPFullSync().
    
New config option: asyncShaderCompilation (enabled by default on Android).


gonetz:
It is the second attempt to solve "stuttering" problem caused by synchronous compilation of heavy GLideN64 shaders.
When display list calls gDPSetCombine with completely new parameters, compilation of the corresponding shader is started but the result is not waited, and the current shader is used instead. So, you may see a wrong shader for a frame or two, until the shader get compiled. On my desktop all pending shaders (up to 50 per frame) were ready to the end of the display list, so wrong shader was used for only one frame.
I think this is the best I can do. It should eliminate all stuttering caused by shader compilation, so the plugin should run as smoothly as other plugins, provided your hardware is powerful enough to run GLideN64 shaders.

